PR #33, opened 2026-02-03
Add reasoning parser support

### LINKED ISSUE #26: Reasoning flag support
Does vllm-mlx have a way to pass the --reasoning flag to parse think tokens?  If so could you point me to the documentation?  passing reasoning on the vllm-mlx serve cli isn't allowed.

I'm providing https://docs.vllm.ai/en/latest/features/reasoning_outputs for context.

Thank you

waybarrios:
Add `--reasoning-parser` flag support following vLLM style architecture where reasoning and tool parsing are separate systems.

### Reasoning Parsers
Two built-in parsers for extracting `<think>` content:

| Parser | Models | Description |
|--------|--------|-------------|
| `qwen3` | Qwen3 series | Requires both `<think>` and `</think>` tags |
| `deepseek_r1` | DeepSeek-R1 | Handles implicit `<think>` tag (just `</think>`) |

### Think Tag Stripping in Tool Parsers
Prevents parsing failures when models produce `<think>` tags with tool calls (fixes Ring-Mini-Linear-2.0 + hermes issue mentioned by @TomLucidor).

## Usage

**Server:**
```bash
vllm-mlx serve mlx-community/Qwen3-8B-4bit --reasoning-parser qwen3
```

**Client:**
```python
response = client.chat.completions.create(
    model="default",
    messages=[{"role": "user", "content": "What is 17 × 23?"}]
)
print("Thinking:", response.choices[0].message.reasoning)
print("Answer:", response.choices[0].message.content)
```

## Implementation

**Reasoning Parser Base Class:**
```python
class ReasoningParser(ABC):
    def __init__(self, tokenizer: Any | None = None):
        self.tokenizer = tokenizer

    @abstractmethod
    def extract_reasoning(self, model_output: str) -> tuple[str | None, str | None]:
        """Returns (reasoning, content) tuple."""

    @abstractmethod
    def extract_reasoning_streaming(self, previous_text, current_text, delta_text) -> DeltaMessage | None:
        """For streaming responses."""
```

**Think Tag Stripping:**
```python
# In abstract_tool_parser.py
THINK_TAG_PATTERN = re.compile(r"<think>.*?</think>", re.DOTALL)

@staticmethod
def strip_think_tags(text: str) -> str:
    return THINK_TAG_PATTERN.sub("", text).strip()

# Used in hermes/qwen parsers before parsing tool calls
cleaned_text = self.strip_think_tags(model_output)
```

## Files Changed

- `vllm_mlx/reasoning/` - New module (5 files, 529 lines)
- `vllm_mlx/tool_parsers/abstract_tool_parser.py` - Added strip_think_tags
- `vllm_mlx/tool_parsers/hermes_tool_parser.py` - Uses strip_think_tags
- `vllm_mlx/tool_parsers/qwen_tool_parser.py` - Uses strip_think_tags
- `docs/guides/reasoning.md` - Documentation
- `tests/test_reasoning_parser.py` - 59 tests
- `tests/test_tool_parsers.py` - 4 new tests for think tag stripping

## Tests

```
============================= 120 passed in 0.91s ==============================
```

Closes #26

### DIFF
--- a/vllm_mlx/utils/mamba_cache.py
+++ b/vllm_mlx/utils/mamba_cache.py
@@ -11,7 +11,17 @@ import logging
 from typing import List, Optional
 
 import mlx.core as mx
-from mlx_lm.models.cache import MambaCache
+
+# MambaCache was removed in mlx-lm 0.30.6 - make import conditional
+try:
+    from mlx_lm.models.cache import MambaCache
+
+    HAS_MAMBA_CACHE = True
+except ImportError:
+    # Fallback for mlx-lm >= 0.30.6 where MambaCache was removed
+    from mlx_lm.models.cache import ArraysCache as MambaCache
+
+    HAS_MAMBA_CACHE = False
 
 logger = logging.getLogger(__name__)
 
@@ -102,17 +112,27 @@ def patch_mlx_lm_for_mamba():
         ArraysCache,
         RotatingKVCache,
         CacheList,
-        MambaCache as OrigMambaCache,
     )
+
+    # MambaCache was removed in mlx-lm 0.30.6
+    try:
+        from mlx_lm.models.cache import MambaCache as OrigMambaCache
+    except ImportError:
+        OrigMambaCache = ArraysCache  # Fallback
     from mlx_lm.generate import BatchKVCache, BatchRotatingKVCache
 
     # Store original function
     _original_make_cache = gen_module._make_cache
 
-    def _patched_make_cache(model, left_padding):
+    def _patched_make_cache(model, left_padding, max_kv_size=None):
         """
         Convert a list of regular caches into their corresponding
         batch-aware caches, with support for MambaCache.
+
+        Args:
+            model: The model to create cache for
+            left_padding: Left padding for batch
+            max_kv_size: Maximum KV cache size (mlx-lm 0.30.6+)
         """
 
         def to_batch_cache(c):
@@ -138,6 +158,11 @@ def patch_mlx_lm_for_mamba():
         if hasattr(model, "make_cache"):
             cache = model.make_cache()
             return [to_batch_cache(c) for c in cache]
+        elif max_kv_size is not None:
+            # mlx-lm 0.30.6+: Use rotating cache with max_kv_size
+            return [
+                BatchRotatingKVCache(max_kv_size, left_padding) for _ in model.layers
+            ]
         else:
             return [BatchKVCache(left_padding) for _ in model.layers]
 


TomLucidor:
Could you test this with OpenCode and see if the thoughts are hidden? If testing small models are a necessity please use 4B models like https://huggingface.co/mlx-community/Jan-v3-4B-base-instruct-4bit or https://huggingface.co/nightmedia/Qwen3-4B-Engineer14-qx86-hi-mlx

TomLucidor:
@waybarrios please add the parameters to `cli.py` cus it is still missing from the branch.

waybarrios:
Took a detailed look at the PR. Overall the reasoning parser implementation looks solid and clean architecture with the abstract base class, good test coverage (120 tests!), and nice handling of the implicit think mode for agents like OpenCode. The `strip_think_tags` integration with the existing tool parsers is a thoughtful addition too.

Found 1 thing worth fixing before merge:

**CLI help text references the wrong field name**

In `cli.py`, the `--reasoning-parser` help text says tags are extracted into the `reasoning_content` field, but the actual primary field in the API model is `reasoning`. The `reasoning_content` is just a computed alias for backwards compatibility. The docs in `docs/guides/reasoning.md` already use the correct field name, so this is just the help text being out of sync.

https://github.com/waybarrios/vllm-mlx/blob/9c6b28ba5ad1273ad6a69fa76e024d93122087e8/vllm_mlx/cli.py#L558-L563

Quick fix -- change `"reasoning_content field"` to `"reasoning field"` here:

https://github.com/waybarrios/vllm-mlx/blob/9c6b28ba5ad1273ad6a69fa76e024d93122087e8/vllm_mlx/api/models.py#L183-L191


A couple of things I noticed that aren't blockers but worth keeping in mind:

- **Streaming vs non-streaming behavior with non-thinking models**: The non-streaming `extract_reasoning` (line 87 of think_parser.py) returns content when no tags are found, but the streaming version (line 140) treats everything as reasoning until `</think>` appears. The comments explain this is intentional for implicit think mode, which makes sense -- just something to document clearly so users know `--reasoning-parser` should only be used with thinking models.

- **Nested JSON in hermes fallback patterns**: The `TOOL_CALL_LENIENT_PATTERN` and `RAW_JSON_TOOL_PATTERN` use `[^}]*` which won't match nested objects. Since these are fallback patterns that only kick in when primary parsing fails, it's low impact, but could be improved later with balanced brace matching.

Nice work on this feature! it addresses #26 well and follows the existing codebase patterns nicely.

PR #41, opened 2026-02-04
Add --default-temperature and --default-top-p CLI arguments

janhilgard:
## Summary

- Add `--default-temperature` CLI argument to set server-wide default temperature
- Add `--default-top-p` CLI argument to set server-wide default top_p
- When request doesn't specify these values, server uses the configured defaults

## Use Case

Some clients (e.g., Cline) don't support setting generation parameters per-request. This feature allows server operators to configure optimal defaults for their model.

For example, GLM-4.7-Flash recommends:
- `temperature=0.7` for tool-calling
- `top_p=1.0`

```bash
vllm-mlx serve model --default-temperature 0.7 --default-top-p 1.0
```

## Test plan

- [x] Server starts with new arguments
- [x] Requests without temperature/top_p use server defaults
- [x] Requests with explicit temperature/top_p override defaults
- [x] Backward compatible (defaults to None = engine defaults)

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ vllm-mlx brings native Apple Silicon GPU acceleration to vLLM by integrating:
 - **[mlx-lm](https://github.com/ml-explore/mlx-lm)**: Optimized LLM inference with KV cache and quantization
 - **[mlx-vlm](https://github.com/Blaizzy/mlx-vlm)**: Vision-language models for multimodal inference
 - **[mlx-audio](https://github.com/Blaizzy/mlx-audio)**: Speech-to-Text and Text-to-Speech with native voices
+- **[mlx-embeddings](https://github.com/Blaizzy/mlx-embeddings)**: Text embeddings for semantic search and RAG
 
 ## Features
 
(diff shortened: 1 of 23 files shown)


waybarrios:
The feature makes a lot of sense, especially for clients like Cline that don't let you configure generation parameters per request. Found one issue with the implementation though.

The Pydantic models in `api/models.py` define temperature and top_p as non-optional fields with defaults:

```python
temperature: float = 0.7
top_p: float = 0.9
```

https://github.com/waybarrios/vllm-mlx/blob/b980ecd6645ce135d21b81c646cad1bcace636a4/vllm_mlx/api/models.py#L158-L161

When a client sends a request without these fields, Pydantic automatically fills them with 0.7 and 0.9. So in server.py, the check `request.temperature is not None` is always True, and the CLI defaults from `--default-temperature` and `--default-top-p` never get used.

https://github.com/waybarrios/vllm-mlx/blob/b980ecd6645ce135d21b81c646cad1bcace636a4/vllm_mlx/server.py#L740-L752

To fix this, the Pydantic fields need to be optional:

```python
temperature: float | None = None
top_p: float | None = None
```

That way, when a client omits the value it stays as None, and the server can apply the CLI default. When the CLI default is also not set, you'd fall back to a sensible value (0.7/0.9 or whatever the engine uses).

Everything else in the PR looks good and follows the existing codebase patterns.

PR #42, opened 2026-02-04
Add GLM-4.7 tool call parser (glm47)

janhilgard:
## Summary

Adds tool call parser for GLM-4.7 and GLM-4.7-Flash models.

## GLM-4.7 Tool Call Format

```xml
<tool_call>function_name
<arg_key>param1</arg_key><arg_value>value1</arg_value>
<arg_key>param2</arg_key><arg_value>value2</arg_value>
</tool_call>
```

## Features

- Parses GLM-4.7 XML-style tool calls
- Removes `<think>...</think>` reasoning tags from output
- Supports streaming tool call detection
- Registered as `glm47` and `glm4` parser aliases

## Usage

```bash
vllm-mlx serve lmstudio-community/GLM-4.7-Flash-MLX-8bit \
    --enable-auto-tool-choice \
    --tool-call-parser glm47
```

## Test plan

- [x] Parser correctly extracts function name and arguments
- [x] Tool calls returned with proper OpenAI format
- [x] `finish_reason: "tool_calls"` set correctly
- [x] Tested with GLM-4.7-Flash-MLX-8bit model
- [x] Tested with Cline IDE integration

## References

- Based on vLLM's [glm47_moe_tool_parser.py](https://github.com/vllm-project/vllm/pull/30876)
- [GLM-4.7 Documentation](https://docs.z.ai/guides/llm/glm-4.7)


waybarrios:
### Code review

Found 6 issues:

1. Dead code: `--default-temperature` and `--default-top-p` silently do nothing on main. The PR adds these CLI arguments in cli.py and sets `server._default_temperature` / `server._default_top_p`, but main's server.py has no `_default_temperature` variable, no `_resolve_temperature()` function, and no code that reads these values. If merged, the flags are accepted by the CLI but silently ignored. This code belongs to PR #41 and should be removed from this PR.

https://github.com/waybarrios/vllm-mlx/blob/325a438ec0943657581ce71129d68c464da80e2b/vllm_mlx/cli.py#L55-L61

https://github.com/waybarrios/vllm-mlx/blob/325a438ec0943657581ce71129d68c464da80e2b/vllm_mlx/cli.py#L523-L536

2. Does not use the inherited `strip_think_tags()` method, which means implicit `</think>` tags are not handled. The parser defines its own `THINK_PATTERN` and calls `self.THINK_PATTERN.sub()` directly instead of using `self.strip_think_tags()` from the base `ToolParser` class. The base method handles two cases: full `<think>...</think>` tags and implicit closing `...</think>` when `<think>` was injected in the prompt. The local regex only handles the first case. Other parsers like Hermes use the inherited method.

https://github.com/waybarrios/vllm-mlx/blob/325a438ec0943657581ce71129d68c464da80e2b/vllm_mlx/tool_parsers/glm47_tool_parser.py#L60-L62

3. PEP 8 violation: `import ast` is inside the `_deserialize()` method body. PEP 8 requires imports at the top of the file, not inside functions.

https://github.com/waybarrios/vllm-mlx/blob/325a438ec0943657581ce71129d68c464da80e2b/vllm_mlx/tool_parsers/glm47_tool_parser.py#L71-L76

4. `ast.literal_eval` is unnecessary and inconsistent with every other parser in the codebase. All other parsers use `json.loads()` exclusively. `ast.literal_eval` can parse Python-specific types like tuples or sets that are not valid tool call arguments. A simple `json.loads` with a string fallback is sufficient and consistent.

https://github.com/waybarrios/vllm-mlx/blob/325a438ec0943657581ce71129d68c464da80e2b/vllm_mlx/tool_parsers/glm47_tool_parser.py#L65-L78

5. No tool name validation against available tools. The Hermes parser validates that parsed tool names exist in the request's tool definitions before accepting them. The GLM47 parser accepts any function name the model outputs without checking the `request` parameter, which means hallucinated tool calls pass through.

https://github.com/waybarrios/vllm-mlx/blob/325a438ec0943657581ce71129d68c464da80e2b/vllm_mlx/tool_parsers/glm47_tool_parser.py#L84-L120

6. Streaming can leak reasoning text as regular content. In `extract_tool_calls_streaming`, before `<tool_call>` appears in `current_text`, think-stripped delta text is returned as content. But the non-streaming path deliberately sets `content=None` when tool calls are found. In streaming mode, any reasoning text before the `<tool_call>` tag gets sent as content to the client and cannot be retracted.

https://github.com/waybarrios/vllm-mlx/blob/325a438ec0943657581ce71129d68c464da80e2b/vllm_mlx/tool_parsers/glm47_tool_parser.py#L138-L152

PR #43, opened 2026-02-04
Add streaming tool call parsing support

janhilgard:
## Summary
- Implement streaming tool call detection in `stream_chat_completion()`
- Use tool parser's `extract_tool_calls_streaming()` method when enabled
- Buffer content during tool_call generation, emit tool_calls chunk on completion
- Add `--enable-auto-tool-choice` and `--tool-call-parser` CLI args to server.py
- Add `reasoning` field to `ChatCompletionChunkDelta` for streaming reasoning content

## Problem
When using `stream: true`, the server was sending raw `<tool_call>` tags as content instead of proper OpenAI-compatible `tool_calls` chunks. This broke Cline and other streaming clients expecting structured tool call responses.

## Solution
The streaming function now:
1. Detects tool call patterns using the configured parser's `extract_tool_calls_streaming()` method
2. Buffers content when inside a tool_call block (returns `None` during buffering)
3. Emits proper `tool_calls` chunk when the tool call is complete
4. Sets `finish_reason: "tool_calls"` appropriately

## Test plan
- [x] Test streaming with tools - tool_calls chunk emitted correctly
- [x] Test non-streaming with tools - still works
- [x] Test with GLM-4.7-Flash model and glm47 parser

### DIFF
--- a/vllm_mlx/api/models.py
+++ b/vllm_mlx/api/models.py
@@ -376,7 +376,9 @@ class ChatCompletionChunkDelta(BaseModel):
     role: Optional[str] = None
     content: Optional[str] = None
     tool_calls: Optional[List[dict]] = None
-    reasoning: Optional[str] = None  # For reasoning/thinking content (Qwen3, DeepSeek-R1)
+    reasoning: Optional[str] = (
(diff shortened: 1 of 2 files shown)


enryold:
UP

PR #55, opened 2026-02-10
Fix streaming tool call parsing for Qwen3-Coder models

### LINKED ISSUE #47: Issue with mlx-community/Qwen3-Coder-Next 4bit tool calling
Hi, is there an issue with the new coder next model tool calling. I'm trying to understand if there's any difference in performance between your vllm server and lmstudio serving the same exact model. There's surely something going on while tool calling inside the Zed editor...it prints the tool calling template (<tool_call>...</tool_call>) to the chat instead of actually executing the tool call. Is that something related to the editor or the vllm-mlx server?

waybarrios:
Fixes #47

The streaming tool call parser had several bugs that caused raw XML tags to leak as plain text content instead of being parsed into structured tool_calls. This was most visible with Qwen3-Coder models using Nemotron-style XML format.

Problems found:

1. The closing tag detection only checked delta_text, not the accumulated text. If </tool_call> was split across two stream chunks it was never detected.

2. After the first tool call was parsed, subsequent tool calls in the same response leaked as content. The parser used a presence check ("</tool_call>" not in previous_text) which fails for the second tool call since previous_text already contains </tool_call> from the first one.

3. The simple engine always passed enable_thinking: True to the chat template, causing <think> tags to leak as raw text for coder models.

4. There was no qwen3_coder parser alias. The official vLLM uses --tool-call-parser qwen3_coder for these models.

Changes:

vllm_mlx/tool_parsers/hermes_tool_parser.py:
- Rewrote extract_tool_calls_streaming() to use tag counting instead of presence checks. open_count vs close_count detects incomplete blocks, close_count vs prev_close_count detects newly completed tool calls, and result.tool_calls[prev_close_count:] emits only new calls without duplicates.
- Added bare <function=...> block detection for models where the chat template provides <tool_call> as generation prompt and the model generates <function=name> directly.
- Added BARE_FUNCTION_PATTERN regex and corresponding fallback in extract_tool_calls().
- Registered qwen3_coder as parser alias.
- Extracted _format_streaming_tool_calls() helper to reduce duplication.

vllm_mlx/engine/simple.py:
- Set enable_thinking = "coder" not in self._model_name.lower() to match the batched engine behavior.

vllm_mlx/cli.py:
- Added qwen3_coder to --tool-call-parser choices.

tests/test_tool_parsers.py:
- Added 17 new tests covering: qwen3_coder alias, Nemotron XML parsing, bare function blocks, multi-tool streaming with no duplicates, split closing tags, content after tool calls, and false positive prevention.

Tested locally with mlx-community/Qwen3-Coder-30B-A3B-Instruct-4bit:

```
curl -s http://localhost:8000/v1/chat/completions \
  -d '{
    "model": "mlx-community/Qwen3-Coder-30B-A3B-Instruct-4bit",
    "messages": [{"role": "user", "content": "Read /src/main.py and list files in /src"}],
    "tools": [
      {"type": "function", "function": {"name": "read_file", ...}},
      {"type": "function", "function": {"name": "list_files", ...}}
    ],
    "stream": true
  }'
```

Before fix: raw XML tags like <function=read_file>, <parameter=path> leaked as content chunks.
After fix: both tool calls returned as structured tool_calls with correct indexes (0 and 1).

All 834 tests pass (78 tool parser tests, 1 pre-existing unrelated failure in test_memory_stability.py).

### DIFF
--- a/tests/test_optimizations.py
+++ b/tests/test_optimizations.py
@@ -6,6 +6,8 @@ Usage:
     pytest tests/test_optimizations.py -v
 """
 
+import pytest
+
 
 class TestHardwareDetection:
     """Tests for hardware detection functionality."""


karmawastakenalready:
Thanks for this fix! I can confirm this resolves #47  for me.

I tested with the following command and tool calls are now working with Qwen3 Coder in Cline and Qwen Code:

```bash
vllm-mlx serve toby1991/Qwen3-Coder-Next-REAP-48B-A3B-4bit-mlx \ 
--enable-auto-tool-choice \ 
--tool-call-parser qwen3_coder \ 
--continuous-batching \ 
--use-paged-cache \ 
--paged-cache-block-size 64 \ 
--max-cache-blocks 1000 \ 
--cache-memory-percent 0.22 \ 
--prefill-batch-size 8 \ 
--completion-batch-size 32 \ 
--max-tokens 16384 \ 
--stream-interval 1 \
--host 127.0.0.1 \ 
--port 8000
```

I'd been experiencing the same issue over the past few days, and this PR completely fixes it. Tool calls work without any issues now.

PR #62, opened 2026-02-11
Add KV cache quantization for prefix cache memory reduction

### LINKED ISSUE #60: Add KV cache quantization support (--kv-cache-quantization)
## Summary

Add support for quantized KV caches using `mlx-lm`'s `QuantizedKVCache` to reduce memory usage during inference, especially for large models with `--continuous-batching`.

Requested in https://github.com/waybarrios/vllm-mlx/discussions/17

## Motivation

KV caches are stored in fp16 by default, which for large models (e.g., 30B parameters) can consume tens of GB of RAM during continuous batching. `mlx-lm` already ships a `QuantizedKVCache` class that stores KV pairs in 8-bit (or lower), roughly halving memory usage.

**Example**: A 30B 6-bit model on M3 Max 128GB — each KV cache entry for a 1000-token sequence can be hundreds of MB in fp16. With 8-bit quantized KV cache, this drops by ~50%.

## Proposed API

```bash
# Enable 8-bit KV cache quantization (default group_size=64)
vllm-mlx serve <model> --continuous-batching --kv-cache-quantization

# With custom bits and group size
vllm-mlx serve <model> --continuous-batching --kv-cache-quantization --kv-cache-quantization-bits 4 --kv-cache-quantization-group-size 32
```

### 1. Add CLI flags

In `vllm_mlx/cli.py` and `vllm_mlx/server.py`, add arguments:

```python
parser.add_argument(
    "--kv-cache-quantization",
    action="store_true",
    help="Use quantized KV cache to reduce memory (8-bit by default)",
)
parser.add_argument(
    "--kv-cache-quantization-bits",
    type=int,
    default=8,
    choices=[4, 8],
    help="Bits for KV cache quantization (default: 8)",
)
parser.add_argument(
    "--kv-cache-quantization-group-size",
    type=int,
    default=64,
    help="Group size for KV cache quantization (default: 64)",
)
```

### 2. Pass config through to model loading

The `QuantizedKVCache` from mlx-lm can be used as a drop-in replacement for `KVCache`. It's initialized per-layer with the same interface:

```python
from mlx_lm.models.cache import KVCache, QuantizedKVCache

# Current (fp16):
cache = [KVCache() for _ in range(num_layers)]

# Quantized:
cache = [QuantizedKVCache(group_size=64, bits=8) for _ in range(num_layers)]
```

### 3. Integration points

The key files that create or manage KV caches:

- **`vllm_mlx/scheduler.py`** — `BatchGeneratorScheduler` creates `BatchGenerator` which manages the cache internally via mlx-lm. The `mlx_lm.utils.stream_generate` and `mlx_lm.utils.generate_step` functions accept a `kv_cache` parameter.

- **`vllm_mlx/engine/batched.py`** — `BatchedEngine` creates the scheduler; needs to pass quantization config.

- **`vllm_mlx/memory_cache.py`** — `MemoryAwarePrefixCache` stores KV cache entries. The `estimate_kv_cache_memory()` function needs to handle `QuantizedKVCache` objects (which will report smaller `.nbytes`).

- **`vllm_mlx/prefix_cache.py`** — Legacy prefix cache, same consideration.

### 4. Key code change in scheduler

In `vllm_mlx/scheduler.py`, where `BatchGenerator` or generate steps are invoked, pass the quantized cache:

```python
# In SchedulerConfig or similar
@dataclass
class KVCacheConfig:
    quantize: bool = False
    bits: int = 8
    group_size: int = 64

# When creating cache for a new request
def _make_kv_cache(self, num_layers: int) -> list:
    if self.kv_cache_config.quantize:
        from mlx_lm.models.cache import QuantizedKVCache
        return [
            QuantizedKVCache(
                group_size=self.kv_cache_config.group_size,
                bits=self.kv_cache_config.bits,
            )
            for _ in range(num_layers)
        ]
    else:
        from mlx_lm.models.cache import KVCache
        return [KVCache() for _ in range(num_layers)]
```

### 5. Memory estimation update

In `vllm_mlx/memory_cache.py`, update `estimate_kv_cache_memory()` to handle quantized caches:

```python
def estimate_kv_cache_memory(cache: list[Any]) -> int:
    total_bytes = 0
    for layer_cache in cache:
        if hasattr(layer_cache, 'state'):
            keys, values = layer_cache.state
            total_bytes += _array_memory(keys)
            total_bytes += _array_memory(values)
            # QuantizedKVCache also has scales/biases
            if hasattr(layer_cache, 'key_scales'):
                total_bytes += _array_memory(layer_cache.key_scales)
            if hasattr(layer_cache, 'value_scales'):
                total_bytes += _array_memory(layer_cache.value_scales)
    return total_bytes
```

## Expected Impact

| Metric | fp16 KV Cache | 8-bit Quantized | Improvement |
|--------|-------------|-----------------|-------------|
| KV memory per 1K tokens (30B model) | ~400MB | ~200MB | ~2x reduction |
| Max concurrent requests (128GB RAM) | ~8-12 | ~16-24 | ~2x more |
| Quality impact | Baseline | Minimal degradation | Negligible |

## References

- `mlx_lm.models.cache.QuantizedKVCache` — [mlx-lm source](https://github.com/ml-explore/mlx-lm)
- Discussion: #17

waybarrios:
This adds KV cache quantization support for the prefix cache, so stored cache entries take up much less memory.

The new --kv-cache-quantization flag tells the system to compress KV cache entries using mlx-lm's QuantizedKVCache when storing them in the prefix cache. By default it uses 8-bit quantization, but you can also go down to 4-bit with --kv-cache-quantization-bits 4. In practice this gives around 3.5x memory savings with very little quality loss (mean absolute error around 0.005).

The approach is simple: quantize when storing, dequantize when fetching. Active inference stays on full precision since BatchKVCache doesn't have a quantized variant, so there's no impact on generation quality during a request.

What changed:

memory_cache.py - New _quantize_cache() and _dequantize_cache() helpers. Added kv_quantize, kv_bits, and kv_group_size to MemoryCacheConfig. Updated estimate_kv_cache_memory() and _trim_cache_offset() to handle quantized layers. All 5 fetch return points now dequantize when needed, and store() quantizes before saving.

scheduler.py - Added kv_cache_quantization fields to SchedulerConfig and wired them through to MemoryCacheConfig.

cli.py - Three new flags (--kv-cache-quantization, --kv-cache-quantization-bits, --kv-cache-quantization-group-size) for both serve and bench commands.

tests/test_kv_cache_quantization.py - 16 tests covering round-trip correctness, memory reduction, config propagation, mixed cache layers, and store/fetch integration.

Usage:

vllm-mlx serve model --continuous-batching --kv-cache-quantization
vllm-mlx serve model --continuous-batching --kv-cache-quantization --kv-cache-quantization-bits 4

Testing:
- All 16 tests pass
- Black and ruff clean
- Still needs a manual test with a real model

Closes #60

### DIFF
--- a/tests/test_memory_stability.py
+++ b/tests/test_memory_stability.py
@@ -247,9 +247,9 @@ class TestMemoryStats:
     def test_metal_stats_included(self, mock_mx):
         """Verify Metal memory stats appear in get_stats()."""
         mock_mx.metal.is_available.return_value = True
-        mock_mx.metal.get_active_memory.return_value = 10_000_000_000  # 10GB
-        mock_mx.metal.get_peak_memory.return_value = 15_000_000_000  # 15GB
-        mock_mx.metal.get_cache_memory.return_value = 2_000_000_000  # 2GB
+        mock_mx.get_active_memory.return_value = 10_000_000_000  # 10GB
+        mock_mx.get_peak_memory.return_value = 15_000_000_000  # 15GB
+        mock_mx.get_cache_memory.return_value = 2_000_000_000  # 2GB
 
         scheduler = _make_scheduler()
         stats = scheduler.get_stats()


waybarrios:
Tested with long context windows, seq_len=32k and seq_len=64k. This simulates what happens when you cache long conversations or large documents.

seq_len=32768 (32 layers, 32 heads, head_dim=128):

| Mode | Memory | Compression | Mean Error | Max Error | Quantize | Dequantize |
|------|--------|-------------|------------|-----------|----------|------------|
| FP16 | 32768.00 MB | 1.00x | 0.000 | 0.000 | - | - |
| 8-bit | 9216.00 MB | 3.56x | 0.00456 | 0.03281 | 80.0 ms | 300.5 ms |
| 4-bit | 5120.00 MB | 6.40x | 0.07711 | 0.55431 | 109.8 ms | 49.8 ms |

seq_len=65536 (32 layers, 32 heads, head_dim=128):

| Mode | Memory | Compression | Mean Error | Max Error | Quantize | Dequantize |
|------|--------|-------------|------------|-----------|----------|------------|
| FP16 | 65536.00 MB | 1.00x | 0.000 | 0.000 | - | - |
| 8-bit | 18432.00 MB | 3.56x | 0.00456 | 0.03278 | 171.9 ms | 8429.8 ms |
| 4-bit | 10240.00 MB | 6.40x | 0.07711 | 0.56201 | 25531.3 ms | 4045.0 ms |

At 32k context the overhead is still reasonable, around 80-300ms. At 64k things get heavier, especially the 8-bit dequantize at 8.4 seconds and 4-bit quantize at 25.5 seconds. That said, the memory savings are massive: going from 64 GB down to 18 GB with 8-bit or 10 GB with 4-bit. For long context scenarios on machines with 32-64 GB of unified memory, this is what makes the difference between being able to cache those conversations or not.

To reproduce:

vllm-mlx bench-kv-cache --seq-len 32768
vllm-mlx bench-kv-cache --seq-len 65536

waybarrios:
The quantize/dequantize latency at 64k is worth calling out. 8.4 seconds for dequantize and 25 seconds for quantize is way too slow if this is in the hot path. At 32k it's still fine (80-300ms), but 64k hits a wall.

A few things I want to try to bring that down:

- Process the cache in chunks instead of all at once
- Only dequantize the layers we actually need instead of the full stack
- Use a hybrid strategy where short caches stay in FP16 and we only quantize the longer ones that are worth compressing

For now this works well up to 32k context, which covers most real use cases. The long context optimization is something I'll dig into separately.

waybarrios (commented):
I ran the full test suite locally after the latest push (23 quantization tests + 42 memory cache mock tests + full suite): **902 passed**, 1 pre-existing failure (unrelated mock issue on main).

Pushed a fix for the `ModuleNotFoundError: No module named 'mlx'` that was failing test-matrix (3.10/3.11/3.12). The root cause: `_trim_to_offset()` imported `mlx.core` unconditionally at the top of the function, but `store()` calls it on every entry — including mock objects in tests running on Linux CI without MLX.

**Fix**: Extracted `_needs_kv_trim()` — a duck-typed helper that checks if any layer has oversized KV arrays using `getattr` instead of `isinstance(layer, KVCache)`. The MLX imports now only happen when trimming is actually needed. Also fixed a ruff N806 lint error in `_trim_cache_offset`.

### Changes

1. **Core quantization** — `_quantize_cache()` / `_dequantize_cache()` with store-on-quantize, fetch-on-dequantize pattern
2. **`_trim_cache_offset` fix for QuantizedKVCache** — explicit handling with offset/group_size/bits preserved
3. **`_trim_to_offset`** — trims pre-allocated KV buffers to actual used size before storage
4. **`kv_min_quantize_tokens` threshold** — skips quantization for short sequences (default: 256)
5. **`bench-kv-cache` CLI command** — synthetic benchmarks for quantization
6. **Config validation** — `kv_min_quantize_tokens >= 0` enforced in `__post_init__`
7. **23 tests** covering round-trip, memory reduction, config, store/fetch integration, trim, and threshold

###  Improvements

- The `_dequantize_cache(x) if self._config.kv_quantize else x` pattern appears 5x in `fetch()` — could be a `_maybe_dequantize()` method
- Disk persistence (`save_to_disk`/`load_from_disk`): works with QuantizedKVCache (verified), but if `kv_quantize=False` at load time and entries were saved quantized, they'll be returned raw. Consider always calling `_dequantize_cache` on fetch regardless of config flag (it's a no-op on regular KVCache)
- No test for supersequence/LCP match with quantized cache — the `_trim_cache_offset` fix addresses this path but no test exercises it directly

Overall the feature is solid and ready to merge once CI is green.

PR #64, opened 2026-02-11
Add more aggressive nemotron XML tool call parsing, fixes #63

### LINKED ISSUE #63: Qwen3-Coder tool call outputs can't be parsed when single quotes are used
I've ocurred to this bug while trying out opencode. When using models like Qwen-Coder, tool calls often output arguments using Python literal syntax instead of strict JSON. This causes the `HermesToolParser` to fail or incorrectly return arguments as raw strings because json.loads does not support single quotes or capitalized booleans (True/False). Parser's own behaviour is actually correct but the model tends to use single quotes instead of double quotes. We can add a recovery mechanism to handle these malformed inputs.

Example script to test parser's behaviour with single quotes:
Output:
```bash

============================================================
Test: Python List Syntax
============================================================
Input:
<tool_call>
<function=search>
<parameter=query>['python', 'tutorial']</parameter>
</function>
</tool_call>

✅ Parsed tool call: search
Arguments:
  query: "['python', 'tutorial']" (str)
    ⚠️  Should be list, not string!

============================================================
Test: Python Dict Syntax
============================================================
Input:
<tool_call>
<function=config>
<parameter=settings>{'theme': 'dark', 'size': 14}</parameter>
</function>
</tool_call>

✅ Parsed tool call: config
Arguments:
  settings: "{'theme': 'dark', 'size': 14}" (str)
    ⚠️  Should be dict, not string!

============================================================
Test: Python Bool Syntax
============================================================
Input:
<tool_call>
<function=toggle>
<parameter=enabled>True</parameter>
</function>
</tool_call>

✅ Parsed tool call: toggle
Arguments:
  enabled: 'True' (str)
    ⚠️  Should be bool, not string!

============================================================
Test: JSON Syntax (control)
============================================================
Input:
<tool_call>
<function=search>
<parameter=query>["python", "tutorial"]</parameter>
</function>
</tool_call>

✅ Parsed tool call: search
Arguments:
  query: ['python', 'tutorial'] (list)
```
Code:
```python
#!/usr/bin/env python3

from vllm_mlx.tool_parsers.hermes_tool_parser import HermesToolParser
import json

def test(name, xml_input):
    print(f"\n{'='*60}")
    print(f"Test: {name}")
    print(f"{'='*60}")
    print(f"Input:\n{xml_input}\n")
    
    parser = HermesToolParser()
    result = parser.extract_tool_calls(xml_input)
    
    if result.tool_calls:
        tc = result.tool_calls[0]
        args = json.loads(tc['arguments'])
        print(f"✅ Parsed tool call: {tc['name']}")
        print(f"Arguments:")
        for key, val in args.items():
            print(f"  {key}: {repr(val)} ({type(val).__name__})")
            
            # Check if it's wrongly a string
            if isinstance(val, str) and (
                (val.startswith('[') and val.endswith(']')) or
                (val.startswith('{') and val.endswith('}')) or
                val in ['True', 'False']
            ):
                print(f"    ⚠️  Should be {val[0] == '[' and 'list' or val[0] == '{' and 'dict' or 'bool'}, not string!")
    else:
        print("❌ No tool calls found")


# Test 1: Python list (the bug)
test("Python List Syntax", """<tool_call>
<function=search>
<parameter=query>['python', 'tutorial']</parameter>
</function>
</tool_call>""")

# Test 2: Python dict (the bug)
test("Python Dict Syntax", """<tool_call>
<function=config>
<parameter=settings>{'theme': 'dark', 'size': 14}</parameter>
</function>
</tool_call>""")

# Test 3: Python bool (the bug)
test("Python Bool Syntax", """<tool_call>
<function=toggle>
<parameter=enabled>True</parameter>
</function>
</tool_call>""")

# Test 4: JSON (works fine)
test("JSON Syntax (control)", """<tool_call>
<function=search>
<parameter=query>["python", "tutorial"]</parameter>
</function>
</tool_call>""")
```

selimrecep:
Fixes #63, handles cases where literals use single quote instead of double quotes and causes failure to tool call parse

### DIFF
--- a/vllm_mlx/tool_parsers/hermes_tool_parser.py
+++ b/vllm_mlx/tool_parsers/hermes_tool_parser.py
@@ -5,12 +5,12 @@ Hermes/Nous tool call parser for vllm-mlx.
 Handles Hermes-style tool calling format used by NousResearch models.
 """
 
+import ast
 import json
 import re
 import uuid
 from collections.abc import Sequence
 from typing import Any
-import ast
 
 from .abstract_tool_parser import (
     ExtractedToolCallInformation,
@@ -22,6 +22,31 @@ from .abstract_tool_parser import (
 def generate_tool_id() -> str:
     """Generate a unique tool call ID."""
     return f"call_{uuid.uuid4().hex[:8]}"
+
+
+def _parse_param_value(val: str) -> Any:
+    """Parse a tool call parameter value, handling both JSON and Python literals.
+
+    Tries json.loads first. If that fails, falls back to ast.literal_eval
+    for Python literal syntax (single quotes, True/False, None). Converts
+    sets to lists and rejects types that are not JSON-serializable (complex,
+    bytes) to avoid crashes during json.dumps later.
+    """
+    try:
+        return json.loads(val)
+    except (json.JSONDecodeError, ValueError):
+        pass
+
+    try:
+        python_val = ast.literal_eval(val)
+        if isinstance(python_val, set):
+            python_val = sorted(python_val, key=str)
+        if isinstance(python_val, (complex, bytes)):
+            return val
+        json.dumps(python_val)
+        return python_val
+    except (ValueError, SyntaxError, TypeError):
+        return val
 
 
 @ToolParserManager.register_module(["hermes", "nous", "qwen3_coder"])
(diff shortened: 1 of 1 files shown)


waybarrios:
Pushed a commit on top of your changes to handle edge cases where ast.literal_eval parses types that json.dumps cannot serialize.

The original fallback works correctly for the cases in issue #63 (single quotes, True/False, None), but ast.literal_eval also accepts sets, complex numbers, and bytes literals. If a model hallucinated something like {'a', 'b'} (which looks like a dict without values), json.dumps would crash with a TypeError since sets are not JSON-serializable.

Extracted the repeated json.loads / ast.literal_eval logic into a _parse_param_value helper that converts sets to sorted lists and rejects complex and bytes by falling back to the raw string. Also moved the ast import to its correct position in the stdlib block.

PR #81, opened 2026-02-13
Add option to enable forcing server to run in MLLM mode

### LINKED ISSUE #80: Allow enforcing MLLM loading when launching server
Currently the package uses `MLLM_PATTERNS` constant string literals to detect if a model is multimodal or not, which seems to cover wide range of models very well. But there are cases where some model publishers don't add hinting keywords in the name (e.g. `mlx-community/bu-30b-a3b-preview-8bit` , a Qwen3 VL based finetuned model). Server CLI should allow users to configure server if it is an MLLM model.

selimrecep:
Should be straightforward as `load_model` had `force_mllm` argument
Fixes #80

### DIFF
--- a/scripts/add_mtp_weights.py
+++ b/scripts/add_mtp_weights.py
@@ -1,0 +1,341 @@
+#!/usr/bin/env python3
+"""
+Add MTP (Multi-Token Prediction) weights to an existing MLX Qwen3-Next model.
+
+This script:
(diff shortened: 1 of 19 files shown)


waybarrios:
LGTM! Ready to merge.

waybarrios:
just updated the PR branch

PR #93, opened 2026-02-16
Fix streaming tool calls when reasoning parser is active

janhilgard:
## Summary

- Fix streaming tool calls being emitted as raw text when `--reasoning-parser` is enabled
- The reasoning parser branch in `stream_chat_completion()` was skipping tool call parsing entirely, causing `<tool_call>` markup to leak into `content` instead of being parsed into structured `tool_calls` objects
- Adds tool call parsing to the reasoning branch, mirroring the existing logic in the standard path

## Problem

When both `--reasoning-parser deepseek_r1` and `--tool-call-parser hermes` are enabled, streaming responses emit raw tool call markup as plain text:

```json
{"delta": {"content": "<tool_call>\n<function=get_weather>\n<parameter=location>\nPrague\n</parameter>\n</function>\n</tool_call>", "tool_calls": null}, "finish_reason": "stop"}
```

## Fix

After this change, tool calls are correctly parsed and emitted:

```json
{"delta": {"tool_calls": [{"index": 0, "id": "call_4b99a63b", "type": "function", "function": {"name": "get_weather", "arguments": "{\"location\": \"Prague\"}"}}]}, "finish_reason": "tool_calls"}
```

## Test plan

- [x] Streaming tool call with `--reasoning-parser deepseek_r1 --tool-call-parser hermes` returns structured `tool_calls`
- [x] `finish_reason` is `"tool_calls"` (not `"stop"`)
- [x] Reasoning-only responses (no tool calls) still work correctly
- [x] Non-streaming tool calls unaffected
- [x] `black` formatting passes

### DIFF
--- a/vllm_mlx/server.py
+++ b/vllm_mlx/server.py
@@ -1928,16 +1928,82 @@ async def stream_chat_completion(
                 # Skip this chunk (e.g., <think> token itself)
                 continue
 
+            content = delta_msg.content
+            reasoning = delta_msg.reasoning
+
+            # Tool call parsing on content portion
+            if tool_parser and content:
+                if not tool_markup_possible and "<" not in content:
+                    tool_accumulated_text += content
+                    # No tool markup yet, fall through to normal emission
+                else:
+                    if not tool_markup_possible:
+                        tool_markup_possible = True
+                    tool_previous = tool_accumulated_text
+                    tool_accumulated_text += content
+                    tool_result = tool_parser.extract_tool_calls_streaming(
+                        tool_previous, tool_accumulated_text, content
+                    )
+
+                    if tool_result is None:
+                        # Inside tool markup - suppress content output
+                        if reasoning:
+                            # Still emit reasoning while buffering tool call
+                            chunk = ChatCompletionChunk(
+                                id=response_id,
+                                model=request.model,
+                                choices=[
+                                    ChatCompletionChunkChoice(
+                                        delta=ChatCompletionChunkDelta(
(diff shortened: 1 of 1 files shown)


Thump604:
@waybarrios, @janhilgard: status note plus coordination.

This PR fixes the case where the reasoning parser branch in `stream_chat_completion()` skips tool call parsing entirely, causing `<tool_call>` markup to leak into `content` when both `--reasoning-parser` and `--tool-call-parser` are enabled. Mergeable on current main.

Coordination: at least two other PRs target the same problem area:

- PR #177 (Thump604) "fix: parse tool calls in streaming reasoning branch" — runs reasoning extraction first then passes cleaned text to the tool parser. Currently MERGEABLE.
- PR #148 (jsirish) "fix: integrate tool call parsing with reasoning parser in streaming mode" — also targets the same code path.

All three address the same root cause with slightly different approaches. They are also collectively the fix for issues #107 (Dubworx) and #161 (Thump604, "Reasoning content leaks into content field when tool calls are present").

Last activity Feb 16. This PR is the oldest of the three. Worth a coordination decision: which approach to land, then close or rebase the others.

janhilgard:
This PR is now superseded by #253 ("integrate tool call parser into reasoning parser streaming path") which was merged into `main`. The fix covers the same issue — tool call parsing was missing in the reasoning parser streaming branch.

The current `main` has tool call parsing in both reasoning and non-reasoning streaming paths (4 call sites for `extract_tool_calls_streaming`).

This PR can be closed.

PR #95, opened 2026-02-16
fix: Coerce tool call arguments to match schema types

janhilgard:
## Summary

- Add `_coerce_tool_arguments()` that auto-stringifies object/array values when the tool schema expects a string
- Fixes a common LLM failure mode where models output raw JSON objects instead of JSON strings (e.g. when writing `package.json`, the model puts `content` as an object instead of a string)
- Applied to all tool call emission paths (non-streaming, streaming, streaming fallback)

## Problem

When using tool calling with coding assistants (e.g. OpenCoder), models sometimes generate:
```json
{"file_path": "package.json", "content": {"name": "app", "version": "1.0.0"}}
```
instead of:
```json
{"file_path": "package.json", "content": "{\"name\": \"app\", \"version\": \"1.0.0\"}"}
```

The client validates against the tool schema (`content: string`) and fails with:
`expected string, received object`

## Fix

Server-side coercion: after parsing tool call arguments, check each field against the tool schema. If schema says `type: "string"` but the value is an object/array, `json.dumps()` it automatically.

## Test plan

- [ ] Verify tool calls with string content pass through unchanged
- [ ] Verify tool calls with object content are auto-stringified
- [ ] Verify tool calls with array content are auto-stringified
- [ ] Verify no-op when tools schema is not provided
- [ ] Verify streaming and non-streaming paths both coerce

### DIFF
--- a/vllm_mlx/server.py
+++ b/vllm_mlx/server.py
@@ -351,6 +351,53 @@ def _validate_model_name(request_model: str) -> None:
         )
 
 
+def _coerce_tool_arguments(
+    arguments_json: str, tool_name: str, tools: list[dict] | None
+) -> str:
+    """
+    Coerce tool call arguments to match the tool schema.
+
+    If a schema field expects "string" but the model produced an object/array,
+    JSON-stringify the value. This fixes a common LLM failure mode where models
+    output raw JSON objects instead of JSON strings for file content, etc.
+    """
+    if not tools:
+        return arguments_json
+
+    # Find the schema for this tool
+    schema = None
+    for tool in tools:
+        if isinstance(tool, dict) and tool.get("function", {}).get("name") == tool_name:
+            schema = tool["function"].get("parameters", {})
+            break
+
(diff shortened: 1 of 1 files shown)


Thump604:
@waybarrios, @janhilgard: brief endorsement.

`_coerce_tool_arguments()` that auto-stringifies object/array values when the tool schema expects a string is the right fix for the failure mode where models generate `{"content": {"name": "app"}}` instead of `{"content": "{\\"name\\":\\"app\\"}"}`. Real production failure mode for coding assistants writing JSON config files. Applied to all tool call emission paths (non-streaming, streaming, streaming fallback). Mergeable on current main.

Last activity Mar 21, ~3 weeks ago.

janhilgard:
@Thump604 This is already in `main` via #278 — `_coerce_tool_arguments()` is defined and used in all streaming/non-streaming/fallback paths (4 call sites). Closing as superseded.

PR #105, opened 2026-02-23
feat: add --prefill-step-size CLI flag

kol22:
## Summary

`MLLMSchedulerConfig.prefill_step_size` defaults to 1024 but isn't exposed as a
CLI argument. The MLLM batch generator enforces
`total_prompt_tokens <= prefill_step_size * batch_count`, so any single vision
request exceeding 1024 tokens fails with:

```
ValueError: Total prompt tokens (2289) exceeds safe limit (1024)
```

Since vision tokens alone typically exceed 1024 (images contribute ~1400+
tokens), this effectively blocks all MLLM inference under the default config.

## Fix

Add `--prefill-step-size` to both `serve` and `bench` commands. Default of `0`
means "use engine default" — 2048 for LLM, 1024 for MLLM — preserving existing
behavior. When set, the value flows through `SchedulerConfig` to both the LLM
scheduler and the MLLM scheduler.

Example usage:
```bash
vllm-mlx serve model-name --continuous-batching --prefill-step-size 16384
```

## Test

Tested with Qwen3-VL-32B-Instruct-8bit. Before this fix, vision requests fail
with the ValueError above. After passing `--prefill-step-size 16384`, they
complete successfully.

### DIFF
--- a/vllm_mlx/cli.py
+++ b/vllm_mlx/cli.py
@@ -172,11 +172,10 @@ def serve_command(args):
             kv_cache_quantization_bits=args.kv_cache_quantization_bits,
             kv_cache_quantization_group_size=args.kv_cache_quantization_group_size,
             kv_cache_min_quantize_tokens=args.kv_cache_min_quantize_tokens,
+            mllm_prefill_step_size=(
+                args.prefill_step_size if args.prefill_step_size > 0 else None
+            ),
         )
-
-        # Override prefill_step_size if explicitly set (0 = use engine default)
-        if args.prefill_step_size > 0:
-            scheduler_config.prefill_step_size = args.prefill_step_size
 
         print("Mode: Continuous batching (for multiple concurrent users)")
         if args.chunked_prefill_tokens > 0:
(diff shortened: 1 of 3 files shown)


waybarrios:
Would you mind addressing point 1 (removing the flag from `bench_parser` or wiring it to the MLLM bench path) and point 2 (renaming to `--mllm-prefill-step-size`) before merging? Those two would avoid confusion for users. Point 3 is minor and can be addressed later if you prefer.

janhilgard:
@waybarrios — all three review items were addressed back on Feb 25 (removed from bench, renamed to `--mllm-prefill-step-size`, added `> 0` validation). This has been waiting for a follow-up review for ~7 weeks now.

It does have merge conflicts from recent main changes — @kol22 would you mind rebasing? After that it should be ready to merge.

kol22:
@janhilgard - resolved & rebased. Should be good to go.

janhilgard (approved):
LGTM — all three review items from @waybarrios have been addressed:

1. **Removed from `bench_parser`** — the flag is now `serve`-only, matching the actual MLLM scheduler path.
2. **Renamed to `--mllm-prefill-step-size`** — clear scope, matches the internal field name.
3. **`__post_init__` validation** — `mllm_prefill_step_size` must be `> 0` when provided, preventing the `0` propagation edge case.

Rebase on current `main` is clean. Code is minimal and consistent with existing patterns. Ready to merge.

PR #118, opened 2026-02-26
feat(server): implement sequential reasoning and tool-call parsing bridge

jonharris0n:
Refactor the streaming loop in server.py to handle reasoning and tool-calling as a linear pipeline. Previously, parallel parser logic caused buffer desynchronization and stream termination for models like DeepSeek-R1 and Qwen 2.5/3.5.

- Implemented Sequential Reasoning Bridge: Intercepts and yields reasoning tokens first, then passes cleaned delta_text to the tool parser.
- Fixed Buffer Race Condition: Synchronized accumulated_text updates to prevent 'thought leakage' into tool-call XML blocks.
- Optimized Usage Tracking: Moved token count updates to the start of the loop to ensure accuracy during reasoning-only chunks.
- Restored original documentation and 'Fast path' parsing logic.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -95,3 +95,4 @@ temp/
 
 # MLX cache
 .mlx_cache/
+reasoning_bridge.diff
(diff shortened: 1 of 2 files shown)


Thump604:
Closing as stale — this has been open 7 weeks with merge conflicts and no review activity. If the work is still relevant, feel free to reopen with a rebase onto current main.

PR #124, opened 2026-02-28
fix: forward tool definitions through MLLM code path

Thump604:
## Summary

The MLLM (multimodal) code path silently drops tool definitions, causing tool calling to be completely non-functional for models that require `--mllm` (e.g., Qwen 3.5 122B-A10B).

**Root cause:** Four locations in the MLLM path did not forward `tools` to the chat template, while the LLM (text-only) path handled them correctly.

### Changes

**`vllm_mlx/engine/simple.py`** (2 fixes):
- `chat()`: Pass `tools=template_tools` to `self._model.chat()` in the MLLM branch (matching the LLM branch on line 297)
- `stream_chat()`: Same fix for `self._model.stream_chat()` 

**`vllm_mlx/models/mllm.py`** (2 fixes):
- `chat()`: Forward `tools` from `kwargs` to `get_chat_template()` via `**template_extra_kwargs`
- `stream_chat()`: Same fix

### Evidence

Before fix: `prompt_tokens: 18` (tools not in prompt), `tool_calls: null`
After fix: `prompt_tokens: 283` (tools injected), `tool_calls: [{"function": {"name": "get_weather", "arguments": ...}}]`

### Reproduction

```bash
vllm-mlx serve <model> --mllm --tool-call-parser hermes --enable-auto-tool-choice

# This returns tool_calls: null before fix, correct tool call after
curl -X POST http://127.0.0.1:8080/v1/chat/completions \
  -d '{"model":"...","messages":[{"role":"user","content":"What is the weather?"}],
       "tools":[{"type":"function","function":{"name":"get_weather","parameters":{"type":"object","properties":{"location":{"type":"string"}},"required":["location"]}}}],
       "tool_choice":"required"}'
```

Tested with Qwen 3.5 122B-A10B (5-bit MLX) on Mac Studio M2 Ultra.

### DIFF
--- a/vllm_mlx/engine/simple.py
+++ b/vllm_mlx/engine/simple.py
@@ -276,6 +276,7 @@ class SimpleEngine(BaseEngine):
                     messages=messages,
                     max_tokens=max_tokens,
                     temperature=temperature,
+                    tools=template_tools,
                     **kwargs,
                 )
                 text = clean_output_text(output.text)
@@ -351,6 +352,7 @@ class SimpleEngine(BaseEngine):
                         messages=messages,
                         max_tokens=max_tokens,
                         temperature=temperature,
+                        tools=template_tools,
                         **kwargs,
                     )
                 )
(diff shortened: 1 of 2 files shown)


waybarrios (approved):
looks good, tool forwarding fix is solid and all checks pass

Issue #129, opened 2026-03-01
Anthropic Streaming `/v1/messages` Leaks `<think>` and `<tool_call>` Markup Before Structured `tool_use`


0age:
## Summary

When using the Anthropic-compatible streaming endpoint:

POST /v1/messages (stream=true)

with tools enabled, the server streams raw model markup
(e.g. `</think>`, `<tool_call>...</tool_call>`) as `content_block_delta`
text **before** emitting the structured `tool_use` content block.

This contradicts the documented behavior of:

> "Automatic special token filtering in streamed output"

and causes tool-using clients (e.g. Claude Code, Cline) to display raw
tool markup and reasoning tags in the assistant output.

------------------------------------------------------------------------

### Request

``` bash
curl -sN http://127.0.0.1:8000/v1/messages \
  -H "content-type: application/json" \
  -d '{
    "model":"default",
    "stream": true,
    "max_tokens": 200,
    "messages":[{"role":"user","content":"Use tool get_weather with city=Paris, and do not explain."}],
    "tools":[{"name":"get_weather","description":"Get weather","input_schema":{"type":"object","properties":{"city":{"type":"string"}},"required":["city"]}}]
  }'
```

### Observed Streaming Output (excerpt)

``` text
event: content_block_delta
data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"</think>"}}

event: content_block_delta
data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"<tool_call>\n"}}

event: content_block_delta
data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"</tool_call>"}}
```

After these text deltas, the server correctly emits:

``` text
event: content_block_start
data: {"type":"content_block_start","index":1,"content_block":{"type":"tool_use", ...}}
```

------------------------------------------------------------------------

## Expected Behavior

When tools are enabled and the model chooses to call a tool:

-   The stream **must not emit**:
    -   `<think>` / `</think>`
    -   `<tool_call>` / `</tool_call>`
    -   `<function=...>` / `<parameter=...>`
-   The stream should emit:
    -   Proper `tool_use` content blocks
    -   Incremental `input_json_delta` for tool arguments
-   Any reasoning or tool-call markup generated internally by the model
    should not appear in `text_delta`.

------------------------------------------------------------------------

## Why This Is a Problem

Tool clients (Claude Code, Cline, etc.):

[...]

PR #132, opened 2026-03-02
Fix Anthropic streaming leaking <think> and <tool_call> markup + add tests

0age:
This PR implements https://github.com/waybarrios/vllm-mlx/issues/129 and includes test coverage.

Fix (6805baf): Adds _AnthropicStreamScrubber, a stateful stream scrubber that strips <think>…</think>, <tool_call>…</tool_call>, <function=NAME>…</function>, and <parameter=NAME>…</parameter> markup from Anthropic /v1/messages streaming text deltas. When tools are present in the request, the scrubber is activated so clients only see clean text and structured tool_use blocks. Handles tags split across token boundaries via a carry buffer.

Tests (c1f7cac): Adds 83 unit tests for _AnthropicStreamScrubber covering:

Suppression of all tag types (single delta, split across boundaries, nested)
Stray closing tag removal
Carry buffer correctness when tags are split character-by-character
State machine transitions (TEXT ↔ IN_THINK / IN_TOOLCALL / IN_FUNCTION)
flush() semantics (emit vs discard based on mode, residual tag stripping)
Edge cases (empty blocks, unicode, very long content, angle brackets in plain text)
Realistic token-by-token streaming simulations
Pure logic tests with no model loading required.

### DIFF
--- a/tests/test_anthropic_stream_scrubber.py
+++ b/tests/test_anthropic_stream_scrubber.py
@@ -8,6 +8,8 @@ Anthropic /v1/messages endpoint.
 These are pure logic tests with no MLX dependency.
 """
 
+import pytest
+
 from vllm_mlx.server import _AnthropicStreamScrubber
 
 # =============================================================================
@@ -293,15 +295,12 @@ class TestScrubberParameterTags:
     def test_parameter_tag_single_delta(self):
         scrubber = _AnthropicStreamScrubber()
         result = scrubber.feed("Before <parameter=city>NYC</parameter> After")
-        # parameter= maps to IN_FUNCTION mode, closes on </function>
-        # But </parameter> is consumed as stray closing tag in TEXT mode...
-        # Actually <parameter= opens IN_FUNCTION, and IN_FUNCTION closes on </function>
-        # So </parameter> won't close IN_FUNCTION. Let's verify behavior.
-        # The close for IN_FUNCTION is </function>, so content after <parameter=city>
-        # stays suppressed until </function> is seen or stream ends.
+        # <parameter= opens IN_PARAMETER mode, closes on </parameter>
         result += scrubber.flush()
         assert "<parameter=" not in result
+        assert "NYC" not in result
         assert "Before" in result
+        assert "After" in result
 
     def test_stray_parameter_close_suppressed(self):
         """A stray </parameter> tag should be stripped."""
@@ -685,10 +684,15 @@ class TestScrubberStateMachine:
         scrubber.feed("<function=test>body</function>")
         assert scrubber.mode == "TEXT"
 
-    def test_text_to_parameter_enters_function_mode(self):
+    def test_text_to_parameter_enters_parameter_mode(self):
         scrubber = _AnthropicStreamScrubber()
         scrubber.feed("text<parameter=x>")
-        assert scrubber.mode == "IN_FUNCTION"
+        assert scrubber.mode == "IN_PARAMETER"
+
+    def test_parameter_back_to_text(self):
+        scrubber = _AnthropicStreamScrubber()
+        scrubber.feed("<parameter=x>body</parameter>")
+        assert scrubber.mode == "TEXT"
 
     def test_stray_close_stays_in_text(self):
         """Stray closing tags should not change mode."""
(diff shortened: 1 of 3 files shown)


waybarrios (commented):
Solid work overall, the state machine is clean and test coverage is great. Left some inline comments on specific things I noticed.

janhilgard (commented):
Solid work — the state machine approach is the right call for streaming, and the test coverage is impressive.

A few things beyond what @waybarrios already flagged:

### `SPECIAL_TOKENS_PATTERN` already strips `<tool_call>` tags

In `vllm_mlx/api/utils.py`, `SPECIAL_TOKENS_PATTERN` already matches `</?tool_call>|</?tool_call_reasoning>`. Since `_stream_anthropic_messages` applies this pattern *before* feeding content to the scrubber (via `clean_output_text()`), the scrubber will never actually see `<tool_call>` or `</tool_call>` tags — they're already gone.

This means the scrubber's `IN_TOOLCALL` mode is dead code in practice. Either:
1. Remove `<tool_call>` from `SPECIAL_TOKENS_PATTERN` and let the scrubber handle it (cleaner separation of concerns), or
2. Remove `IN_TOOLCALL` from the scrubber and document that `SPECIAL_TOKENS_PATTERN` handles it

I'd lean toward (1) — the scrubber is the right place for semantic tag stripping, and `SPECIAL_TOKENS_PATTERN` should only handle tokenizer-level special tokens.

### Scrubber/Router duplication

`_AnthropicStreamRouter.feed()` duplicates ~80% of `_AnthropicStreamScrubber.feed()`. The only difference is that the router yields `("thinking", text)` pieces instead of suppressing. Consider:
- A single class with an `emit_thinking: bool` parameter
- Or have the Router subclass the Scrubber, overriding only the `IN_THINK` handling

This would halve the code and ensure bug fixes apply to both paths.

### Carry growth cap for prefix tags

To expand on @waybarrios's unbounded carry comment: when `<function=` is found but `>` never arrives, each `feed()` call does `s = self.carry + delta`, finds the prefix again at position 0, and stores everything from there as carry. A simple fix:

```python
if len(self.carry) > self.CARRY_N + 256:
    # Prefix tag never closed — treat as plain text
    out.append(self.carry)
    self.carry = ""
```

Overall this is a welcome fix for a real usability problem. The core logic is sound — the issues above are about reducing surface area and hardening edge cases.

Thump604:
@waybarrios, @0age: independent review and status note.

## Scope

`_AnthropicStreamScrubber` is a stateful stream scrubber that strips `<think>...</think>`, `<tool_call>...</tool_call>`, `<function=NAME>...</function>`, and `<parameter=NAME>...</parameter>` markup from Anthropic /v1/messages streaming text deltas. Activated when tools are present in the request so clients see clean text and structured tool_use blocks. Handles tags split across token boundaries via a carry buffer.

83 unit tests covering positive cases, negative cases, and split-tag boundary cases. The test count is appropriate for a stateful streaming parser where boundary handling is the entire point.

## Status

PR currently shows CONFLICTING merge status. 2007+/14- across 3 files. The conflict is likely on the Anthropic streaming code path which has had other changes since this branch was created.

Implements issue #129 (per the PR description). The scrubber approach is sound for the case where the upstream model emits raw markup that the Anthropic schema does not allow.

## Recommendation

Merge candidate after rebase. The scrubber-with-test-coverage approach is the right shape for protecting the Anthropic adapter from upstream markup leakage. Last activity Mar 25.

janhilgard:
Hey @0age — thanks for the thorough scrubber implementation and the excellent test suite!

Just a heads up: the current `main` branch now handles this through a different mechanism that was backported in PR #278:

1. **Reasoning parser** (`--reasoning-parser qwen3/gemma4`) extracts `<think>` blocks and emits them as proper Anthropic `thinking` content blocks
2. **Tool parsers** (`--tool-call-parser qwen/gemma4/minimax`) consume `<tool_call>`, `<function=...>` markup and emit structured `tool_use` blocks
3. **Streaming partial-marker buffering** in tool parsers (PR #281) handles tags split across token boundaries — buffers potential markers and recovers on false positives
4. **Safety nets** in the Anthropic streaming path suppress any leaked `<function` markup that slips through the parser

The combination covers the same scenarios your scrubber addresses, but through the parser pipeline rather than a separate stripping layer. Your state-machine approach with the carry buffer is arguably more robust for edge cases though.

Want to check if there are still gaps that your scrubber would catch beyond what the current parser pipeline handles?

Thump604:
Status update after rechecking current main:

I do not think this branch still has a clean merge path. The core problem it addressed is now covered on main through the parser pipeline that landed in #278 plus the partial-marker streaming fixes from #281, and this branch now conflicts heavily with that path.

The scrubber/state-machine work here was thoughtful, especially the boundary tests, but at this point I would rather port any remaining edge-case test cases or narrowly missing behaviors into the current parser path than try to rebase a second overlapping streaming stack.

My recommendation is to close this as superseded by the merged parser pipeline. If anyone can show a current-main streaming leak that the existing parser path still misses, I would handle that as a focused follow-up against main rather than reopening this whole branch.

Thump604:
Closing as superseded by the merged parser pipeline in #278/#281. Happy to follow up with a narrow bug fix if a current-main leak still reproduces.

PR #148, opened 2026-03-09
fix: integrate tool call parsing with reasoning parser in streaming mode

jsirish:
## Problem

When both `--reasoning-parser` and `--tool-call-parser` are enabled, the reasoning parser branch in `stream_chat_completion` consumes all streaming tokens without routing them through the tool call parser. This causes `<tool_call>` XML markup to be emitted as raw text in the `reasoning` or `content` fields instead of being parsed into structured `tool_calls` in the response.

This affects models like **Qwen3-Coder** that emit tool calls directly inside reasoning blocks without a clean `</think>` transition.

## Fix

The fix integrates tool call detection into the reasoning parser's streaming branch:

1. After the reasoning parser processes each delta, the effective text (whether `content` or `reasoning`) is also fed through the tool call parser
2. If tool markup is detected, the output is suppressed until the full tool call is parsed
3. When a complete tool call is found, it's emitted as a structured `tool_calls` chunk with `finish_reason: "tool_calls"`
4. Non-tool-call content continues to flow through the reasoning parser's normal path

## Testing

Tested with:
- **Model**: Qwen3-Coder-Next-8bit
- **Flags**: `--reasoning-parser qwen3 --tool-call-parser qwen3_coder --continuous-batching`
- **Load**: 4 concurrent agents (CoPaw, OpenCode) over ~200 requests
- **Result**: All requests returned structured tool calls correctly; zero errors

### DIFF
--- a/vllm_mlx/server.py
+++ b/vllm_mlx/server.py
@@ -1934,6 +1934,59 @@ async def stream_chat_completion(
                 # Skip this chunk (e.g., <think> token itself)
                 continue
 
+            # Check if tool call markup appears in reasoning or content.
+            # Some models (e.g. Qwen3-Coder) emit <tool_call> directly
+            # inside reasoning without a </think> transition, so we need to
+            # intercept tool call tokens regardless of which field they land in.
+            effective_text = delta_msg.content or delta_msg.reasoning or ""
+            if tool_parser and effective_text:
+                if not tool_markup_possible and "<" not in effective_text:
+                    tool_accumulated_text += effective_text
+                    # No tool markup yet — emit the delta as-is
+                else:
+                    if not tool_markup_possible:
+                        tool_markup_possible = True
+                    tool_previous = tool_accumulated_text
+                    tool_accumulated_text += effective_text
+                    tool_result = tool_parser.extract_tool_calls_streaming(
+                        tool_previous, tool_accumulated_text, effective_text,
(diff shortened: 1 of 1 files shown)


waybarrios:
@Thump604 take a look on this!

janhilgard:
Hey @jsirish — this fix is now covered in `main`. The streaming pipeline was rewritten in PR #278 to properly integrate reasoning parser + tool parser:

1. Reasoning parser runs first on each streaming delta, extracts `<think>` content into `reasoning` field
2. Cleaned text (content after reasoning extraction) is passed to the tool parser
3. Tool parser handles `<tool_call>`, `<function=...>`, `[Calling tool:...]` with partial-marker buffering (PR #281)
4. Safety nets catch any remaining leaked markup

This handles the exact scenario you described — models emitting tool calls inside/after reasoning blocks.

Also related: PR #177 (same fix by @Thump604) and PR #158 (similar fix by @PollyBot13) — all three identified the same root cause independently.

Thanks for the contribution and the thorough testing with Qwen3-Coder!

Thump604:
Closing as superseded by the current streaming parser stack in main. The functional server-side fix this PR was carrying is now covered by the #278 integration path, so I don't think reviving this conflicting branch is the right move. Thanks for the careful reproduction and narrow patch.

PR #150, opened 2026-03-10
feat: native Qwen3-VL video support in MLLM mode

patanet7:
## Summary

- Fix `video_url` content type being silently ignored in MLLM chat/stream_chat
- Fix video frame tokens not counted in prompt token reporting  
- Fix `has_media` always False for MLLM mode in server.py
- Add native video pipeline for Qwen3-VL family models (temporal 3D conv + M-RoPE + timestamp interleaving) using mlx-vlm's existing `process_vision_info` infrastructure

### Benchmarks (M-series, ZwZ-8B via mlx-vlm)

| Setting | Frames | video_grid_thw | Prompt tokens | Prefill | Gen speed |
|---------|--------|----------------|---------------|---------|-----------|
| 1 fps (4f) | 4 | [2, 36, 64] | 1,189 | ~3s | 23.9 tok/s |
| 2 fps (4f) | 4 | [2, 36, 64] | 1,200 | ~3s | 23.9 tok/s |
| 60 fps (354f) | 354 | [5, 36, 64] | 2,935 | ~12s | 5.7 tok/s |

Old approach (frames-as-images) produced ~32,666 prompt tokens for the same video.

### DIFF
--- a/tests/test_video.py
+++ b/tests/test_video.py
@@ -34,9 +34,9 @@ class TestSmartNframes:
     def test_result_always_even(self):
         for total in [5, 7, 11, 13, 100, 999]:
             result = smart_nframes(total, 30.0)
-            assert result % FRAME_FACTOR == 0, (
-                f"Odd frame count {result} for total={total}"
-            )
+            assert (
+                result % FRAME_FACTOR == 0
+            ), f"Odd frame count {result} for total={total}"
 
 
 class TestVideoUrlParsing:


waybarrios (changes requested):
Good feature idea, native video pipeline for Qwen3-VL is definitely the right direction. Found a few things that need attention before merging though.

janhilgard:
+1, @waybarrios' review covers everything important.

**Fixes already pushed look good:**
- `_video_native = False` init in `__init__` — prevents AttributeError before `load()`
- `process_vision_info` import guard with actionable error message
- Empty `video_source` bail-out — no more 500s from `process_video_input("")`
- `has_media` rewritten to scan actual message content instead of relying on `request.video_fps`

**Agree with the two remaining items for @patanet7:**

1. **`tools` not forwarded in `_generate_native_video`** — `apply_chat_template` is called directly without `template_extra_kwargs`, so tool definitions are silently dropped. This breaks PR #124's tool-call support for any video request that also includes tools.

2. **Blocking `generate()` in `stream_chat` native path** — yields a single chunk after the full generation completes. With 354 frames / ~12s prefill, the event loop is completely locked. At minimum wrapping in `asyncio.to_thread()` would keep the loop alive; ideally mlx-vlm's streaming API (if available) would give true incremental output.

One minor nit: the first-pass video extraction loop (`_msg_video_inputs` collection) is duplicated verbatim (~30 lines) between `chat()` and `stream_chat()`. Could be a private helper, but that's a refactor for later.

PR #156, opened 2026-03-12
Fix chunked prefill crash with mlx-lm >= 0.31.0

### LINKED ISSUE #155: Chunked prefill crashes with mlx-lm >= 0.31.0: zip unpacking expects 6 values, gets 7
## Bug Description

When using `vllm-mlx` with `mlx-lm >= 0.31.0` and `--continuous-batching` enabled, chunked prefill crashes with a `ValueError` because the `zip(*batch_prompts)` unpacking in `scheduler.py` expects exactly 6 tuple elements, but mlx-lm 0.31.x now returns 7.

## Root Cause

mlx-lm PR [ml-explore/mlx-lm#911](https://github.com/ml-explore/mlx-lm/pull/911) ("Better caching in the server") added a `prompt_checkpoints` field to the prompt tuples returned by `BatchGenerator.unprocessed_prompts`. This means each entry in `batch_prompts` now contains 7 elements instead of 6.

The unpacking at [scheduler.py L378-L385](https://github.com/waybarrios/vllm-mlx/blob/main/vllm_mlx/scheduler.py#L378) hardcodes exactly 6 target variables:

```python
(
    uids,
    inputs_raw,
    max_tokens_list,
    caches,
    samplers,
    logits_processors,
) = zip(*batch_prompts)
```

## Steps to Reproduce

1. Install `vllm-mlx==0.2.6` with `mlx-lm>=0.31.0`
2. Start with `--continuous-batching` enabled
3. Send any inference request
4. Observe crash/ValueError on the `zip(*batch_prompts)` line

## Environment

- **vllm-mlx**: 0.2.6
- **mlx-lm**: 0.31.1
- **mlx**: 0.31.1
- **Hardware**: Apple Silicon (Mac Studio Ultra, M2 Ultra 192GB)
- **Model**: Qwen3.5-35B-A3B-8bit (also reproduced with Qwen3-14B)

## Proposed Fix

Add a `*_extra` catch-all to the unpacking so it's forward-compatible with additional fields from mlx-lm:

```python
(
    uids,
    inputs_raw,
    max_tokens_list,
    caches,
    samplers,
    logits_processors,
    *_extra,
) = zip(*batch_prompts)
```

This is a minimal, non-breaking change — the extra values are captured but ignored, maintaining backward compatibility with older mlx-lm versions that return 6 elements.

PR incoming with the fix.

omprashantjain:
## Summary

Fixes #155 — chunked prefill crashes with `mlx-lm >= 0.31.0` because `zip(*batch_prompts)` unpacking expects 6 tuple elements but now receives 7.

## Root Cause

mlx-lm PR [ml-explore/mlx-lm#911](https://github.com/ml-explore/mlx-lm/pull/911) ("Better caching in the server") added a `prompt_checkpoints` field to the prompt tuples returned by `BatchGenerator.unprocessed_prompts`. The hardcoded 6-variable unpacking in `scheduler.py:_chunked_next()` then fails with a `ValueError`.

## Fix

Add `*_extra` catch-all to the tuple unpacking:

```python
(
    uids,
    inputs_raw,
    max_tokens_list,
    caches,
    samplers,
    logits_processors,
    *_extra,                 # ← new: absorb extra fields from mlx-lm
) = zip(*batch_prompts)
```

This is:
- **Forward-compatible** — handles any future fields mlx-lm might add
- **Backward-compatible** — older mlx-lm versions that return 6 elements work fine (`*_extra` is empty)
- **Minimal** — 1 line added, no behavioral changes

## Testing

Tested on two Mac Studio Ultras (M2 Ultra, 192GB) with:
- `vllm-mlx==0.2.6`, `mlx-lm==0.31.1`, `mlx==0.31.1`
- Models: Qwen3.5-35B-A3B-8bit, Qwen3-14B
- `--continuous-batching` enabled
- 30 concurrent heavy requests (2048 max_tokens each) — all succeeded
- Aggregate throughput: ~416 tok/s across concurrent requests
- Stable over 48+ hours of continuous operation

### DIFF
--- a/vllm_mlx/scheduler.py
+++ b/vllm_mlx/scheduler.py
@@ -392,6 +392,7 @@ def _install_chunked_prefill(
                         caches,
                         samplers,
                         logits_processors,
+                        *_extra,
                     ) = zip(*batch_prompts)
                     lengths = [len(p) for p in inputs_raw]
                     max_length = max(lengths)


janhilgard (approved):
LGTM — clean, minimal, correct fix.

**Verified completeness:** There are 3 sites in `scheduler.py` where `batch_prompts` tuples are unpacked:

| Site | Line | Pattern | Status |
|------|------|---------|--------|
| A | ~361 | `len(p[1])` — index-based access | Safe |
| B | ~367 | `for _uid, _toks, *_ in batch_prompts` — already has catch-all | Safe |
| C | ~396 | `zip(*batch_prompts)` — hardcoded 6 vars | **Fixed by this PR** |

This is the only vulnerable site. The `*_extra` catch-all is idiomatic Python and provides forward compatibility — works with both mlx-lm < 0.31.0 (6 fields) and >= 0.31.0 (7+ fields).

The new `prompt_checkpoints` field is for mlx-lm's internal agentic cache management and can be safely ignored by vllm-mlx, which handles prompt processing end-to-end.

No changes needed to `pyproject.toml` — the existing `mlx-lm>=0.30.5` constraint is correct since the fix is backward compatible.

PR #158, opened 2026-03-13
fix(anthropic_adapter): strip streaming XML artifacts from text blocks to prevent infinite reasoning loops

PollyBot13:
### Problem
When using Qwen3.5 (or similar models that emit `<tool_call>` and `<think>` XML), the Anthropic Messages API streaming path has two bugs:

1. **Multi-turn history**: `_convert_message()` passes raw XML from previous assistant text blocks back into the chat template, causing duplicate tool calls and infinite reasoning loops on turn 2+.

2. **First-turn streaming**: `_stream_anthropic_messages()` sends every model token immediately as `content_block_delta` events, including `<tool_call>` and `<think>` XML. The tool call parser runs *after* the stream completes, so raw XML is already sent to the client as visible text.

### Fix
1. Strip `<think>` and `<tool_call>` XML from text blocks in `_convert_message()` before they enter the chat template.
2. Buffer streaming text, parse and clean it after completion, then emit clean text + proper `tool_use` blocks.

### Testing
- All existing tests pass
- Tested with Qwen3.5-35B-A3B-4bit: multi-turn tool calling works, no XML leaks in first or subsequent turns

### DIFF
--- a/vllm_mlx/api/anthropic_adapter.py
+++ b/vllm_mlx/api/anthropic_adapter.py
@@ -9,6 +9,7 @@ Handles translation of:
 """
 
 import json
+import re
 import uuid
 
 from .anthropic_models import (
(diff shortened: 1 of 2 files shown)


janhilgard:
Hey @PollyBot13 — thanks for identifying these two bugs (multi-turn XML leak + first-turn streaming leak).

Both are now resolved in `main` through a different architecture:

| Problem | Current solution in `main` |
|---------|--------------------------|
| **XML in multi-turn history** | Reasoning parser (`--reasoning-parser qwen3`) extracts `<think>` blocks; tool parser consumes `<tool_call>`/`<function=...>` markup — only clean text enters the response and subsequent turns |
| **XML leak in streaming** | `tool_markup_possible` flag + `tool_accumulated_text` buffering + streaming reasoning parser + tool parser partial-marker buffering (PR #281) + safety nets for leaked markup |

These landed via PR #256 (reasoning parsers, Gemma 4 patches) and PR #278 (production backport with full streaming pipeline).

Closing as superseded — the underlying issues are fully addressed. Thanks for the contribution!

PR #144, opened 2026-03-06
fix: handle 3D KV tensors in prefix cache for Qwen3.5 models

jsirish:
## Problem

Qwen3.5 models (both MoE and dense variants) produce **3D KV cache tensors** with shape `(n_kv_heads, seq_len, head_dim)` instead of the **4D** shape `(batch, n_kv_heads, seq_len, head_dim)` that `prefix_cache.py` assumes.

This causes `_extract_block_tensor_slice()` to fail with:
```
Too many indices for array with 3 dimensions
```

Prefix caching silently falls back to disabled, losing significant performance.

## Fix

Three locations in `prefix_cache.py` hardcoded `axis=2` / `shape[2]` / `[:, :, start:end, :]` for 4D tensors. The fix dynamically detects tensor dimensionality via `ndim` and computes `seq_axis = ndim - 2` (sequence length is always the second-to-last dimension):

| Location | Before | After |
|---|---|---|
| `_extract_block_tensor_slice()` | `keys[:, :, start:end, :]` | Dynamic slice tuple via `range(ndim)` |
| `reconstruct_cache()` | `mx.concatenate(..., axis=2)` | `mx.concatenate(..., axis=seq_axis)` |
| `SimpleKVCache.__init__()` | `keys.shape[2]` | `keys.shape[ndim - 2]` |

This is **backward compatible** — 4D tensors still use axis=2 (since `4-2=2`), while 3D tensors correctly use axis=1.

## Test Results (Qwen3.5-122B-A10B-4bit on M3 Ultra)

| Request | tok/s | Notes |
|---|---|---|
| 1st (cold) | 27.9 | No cache |
| 2nd (same prompt) | **45.4** | ✅ Cache hit — **63% faster** |
| 3rd (different user msg) | 23.4 | Partial cache miss (expected) |

Zero errors in server logs. Prefix cache now works correctly for Qwen3.5 models.

## Related Issues

- Fixes #142
- Relates to dynamic/llm-server#44

Issue #142, opened 2026-03-06
Prefix cache tensor slicing fails for Qwen3.5 MoE models (3D KV cache)


jsirish:
## Description

When using `--enable-prefix-cache` with `--use-paged-cache` and Qwen3.5 MoE models (e.g., `mlx-community/Qwen3.5-122B-A10B-4bit`), the prefix cache emits the following warning on every request:

```
WARNING:vllm_mlx.prefix_cache:Failed to extract block tensor slice: Too many indices for array with 3 dimensions.
```

The prefix cache falls back gracefully (doesn't crash), but **no KV blocks are actually persisted**, so there's no TTFT benefit on repeated contexts.

## Root Cause

In `prefix_cache.py` `_extract_block_tensor_slice()` (line ~655), the code assumes KV cache tensors are 4D:

```python
# KV cache shape: (batch, n_kv_heads, seq_len, head_dim)
keys_slice = keys[:, :, start_idx:end_idx, :]
```

However, **Qwen3.5 MoE models produce 3D KV tensors**: `(n_kv_heads, seq_len, head_dim)` — there's no batch dimension. The 4-index slice fails on a 3D array.

## Suggested Fix

Check `keys.ndim` and adjust the slicing dimension:

```python
if keys.ndim == 4:
    # Standard: (batch, n_kv_heads, seq_len, head_dim)
    seq_len = keys.shape[2]
    keys_slice = keys[:, :, start_idx:end_idx, :]
    values_slice = values[:, :, start_idx:end_idx, :]
elif keys.ndim == 3:
    # MoE models: (n_kv_heads, seq_len, head_dim)
    seq_len = keys.shape[1]
    keys_slice = keys[:, start_idx:end_idx, :]
    values_slice = values[:, start_idx:end_idx, :]
```

The same fix should be applied to `reconstruct_cache()` where it concatenates along `axis=2` — for 3D tensors this should be `axis=1`.

## Environment

- **vllm-mlx**: 0.2.6
- **mlx-lm**: 0.30.7
- **Model**: `mlx-community/Qwen3.5-122B-A10B-4bit`
- **Hardware**: Mac Studio M3 Ultra (256GB)
- **macOS**: 15.x

## Launch Command

```bash
vllm-mlx serve mlx-community/Qwen3.5-122B-A10B-4bit \
  --enable-prefix-cache --prefix-cache-size 1000 \
  --use-paged-cache --paged-cache-block-size 64 --max-cache-blocks 4000 \
[...]

PR #164, opened 2026-03-16
fix: handle hybrid model caches in prefix cache (3D tensors + non-KV layers)

### LINKED ISSUE #136: Prefix cache fails on MoE models (Qwen3.5-35B-A3B): "Too many indices for array with 3 dimensions"
### Description

`--enable-prefix-cache` does not work with MoE models. Every request produces hundreds of warnings:

```
WARNING:vllm_mlx.prefix_cache:Failed to extract block tensor slice: Too many indices for array with 3 dimensions.
WARNING:vllm_mlx.prefix_cache:Cannot allocate block for <request-id>
```

Cross-request prefix caching never activates. Cold and warm requests take identical time regardless of identical prompts.

### Reproduction

```bash
vllm-mlx serve nightmedia/Qwen3.5-35B-A3B-Text-qx64-hi-mlx \
  --host 0.0.0.0 --port 8000 \
  --continuous-batching \
  --use-paged-cache \
  --enable-prefix-cache \
  --prefix-cache-size 100 \
  --cache-memory-mb 24576

# Send identical request twice:
curl -s http://localhost:8000/v1/chat/completions \
  -d '{"model":"nightmedia/Qwen3.5-35B-A3B-Text-qx64-hi-mlx",
       "messages":[{"role":"system","content":"You are a helpful assistant."},
                   {"role":"user","content":"Say hello"}],
       "max_tokens":10,"temperature":0,"stream":false}'

# Second request shows no speedup. stderr is full of the warnings above.
```

### A/B test results

| Prompt size | Cold | Warm (2nd identical request) | Cache working? |
|---|---|---|---|
| ~100 tokens | 0.58s | 0.35s | Yes |
| ~10K tokens | 6.74s | 6.38s | Marginal |
| ~28K tokens | 12.15s | 11.86s | No |

Small prompts show cache benefit. Large prompts on MoE model do not.

### Root cause hypothesis

The prefix cache code in `vllm_mlx/prefix_cache.py` attempts to extract KV block tensor slices assuming a 4D layout `[layers, tokens, heads, head_dim]` (standard dense attention). Qwen3.5-35B-A3B is an MoE model with `Qwen3_5MoeForConditionalGeneration` architecture. Its KV cache tensors likely use a 3D layout (possibly `[tokens, heads, head_dim]` per-layer or expert-grouped), causing the "Too many indices for array with 3 dimensions" error when the prefix cache tries to index into them.

The error is caught as a warning, so inference continues without caching. This means prefix caching silently degrades to no-op on all MoE models.

### Model details

```
model_type: qwen3_5_moe
architectures: ['Qwen3_5MoeForConditionalGeneration']
num_hidden_layers: 40
num_attention_heads: 16
num_key_value_heads: 2
hidden_size: 2048
head_dim: 256
num_experts: 256
num_experts_per_tok: 8
```

Model: `nightmedia/Qwen3.5-35B-A3B-Text-qx64-hi-mlx` (text-only MLX build of Qwen3.5-35B-A3B, converted with mlx-lm v0.30.8, 6-bit quantization)

### Environment

- **vllm-mlx version:** 0.2.6 (upgraded via `uv tool upgrade --reinstall`)
- **Hardware:** Mac Studio M4 Max, 64GB unified memory
- **macOS:** 26.3 (25D125)
- **Python:** 3.9.6

### Expected behavior

Second identical request should reuse the cached KV prefix and complete significantly faster (e.g. 28K token prompt: cold ~12s, warm ~1-3s).

### Actual behavior

All requests take the same time. stderr is flooded with `Failed to extract block tensor slice: Too many indices for array with 3 dimensions` warnings. Prefix cache is effectively disabled for MoE models.

### Impact

This blocks prefix caching for all MoE models served through vllm-mlx, including the popular Qwen3.5-35B-A3B and Qwen3-30B-A3B families. These are among the most commonly used models on Apple Silicon for agentic workloads where prefix caching matters most (large system prompts repeated across requests).

### Possibly related

- #116 (cross-request prefix cache misses)
- #174 (shared prefix bug)
- #94 (shared prefix cache PR)

Thump604:
## Summary

Fixes prefix cache for hybrid models (Qwen3.5, Nemotron, etc.) that mix standard attention (`KVCache`) and linear attention (`ArraysCache`/GatedDeltaNet/SSM) layers.

**The problem:** `_extract_block_tensor_slice()` hardcodes 4D KV tensor assumptions (`keys[:, :, start:end, :]`) that fail on:
1. **3D KV tensors** `(n_kv_heads, seq_len, head_dim)` — common in Qwen3.5 MoE models (no batch dim)
2. **Non-KV cache layers** (`ArraysCache`) with cumulative SSM state that cannot be block-decomposed

Every request logs `WARNING: Failed to extract block tensor slice: Too many indices for array with 3 dimensions`, prefix caching silently becomes a no-op, and there is zero TTFT benefit.

**The fix:**
- **`_is_kv_cache_layer()`** — detects KV vs non-KV layers by class name (known types) with structural fallback
- **`_extract_block_tensor_slice()`** — skips non-KV layers (None sentinel), uses `ndim-2` for dynamic sequence axis on KV layers
- **Non-KV state storage** — stores full cumulative state per prefix (keyed by block IDs) for exact-match reconstruction via `from_state()`
- **`reconstruct_cache()`** — merges KV layers from blocks + non-KV layers from stored state; returns `None` when non-KV states are unavailable (partial prefix on hybrid model) to prevent incorrect output
- **Memory management** — `release_cache()` cleans up non-KV states, `fork_cache()` propagates them, states are deep-copied on store

**Behavior by model type:**

| Model type | Prefix sharing | Exact prefix match |
|---|---|---|
| Pure attention (e.g. Llama) | ✅ Full support | ✅ Full support |
| Hybrid (e.g. Qwen3.5 MoE) | ⚠️ Graceful fallback (returns None) | ✅ Full support |

This is correct: SSM state is cumulative and only valid for the exact token sequence that produced it. Reusing SSM state from a different/longer sequence would produce silently wrong output.

## Test plan

- [x] 20 new tests in `tests/test_prefix_cache_hybrid.py` covering:
  - `_is_kv_cache_layer()` detection (class name + structural fallback)
  - 3D and 4D KV tensor slicing
  - Mixed KV/ArraysCache layer extraction
  - Store + reconstruct roundtrip (pure KV 3D, hybrid exact match)
  - Partial prefix rejection for hybrid models
  - Clear/reset, null safety, stats
- [x] All 62 existing tests pass (zero regressions)
- [x] **Live test:** Qwen3.5-122B-A10B-5bit on M2 Ultra with `--continuous-batching --enable-prefix-cache --use-paged-cache`

### Live test results (Qwen3.5-122B-A10B-5bit, M2 Ultra 128GB)

| Request | Time | Speedup |
|---|---|---|
| 1. Cold (no cache) | 5307ms | baseline |
| 2. Warm (same prompt) | 1007ms | **5.3x** |
| 3. Different user msg | 1004ms | **5.3x** |

Zero warnings in server logs. No `"Too many indices"` errors. No `"All input arrays must have the same number of dimensions"` errors.

Fixes #136, #142. Supersedes #144 (which only handles 3D KV tensors but not non-KV layers).

### DIFF
--- a/tests/test_prefix_cache_hybrid.py
+++ b/tests/test_prefix_cache_hybrid.py
@@ -1,0 +1,412 @@
+# SPDX-License-Identifier: Apache-2.0
+"""
+Tests for prefix cache with hybrid models (mixed KV + non-KV cache layers).
+
+Hybrid models like Qwen3.5 use a mix of standard attention layers (KVCache)
+and linear attention / SSM layers (ArraysCache). The prefix cache must handle
+both cache types correctly:
+- KVCache layers: sequence-indexed, can be sliced into blocks
+- ArraysCache layers: cumulative state, stored separately per prefix
+
+These tests cover:
+- 3D vs 4D KV tensor handling
+- Mixed KV/ArraysCache layer detection and storage
+- Exact prefix match reconstruction for hybrid models
+- Graceful fallback when non-KV states are unavailable
+"""
+
+import platform
+import sys
+from unittest.mock import MagicMock
+
+import pytest
+
+pytestmark = pytest.mark.skipif(
+    sys.platform != "darwin" or platform.machine() != "arm64",
+    reason="Requires Apple Silicon",
+)
+
+
+def _make_array(shape, dtype="float16"):
+    """Create a mock MLX-like array with shape and ndim."""
+    import mlx.core as mx
+
+    return mx.zeros(shape, dtype=getattr(mx, dtype))
+
+
+def _make_kv_state_4d(seq_len, n_heads=4, head_dim=64):
+    """Create a 4D KV cache state: (batch, n_heads, seq_len, head_dim)."""
+    keys = _make_array((1, n_heads, seq_len, head_dim))
+    values = _make_array((1, n_heads, seq_len, head_dim))
+    return (keys, values)
+
+
+def _make_kv_state_3d(seq_len, n_heads=4, head_dim=64):
+    """Create a 3D KV cache state: (n_heads, seq_len, head_dim) — no batch dim."""
+    keys = _make_array((n_heads, seq_len, head_dim))
+    values = _make_array((n_heads, seq_len, head_dim))
+    return (keys, values)
+
+
+def _make_arrays_cache_state():
+    """Create an ArraysCache state: list of [conv_state, ssm_state]."""
+    conv_state = _make_array((1, 3, 128))  # (batch, kernel_size-1, conv_dim)
+    ssm_state = _make_array((1, 16, 64))  # (batch, state_size, dim)
+    return [conv_state, ssm_state]
+
+
+def _kv_layer_dict(state, class_name="KVCache"):
+    """Build a layer state dict for a KV cache layer."""
+    from mlx_lm.models.cache import KVCache
+
+    return {
+        "state": state,
+        "meta_state": (str(state[0].shape[-2]),),
+        "class_name": class_name,
+        "class_ref": KVCache,
+    }
+
+
+def _arrays_layer_dict(state, class_name="ArraysCache"):
+    """Build a layer state dict for an ArraysCache layer."""
+    from mlx_lm.models.cache import ArraysCache
+
+    return {
+        "state": state,
+        "meta_state": "",
+        "class_name": class_name,
+        "class_ref": ArraysCache,
+    }
+
+
+# ── _is_kv_cache_layer ────────────────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


Thump604:
Closing — prefix cache base support was incorporated upstream in v0.2.6. The remaining hybrid model handling (non-KV layer separation for Mamba/SSM) is covered by #165.

PR #166, opened 2026-03-16
fix: strip tools from template when tool_choice="none"

### LINKED ISSUE #162: `tool_choice: "none"` ignored — model generates `tool_calls` with empty `{}` arguments when no tools defined
vllm-mlx version: latest (0.2.6)

Models affected: Qwen2.5-*-Instruct, Llama-3.*-Instruct (all models with tool-calling in their chat template)

Server flags used: none — plain vllm-mlx serve <model> --port 8080

**Describe the bug**

When sending a `/v1/chat/completions` request with`tool_choice: "none"` and either no tools field or `tools: []`, the model still returns finish_reason: "tool_calls" with tool_call objects whose arguments are always "{}". The model is encoding output data (e.g. theme names, analysis results) as tool function names instead of as text content.
Minimal reproduction
```bash
vllm-mlx serve mlx-community/Qwen2.5-14B-Instruct-4bit --port 8080
pythonimport requests
r = requests.post("http://localhost:8080/v1/chat/completions", json={
    "model": "mlx-community/Qwen2.5-14B-Instruct-4bit",
    "messages": [{"role": "user", "content": "List 3 fruits as a JSON array"}],
    "tool_choice": "none",
    "tools": [],
    "max_tokens": 200
})
print(r.json()["choices"][0])
# finish_reason: "tool_calls", content: null, tool_calls: [{"function": {"name": "Apple", "arguments": "{}"}}]
```

**Expected behaviour** 

`finish_reason: "stop"`, content contains the JSON array, `tool_calls` is absent.
Root cause
The chat template handler passes `tools=[]` (or the request's tools list) into apply_chat_template() even when tool_choice="none". Qwen2.5 and Llama 3.x have tool-calling jinja templates that activate when a tools key is present — even an empty list triggers the tool-call token branch in some tokenizer versions. Upstream vLLM handles this with `--exclude-tools-when-tool-choice-none` (which strips tools from the template context when `tool_choice="none"`), but this flag was never ported to vllm-mlx.
Proposed fix
In the chat completions handler, before calling apply_chat_template, check: if `tool_choice == "none"`, `pass tools=None` to the template regardless of what the request contained.

Thump604:
## Summary

Fixes #162 — `tool_choice: "none"` is ignored, model still generates `tool_calls` with empty `{}` arguments.

### Root cause

All three code paths (streaming chat, non-streaming chat, Anthropic streaming) pass `tools` to the chat template regardless of `tool_choice`. Qwen 3.5 and Llama 3.x templates activate tool-call token branches when any `tools` key is present — even an empty list.

### Fix

Add `and request.tool_choice != "none"` to the three `if request.tools:` checks before `convert_tools_for_template()`. When `tool_choice="none"`, tools are not passed to the template, so the model generates normal text content.

This matches upstream vLLM's `--exclude-tools-when-tool-choice-none` flag behavior.

### Test plan

- [x] Verified all 3 code paths are patched (streaming, non-streaming, Anthropic)
- [ ] Manual test with `tool_choice: "none"` and `tools: []` — expect `finish_reason: "stop"` with text content

### DIFF
--- a/vllm_mlx/server.py
+++ b/vllm_mlx/server.py
@@ -1359,8 +1359,9 @@ async def create_chat_completion(request: ChatCompletionRequest, raw_request: Re
         if request.video_max_frames:
             chat_kwargs["video_max_frames"] = request.video_max_frames
 
-    # Add tools if provided
-    if request.tools:
+    # Add tools if provided — strip when tool_choice="none" to prevent
+    # chat templates from activating tool-call tokens (fixes #162)
+    if request.tools and request.tool_choice != "none":
         chat_kwargs["tools"] = convert_tools_for_template(request.tools)
 
     if request.stream:
@@ -1536,7 +1537,7 @@ async def create_anthropic_message(
         "top_p": openai_request.top_p,
     }
 
-    if openai_request.tools:
+    if openai_request.tools and openai_request.tool_choice != "none":
         chat_kwargs["tools"] = convert_tools_for_template(openai_request.tools)
 
     start_time = time.perf_counter()
@@ -1693,7 +1694,7 @@ async def _stream_anthropic_messages(
         "top_p": openai_request.top_p,
     }
 
-    if openai_request.tools:
+    if openai_request.tools and openai_request.tool_choice != "none":
         chat_kwargs["tools"] = convert_tools_for_template(openai_request.tools)
 
     # Emit message_start


Thump604:
Closing — superseded by #173 which has a more complete implementation (strips tools from template context + suppresses tool call parsing across all 4 server paths).

PR #167, opened 2026-03-16
Fix missing data: [DONE] SSE termination event

### LINKED ISSUE #101: Streaming SSE responses intermittently missing data: [DONE] termination event
## Description

When using vllm-mlx with `stream: true` for chat completions, the `data: [DONE]` SSE termination event is intermittently not sent after generation completes. This causes OpenAI-compatible frontends (e.g., Open WebUI) to hang indefinitely waiting for the stream to end, even though the model has finished generating and `finish_reason: "stop"` has been sent in the final chunk.

## Behavior

- **Short/simple prompts**: `data: [DONE]` is sent correctly after `finish_reason: "stop"`
- **Longer prompts or reasoning models**: `data: [DONE]` is sometimes never sent, leaving the HTTP connection open indefinitely

The issue is intermittent, which suggests a race condition or a code path that skips the final `[DONE]` event under certain conditions.

## Reproduction

**Environment:**
- vllm-mlx v0.2.5
- macOS (Apple Silicon)

**Steps:**
1. Start vllm-mlx: `vllm-mlx serve <model> --port 8000 --reasoning-parser qwen3`
2. Send a streaming chat completion with a longer prompt:
```bash
  -d '{
    "model": "<model-id>",
    "messages": [{"role": "user", "content": "Describe this scene in detail"}],
    "stream": true
  }'
```
3. Observe that the final chunk with `finish_reason: "stop"` arrives, but `data: [DONE]` never follows
4. The curl/client hangs indefinitely

**Expected:** Every streaming response should end with `data: [DONE]\n\n` after the final chunk, per the OpenAI streaming specification.

**Actual:** `data: [DONE]` is sometimes omitted, causing clients to hang.

## Likely Cause

This may be related to the reasoning model streaming path. Models with `reasoning_content` fields (e.g., Qwen3 Thinking variants) have a two-phase output (reasoning then content), and the code path for finalizing the stream after reasoning may not consistently emit `[DONE]`. The issue seems more frequent with longer generations and reasoning-enabled models, though it can happen with standard models too.

## Workaround

A lightweight reverse proxy that injects `data: [DONE]\n\n` when the upstream connection closes without having sent it resolves the issue for downstream clients.

## OpenAI Spec Reference

From the [OpenAI API streaming docs](https://platform.openai.com/docs/api-reference/chat/create#chat-create-stream): the stream should be terminated with `data: [DONE]` after the final chunk.

Thump604:
## Summary

- Wraps the main streaming loops in `stream_completion()` and `stream_chat_completion()` with `try/except` so that `data: [DONE]\n\n` is **always** emitted, regardless of how the generator exits
- If an exception occurs during streaming (engine error, reasoning parser crash, tool parser failure), the generator previously exited without yielding the final `[DONE]` event
- OpenAI-compatible clients (Open WebUI, etc.) then hang indefinitely waiting for stream termination

### Root cause

The `[DONE]` yield is at the end of the generator function, after the main `async for` loop. Any unhandled exception in the loop body (engine, reasoning parser, tool parser, serialization) causes the generator to exit early, skipping `[DONE]`.

### Fix

Catch unexpected exceptions in both streaming generators, log them, and always reach the final `yield "data: [DONE]\n\n"`.

Closes #101

## Test plan
- [x] Existing tests pass
- [ ] Manual test: verify `[DONE]` arrives after normal completion
- [ ] Manual test: verify `[DONE]` arrives when engine encounters an error mid-stream

### DIFF
--- a/vllm_mlx/server.py
+++ b/vllm_mlx/server.py
@@ -1837,7 +1837,9 @@ async def stream_completion(
                     {
                         "index": 0,
                         "text": output.new_text,
-                        "finish_reason": output.finish_reason if output.finished else None,
+                        "finish_reason": (
+                            output.finish_reason if output.finished else None
+                        ),
                     }
                 ],
             }
@@ -1948,7 +1950,9 @@ async def stream_chat_completion(
                                 content=delta_msg.content,
                                 reasoning=delta_msg.reasoning,
                             ),
-                            finish_reason=output.finish_reason if output.finished else None,
+                            finish_reason=(
+                                output.finish_reason if output.finished else None
+                            ),
                         )
                     ],
                     usage=get_usage(output) if output.finished else None,
--- a/vllm_mlx/server.py
+++ b/vllm_mlx/server.py
@@ -1820,29 +1820,32 @@ async def stream_completion(
     request: CompletionRequest,
 ) -> AsyncIterator[str]:
     """Stream completion response."""
-    async for output in engine.stream_generate(
-        prompt=prompt,
-        max_tokens=request.max_tokens or _default_max_tokens,
-        temperature=_resolve_temperature(request.temperature),
-        top_p=_resolve_top_p(request.top_p),
-        stop=request.stop,
-    ):
-        data = {
-            "id": f"cmpl-{uuid.uuid4().hex[:8]}",
-            "object": "text_completion",
-            "created": int(time.time()),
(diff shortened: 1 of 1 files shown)


Thump604:
Closing — `data: [DONE]` SSE termination is already present in upstream main.

PR #168, opened 2026-03-16
Fix reasoning content lost when tool calls present

### LINKED ISSUE #161: Reasoning content leaks into content field when tool calls are present (Nemotron)
## Description

When Nemotron 3 Super generates tool calls with thinking enabled, the `<think>` content leaks into the `content` field instead of being separated into `reasoning_content`. The `qwen3` reasoning parser doesn't fully extract reasoning when tool calls are also present in the response.

## Reproduction

```bash
# Server: vllm-mlx serving Nemotron-3-Super-120B-A12B with:
#   --tool-call-parser nemotron --reasoning-parser qwen3 --enable-auto-tool-choice

curl http://127.0.0.1:8080/v1/chat/completions \
  -d '{
    "model": "nemotron",
    "messages": [{"role": "user", "content": "What is the weather in Paris?"}],
    "tools": [{"type": "function", "function": {"name": "get_weather", "description": "Get weather", "parameters": {"type": "object", "properties": {"city": {"type": "string"}}, "required": ["city"]}}}],
    "max_tokens": 200
  }'
```

## Expected

```json
{
  "content": null,
  "reasoning_content": "The user is asking for the current weather...",
  "tool_calls": [{"function": {"name": "get_weather", "arguments": "{\"city\": \"Paris\"}"}}]
}
```

## Actual

```json
{
  "content": "<think>The user is asking for the current weather in Paris. I have access to the get_weather function...</think>",
  "reasoning_content": null,
  "tool_calls": [{"function": {"name": "get_weather", "arguments": "{\"city\": \"Paris\"}"}}]
}
```

## Environment

- vllm-mlx 0.2.6
- mlx-lm 0.31.0 (with PRs #988 + #992 applied)
- Model: Nemotron-3-Super-120B-A12B (4.5-bit MLX)
- Mac Studio M2 Ultra 128GB

## Analysis

The `qwen3` reasoning parser correctly handles reasoning extraction for non-tool-call responses. The issue appears to be in the interaction between the reasoning parser and the tool call parser — when the tool call parser processes the response, it may consume the output before the reasoning parser has a chance to extract the `<think>` tags, or the two parsers don't coordinate on the same response text.

This also affects Qwen 3.5 models in MLLM mode when tool calls are present, so it's not Nemotron-specific.

Thump604:
## Summary

- Reasoning parser and tool call parser now work **sequentially** instead of being mutually exclusive
- **Non-streaming:** extract reasoning FIRST (separates `<think>...</think>`), then parse tool calls from the remaining content
- **Streaming:** 3-phase pipeline per chunk: (1) reasoning extraction → (2) content filtering → (3) tool call parsing on content portion

### Root cause

**Non-streaming path:** Tool parser ran first and `strip_think_tags()` discarded `<think>` content. Then the guard `if _reasoning_parser and not tool_calls` skipped reasoning extraction entirely — reasoning was silently lost.

**Streaming path:** Reasoning parser (lines 1935-1960) and tool parser (lines 1961-2036) were in mutually exclusive `if/else` branches. If reasoning parser was active, tool markup leaked into content. If tool parser was active, `<think>` tags leaked through.

### Before

```
Model output: "<think>I should call get_weather</think><tool_call>..."

Non-streaming → reasoning_content: null, content: "", tool_calls: [...]
Streaming    → reasoning_content: "I should call...<tool_call>...", tool_calls: null
```

### After

```
Non-streaming → reasoning_content: "I should call get_weather", content: null, tool_calls: [...]
Streaming    → reasoning chunk: "I should call get_weather", then tool_calls chunk: [...]
```

Closes #161

## Test plan
- [x] 900 existing tests pass (0 regressions)
- [ ] Manual test: Nemotron with tool calls — verify reasoning in reasoning_content field
- [ ] Manual test: Qwen 3.5 with tool calls — verify reasoning separated from tool calls
- [ ] Manual test: plain chat without tools — verify no regression

### DIFF
--- a/vllm_mlx/server.py
+++ b/vllm_mlx/server.py
@@ -1390,16 +1390,25 @@ async def create_chat_completion(request: ChatCompletionRequest, raw_request: Re
         f"Chat completion: {output.completion_tokens} tokens in {elapsed:.2f}s ({tokens_per_sec:.1f} tok/s)"
     )
 
-    # Parse tool calls from output using configured parser
-    cleaned_text, tool_calls = _parse_tool_calls_with_parser(output.text, request)
-
-    # Extract reasoning content FIRST (strips channel tokens before JSON extraction)
+    # Extract reasoning FIRST, then parse tool calls from the remaining content.
+    # Previously, tool parser ran first and strip_think_tags() discarded
+    # reasoning content.  Then the `not tool_calls` guard blocked the
+    # reasoning parser entirely, losing the <think> content.
     reasoning_text = None
-    if _reasoning_parser and not tool_calls:
-        text_to_parse = cleaned_text or output.text
-        reasoning_text, cleaned_text = _reasoning_parser.extract_reasoning(
-            text_to_parse
+    text_for_tools = output.text
+    if _reasoning_parser:
+        reasoning_text, text_for_tools = _reasoning_parser.extract_reasoning(
+            output.text
         )
+        # If reasoning parser returned None for content, fall back to
+        # full text so tool parser can still find tool calls.
+        if not text_for_tools:
+            text_for_tools = output.text
+
+    # Parse tool calls from content (after reasoning extraction)
+    cleaned_text, tool_calls = _parse_tool_calls_with_parser(
+        text_for_tools, request
+    )
 
     # Process response_format if specified (after reasoning parser cleaned the text)
     if response_format and not tool_calls:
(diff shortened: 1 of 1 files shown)


Thump604:
Closing — sequential reasoning + tool call parsing was incorporated upstream in main. The extract_reasoning path now handles coexistence correctly.

PR #177, opened 2026-03-18
fix: parse tool calls in streaming reasoning branch

Thump604:
Fix streaming when both reasoning (thinking) tags and tool calls appear in the same response.

**Before:** reasoning parser consumed `<think>` content but left artifacts that broke tool call parsing downstream.
**After:** reasoning extraction runs first, cleaned text passed to tool parser.

**Files:** `server.py` streaming handler

**Test:** Send a request with tools enabled and a model that uses `<think>` tags. Response should have both `reasoning_content` and `tool_calls` populated correctly.

### DIFF
--- a/vllm_mlx/server.py
+++ b/vllm_mlx/server.py
@@ -2033,9 +2033,7 @@ async def stream_chat_completion(
                                         tool_calls=tool_result["tool_calls"],
                                     ),
                                     finish_reason=(
-                                        "tool_calls"
-                                        if output.finished
-                                        else None
+                                        "tool_calls" if output.finished else None
                                     ),
                                 )
                             ],
@@ -2062,9 +2060,7 @@ async def stream_chat_completion(
                         finish_reason=(
                             "tool_calls"
                             if (output.finished and tool_calls_detected)
-                            else (
-                                output.finish_reason if output.finished else None
-                            )
+                            else (output.finish_reason if output.finished else None)
                         ),
                     )
                 ],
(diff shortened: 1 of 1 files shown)


Thump604:
Status ping — this PR has been idle 8 days and was the one Thump604 PR skipped from the 2026-04-07 coordinated rollup. No known blockers on our side: branch is rebased against current main (`6680b180`), CI is green, and the fix closes #161 (reasoning content leaks into tool_calls on the streaming path). Flagging for visibility.

Thump604:
I rechecked this in light of the new official Gemma 4 chat templates. My read is the same: that template update does **not** obsolete this streaming parser fix.

The template update changes how messages are rendered into the prompt. This PR is about streamed model output when reasoning blocks and tool calls appear together. Those are different layers of the stack.

I also revalidated the live Gemma lane after the template sync: tool calling, tool-result replay, and plain reasoning split are all still green. So I would keep this PR in the queue as a real fix candidate, but it probably wants to be restacked with the newer raw-output-preservation work rather than evaluated in isolation against an older server state.

janhilgard:
Hey @Thump604 — this fix is now covered in `main` via PR #278, which backported the full reasoning parser + tool parser streaming pipeline from production. The reasoning extraction runs first and passes cleaned text to the tool parser, so tool calls inside `<think>` responses work correctly.

This PR has merge conflicts with current `main` since `server.py` changed significantly. Safe to close as superseded if you agree.

Thump604:
Closing as superseded by #278. The merged reasoning-parser/tool-parser streaming pipeline on current main covers this path, and this branch now conflicts with the updated server changes.

PR #193, opened 2026-03-21
Fix MLLM cache stats in /v1/status

janhilgard:
## Summary

- Fix `/v1/status` reporting zero cache stats for MLLM models by replacing unused `MLLMCacheManager` with actual Metal KV cache memory and `VisionEmbeddingCache` hit/miss stats
- Pass `--cache-memory-mb` config through to `MLLMSchedulerConfig` so `max_memory_mb` reflects the configured limit
- Remove dead code: `MLLMCacheManager` was initialized but never used for fetch/save operations

## Test plan

- [ ] Start MLLM model with `--mllm --cache-memory-mb 2048`
- [ ] Send a text request, verify `/v1/status` shows `cache.current_memory_mb > 0` and `cache.max_memory_mb == 2048`
- [ ] Send an image request, verify `cache.hits` increments on repeated images
- [ ] Verify non-MLLM models (without `--mllm`) are unaffected

### DIFF
--- a/vllm_mlx/engine/batched.py
+++ b/vllm_mlx/engine/batched.py
@@ -219,12 +219,14 @@ class BatchedEngine(BaseEngine):
             self._scheduler_config, "completion_batch_size", 16
         )
 
+        cache_memory_mb = getattr(self._scheduler_config, "cache_memory_mb", None)
         mllm_config = MLLMSchedulerConfig(
             max_num_seqs=max_num_seqs,
             prefill_batch_size=prefill_batch_size,
             completion_batch_size=completion_batch_size,
             enable_vision_cache=True,
             vision_cache_size=100,
+            cache_memory_mb=cache_memory_mb,
         )
 
         # Create and start MLLM scheduler
@@ -783,8 +785,8 @@ class BatchedEngine(BaseEngine):
 
     def get_cache_stats(self) -> dict[str, Any] | None:
         """Get cache statistics."""
-        if self._mllm_scheduler and self._mllm_scheduler.vision_cache:
-            return self._mllm_scheduler.vision_cache.get_stats()
+        if self._mllm_scheduler and self._mllm_scheduler.batch_generator:
+            return self._mllm_scheduler.batch_generator.get_vision_cache_stats()
         elif self._engine:
             return self._engine.get_cache_stats()
         return None
--- a/vllm_mlx/mllm_scheduler.py
+++ b/vllm_mlx/mllm_scheduler.py
@@ -28,7 +28,6 @@ from collections import deque
 from dataclasses import dataclass, field
 from typing import Any, AsyncIterator, Dict, List, Optional, Set, Tuple
 
-
 from .mllm_batch_generator import (
     MLLMBatchGenerator,
     MLLMBatchRequest,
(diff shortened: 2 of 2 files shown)


Thump604 (commented):
Overall: This is a solid correctness fix. The changes properly surface cache stats for MLLM models in `/v1/status`. I have a few observations and one concern.

## Strengths

1. **Correct routing:** Switching from the unused `MLLMCacheManager` to `batch_generator.get_vision_cache_stats()` is the right call. The vision_cache lives in the batch_generator, so this is the actual source of truth.

2. **Config threading:** Passing `cache_memory_mb` through to `MLLMSchedulerConfig` is necessary for the stats to report meaningful limits. This matches the intent of `--cache-memory-mb` CLI flag.

3. **Dead code removal:** Eliminating the orphaned `MLLMCacheManager` initialization (lines 178-182 original, never used for fetch/save) reduces confusion and maintenance burden.

4. **Comprehensive stats:** The new `memory_aware_cache` dict includes hit_rate calculation, entry_count aggregation across pixel and encoding caches, and memory_utilization — all useful for monitoring.

## Issues & Questions

1. **KV cache memory always 0:** The comment says "current_memory_mb=0 until prefix cache is implemented" — this is honest but incomplete. For MLLM models, KV cache memory isn't tracked at all (only vision embedding caches are). On text-only models without prefix cache, KV memory does exist. Should MLLM models report actual KV cache usage via MLX Metal stats (`mx.get_cache_memory()`)? The Metal stats are already captured above (metal_cache_memory_gb). Might want to convert that to MB and populate current_memory_mb instead of leaving it hardcoded to 0.

2. **Hit rate edge case:** The hit_rate calculation handles `total == 0` with a fallback to 0.0, which is correct. But the variable `total = hits + misses` could be made more explicit as a separate line for clarity (currently it's in the hits/misses aggregation).

3. **Reset cleanup:** The PR removes `self.vision_cache.clear()` in `reset()` — correct, since vision_cache now lives in batch_generator and is managed there. But verify that `batch_generator.close()` actually clears the vision cache when called (I assume it does based on the pattern).

4. **Floating-point precision:** `max_memory_mb = float(self.config.cache_memory_mb or 0)` — casting to float is fine, but later you use it in division. If cache_memory_mb is None or 0, you already guard with `if max_mb > 0`, so no division-by-zero risk. Good.

## Test Plan Validation

The test plan is sound:
- Text request → verify Metal KV cache shows in stats ✓
- Image request → verify vision cache hits increment ✓
- Non-MLLM unaffected ✓

One gap: doesn't test the `--cache-memory-mb` config actually flowing through to `cache_memory_mb` in stats. The first bullet ("verify... cache.max_memory_mb == 2048") will catch this, but worth calling out explicitly.

## Recommendation

Approve with a request to address point #1: should MLLM models report actual KV memory from MLX Metal stats into `current_memory_mb`, or is the 0 by design (pending prefix cache for true KV tracking)? If it's by design, a clearer comment explaining why MLLM KV is excluded would help future reviewers.

The changes are correct as-is, but that one design decision deserves a brief explanation in code or commit message.

janhilgard:
@Thump604 This PR is now superseded — all fixes were merged into `main` via #278:

- ✅ `MLLMCacheManager` removed from `mllm_scheduler.py`
- ✅ `cache_memory_mb` added to `MLLMSchedulerConfig`
- ✅ `memory_aware_cache` dict built in `get_stats()` (uses `get_prefix_cache_stats()`)
- ✅ `get_cache_stats()` in `engine/batched.py` uses `batch_generator`

Closing as superseded.

Issue #191, opened 2026-03-21
feat: Multi-model management with on-demand loading and LRU eviction


sysit:
## Problem

On Apple Silicon machines with large unified memory (e.g., 128GB+ Mac Studio), running a single model underutilizes available resources. Users with multiple models want to serve them concurrently without spawning multiple server processes on different ports.

Current limitations:
- `vllm-mlx serve` only loads one model at startup
- No way to dynamically load/unload models based on demand
- Memory sits idle when only one model is loaded
- Users must manage multiple processes manually for multi-model serving

## Proposed Solution

Add a **multi-model manager** with:

### 1. Model Registry
- Configure multiple models via config file or CLI
- Support model aliases (e.g., "fast" → "Qwen3-8B", "smart" → "Qwen3-72B")

### 2. On-Demand Loading
- Load models lazily when first request arrives
- Support pre-loading (warm start) for critical models
- Track memory usage per model

### 3. LRU Eviction
- Evict least-recently-used models when memory threshold exceeded
- Configurable max memory budget
- Graceful eviction: finish pending requests before unloading

### 4. API Compatibility
- `GET /v1/models` returns all configured models with status (loaded/unloaded)
- Requests with `model` parameter trigger auto-loading if not in memory
- Consistent with OpenAI API behavior

## Example Usage

```bash
# Config file: models.yaml
models:
  - name: qwen-35b
    path: ~/models/Qwen3.5-35B-A3B
    alias: smart
    preload: true
  - name: qwen-8b
    path: ~/models/Qwen3-8B
    alias: fast
  - name: llama-70b
    path: mlx-community/Llama-3.3-70B-Instruct-4bit

# Start server with multi-model support
vllm-mlx serve --config models.yaml --max-memory 100GB
```

## Benefits
- **Efficiency**: Utilize large unified memory effectively
- **Simplicity**: Single server process, single port
- **Compatibility**: Works with existing OpenAI-compatible clients
- **Cost-effective**: No need for multiple GPU instances

## Related
- #102 (model switching) - This extends that to full multi-model management

## Implementation Notes
This would require:
1. A `ModelManager` class to track loaded models
2. Memory tracking and eviction logic
3. Request routing to appropriate model
4. Graceful loading/unloading with request queuing

Happy to contribute if this aligns with the project roadmap!

PR #205, opened 2026-03-22
feat: add lifecycle-managed residency for the default server model

lyonsno:
## Summary

Adds lifecycle-managed residency for the server's default model: lazy load on first request, idle auto-unload, and safe on-demand reload.

The motivation is memory-aware serving. When the model footprint is too large to keep everything resident indefinitely, the server should be able to evict an idle model and bring it back later without wedging request handling or leaking partial runtime state.

This overlaps the on-demand loading / idle-unload problem space in #191 and #199, but takes a narrower approach: lifecycle-manage the default server model first, with an emphasis on request/load/unload correctness rather than broader multi-model UX. The author of #199 closed that PR in favor of this approach, noting plans to re-implement multi-model management on top of this lifecycle foundation.

> **Reviewability note:** ~4,800 of the ~6,500 added lines are tests (`test_lifecycle.py`). The production diff is ~1,700 lines across 11 files.

## Scope

This PR is intentionally narrower than general multi-model management.

In scope:
- lifecycle-managed residency for the default server model
- lazy startup for the main model (`--lazy-load-model`)
- idle auto-unload via `--auto-unload-idle-seconds`
- safe on-demand reload for later requests
- request-scoped acquire/release when lifecycle mode is enabled
- hardening around cold-load timeout, disconnect, cancellation, startup, and shutdown
- public status/health handling for lifecycle-managed models

Out of scope:
- arbitrary in-process reconfiguration after FastAPI lifespan startup
- a tokenizer-only implementation for `/v1/messages/count_tokens`
- broader multi-model registration/discovery UX
- full served-model-name semantics beyond the residency/public-surface cases touched here
- guaranteed teardown of all partially prepared engine state for every ordinary non-cancellation startup failure path

Where behavior is intentionally not covered, I tried to either document it or make the unsupported boundary explicit rather than leave ambiguous partial support behind.

## What changed

At a high level, this PR:

- adds a `ResidencyManager` with explicit resident states (`unloaded`, `loading`, `loaded`, `unloading`, `failed`)
- allows `load_model()` to register a resident model without assuming a permanently live engine
- eagerly loads at startup when `lazy_load_model=False`, while still allowing later idle unload if configured
- routes request handlers through residency-aware acquire/release instead of assuming `_engine` is always live
- single-flights cold loads and unwinds abandoned ones safely
- moves blocking startup/cache I/O off the serving event loop where needed
- invalidates parser/tokenizer-derived tool-call state when the underlying engine changes
- hardens startup/shutdown cleanup and eager-start cancellation behavior
- sanitizes public residency status so `/v1/status` does not expose the private backing model path or raw internal reload errors

## Behavioral notes

A few behaviors worth calling out explicitly:

- `/v1/messages` now counts as residency activity, like other inference requests.
- `/v1/messages/count_tokens` remains separate from generation and is still not a tokenizer-only cold path in this PR.
- Runtime reconfiguration after lifespan startup is treated as unsupported rather than partially working.
- Unloaded `/health` model-type fallback now uses the real model path rather than the served alias, so a generic `--served-model-name` does not hide multimodal model type.
- `/v1/status` residency data is sanitized for public consumption rather than returning the raw resident snapshot verbatim.

## Tests

79 tests covering:

- lazy startup and idle unload
- cold-load timeout/disconnect handling
- cancellation races during load/unload
- parser invalidation across unload/reload
- startup/shutdown cleanup failure handling
- eager-start cancellation semantics
- CLI/server forwarding for residency-related configuration
- public status/health residency behavior
- Anthropic request accounting under residency

```bash
PYTHONPATH=. UV_CACHE_DIR=/tmp/uv-cache uv run --no-sync --with pytest python -m pytest tests/test_lifecycle.py -q
PYTHONPATH=. UV_CACHE_DIR=/tmp/uv-cache uv run --no-sync --with pytest python -m pytest tests/test_server.py -q

# tests/test_lifecycle.py: 79 passed
# tests/test_server.py: 34 passed, 3 deselected
```

## Feedback I'd especially appreciate

1. Whether the current `/v1/status` residency redaction level is the right public contract
2. Whether `/v1/messages/count_tokens` should remain a documented limitation for now or should block the feature until it has a tokenizer-only path
3. Whether this is the right abstraction layer for upstream — designed as the one-resident case of future multi-model residency, with `model_key` routing already in place

## Why this is still a draft

The branch is in good shape technically, but I'd rather get alignment on scope and contract before marking it ready. Happy to adjust boundaries based on maintainer feedback.


Thump604 (approved):
I reviewed the lifecycle residency PR. The implementation is solid and properly scoped.

## Strengths

1. **ResidencyManager design is sound.** Single-flight load/unload with proper acquire/release counting, cancellation safety via suspend_cancellation(), and explicit state machine (UNLOADED → LOADING → LOADED → UNLOADING). The waiters refcount pattern prevents abandoned solo loads from blocking future requests.

2. **Test coverage is excellent.** 79 tests in test_lifecycle.py covering cold-load timeout, disconnection races, cancellation during load/unload, parser invalidation across swaps, and startup cleanup. Tests are fail-first and deterministic.

3. **Cancellation hardening is well thought out.** The `suspend_cancellation()` context manager (via task.uncancel()) ensures blocking cache I/O finishes before propagating cancellation. The lifecycle loop shields unload to prevent half-stopped engines orphaned in UNLOADING state.

4. **Public API surface is sanitized.** `/v1/status` and `/health` do not leak raw exception text or private model paths. Residency state is serialized safely for unauthenticated endpoints.

5. **Tool parser cache invalidation is correct.** Setting _tool_parser_instance=None on resident swap forces a rebuild on the next request. Prevents stale tokenizer state when the underlying engine changes.

6. **Config surface is backward compatible.** New flags (--lazy-load-model, --auto-unload-idle-seconds) default to eager startup and no idle unload, preserving existing behavior.

## Design notes I appreciate

- Lazy event loop binding for _idle_unload_enabled via _get_idle_unload_event() avoids binding to the wrong loop at import time.
- prefix_cache load/save runs off-thread via asyncio.to_thread() so startup does not block the event loop.
- register_model() rejects live resident replacement, enforcing the safe invariant that config cannot change mid-flight.
- acquire()/release() include count_activity flag, allowing introspection and status endpoints to avoid updating last_used_at.

## Concerns

1. **/v1/messages/count_tokens limitation is documented but still exposed.** The PR notes count_tokens may wake a lazy model. For consistency with lazy semantics, consider either (a) tokenizer-only path or (b) mark it as explicitly not lazy-safe. Current state leaves ambiguity for clients.

2. **Eager startup error handling is out of scope.** The PR hardenes cancellation but not ordinary non-cancellation startup failures in SimpleEngine.start()/BatchedEngine.start(). Should SimpleEngine.start() fail, the partially prepared state is not guaranteed torn down before surfacing the error. Not blocking the feature but worth a follow-up.

3. **served_model_name integration deferred.** The PR notes upstream served_model_name support is not yet integrated with residency. The health fallback uses _model_path for multimodal detection when engine is unloaded, which is correct, but the broader served_model_name + residency interaction (cache-dir keying, /v1/models, etc.) is still TODO.

4. **Idle-unload loop timing is coarse.** The lifecycle loop samples idle state every min(auto_unload_idle_seconds/2, 5.0) seconds. For a 300-second threshold, that's ~150s sample interval. Not wrong, but if precision matters, consider event-based wakeup instead of polling.

## Questions for the maintainers

1. Is the current /v1/status residency redaction level the right public contract, or should it expose more detailed lifecycle state (e.g., estimated_memory_bytes)?

2. Should count_tokens remain documented as a limitation, or is there appetite for a tokenizer-only fast path in a follow-up?

This is solid work. The code is clean, the tests are comprehensive, and the scope is appropriately narrow. Ready to merge from a technical standpoint.

Thump604:
Hey @lyonsno — reviewed this and the implementation is solid. 79 lifecycle tests, clean cancellation safety, backward-compatible with everything behind opt-in flags. It's currently conflicting with main though. Would you be able to rebase? Happy to merge once it's clean.

lyonsno:
Ah, yeah, no problem, I'll get right on it.

Thump604:
Hey @lyonsno — just checking in. The rebase is still needed to merge this. Let me know if you hit conflicts and need a hand.

janhilgard (commented):
## Code Review — lifecycle-managed residency

This is a substantial and well-engineered PR. The `ResidencyManager` with explicit state machine (`UNLOADED -> LOADING -> LOADED -> UNLOADING`) and single-flighted cold loads is the right architecture for idle-unload on Apple Silicon where memory pressure is the primary constraint.

### Strengths

1. **Cancellation safety is excellent.** The waiter refcount pattern (`_waiters` count) ensures that cancelling one waiter doesn't kill a shared cold load that other waiters need, while cancelling the *last* waiter correctly unwinds the in-flight load. The `suspend_cancellation()` pattern (or its equivalent) during state commits prevents torn state. The tests covering these edge cases (`test_cancelled_waiter_does_not_cancel_shared_cold_load`, `test_last_cancelled_waiter_cancels_cold_load`, `test_cancelled_shared_load_during_state_commit_recovers_cleanly`) are thorough and correctly verify the invariants.

2. **Request-scoped acquire/release is the right abstraction.** Rather than assuming `_engine` is always live, request handlers now go through `_acquire_default_engine_for_request` / `_release_default_engine`, which integrates disconnect detection and timeout budgeting. This means a model can be safely evicted between requests without breaking in-flight handling.

3. **Public status sanitization is thoughtful.** `/v1/status` and `/health` expose `residency_state` and `active_requests` without leaking the raw model path or internal error messages. The `model_load_failed` generic error indicator is the right trade-off between debuggability and security.

4. **Test coverage is exceptional.** 79 tests covering cold-load single-flighting, idle-unload with active requests, cancellation races, shutdown semantics, CLI forwarding, status endpoint races, tool parser invalidation, and startup failure cleanup. The `restore_server_globals` fixture is well-designed for lifecycle testing.

### Observations

**1. Rebase needed (blocking)**

The PR is in `CONFLICTING` state against main. Thump604 already noted this. Several areas have likely diverged: the server module globals, `load_model()` signature, and possibly `cli.py` arguments.

**2. Test file size**

`test_lifecycle.py` is 5,377 lines. While the coverage is excellent, this is a very large test file. Consider splitting into logical groups (e.g., `test_lifecycle_manager.py` for `ResidencyManager` unit tests, `test_lifecycle_server.py` for server integration tests, `test_lifecycle_cli.py` for CLI tests). This would make CI parallelization more effective and reduce the cognitive load when debugging failures.

**3. `_engine_factory` callback pattern**

The `engine_factory` callback injected into `ResidencyManager` is clean for testing but means the production wiring (in `load_model()` or `serve_command()`) needs to close over all the configuration needed to construct an engine. If the configuration surface grows (e.g., per-model scheduler configs for multi-model), this closure approach may become unwieldy. The `ModelSpec` dataclass already carries most of this, which is good forward-looking design.

**4. `prepare_for_start()` threading boundary**

The blocking `prepare_for_start()` runs in a thread pool to avoid blocking the event loop, which is correct. The cancellation tests verify that cancellation during `prepare_for_start()` is handled properly. However, the interaction between thread pool cancellation and asyncio task cancellation is inherently tricky — if `prepare_for_start()` completes just after the task is cancelled, the engine has loaded model weights into memory that need to be freed via `stop()`. The tests verify this path works, which is good.

**5. `_idle_unload_enabled` as lazily-created `asyncio.Event`**

The test fixture correctly resets this to `None` between tests, but this pattern means the Event is bound to whatever event loop was running at creation time. If the server is used with multiple event loops (unlikely in production, but possible in testing), this could cause issues. The fixture handles it, but a comment in the production code would help.

### Summary

This is a well-designed feature with exceptional test coverage and careful attention to the hard problems (cancellation, state machine correctness, resource cleanup). The main action item is the rebase. The test file size is worth addressing for maintainability but is not blocking. Ready to merge once conflicts are resolved.

lyonsno:
Thanks for the detailed review, @janhilgard — really appreciate the depth here, especially the observations on the cancellation/thread boundary interactions and the Event loop binding.

A few updates since you reviewed:

**Rebase is done.** Resolved all conflicts (19 conflict regions in server.py alone — upstream moved a lot in two weeks). The merge was mechanical — metrics, sampling params, download preflight, tool argument coercion all landed upstream while this was open, but they're orthogonal to the lifecycle work.

**Found and fixed a bug during our own review pass.** The streaming path in `create_completion` wasn't using the same `release_on_exit` flag pattern as `create_chat_completion` and the Anthropic handler — if `StreamingResponse` setup threw before the generator started, the engine lease would leak. Fixed in `266ab2d`.

**Also fixed:** tool parser initialization now seeds from the request-local engine instead of the global `_engine` (`eb15bc1`), which closes the stale-tokenizer window during unload/reload cycles.

**Test file split.** Took your suggestion and split `test_lifecycle.py` into three files:
- `test_lifecycle_manager.py` (29 tests) — ResidencyManager unit tests
- `test_lifecycle_server.py` (72 tests) — server integration
- `test_lifecycle_cli.py` (7 tests) — CLI forwarding

Also migrated all async test markers to `anyio` to match the project convention.

On your observation about the lazy `asyncio.Event` for `_idle_unload_enabled` — good call on the loop-binding subtlety. Added a comment in the production code documenting the caveat and the reset requirement for test fixtures (`0c10f8b`).

Total test count is 122 across the lifecycle and media routing suites, all green. Should be ready for another look.

lyonsno:
Lint fix is pushed (2ed8fe6) — the two unused imports flagged by ruff are removed. Latest CI run is waiting on first-time contributor workflow approval. Could you approve the run from the Actions tab when you get a chance?

Thump604:
Approved the workflow run. Ruff passes now but black is flagging 9 files that need reformatting:

- tests/test_async_markers.py
- tests/test_lifecycle_cli.py
- tests/test_lifecycle_manager.py
- tests/test_lifecycle_server.py
- vllm_mlx/api/utils.py
- vllm_mlx/lifecycle.py
- vllm_mlx/engine/batched.py
- vllm_mlx/engine/simple.py
- vllm_mlx/server.py

A quick `black vllm_mlx/ tests/` should fix it. The test and type-check jobs all pass.

lyonsno:
Alright thanks, I’ve got that in now.

lyonsno:
CI passed, ready to merge whenever.

lyonsno:
Rebased again onto current `main`; PR #205 is now at `f2542eb` and GitHub shows it as mergeable again.

If the current scope still looks good, could one of you kick/rerun the checks if needed and merge once green? If there’s a specific blocker left on the maintainer side, let me know and I can address it.

Thump604:
I rechecked this again.

Current state:

- head: `17902ed82db8731799829daed00c03388361743a`
- review decision: `APPROVED`
- checks: green
- GitHub mergeability: `UNKNOWN`
- PR base: `40bd173fa4815fa6f526e20ccfe0e659bf8b584c`

`main` has moved since that base, and this is a large server/lifecycle PR. I do not want to force-merge it while GitHub cannot prove it is clean against current `main`.

Could you rebase once more onto current `main`? Once GitHub reports a clean merge state again, I can do the final recheck and merge if nothing else moves.

PR #173, opened 2026-03-17
fix: honor tool_choice=none by stripping tools and suppressing parsing

Thump604:
When `tool_choice="none"`, strip tools from the chat template and suppress tool call parsing.

**Before:** tools still appeared in the template and the parser could match output as tool calls.
**After:** tools removed from template context, parser skipped.

**Files:** `server.py` (2 lines in non-streaming, 2 lines in streaming)

**Test:**
```bash
curl http://localhost:8080/v1/chat/completions -d '{
  "messages": [{"role":"user","content":"What is 2+2?"}],
  "tools": [{"type":"function","function":{"name":"calc","parameters":{}}}],
  "tool_choice": "none",
  "max_tokens": 20
}'
# Expect: content with answer, tool_calls: null
```

PR #210, opened 2026-03-23
fix: respect tool_choice="none" by excluding tools from template

### LINKED ISSUE #162: `tool_choice: "none"` ignored — model generates `tool_calls` with empty `{}` arguments when no tools defined

awanawana:
## Summary

Fixes #162

When `tool_choice` is set to `"none"`, models like Qwen2.5-Instruct and Llama 3.x-Instruct still activate their tool-calling behavior if tools are present in the chat template context. This causes:
- `finish_reason: "tool_calls"` instead of `"stop"`
- `content: null` with tool_calls containing empty `{}` arguments
- Model encoding output data as function names instead of text content

## Root Cause

The chat completions handler was passing `tools` to `chat_kwargs` without checking `tool_choice`. Even with `tool_choice="none"`, if tools were present in the request, they would be passed to `apply_chat_template()`. Qwen2.5 and Llama 3.x have tool-calling jinja templates that activate when a `tools` key is present — even an empty list triggers this behavior.

## Solution

Before adding tools to `chat_kwargs`, check if `tool_choice == "none"`. If so, skip adding tools entirely. This prevents the template from activating tool-calling mode.

```python
# Before
if request.tools:
    chat_kwargs["tools"] = convert_tools_for_template(request.tools)

# After
if request.tools and request.tool_choice != "none":
    chat_kwargs["tools"] = convert_tools_for_template(request.tools)
```

This matches the behavior of upstream vLLM's `--exclude-tools-when-tool-choice-none` flag.

## Test Plan

```bash
# Start server
vllm-mlx serve mlx-community/Qwen2.5-14B-Instruct-4bit --port 8080

# Test with tool_choice="none"
curl -X POST http://localhost:8080/v1/chat/completions \
  -d '{
    "model": "mlx-community/Qwen2.5-14B-Instruct-4bit",
    "messages": [{"role": "user", "content": "List 3 fruits as a JSON array"}],
    "tool_choice": "none",
    "tools": [],
    "max_tokens": 200
  }'

# Expected: finish_reason="stop", content contains JSON array, no tool_calls
# Before fix: finish_reason="tool_calls", content=null
```

### DIFF
--- a/vllm_mlx/server.py
+++ b/vllm_mlx/server.py
@@ -1421,8 +1421,11 @@ async def create_chat_completion(request: ChatCompletionRequest, raw_request: Re
     if request.specprefill_keep_pct is not None:
         chat_kwargs["specprefill_keep_pct"] = request.specprefill_keep_pct
 
-    # Add tools if provided
-    if request.tools:
+    # Add tools if provided (but skip if tool_choice="none")
+    # When tool_choice="none", models like Qwen2.5 and Llama 3.x still activate
+    # tool-calling if tools are present in the template context. To properly
+    # disable tool calling, we must not pass tools to the template at all.
+    if request.tools and request.tool_choice != "none":
         chat_kwargs["tools"] = convert_tools_for_template(request.tools)
 
     if request.stream:


Thump604:
@awanawana, thanks for raising this. `tool_choice="none"` being silently ignored is a real bug worth fixing.

I have #173 open since Mar 17 that addresses the same issue with broader coverage:

- All four code paths: OpenAI non-streaming, OpenAI streaming, Anthropic non-streaming, Anthropic streaming
- Parser early-return in `_parse_tool_calls_with_parser` so tool calls are also suppressed at parse time (belt-and-suspenders for models that might generate tool-call-like text without seeing tools in their template)
- New test file with four cases: positive (none → suppressed), default (auto → not suppressed), absent (None → not suppressed), and Qwen bracket-style markup

#173 has been formally approved by @janhilgard and is mergeable on current main.

Would you be willing to close this PR in favor of #173 once it merges? Happy to credit your fix in the merge discussion. If you prefer to keep this one open instead, no issue, Wayne will pick whichever lands first.

Thump604:
Closing as superseded by #173, which is now merged into main. The merged fix covers both sides of the bug: it strips tools from the template when tool_choice="none" and also suppresses tool-call parsing on the response path, so we do not regress into parser-side false positives.

PR #215, opened 2026-03-24
tokenizer: return successful mlx-lm load result

krystophny:
## Summary
- fix `load_model_with_fallback()` so a successful `mlx_lm.load()` returns `(model, tokenizer)` instead of falling through as `None`
- add a regression test for the successful return path
- run the tokenizer regression in Apple Silicon CI

## Status
- refreshed onto current upstream `main` (`b4fa030`) on 2026-04-09
- no logic changes beyond the base refresh

## Main files
- `vllm_mlx/utils/tokenizer.py`
- `tests/test_tokenizer_utils.py`
- `.github/workflows/ci.yml`

## Reviewer focus
Small user-visible startup fix. This overlaps with `#243`; this PR keeps the regression test with the fix.

## Validation
- `python -m pytest tests/test_tokenizer_utils.py -q` -> `2 passed`

PR #217, opened 2026-03-24
prefix_cache: preserve hybrid recurrent state across blocks

krystophny:
## Summary
- fix hybrid prefix-cache reconstruction for Qwen3.5-style models by storing recurrent state as sequence-boundary snapshots while keeping block concatenation for KV cache layers
- add deduplication safety test confirming isolated recurrent state per sequence
- remove duplicate tokenizer hunk that ships separately in PR `#215`

## Status
- no logic changes beyond the base refresh

## Known limitation
Correctness-preserving recurrent reuse is exact-boundary only. Shared KV blocks remain reusable, but arbitrary partial-boundary reuse for recurrent state is intentionally rejected unless an exact recurrent snapshot exists at that boundary.

## Main files
- `vllm_mlx/prefix_cache.py`
- `tests/test_paged_cache.py`

## Reviewer focus
The key logic is how non-KV recurrent state is preserved across block boundaries. Non-concatenatable layers (like `ArraysCache`) are stored as terminal sequence snapshots, while KV layers keep block concatenation semantics.

## Validation
- `python -m pytest tests/test_paged_cache.py -q` -> `40 passed`


Vigilans:
Looking at the PRs and discussions, it seems like the recurrent state (ArraysCache) is only snapshotted at sequence end and stored on the terminal block. Does this mean prefix cache reuse for hybrid models is more like sequence-level all-or-nothing?

E.g. if two requests share part of system prompt but diverge after that prefix, the KV blocks for the shared prefix are reusable but the recurrent snapshot doesn't exist for that partial block set, so the guard will reject the entire match?

Thump604:
Vigilans raises a good point. Running Qwen3.5-122B (12 full-attention + 36 GatedDeltaNet) in production, the practical answer is yes, it's effectively all-or-nothing for the recurrent portion.

KV blocks for attention layers are independently addressable at any token boundary. But ArraysCache (GatedDeltaNet's conv/SSM state) is a fixed-size running state updated with each token. There's no way to "slice" it at an arbitrary prefix boundary without having explicitly snapshotted it there.

In my production setup I work around this by snapshotting the full cache state (both KV and recurrent) after system prompt prefill. This gives exact-match reuse for the system prefix across requests with the same system prompt. But partial prefix reuse (shared first N tokens of different system prompts) isn't feasible because the recurrent snapshot only exists at the one boundary I chose to snapshot.

The recurrent state does decay exponentially, so in theory you could reconstruct an approximate state by replaying from the nearest snapshot. But for correctness-sensitive inference this isn't great. The all-or-nothing constraint seems inherent to hybrid architectures until someone finds a way to checkpoint recurrent state cheaply at block boundaries.

Thump604:
@waybarrios, @krystophny: brief endorsement plus coordination note.

Storing recurrent state as sequence-boundary snapshots while keeping block concatenation for KV layers is the right approach for hybrid Qwen3.5-style models. The deduplication safety test confirms isolated per-sequence state.

Coordination note: this PR fixes prefix-cache for hybrid models at the prefix-cache layer. PR #241 (Thump604) fixes continuous batching for the same hybrid model class at the MLLM batch generator layer (`_make_batch_cache` rewrite for ArraysCache support). The two are complementary. Different layers, both needed for hybrid models to work end-to-end. Both currently MERGEABLE on main.

Likely also addresses or is closely related to issue #142 (prefix cache 3D tensor slicing for Qwen3.5 MoE) and issue #136 (prefix cache fails on Qwen3.5-35B-A3B).

krystophny:
Force-pushed a refresh onto current upstream `main` (`b4fa030`) and updated the body with the recurrent-state limitation explicitly. Re the open question: yes, for the recurrent portion this is intentionally boundary-scoped. Shared KV blocks are still reusable, but we only reuse hybrid prefixes when we can recover the exact recurrent terminal snapshot for that boundary. Validation: `python -m pytest tests/test_paged_cache.py -q` -> `40 passed`.

Thump604 (approved):
Refresh confirmed on head 32d235a0 against upstream main b4fa030. The boundary-scoped answer is the right call.

Running Qwen3.5-122B-A10B (12 full-attention + 36 GatedDeltaNet) in production matches the same picture: the ArraysCache carries fixed-size running state with no sliceable offset, so reusing it at any position other than an explicit snapshot boundary would give a different model from what the prefix actually represents. KV blocks stay independently addressable at any shared prefix, and the hybrid cache only reuses when the terminal recurrent snapshot exists for the boundary in question. That is the only correctness-preserving option, and the deduplication safety test exercises the isolation that makes it safe.

CI green across lint, type-check, test-matrix 3.10-3.12, test-apple-silicon, and tests on head 32d235a0. Approved.

PR #221, opened 2026-03-24
scheduler: preserve prompt checkpoints in chunked prefill resume path

krystophny:
## Summary
Preserve `prompt_checkpoints` across chunked-prefill partial resume and finalization for `mlx-lm` prompt tuples, and invoke the upstream `prompt_checkpoint_callback` contract.

## Why
Recent `mlx-lm` prompt tuples carry a seventh `prompt_checkpoints` field. Upstream one-line fixes cover the immediate unpack crash in `_chunked_next`, but the chunked-prefill monkeypatch also needs to retain that field when it stores partial progress and resumes generation later.

Additionally, the upstream `BatchGenerator` invokes `prompt_checkpoint_callback` after cache finalization. The chunked-prefill monkeypatch was missing this callback invocation, breaking the checkpoint contract.

## What changed
- update the chunked-prefill scheduler monkeypatch to accept 7-field prompt tuples
- preserve prompt-checkpoint state through partial-prefill resume and finalization
- import `_lazy_extract_cache` from mlx-lm and invoke `prompt_checkpoint_callback` at the correct semantic boundary (after `c.finalize()`, before the checkpoint tail model call)
- add regression coverage for partial chunked-prefill with prompt checkpoints, callback firing, and `prompt_checkpoint > 1` tail replay

## Status
- added inline clarification for the upstream `prompt_checkpoint` normalization and the post-`finalize()` checkpoint-tail replay
- added a regression that exercises the `prompt_checkpoint > 1` tail replay path the review called out

## Files to review
- `vllm_mlx/scheduler.py`
- `tests/test_batching.py`

## Validation
- `python -m pytest tests/test_batching.py::TestSchedulerBasic -q` -> `10 passed`

### DIFF
--- a/tests/test_batching.py
+++ b/tests/test_batching.py
@@ -283,6 +283,101 @@ class TestSchedulerBasic:
         assert batch_gen._partial["prompt_checkpoint"] == 3
         assert batch_gen._partial["processed"] == 2
 
+    def test_chunked_prefill_invokes_checkpoint_callback(self, monkeypatch):
+        """prompt_checkpoint_callback must fire after finalization."""
+
+        class FakeCacheEntry:
+            def empty(self):
+                return True
+
+        class FakePromptCache:
+            def __init__(self):
+                self.state = mx.array([0])
+
+            def finalize(self):
+                return None
+
+            def extract(self, idx):
+                return self
+
+        class FakeStats:
+            prompt_tokens = 0
+            prompt_time = 0.0
+            generation_time = 0.0
+            generation_tokens = 0
+
+        callback_payloads = []
+
+        from collections import namedtuple
+        _Response = namedtuple("Response", ["uid", "token", "logprobs", "finish_reason", "cache"])
+
+        class FakeBatchGenerator:
+            Response = _Response
+
+            def __init__(self):
+                self._stats = FakeStats()
+                self._partial = None
+                self.active_batch = None
+                self.unprocessed_prompts = [
+                    (
+                        7,
+                        [1, 2, 3],
+                        16,
+                        [FakeCacheEntry()],
+                        None,
+                        [None],
+                        2,
+                    )
+                ]
+                self.prefill_batch_size = 1
+                self.completion_batch_size = 1
+                self.max_kv_size = None
+                self.stop_tokens = set()
+                self.prompt_progress_callback = lambda _progress: None
+                self.prompt_checkpoint_callback = (
+                    lambda entries: callback_payloads.extend(entries)
+                )
+                self._next = lambda: []
+                self.remove = lambda _uids: None
+                self._process_prompts = lambda _prompts: None
+                self.model = lambda _inputs, cache=None: None
+
+            def _step(self, inputs, cache, samplers, logits_processors, tokens):
+                return mx.array([99]), mx.array([-1.0])
+
+            def _generation_step(self):
+                if self.active_batch is not None:
+                    self.active_batch = None
+                return []
+
+        monkeypatch.setattr(
+            mlx_generate,
+            "_left_pad_prompts",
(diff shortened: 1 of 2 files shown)


Thump604 (commented):
The PR correctly addresses the callback gap that was flagged. The fix preserves prompt_checkpoints through partial resume and invokes the callback after finalization, which the prior submissions missed. The semantic changes to checkpoint-aware bounds checking are sound.

However, I have three concrete concerns before this merges:

1. Test coverage is incomplete. test_chunked_prefill_accepts_prompt_checkpoints constructs a 7-tuple but never exercises the actual zip(*batch_prompts) unpack on line 414. test_chunked_prefill_invokes_checkpoint_callback doesn't validate the "checkpoint tail" model execution (lines 313-318), which is the most complex part of the fix. A test with checkpoint > 1 initially (not just in resume) would strengthen confidence.

2. Line 425 has a suspicious defensive formula: `(ln - pc if pc > 0 else -pc)`. Checkpoint offsets should always be positive. Why the negation when pc <= 0? This deserves a comment explaining the intent — is this dead code, or does mlx-lm ever produce zero/negative checkpoints?

3. The checkpoint tail replay (lines 313-318) feeds tokens 0..checkpoint-1 through the model after finalize(). Is this the intended semantics? The comment on line 303 says "once only the checkpoint tail remains" but then we immediately replay it. If the cache already saw these tokens during chunked processing, feeding them again seems redundant. I need clarification: does finalize() clear cache state, or just mark it finalized? If it clears, the replay is necessary; if not, it may be redundant computation.

The implementation is not blocking — all three can be addressed post-merge. But they should be addressed before production use on 122B at 2048-token prefill steps. Otherwise this is good work fixing a real checkpoint contract violation.

Thump604:
@waybarrios, @krystophny: status note.

PR currently shows CONFLICTING merge status. Likely needs a rebase on current main since the chunked prefill code path has had other changes since this branch was created. The fix itself addresses a real upstream contract: recent mlx-lm prompt tuples carry a seventh `prompt_checkpoints` field that the chunked-prefill monkeypatch needs to retain across partial resume.

This PR is likely the right fix for issue #155 (omprashantjain, "Chunked prefill crashes with mlx-lm >= 0.31.0") which describes the symptom of the same underlying mlx-lm format change.

krystophny:
Force-pushed a refresh onto current upstream `main` (`b4fa030`). I addressed the three concrete review points in the refresh: added inline clarification for the upstream checkpoint normalization, added an inline comment on the post-`finalize()` checkpoint-tail replay, and added a regression that exercises the `prompt_checkpoint > 1` tail replay path. Validation: `python -m pytest tests/test_batching.py::TestSchedulerBasic -q` -> `10 passed`.

Thump604 (approved):
Refresh confirmed on head b80073a3 against upstream main b4fa030. The three concrete review points from the prior COMMENTED review are addressed:

1. Inline clarification for the upstream checkpoint normalization is visible in vllm_mlx/scheduler.py.
2. Inline comment on the post-finalize() checkpoint-tail replay path is visible.
3. tests/test_batching.py gains +358 lines of regression coverage, including the prompt_checkpoint > 1 tail replay path that the prior test construction never exercised.

The chunked prefill path now invokes prompt_checkpoint_callback after finalization, which was the blocking issue flagged on the earlier revision. Checkpoint semantics through partial resume match the upstream contract: recent mlx-lm prompt tuples carry a seventh prompt_checkpoints field that the chunked-prefill monkeypatch now retains across partial resume.

CI green across lint, type-check, test-matrix 3.10-3.12, test-apple-silicon, tests. Approved. This is the right fix for issue #155 ("Chunked prefill crashes").

PR #218, opened 2026-03-24
chat: forward chat_template_kwargs on simple-engine paths

krystophny:
## Summary
Honor `chat_template_kwargs` on the simple-engine paths that still ignored it and run the regression coverage in Apple Silicon CI.

## Why
Before this branch, `chat_template_kwargs` was only reliably honored on the batched path and the plain LLM chat path. Simple-engine multimodal chat, multimodal stream chat, and the text-only MTP route still dropped the field.

## What changed
- forward `chat_template_kwargs` through simple-engine multimodal `chat()`
- forward `chat_template_kwargs` through simple-engine multimodal `stream_chat()`
- forward `chat_template_kwargs` into `_stream_generate_text()` for the text-only MTP route
- include `tests/test_chat_template_kwargs.py` in Apple Silicon CI

## Status
- refreshed onto current upstream `main` (`b4fa030`) on 2026-04-09
- no logic changes beyond the base refresh

## Files to review
- `vllm_mlx/engine/simple.py`
- `.github/workflows/ci.yml`
- `tests/test_chat_template_kwargs.py`

## Validation
- `python -m pytest tests/test_chat_template_kwargs.py -q` -> `6 passed`
- note: the older `tests/test_simple_engine.py` validation command now depends on the separate async-harness refresh in `#226` on current upstream, so I kept validation scoped to this PR's dedicated regression file here

PR #222, opened 2026-03-24
simple-engine: keep tool chat on the streaming execution path

krystophny:
## Summary
Make tool-enabled non-stream `SimpleEngine.chat()` aggregate the existing streaming chat path instead of calling the separate blocking chat path.

## Why
On some local models, non-stream tool-enabled chat stalls while the streaming path completes normally. This change keeps one tool-capable execution path for simple-engine chat and still returns a normal non-stream `GenerationOutput`.

## What changed
- tool-enabled non-stream chat now aggregates `stream_chat()` output
- preserve the streamed token ids from the final output
- keep the streamed completion token count and final finish reason
- add regression coverage for the tool-chat path

## Status
- no logic changes beyond the base refresh

## Files to review
- `vllm_mlx/engine/simple.py`
- `tests/test_simple_engine.py`

## Validation
- `python -m pytest tests/test_simple_engine.py::TestSimpleEngineConcurrency::test_chat_with_tools_aggregates_streaming_path -q` -> `1 passed`
- note: the broader async `tests/test_simple_engine.py` run on current upstream depends on the separate harness refresh in `#226`

### DIFF
--- a/tests/test_simple_engine.py
+++ b/tests/test_simple_engine.py
@@ -146,7 +146,6 @@ class TestSimpleEngineConcurrency:
             engine = SimpleEngine("test-model")
             engine._model = mock_llm_model
             engine._loaded = True
-            engine._model.tokenizer.encode = MagicMock(return_value=[7, 8, 9])
             engine.stream_chat = fake_stream_chat  # type: ignore[method-assign]
 
             output = await engine.chat(
@@ -164,14 +163,11 @@ class TestSimpleEngineConcurrency:
             )
 
             assert output.text.startswith("<tool_call>")
-            assert output.tokens == [7, 8, 9]
+            assert output.tokens == []
             assert output.prompt_tokens == 11
             assert output.completion_tokens == 4
             assert output.finish_reason == "stop"
             mock_llm_model.chat.assert_not_called()
-            engine._model.tokenizer.encode.assert_called_once_with(
-                output.text, add_special_tokens=False
-            )
 
     @pytest.mark.anyio
     async def test_lock_serializes_stream_generate(self, mock_model):
--- a/tests/test_simple_engine.py
+++ b/tests/test_simple_engine.py
@@ -134,7 +134,7 @@ class TestSimpleEngineConcurrency:
                 finished=False,
             )
             yield MagicMock(
-                text="<|im_end|><tool_call>{\"name\":\"bash\",\"arguments\":{\"command\":\"pwd\"}}</tool_call>",
+                text='<|im_end|><tool_call>{"name":"bash","arguments":{"command":"pwd"}}</tool_call>',
                 tokens=[],
                 prompt_tokens=11,
                 completion_tokens=4,
--- a/tests/test_simple_engine.py
+++ b/tests/test_simple_engine.py
@@ -11,6 +11,10 @@ class TestSimpleEngineConcurrency:
     """Test SimpleEngine lock behavior with concurrent requests."""
 
     @pytest.fixture
+    def anyio_backend(self):
+        return "asyncio"
+
+    @pytest.fixture
     def mock_model(self):
         """Create a mock model that tracks concurrent calls."""
         model = MagicMock()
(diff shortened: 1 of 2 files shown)


Thump604:
@waybarrios, @krystophny: independent endorsement plus coordination note.

This PR makes tool-enabled non-stream `SimpleEngine.chat()` aggregate the existing streaming chat path instead of calling a separate blocking chat path. The accumulator pattern is the right shape: one source of truth for the chat execution, with non-stream wrapping streaming. Mergeable on current main.

Coordination note: the accumulator pattern in this PR (non-stream wrapping streaming) is also useful for wiring SpecPrefill into the non-streaming path, since SpecPrefill currently engages on the streaming surface only. If both fixes land in the same direction (streaming as the single source of truth for SimpleEngine execution), the result is cleaner than fixing them separately. Recommend this PR first since it is smaller and establishes the pattern.

Thump604 (approved):
The accumulator-over-streaming shape is the right call. I've been sketching the same structure for `SimpleEngine.generate()` (the /v1/completions engine path) as a follow-up, since the bug class is symmetric: streaming has feature X, non-streaming forgets X. Landing this one first means the generate-side PR is a smaller delta against an established pattern rather than introducing the pattern.

One small question, not blocking: should `_stream_chat_tool_llm` eventually become the shared path for all SimpleEngine chat, or do you see it staying tool-specific? Either works; I'm just trying to gauge whether I should extend my `generate()` refactor to go through a shared helper or keep it on its own branch.

Approving based on the design and the test coverage you added.

krystophny:
Force-pushed a refresh onto current upstream `main` (`b4fa030`). No logic change beyond the base refresh. Re the helper question: short term I'd keep `_stream_chat_tool_llm` tool-specific. The pattern I want to standardize is 'non-stream wraps streaming for featureful paths', not forcing all chat paths through one shared helper yet. Validation: `python -m pytest tests/test_simple_engine.py::TestSimpleEngineConcurrency::test_chat_with_tools_aggregates_streaming_path -q` -> `1 passed`.

Thump604:
Refresh confirmed on head 15e3ca5c against upstream main b4fa030. No logic change on the accumulator path; the small vllm_mlx/engine/simple.py delta (+26/0) is the expected shape for the rebase.

On the helper question: keeping _stream_chat_tool_llm tool-specific for now is the pragmatic call. The pattern worth standardizing is "non-stream wraps streaming for featureful paths," not "one shared helper for every chat path." The plain LLM non-tool branch can get its own accumulator independently once #218 lands and the chat_template_kwargs forwarding is stable, rather than trying to collapse everything into one helper in a single PR.

CI green across the full matrix. Prior APPROVED review at f9468bc4 stands for the refreshed head.

Thump604:
I rebased this onto current main locally and validated it cleanly. GitHub compare sees the restack branch correctly, but  is misbehaving on this one from the CLI, so I’m opening the writable replacement branch via the API path instead and will link it here once GitHub returns the PR URL. Local validation on the restack branch:\n- python -m py_compile vllm_mlx/engine/simple.py tests/test_simple_engine.py\n- python -m black --check --fast vllm_mlx/engine/simple.py tests/test_simple_engine.py\n- pytest -q tests/test_simple_engine.py

Thump604:
Replacement restack is up at #285: https://github.com/waybarrios/vllm-mlx/pull/285\n\nThat branch carries the rebased tool-chat aggregation change on a writable branch and passed local validation here:\n- python -m py_compile vllm_mlx/engine/simple.py tests/test_simple_engine.py\n- python -m black --check --fast vllm_mlx/engine/simple.py tests/test_simple_engine.py\n- pytest -q tests/test_simple_engine.py\n\nI’m using the replacement branch because the original computor-org fork could not be updated directly from this auth context.

Thump604:
Closing as superseded by #285, which carries the rebased SimpleEngine tool-chat aggregation change on a writable branch and validated cleanly on current main. The original computor-org fork branch could not be updated directly from this auth context, so the merge path moved to the replacement PR.

PR #232, opened 2026-03-29
fix: suppress tool call XML from streaming text content (#129)

### LINKED ISSUE #129: Anthropic Streaming `/v1/messages` Leaks `<think>` and `<tool_call>` Markup Before Structured `tool_use`

sjswerdloff:
Tool call XML (e.g. <minimax:tool_call>, <tool_call>) was leaking into streaming text deltas via the /v1/messages endpoint. The raw markup appeared in the client's conversation context alongside the structured tool_use block, doubling token consumption for every tool call.

Add StreamingToolCallFilter that buffers streaming text and suppresses content inside tool call blocks. Handles tags split across multiple deltas, multiple tool calls per response, and preserves <think> blocks.

Supports MiniMax (<minimax:tool_call>) and Qwen (<tool_call>) formats.

14 unit tests included.

Fixes #129

### DIFF
--- a/tests/test_streaming_tool_filter.py
+++ b/tests/test_streaming_tool_filter.py
@@ -6,7 +6,6 @@ from vllm_mlx.api.utils import StreamingToolCallFilter
 
 
 class TestStreamingToolCallFilter(unittest.TestCase):
-
     def test_normal_text_passes_through(self):
         f = StreamingToolCallFilter()
         assert f.process("Hello world") == "Hello world"
@@ -110,6 +109,35 @@ class TestStreamingToolCallFilter(unittest.TestCase):
         assert f.process("") == ""
         assert f.process("text") == "text"
         assert f.process("") == ""
+
+    def test_calling_tool_bracket_suppressed(self):
+        """Qwen3 bracket-style: [Calling tool: func({...})]\n"""
+        f = StreamingToolCallFilter()
+        result = f.process('[Calling tool: search({"q": "test"})]\n')
+        assert result == ""
+
+    def test_calling_tool_multiline_json(self):
+        """Multi-line JSON args in bracket-style tool call."""
+        f = StreamingToolCallFilter()
+        r1 = f.process('[Calling tool: search({"q": "test",')
+        r2 = f.process(' "limit": 5})]\n')
+        r3 = f.process("After")
+        assert r1 + r2 + r3 == "After"
+
+    def test_buffer_cap_on_unclosed_block(self):
+        """Buffer should be capped if tool call block never closes."""
+        from vllm_mlx.api.utils import _MAX_TOOL_BUFFER_BYTES
+
+        f = StreamingToolCallFilter()
+        f.process("<minimax:tool_call>")
+        # Feed data exceeding the cap
+        chunk = "x" * 10000
+        for _ in range(_MAX_TOOL_BUFFER_BYTES // 10000 + 2):
+            f.process(chunk)
+        # After cap, filter should have exited the block
+        assert not f._in_block
+        # New text should pass through
+        assert f.process("after") == "after"
 
 
 if __name__ == "__main__":
--- a/vllm_mlx/api/utils.py
+++ b/vllm_mlx/api/utils.py
@@ -3,9 +3,12 @@
 Utility functions for text processing and model detection.
 """
 
+import logging
 import re
 
 from .models import Message
+
+logger = logging.getLogger(__name__)
 
 # =============================================================================
 # Special Token Patterns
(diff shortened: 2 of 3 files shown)


Thump604:
Thanks for the detailed response. The upstream-of-parsers flow makes sense and the #177 compatibility analysis is correct — different pipeline stages, no conflict. LGTM.

sjswerdloff:
@Thump604 Thank you for the thorough review. Addressed all concerns:

**1. StreamingThinkRouter unit tests** — Added 18 tests covering:
- `start_in_thinking` mode and transition to text
- Partial tag handling (held back across deltas, false alarms)
- Multiple think blocks, token-by-token streaming
- Flush behavior and state reset

**2. Integration tests** — Added 12 tests for the full pipeline:
- `tool_filter → think_router → SSE events`
- Pure text, thinking→text, start_in_thinking→text
- Tool call suppression with accumulated text preserved for parsing
- Mixed thinking + text + tool calls
- Block index increment verification

**3. Block transition refactored** — Extracted `_emit_content_pieces()` helper that replaces the 3x repeated logic. Returns updated state (block_type, index) so the caller stays clean.

**4. `start_in_thinking` detection** — Noted as assumption. The current heuristic works for MiniMax, Qwen3, DeepSeek-R1 which all inject `<think>` into `add_generation_prompt`. A model-specific override would be the right next step if new patterns emerge.

**5. `<function=` newline delimiter** — Kept as-is since the full text is captured separately for parsing. The filter only needs to suppress it from the streaming delta, not parse it.

44 tests passing across filter (14), router (18), and integration (12).

— Clement

PR #233, opened 2026-03-29
Add TurboQuant KV cache compression for prefix cache (4.6x)

arozanov:
## Summary

Adds `--turbo-kv-bits` option (1-4) to compress prefix cache entries using [TurboQuant](https://arxiv.org/abs/2504.19874) (PolarQuant: randomized Hadamard rotation + Lloyd-Max codebook quantization). At 3-bit, this gives **4.6x compression** vs FP16, compared to ~2x from the existing `--kv-cache-quantization`.

This is useful for Apple Silicon where memory is the bottleneck — more prefix cache entries fit in RAM, improving cache hit rates on long-context workloads.

## Usage

```bash
vllm-mlx serve model --turbo-kv-bits 3
```

Replaces `--kv-cache-quantization` when set. Falls back to standard quantization if TurboQuant is not available.

## Changes

- **`memory_cache.py`**: `_turbo_quantize_cache()` / updated `_dequantize_cache()`, `estimate_kv_cache_memory()` support, `_trim_cache_offset()` support, `needs_dequantize` property on config, validation
- **`scheduler.py`**: `turbo_kv_bits` field in `SchedulerConfig`, propagation to `MemoryCacheConfig`
- **`cli.py`**: `--turbo-kv-bits` argument for `serve` and `bench` commands

## Dependency

Requires mlx-lm with TurboQuant KV cache support: ml-explore/mlx-lm#1067

## Test plan

- [x] Roundtrip: quantize → dequantize preserves data (cosine sim 0.98+)
- [x] `from_state` deserialization → dequantize (auto-init quantizer)
- [x] `estimate_kv_cache_memory` returns correct bytes for TurboQuant entries
- [x] `_trim_cache_offset` creates shallow copy (does not mutate stored entry)
- [x] `has_non_trimmable` correctly identifies TurboQuant as trimmable
- [x] `needs_dequantize` property gates all fetch paths
- [x] Config validation rejects invalid bit widths
- [x] dtype preserved through quantize/dequantize cycle (float16, bfloat16, float32)
- [x] CLI `--turbo-kv-bits` propagates to SchedulerConfig → MemoryCacheConfig


janhilgard (commented):
## Code Review

Clean, well-structured code that follows existing patterns (`_quantize_cache` / `_dequantize_cache`). The `needs_dequantize` property is an elegant abstraction. A few concerns:

### 🔴 Breaking change: `is_trimmable()` (blocking)

The `has_non_trimmable` check changed from duck-typing (`hasattr(lc, "offset") and hasattr(lc, "keys")`) to `hasattr(lc, "is_trimmable") and lc.is_trimmable()`. Problem: **existing `KVCache` and `QuantizedKVCache` don't have an `is_trimmable()` method**, so after this change ALL cache layers will be marked as non-trimmable — this completely breaks supersequence and LCP matching for all users, even without `--turbo-kv-bits`.

Suggestion — keep the old duck-typing as fallback:
```python
has_non_trimmable = any(
    not (
        (hasattr(lc, "is_trimmable") and lc.is_trimmable())
        or (hasattr(lc, "offset") and hasattr(lc, "keys"))
    )
    for lc in cache
)
```

### 🟡 Private API access in dequantization

`_dequantize_cache()` accesses many private attributes of TurboQuantKVCache (`_k_q`, `_v_q`, `_k_dim`, `_v_dim`, `_dtype`, `_full_dequant()`, `_ensure_quantizer()`). This is fragile — private API can change without notice. Does mlx-lm PR #1067 expose a public `.dequantize()` or `.to_kvcache()` method? If not, it would be worth proposing one upstream.

### 🟡 Shallow copy risk in `_trim_cache_offset`

`tc.__dict__.update(layer_cache.__dict__)` shares references to all internal objects. The invalidation of `_k_deq_buf` / `_v_deq_buf` assumes specific implementation details. If TurboQuantKVCache adds more cache buffers later, they'll be stale. Does the upstream class provide a `copy()` or `trim()` method?

### 🟡 `estimate_kv_cache_memory` — `.state` property

Accessing `.state` may trigger lazy evaluation if it returns dequantized tensors. Safer to iterate directly over packed arrays (`k_packed`, `v_packed`, `k_norms`, `v_norms`).

### ℹ️ Upstream dependency

PR depends on `ml-explore/mlx-lm#1067` which is not merged yet. Worth noting as a prerequisite in the description so this doesn't get merged prematurely.

janhilgard (commented):
## Code Review

Clean, well-structured code that follows existing patterns (`_quantize_cache` / `_dequantize_cache`). The `needs_dequantize` property is an elegant abstraction. A few concerns:

### 🔴 Breaking change: `is_trimmable()` (blocking)

The `has_non_trimmable` check changed from duck-typing (`hasattr(lc, "offset") and hasattr(lc, "keys")`) to `hasattr(lc, "is_trimmable") and lc.is_trimmable()`. Problem: **existing `KVCache` and `QuantizedKVCache` don't have an `is_trimmable()` method**, so after this change ALL cache layers will be marked as non-trimmable — this completely breaks supersequence and LCP matching for all users, even without `--turbo-kv-bits`.

Suggestion — keep the old duck-typing as fallback:
```python
has_non_trimmable = any(
    not (
        (hasattr(lc, "is_trimmable") and lc.is_trimmable())
        or (hasattr(lc, "offset") and hasattr(lc, "keys"))
    )
    for lc in cache
)
```

### 🟡 Private API access in dequantization

`_dequantize_cache()` accesses many private attributes of TurboQuantKVCache (`_k_q`, `_v_q`, `_k_dim`, `_v_dim`, `_dtype`, `_full_dequant()`, `_ensure_quantizer()`). This is fragile — private API can change without notice. Does mlx-lm PR #1067 expose a public `.dequantize()` or `.to_kvcache()` method? If not, it would be worth proposing one upstream.

### 🟡 Shallow copy risk in `_trim_cache_offset`

`tc.__dict__.update(layer_cache.__dict__)` shares references to all internal objects. The invalidation of `_k_deq_buf` / `_v_deq_buf` assumes specific implementation details. If TurboQuantKVCache adds more cache buffers later, they'll be stale. Does the upstream class provide a `copy()` or `trim()` method?

### 🟡 `estimate_kv_cache_memory` — `.state` property

Accessing `.state` may trigger lazy evaluation if it returns dequantized tensors. Safer to iterate directly over packed arrays (`k_packed`, `v_packed`, `k_norms`, `v_norms`).

### ℹ️ Upstream dependency

PR depends on `ml-explore/mlx-lm#1067` which is not merged yet. Worth noting as a prerequisite in the description so this doesn't get merged prematurely.

arozanov:
> ## Code Review
> Clean, well-structured code that follows existing patterns (`_quantize_cache` / `_dequantize_cache`). The `needs_dequantize` property is an elegant abstraction. A few concerns:
> 
> ### 🔴 Breaking change: `is_trimmable()` (blocking)
> The `has_non_trimmable` check changed from duck-typing (`hasattr(lc, "offset") and hasattr(lc, "keys")`) to `hasattr(lc, "is_trimmable") and lc.is_trimmable()`. Problem: **existing `KVCache` and `QuantizedKVCache` don't have an `is_trimmable()` method**, so after this change ALL cache layers will be marked as non-trimmable — this completely breaks supersequence and LCP matching for all users, even without `--turbo-kv-bits`.
> 
> Suggestion — keep the old duck-typing as fallback:
> 
> ```python
> has_non_trimmable = any(
>     not (
>         (hasattr(lc, "is_trimmable") and lc.is_trimmable())
>         or (hasattr(lc, "offset") and hasattr(lc, "keys"))
>     )
>     for lc in cache
> )
> ```
> 
> ### 🟡 Private API access in dequantization
> `_dequantize_cache()` accesses many private attributes of TurboQuantKVCache (`_k_q`, `_v_q`, `_k_dim`, `_v_dim`, `_dtype`, `_full_dequant()`, `_ensure_quantizer()`). This is fragile — private API can change without notice. Does mlx-lm PR #1067 expose a public `.dequantize()` or `.to_kvcache()` method? If not, it would be worth proposing one upstream.
> 
> ### 🟡 Shallow copy risk in `_trim_cache_offset`
> `tc.__dict__.update(layer_cache.__dict__)` shares references to all internal objects. The invalidation of `_k_deq_buf` / `_v_deq_buf` assumes specific implementation details. If TurboQuantKVCache adds more cache buffers later, they'll be stale. Does the upstream class provide a `copy()` or `trim()` method?
> 
> ### 🟡 `estimate_kv_cache_memory` — `.state` property
> Accessing `.state` may trigger lazy evaluation if it returns dequantized tensors. Safer to iterate directly over packed arrays (`k_packed`, `v_packed`, `k_norms`, `v_norms`).
> 
> ### ℹ️ Upstream dependency
> PR depends on `ml-explore/mlx-lm#1067` which is not merged yet. Worth noting as a prerequisite in the description so this doesn't get merged prematurely.

Thanks for the thorough review!

Fixed:
- is_trimmable() regression: added duck-typing fallback for existing KVCache/QuantizedKVCache
- estimate_kv_cache_memory: now iterates packed arrays directly (k_packed, v_packed, k_norms, v_norms) instead of .state to avoid triggering lazy dequantization

For the private API and shallow copy concerns: added public dequantize() and copy() methods to TurboQuantKVCache upstream in mlx-lm #1067. Will update this PR to use them once that's merged.
Upstream dependency noted in the description.

janhilgard:
Thanks for the quick fixes! The `is_trimmable()` fallback and direct packed array iteration look good.

Happy to hear about the public `dequantize()` and `copy()` methods upstream — that'll make the integration much cleaner. No further concerns from my side, just waiting on mlx-lm #1067 to land.

Thump604:
@waybarrios, @arozanov: brief positive note.

Memory-bound prefix caching is exactly the right pressure point for Apple Silicon (where unified memory is the long-context bottleneck), and 4.6x compression on prefix entries is meaningful relative to the existing ~2x from `--kv-cache-quantization`.

Two questions for completeness, not blocking:

1. What is the quality impact at 3-bit (PolarQuant) on representative tasks? The TurboQuant paper has ablation numbers but the empirical impact on long-context QA or needle-in-haystack on Qwen 3.5 / Gemma 4 would be useful for users to decide whether to enable it.
2. Is the `--turbo-kv-bits` flag mutually exclusive with `--kv-cache-quantization`, or are they layered?

Mergeable on current main.

waybarrios (commented):
### Code review

Found 4 issues:

---

#### 1. PR needs rebase — `_trim_cache_offset` and `_dequantize_cache` were rewritten on main

After this PR branched, the `_QuantizedCacheWrapper` refactor landed on main (commit `6f0efc2`), which completely rewrote both `_trim_cache_offset` and `_dequantize_cache`. The PR still imports and checks against `QuantizedKVCache` from mlx-lm, but main now uses an internal `_QuantizedCacheWrapper` class. This will cause merge conflicts and silent bugs if resolved incorrectly.

What the PR expects:
```python
from mlx_lm.models.cache import QuantizedKVCache
# ...
if QuantizedKVCache is not None and isinstance(layer_cache, QuantizedKVCache):
```

What main now has:
```python
if isinstance(layer_cache, _QuantizedCacheWrapper):
    # completely different structure with orig_type/orig_attrs
```

CI also confirms `black --check` fails on `memory_cache.py`. A full rebase against current main is needed.

https://github.com/waybarrios/vllm-mlx/blob/9e909a8e46b66de3e6a678ea6f44bff8254d3b15/vllm_mlx/memory_cache.py#L271-L320

---

#### 2. Shallow copy in `_trim_cache_offset` shares mutable quantizer state with stored cache

The TurboQuantKVCache branch uses `__dict__.update` which shallow-copies all references. The mutable `_k_q`/`_v_q` quantizer objects end up shared between the trimmed copy and the original stored entry:

```python
# This creates shared references to _k_q, _v_q (mutable quantizer objects)
tc = TurboQuantKVCache.__new__(TurboQuantKVCache)
tc.__dict__.update(layer_cache.__dict__)  # shallow copy
tc.offset = max(layer_cache.offset - trim_by, 0)
tc._k_deq_buf = None   # only buffers are reset
tc._v_deq_buf = None
# but _k_q and _v_q are NOT copied — they're shared with the original
```

Later, `_dequantize_cache` calls `layer._ensure_quantizer(...)` which mutates quantizer state in-place. Since `_k_q`/`_v_q` are shared, this corrupts the stored cache entry — violating the "do NOT mutate original" comment.

Fix: either deep-copy the quantizer objects, or use the upstream `copy()` method once mlx-lm#1067 lands:

```python
# Option A: deep copy quantizers
import copy
tc._k_q = copy.deepcopy(layer_cache._k_q) if layer_cache._k_q is not None else None
tc._v_q = copy.deepcopy(layer_cache._v_q) if layer_cache._v_q is not None else None

# Option B (preferred): use upstream public API
tc = layer_cache.copy()
tc.offset = max(layer_cache.offset - trim_by, 0)
```

https://github.com/waybarrios/vllm-mlx/blob/9e909a8e46b66de3e6a678ea6f44bff8254d3b15/vllm_mlx/memory_cache.py#L309-L320

---

#### 3. `_dequantize_cache` accesses 10+ private attributes — should use public API

The dequantization path reaches deep into `TurboQuantKVCache` internals (`_k_q`, `_v_q`, `_k_dim`, `_v_dim`, `_dtype`, `_full_dequant()`, `_ensure_quantizer()`, etc.). This reimplements internal logic that belongs inside the cache class itself, and since `TurboQuantKVCache` comes from an unmerged upstream PR (ml-explore/mlx-lm#1067), these private APIs are highly likely to change.

Current approach:
```python
# 10+ private attribute accesses
if layer._k_q is None:
    layer._ensure_quantizer(layer._k_dim, layer._v_dim)
B, H = layer.k_packed.shape[:2]
dtype = layer._dtype if layer._dtype is not None else mx.float16
k_all = layer._full_dequant(
    layer.k_packed, layer.k_norms, layer._k_q,
    layer._k_dim, B, H, layer.offset, dtype,
)
```

The upstream PR already exposes `dequantize()` and `copy()` public methods. This should be:
```python
elif TurboQuantKVCache is not None and isinstance(layer, TurboQuantKVCache) and not layer.empty():
    result.append(layer.dequantize())
```

https://github.com/waybarrios/vllm-mlx/blob/9e909a8e46b66de3e6a678ea6f44bff8254d3b15/vllm_mlx/memory_cache.py#L445-L462

---

#### 4. `RotatingKVCache` metadata lost on dequantize — regression for sliding-window models

`_turbo_quantize_cache` only handles plain `KVCache`. On dequantize, it always reconstructs a plain `KVCache`:

```python
# _turbo_quantize_cache — only matches plain KVCache
if isinstance(layer, KVCache) and layer.keys is not None:
    compressed.append(layer.to_turbo_quantized(bits=bits))

# _dequantize_cache — always creates plain KVCache, losing RotatingKVCache metadata
kv = KVCache()  # step, max_size, _idx are gone
kv.update_and_fetch(k_all, v_all)
```

This re-introduces the bug fixed by the `_QuantizedCacheWrapper` refactor, which preserves `orig_type`/`orig_attrs` to correctly reconstruct `RotatingKVCache` for sliding-window models (Gemma 4, etc.). The TurboQuant path needs the same preservation pattern:

```python
# Should preserve the original cache type, similar to _QuantizedCacheWrapper
orig_type = type(layer)  # could be RotatingKVCache
orig_attrs = {k: getattr(layer, k) for k in ("step", "max_size", "_idx") if hasattr(layer, k)}
# ... then reconstruct with orig_type and orig_attrs on dequantize
```

https://github.com/waybarrios/vllm-mlx/blob/9e909a8e46b66de3e6a678ea6f44bff8254d3b15/vllm_mlx/memory_cache.py#L406-L420

---

**TL;DR**: The PR needs a rebase against current main (the `_QuantizedCacheWrapper` refactor changed the code this PR modifies). After rebasing, the main concerns are: (1) use public API from upstream instead of private attributes, (2) handle `RotatingKVCache` preservation like the existing quantization path does, and (3) fix the shallow copy to avoid shared mutable state.

Thump604:
I rebased this PR onto current `main` and pushed the updated branch.

The follow-up changes address the review points directly:
- switched the TurboQuant path onto the current wrapper-based `memory_cache.py` shape instead of the older `QuantizedKVCache` branch
- removed the private TurboQuant dequantization path and now use public `copy()` / `dequantize()` when TurboQuant objects are present
- stopped the shallow `__dict__.update(...)` copy for TurboQuant cache trimming
- constrained TurboQuant storage to the plain `KVCache` path and preserved wrapper/original-cache metadata alongside it
- kept the `is_trimmable` fallback and memory-estimation fixes from the later review-addressing commits

Validation I ran after the rebase:
- `black --check vllm_mlx/memory_cache.py vllm_mlx/cli.py vllm_mlx/scheduler.py tests/test_kv_cache_quantization.py`
- `pytest -q tests/test_memory_cache.py tests/test_kv_cache_quantization.py` (`65 passed`)

I also updated the stale quantization assertions in `tests/test_kv_cache_quantization.py` so they match the current wrapper-based implementation on `main`.

arozanov:
Thanks @waybarrios for the detailed review and @Thump604 for the rebase and fixes.

Not in progress, ready for final review. All four issues from the review are addressed in the latest push. Waiting on ml-explore/mlx-lm#1067 upstream before this can land.
@Thump604 on the quality question: at 3-bit on Qwen 3 8B we see less than 0.5 perplexity increase on WikiText-2 and no degradation on MMLU. Can add benchmark numbers to the README if useful.
The --turbo-kv-bits flag is mutually exclusive with --kv-cache-quantization, setting one disables the other with a clear error if both are specified.

Thump604 (changes requested):
The rebase fixed the earlier structural issues, but I still see one merge blocker around the upstream dependency boundary.

Right now the CLI accepts `--turbo-kv-bits`, and the cache path will happily carry that config even when the underlying mlx-lm TurboQuant support is not present. In `_turbo_quantize_cache()` the actual compression is gated by `hasattr(layer, "to_turbo_quantized")`, so on a runtime without mlx-lm#1067 this can silently degrade into "flag accepted, no compression happened".

That is a bad failure mode for a user-facing memory/compression flag. I think the PR needs one of these before merge:
- fail fast at startup / config validation when `--turbo-kv-bits` is set but TurboQuant support is unavailable, or
- gate the CLI flag itself behind detected TurboQuant capability.

Without that, we expose a feature flag whose success path depends on an unmerged upstream capability and can no-op silently.

janhilgard (commented):
## Code Review — TurboQuant KV cache compression

Good work integrating TurboQuant as an alternative compression path for prefix cache entries. The wrapper-based approach (`_TurboQuantCacheWrapper`) mirrors the existing `_QuantizedCacheWrapper` pattern cleanly, and the `needs_dequantize` property is a nice abstraction that eliminates the scattered `kv_quantize` checks.

### Observations

**1. Silent no-op when mlx-lm lacks TurboQuant (addressed in Thump604 review, still open)**

Thump604 already flagged this and it's still the main merge blocker: `_turbo_quantize_cache()` silently falls through when `to_turbo_quantized` is missing. The user sees `TurboQuant: 3-bit prefix cache compression` in CLI output but gets zero compression. An explicit startup check with a clear error ("--turbo-kv-bits requires mlx-lm with TurboQuant support, see ml-explore/mlx-lm#1067") would be much better than runtime degradation.

**2. `_TurboQuantCacheWrapper.is_trimmable()` depends on `copy()` existing**

`is_trimmable()` returns `hasattr(self.layer, "copy")`, which ties trimmability to an upstream API that may or may not exist depending on the mlx-lm version. If `copy()` is absent, the wrapper reports itself as non-trimmable, which means the `has_non_trimmable` check will skip trimming entirely for that cache entry. This is safe (no crash), but means supersequence/LCP trimming silently degrades. Worth a log warning when this happens.

**3. Memory estimation for `_TurboQuantCacheWrapper`**

The `estimate_kv_cache_memory` function has two separate paths for TurboQuant: one for wrappers where `isinstance(layer_cache.layer, _TQ)` and one for bare `TurboQuantKVCache` instances. The wrapper path iterates `layer_cache.layer.state` while the bare path iterates individual packed arrays. This dual-path approach works but is fragile if upstream changes the internal layout. Consider consolidating to a single estimation strategy, or at minimum add a comment explaining why both paths are needed.

**4. `_trim_cache_offset` for `_TurboQuantCacheWrapper` does a shallow copy of internal state**

```python
tc.layer = layer_cache.layer.copy()
tc.layer.offset = max(layer_cache.layer.offset - trim_by, 0)
```

This relies on `copy()` producing a sufficiently deep copy that mutating `tc.layer.offset` doesn't affect the original stored entry. If `copy()` is a shallow copy and the packed arrays are shared, this is correct (offset is a scalar). But if upstream changes `copy()` semantics, stored entries could be corrupted. A defensive comment noting this assumption would help future maintainers.

**5. Test coverage is thorough**

The existing test updates properly replace `QuantizedKVCache` references with `_QuantizedCacheWrapper` throughout. The tests cover the key paths: quantize/dequantize roundtrip, type assertions on stored entries, non-KVCache layer preservation.

### Summary

The implementation is clean and follows established patterns. The upstream dependency on mlx-lm#1067 is the blocking item — once that lands, the main remaining concern is adding a startup-time check so `--turbo-kv-bits` fails fast rather than silently degrading. Everything else is non-blocking.

arozanov:
Pushed b5dd59d. Fail-fast validation is in now: both MemoryCacheConfig.__post_init__ and the CLI serve/bench paths error out with a clear message pointing to ml-explore/mlx-lm#1067 when TurboQuant support is missing, so the flag no longer silently no-ops.

--turbo-kv-bits and --kv-cache-quantization are mutually exclusive now — rejected both in the CLI and 
  in the config ```__post_init__``` so programmatic users get the same guard. Help text updated. 

For the trim concern: _TurboQuantCacheWrapper.is_trimmable() now logs a one-time warning when the underlying TurboQuantKVCache lacks copy(). Trimming falls back to full-prefix matching in that case, so it's correct just less efficient. 

Also added inline comments on the two estimate_kv_cache_memory paths (wrapper vs bare) and on the copy() contract used in _trim_cache_offset.

Rebased onto current main, cli.py conflict resolved. Still blocked on ml-explore/mlx-lm#1067, please don't merge before that lands.

janhilgard (approved):
## Re-review: All blocking items addressed

Both previously flagged concerns are now resolved:

1. **Fail-fast at startup** (Thump604's blocker): `_check_turboquant_capability()` is called in both CLI validation (`cli.py`) and `MemoryCacheConfig.__post_init__()`. If mlx-lm lacks TurboQuant support, the server exits immediately with an actionable error message pointing to ml-explore/mlx-lm#1067. No more silent degradation.

2. **Mutual exclusion**: `--turbo-kv-bits` and `--kv-cache-quantization` are validated as mutually exclusive at CLI level and config level. Clear error messaging.

3. **Trim degradation warning**: `_TurboQuantCacheWrapper.is_trimmable()` logs a one-time warning when `copy()` is unavailable, so operators know supersequence/LCP trimming is disabled.

4. **Code quality**: The implementation cleanly mirrors the existing `_QuantizedCacheWrapper` pattern. `needs_dequantize` property is a good abstraction that replaces scattered `kv_quantize` checks. Memory estimation covers both wrapper and bare paths.

### Minor note

- CI lint (ruff) is currently failing — needs a formatting pass before merge.
- The `kimi_tool_parser.py` change is unrelated cleanup (removes dead try/except, fixes `rsplit`) — fine to include but could be a separate commit for cleaner git history.

LGTM once lint passes.

Thump604:
Hi @arozanov -- this TurboQuant PR has review feedback outstanding and has been open since late March. Are you still planning to address the review comments and rebase? We're interested in the work but want to keep the PR list current. Will check back in two weeks.

Thump604:
Closing this stale TurboQuant PR because it is now conflicted and no longer reviewable as a narrow prefix-cache compression change. The branch history also contains unrelated work, so it is not a good merge base for the current tree.

If TurboQuant is still worth pursuing after the mlx-lm TurboQuant API is stable, the right path is a fresh PR from current main with only the TurboQuant prefix-cache integration, fail-fast capability checks, wrapper-compatible trim/dequantize behavior, and focused tests.

PR #235, opened 2026-03-30

benaskins:
When mlx_lm.load() succeeds on the happy path (no ValueError), the function falls through to implicit return None instead of returning the loaded (model, tokenizer) tuple.

This causes a TypeError in LLMModel.load():
  TypeError: cannot unpack non-iterable NoneType object

The bug only affects models that load successfully without needing the tokenizer or strict=False fallback paths (which do have return statements).


benaskins:
Closing. Dup of PR #230

PR #237, opened 2026-03-31
fix: add missing return in load_model_with_fallback()

keegoid:
## Summary

- `load_model_with_fallback()` in `vllm_mlx/utils/tokenizer.py` is missing a `return model, tokenizer` statement after the `try/except` block
- When `mlx_lm.load()` succeeds without raising a `ValueError` (no fallback path triggered), the function falls through and implicitly returns `None`
- The caller in `batched.py:257` then crashes with `TypeError: cannot unpack non-iterable NoneType object`

This affects models where `load()` succeeds on the first attempt, including `mlx-community/Qwen3-Coder-Next-4bit` with `trust_remote_code=True` (the default in `BatchedEngine`).

## Test plan

- [x] Verified the fix resolves the crash when serving `mlx-community/Qwen3-Coder-Next-4bit` with `--continuous-batching`
- [x] Confirmed inference works end-to-end after the fix (single and concurrent requests)
- [ ] Existing fallback paths (ValueError handlers) are unaffected — they already have explicit `return` statements

### DIFF
--- a/vllm_mlx/utils/tokenizer.py
+++ b/vllm_mlx/utils/tokenizer.py
@@ -67,6 +67,8 @@ def load_model_with_fallback(model_name: str, tokenizer_config: dict = None):
             return _load_strict_false(model_name, tokenizer_config)
         raise
 
+    return model, tokenizer
+
 
 def _load_strict_false(model_name: str, tokenizer_config: dict = None):
     """Load model with strict=False to discard extra weights (e.g., vision tower, MTP)."""


keegoid:
@waybarrios Thanks for reviewing — but I believe the bug is still present on `main`. Let me walk through the evidence:

**Line 54 on current `main` (`4ede902`) is an assignment, not a return:**

```python
# line 54 — actual code on main
```

It is **not** `return load(...)` as described in your comment. You can verify by viewing the file on main: https://github.com/waybarrios/vllm-mlx/blob/main/vllm_mlx/utils/tokenizer.py#L54

**The success path has no return statement.** When `load()` succeeds without raising `ValueError`, execution falls through the try/except and off the end of the function, implicitly returning `None`. The caller at `batched.py:257` then crashes:

```python
self._model, self._tokenizer = load_model_with_fallback(...)
# TypeError: cannot unpack non-iterable NoneType object
```

This is the exact error I hit when loading `mlx-community/Qwen3-Coder-Next-4bit` with `--continuous-batching`. Adding `return model, tokenizer` after line 68 fixes it.

**PR #230** (referenced as "same fix, merged") was also **closed without merging** (`merged: false`). Neither fix has landed on `main`.

I have had two independent AI models (Claude Opus 4.6 and OpenAI Codex/GPT-5.4) review the code and both confirm the bug. Happy to reopen this PR or submit a fresh one — whatever you prefer.

PR #223, opened 2026-03-24
feat: MTP per-request routing in BatchedEngine

Thump604:
## Summary

Port SimpleEngine's MLLM+MTP per-request routing to BatchedEngine for continuous batching.

- **Text-only requests** → mlx_lm TextModel with MTP speculative decoding (1.1-1.3x gen speed)
- **Media requests** → MLLM path (unchanged)

Routing decision is per-request based on `_has_media_content()`. Uses `text_model_from_vlm.py` (upstream from PR #180) to build a zero-copy TextModel from VLM backbone weights (~0 extra RAM).

## Changes (+226 lines in `batched.py`, +107 lines tests)

- `_has_media_content()` helper + `_MEDIA_TYPES` constant (mirrors SimpleEngine)
- `mtp` and `prefill_step_size` params in `BatchedEngine.__init__`
- TextModel build in `_start_mllm()` with Qwen3.5 eos_token fix
- Per-request routing in `chat()` and `stream_chat()`
- `_chat_text_model()` / `_stream_chat_text_model()` for mlx_lm generation under lock
- Cleanup in `stop()`

## Design decisions

- **Focused scope:** No SpecPrefill, no system KV cache — those can come in follow-up PRs
- **Lock-based serialization:** `_text_generation_lock` serializes Metal operations for text model path (same pattern as SimpleEngine's `_generation_lock`)
- **Graceful degradation:** If MTP weights don't exist, `build_text_model` returns None and all requests go to MLLM

## Test plan

- [x] 8 unit tests for `_has_media_content` (text, image, video, audio, multi-turn, mixed)
- [x] Black + ruff clean
- [ ] Integration test with VLM+MTP model (text routing + media routing)

PR #245, opened 2026-04-01
Add MTP speculative decoding for Qwen3.5 models

janhilgard:
## Summary

- Adds runtime MTP (Multi-Token Prediction) support for Qwen3.5 Dense and MoE architectures (27B, 35B-A3B, 122B-A10B)
- Qwen3.5's built-in MTP head predicts token n+2 from hidden states + token n+1, enabling single-user speedup with 75-80% acceptance rate
- **MTP auto-disables when batch > 1** — no throughput penalty under concurrency
- Includes MTP weight extraction script for preparing BF16 weights from HuggingFace source models

## Benchmark Results

**Hardware:** Apple M3 Ultra, 256 GB unified memory
**Setup:** `--mllm --continuous-batching`, 300 tokens/request, same prompt
**MTP verified active** via process args and server logs

### Qwen3.5-35B-A3B-4bit (MoE, ~20 GB)

| Concurrency | Throughput (tok/s) | MTP status | Notes |
|-------------|-------------------|------------|-------|
| **1x** | **85.9** | Active | +18% vs baseline (73.7) |
| **8x** | **307.3** | Auto-skip | Matches no-MTP baseline (302.3) |

**MTP acceptance rate:** 79-80%

### Qwen3.5-122B-A10B-4bit (MoE, ~69 GB)

| Concurrency | No MTP (tok/s) | With MTP (tok/s) | Speedup |
|-------------|----------------|------------------|---------|
| **1x** (single-user) | 46.4 | **52.0** | **1.12x** |

**MTP acceptance rate:** 75-76%

### Qwen3.5-27B-4bit (Dense, ~16 GB)

| Concurrency | No MTP (tok/s) | With MTP (tok/s) | Speedup |
|-------------|----------------|------------------|---------|
| **1x** (single-user) | 30.1 | **35.9** | **1.19x** |

**MTP acceptance rate:** 78-80%
**Note:** Dense model (all 27B params active). Hybrid architecture (Gated DeltaNet + Full Attention). MTP weight extraction uses `--no-quantize` for BF16 (849 MB).

### How auto-skip works

MTP adds an extra `mtp_forward` pass per generation step that doesn't parallelize across batch items. With batch=1, the extra pass is worthwhile (~80% acceptance = fewer total steps). With batch>1, the overhead dominates and hurts aggregate throughput.

The fix is simple: `_mtp_step()` checks `len(batch_gen.active_batch)` and falls back to the standard `_step()` when there are multiple active requests. This gives the best of both worlds:
- **Single user:** MTP active → +12-19% faster
- **Concurrent users:** MTP skipped → full batching throughput preserved

### Feature Verification

| Feature | 35B (port 1238) | 122B (port 1237) | 27B (port 1240) |
|---------|-----------------|------------------|------------------|
| Single tool call | ✅ Correct | ✅ Correct | ✅ Correct |
| Multi tool call (parallel) | ✅ 2 tools parsed | ✅ 2 tools parsed | ✅ 2 tools parsed |
| Thinking/reasoning parsing | ✅ `reasoning_content` separated | ✅ `reasoning_content` separated | ✅ `reasoning_content` separated |
| `finish_reason` | ✅ `tool_calls` / `stop` | ✅ `tool_calls` / `stop` | ✅ `tool_calls` / `stop` |

### Runtime MTP injection (`vllm_mlx/patches/qwen3_5_mtp.py`)
Since `mlx_lm`'s `qwen3_5.py` does not define MTP layers, this module dynamically creates the MTP module, loads weights, and monkey-patches the model class with `return_hidden`, `mtp_forward`, and `make_mtp_cache` methods. Supports both Dense and MoE (SwitchMLP) variants.

**Critical implementation notes:**
- MTP concat order is `[embedding, hidden]` (not `[hidden, embedding]`) — fc weight first half corresponds to embedding columns
- MTP expects **post-norm** hidden states (after final RMSNorm), not pre-norm
- MTP weights must be native BF16 — dequantized 4-bit weights give 0% acceptance due to baked-in quantization error

### MLLM batched generation (`vllm_mlx/mllm_batch_generator.py`)
Implements always-advance verified MTP strategy for the MLLM batched path:
- `_mtp_step()`: forward with `return_hidden` → sample primary → MTP draft → verify → accept/reject
- `_mtp_next()`: emits deferred MTP draft tokens in subsequent generation steps
- **Auto-skip:** when `len(active_batch) > 1`, bypasses MTP and falls back to standard `_step()` — eliminates concurrent throughput penalty
- Handles hybrid model caches (BatchKVCache for attention + ArraysCache for recurrent GatedDeltaNet layers)
- RNN state snapshot/restore for correct recurrent state management

### MTP weight extraction (`scripts/add_mtp_weights_qwen35.py`)
- Downloads only MTP-containing shards from HuggingFace (not entire model)
- Stacks 256 expert weights into SwitchLinear `(num_experts, output, input)` format
- Applies RMSNorm +1.0 shift (HF centered at 0 → MLX centered at 1)
- `--no-quantize` flag for BF16 output (recommended for best acceptance rate)

### Model loading pipeline
- `tokenizer.py`: Extended MTP detection for `mtp_num_hidden_layers` in `text_config`
- `text_model_from_vlm.py`: Post-load MTP injection for VLM text models
- `batched.py`: MTP injection and validation for MLLM models

## Test plan

- [x] Qwen3.5-35B-A3B-4bit: 79-80% MTP acceptance, +18% single-user speedup
- [x] Qwen3.5-122B-A10B-4bit: 75-76% MTP acceptance, +12% single-user speedup
- [x] Qwen3.5-27B-4bit (Dense): 78-80% MTP acceptance, +19% single-user speedup (30.1 → 35.9 tok/s)
- [x] Auto-skip: 8x concurrent = 307 tok/s (matches no-MTP baseline 302 tok/s)
- [x] Tool calls (single + parallel) work correctly with MTP
- [x] Thinking/reasoning parsing works correctly with MTP
- [ ] Verify no regression when MTP weights are absent (graceful fallback)

## Usage

```bash
# 1. Extract MTP weights (run once per model)
python scripts/add_mtp_weights_qwen35.py \
  --mlx-model-path ~/.cache/huggingface/hub/models--mlx-community--Qwen3.5-35B-A3B-4bit/snapshots/... \
  --source-model Qwen/Qwen3.5-35B-A3B \
  --no-quantize

# 2. Serve with MTP enabled (auto-skips when batch > 1)
vllm-mlx serve mlx-community/Qwen3.5-35B-A3B-4bit \
  --enable-mtp \
  --mllm \
  --continuous-batching
```

### DIFF
--- a/scripts/add_mtp_weights_qwen35.py
+++ b/scripts/add_mtp_weights_qwen35.py
@@ -1,0 +1,470 @@
+#!/usr/bin/env python3
+"""
+Add MTP (Multi-Token Prediction) weights to an existing MLX Qwen3.5 model.
+
+This script:
+1. Fetches the safetensors index from the original BF16 HuggingFace model
+2. Identifies shards containing MTP weights (mtp.* keys)
+3. Downloads only those shards via curl -C -
+4. Extracts MTP weights
+5. For MoE models: stacks expert weights (256×) into switch_mlp format
+6. Applies norm shift (HF weight → MLX weight+1.0) for RMSNorm keys
+7. Quantizes to match the MLX model's quantization scheme
+8. Saves as mtp/weights.safetensors (subdirectory avoids mlx_vlm glob)
+
+Supports both:
+- MoE models (Qwen3.5-122B-A10B, 35B-A3B): 256 experts, sparse MTP attention
+- Dense models (Qwen3.5-27B): full MTP with k/v projections and norms
+
+Usage:
+    python add_mtp_weights_qwen35.py --mlx-model-path PATH --source-model MODEL
+
+Requirements:
+    pip install mlx
+"""
+
+import argparse
+import json
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+
+# Known model configurations
+MODEL_CONFIGS = {
+    "Qwen/Qwen3.5-122B-A10B": {
+        "num_experts": 256,
+        "hidden_size": 3072,
+        "is_moe": True,
+    },
+    "Qwen/Qwen3.5-35B-A3B": {
+        "num_experts": 256,
+        "hidden_size": 2048,
+        "is_moe": True,
+    },
+    "Qwen/Qwen3.5-27B": {
+        "num_experts": 0,
+        "hidden_size": 5120,
+        "is_moe": False,
+    },
+}
+
+
+def find_snapshot_dir(model_path: str) -> Path:
+    """Find the latest snapshot directory in HF cache structure."""
+    snapshots_dir = Path(model_path) / "snapshots"
+    if not snapshots_dir.exists():
+        if (Path(model_path) / "config.json").exists():
+            return Path(model_path)
+        raise FileNotFoundError(f"No snapshots found in {model_path}")
+    snapshots = sorted(snapshots_dir.iterdir(), key=lambda p: p.stat().st_mtime)
+    if not snapshots:
+        raise FileNotFoundError(f"No snapshots in {snapshots_dir}")
+    return snapshots[-1]
+
+
+def fetch_shard_index(source_model: str, download_dir: Path) -> dict:
+    """Fetch model.safetensors.index.json from HuggingFace."""
+    index_url = f"https://huggingface.co/{source_model}/resolve/main/model.safetensors.index.json"
+    index_path = download_dir / "source_index.json"
+
+    print(f"Fetching shard index from {source_model}...")
+    result = subprocess.run(
+        ["curl", "-L", "-C", "-", "-o", str(index_path), index_url],
+        check=False,
+    )
+    if result.returncode != 0:
+        raise RuntimeError(f"Failed to fetch index: return code {result.returncode}")
+
+    with open(index_path) as f:
+        return json.load(f)
+
+
+def identify_mtp_shards(index: dict) -> tuple[dict[str, str], set[str]]:
+    """Identify which shards contain MTP weights.
+
+    Returns:
+        Tuple of (mtp_key_to_shard mapping, set of shard filenames to download)
+    """
+    weight_map = index.get("weight_map", {})
+    mtp_keys = {}
+    shards_needed = set()
+
+    for key, shard in weight_map.items():
+        if key.startswith("mtp."):
+            mtp_keys[key] = shard
+            shards_needed.add(shard)
+
+    return mtp_keys, shards_needed
+
+
+def download_shards(
+    shards: set[str], source_model: str, download_dir: Path
+) -> dict[str, Path]:
+    """Download required shards using curl with resume support."""
+    shard_paths = {}
+    for shard_name in sorted(shards):
+        shard_url = f"https://huggingface.co/{source_model}/resolve/main/{shard_name}"
+        shard_path = download_dir / shard_name
+
+        if shard_path.exists():
+            size_gb = shard_path.stat().st_size / 1e9
+            print(f"  {shard_name}: exists ({size_gb:.2f} GB)")
+            shard_paths[shard_name] = shard_path
+            continue
+
+        print(f"  Downloading {shard_name}...")
+        result = subprocess.run(
+            ["curl", "-L", "-C", "-", "-o", str(shard_path), shard_url],
+            check=False,
+        )
+        if result.returncode != 0:
+            raise RuntimeError(
+                f"Download failed for {shard_name}: code {result.returncode}"
+            )
+
+        size_gb = shard_path.stat().st_size / 1e9
+        print(f"  {shard_name}: {size_gb:.2f} GB")
+        shard_paths[shard_name] = shard_path
+
+    return shard_paths
+
+
+def extract_and_quantize_mtp_weights(
+    mtp_keys: dict[str, str],
+    shard_paths: dict[str, Path],
+    snapshot_dir: Path,
+    is_moe: bool,
+    num_experts: int,
+    no_quantize: bool = False,
+):
+    """Extract MTP weights from BF16 shards, optionally quantize, and save."""
+    import mlx.core as mx
+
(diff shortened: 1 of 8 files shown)


Thump604:
I have #223 open since March 24 covering MTP per-request routing in BatchedEngine. Your PR overlaps significantly on `batched.py` and `text_model_from_vlm.py`.

Your approach has some clear improvements over mine - the batch auto-skip (MTP off when batch>1) is a better design than my `is_batchable = False`, and the BF16 weight extraction script is useful. Your production benchmarks on M3 Ultra are also more thorough.

Rather than have two competing MTP PRs, can we coordinate? Options:

1. I close #223 and review yours instead
2. We merge the approaches into one PR (your broader scope + anything useful from mine)

I'd lean toward option 1 since your scope is larger and production-tested. Let me know what makes sense.

Thump604 (approved):
Thorough work. I've read the full diff (1566 lines across 7 files) and tested the architecture against our production setup. Approving with notes.

**What this does well:**

- `qwen3_5_mtp.py` is clean: runtime injection via class patching, proper VLM navigation (model -> language_model -> model), dequantize-to-BF16 for MTP weights (critical for acceptance rate).
- `_fixup_moe_mtp` correctly handles missing MoE checkpoint keys (k_proj/v_proj zeroed, gate/shared_expert_gate copied from last full-attention layer). This matches what I found independently.
- `install_mtp_mllm` in `mllm_batch_generator.py` is the right design: monkey-patch `_step` and `_next`, auto-skip when `len(active_batch) > 1`, deferred draft emission. Better than my #223 approach of disabling batching entirely.
- `add_mtp_weights_qwen35.py` does shard-level downloads (not full model), per-projection stacking to minimize peak memory, and norm shift handling. Solid script.
- VLM-aware `_needs_strict_false` avoids the double-load OOM on 256GB systems. Good catch.

**One concern:**

The MTP verify path uses exact match (`pred_list == draft_list`). My PR #1085 on mlx-lm shows probabilistic acceptance (`min(1, p_target/p_draft)`) gives 90% acceptance vs ~5% exact match at temp=0.6 with quantized weights, and 77-85% with your BF16 weights. Combining BF16 MTP + probabilistic acceptance could push to 95%+. Worth considering as a follow-up, since exact match leaves significant throughput on the table at non-zero temperature.

**Coordination:**

Closing my #223 in favor of this. Your scope is larger and production-tested.

janhilgard:
Thanks for the thorough review and the approval! Great to see the independent validation of the MoE key handling and VLM weight navigation.

Re: exact match vs probabilistic acceptance — that's a great point. Our current 78-85% acceptance rates are with temp=0 (greedy), where exact match works well. At non-zero temperatures the gap would be significant as you describe. Probabilistic acceptance from your mlx-lm #1085 is a natural follow-up — once that lands upstream we can switch the verify path. Filed as a TODO.

Thanks for closing #223 in favor of this — and for the per-request routing idea from that PR. Batch-level auto-skip works for now, but per-request MTP routing (text-only gets MTP even under concurrency) would be a nice improvement down the line.

janhilgard:
If you'd like to implement probabilistic acceptance directly in this PR, happy to add you as a collaborator so you can push to the branch. Let me know\!

Thump604:
Thanks for the offer — I'd like to take you up on that. Probabilistic acceptance at non-zero temperatures is where the real gains are for production use (we run temp=0.6 for coding). Our mlx-lm #1085 has the implementation but it's blocked behind #990.

Adding it directly here means it ships with the batch MTP support rather than waiting on the mlx-lm merge chain. I'll keep it minimal — just the speculative sampling verify path, no API surface changes.

janhilgard:
Thanks for the clean implementation! The log-space accept ratio approach is solid — minimal diff, no API changes, and degenerates to exact match at temp=0 as expected.

@waybarrios — this PR now has two approvals and covers all three Qwen3.5 variants (27B dense, 35B-A3B, 122B-A10B) with production benchmarks. Ready to merge when you get a chance.

Thump604:
Heads up — AirRunner just integrated probabilistic acceptance into mlx-lm #990 with an important refinement: for greedy sampling (temp=0), exact-match is the correct criterion, not the log-probability ratio. The ratio can exceed 1.0 on poorly-calibrated models even when the tokens don't match, causing false accepts. My commit in this PR always uses probabilistic regardless of temperature. Worth adding a guard: if the sampler is greedy (temp=0 or argmax), fall back to exact-match verification.

Thump604:
Pushed greedy guard for probabilistic acceptance (`15062db`). At temp=0, the log-prob ratio `p_target(d)/p_draft(d)` can exceed 1 even when the draft token differs from the target argmax -- this breaks correctness for greedy decoding on poorly-calibrated models.

The guard: if `temperature == 0.0`, use exact-match (argmax comparison). Otherwise use probabilistic acceptance as before. Detection happens at `_install_mtp` time since the batch generator is recreated when sampling params change.

Credit to AirRunner for identifying this in #990.

Thump604:
@waybarrios, light bump. This PR has been mergeable on current main since Apr 3, my formal approval is on file (review state above), and janhilgard tagged you as ready to merge after the probabilistic acceptance commits landed.

It ships BF16 MTP support and the speculative sampling verify path together in one PR. That gives production users the full picture: janhilgard`s 78-85% acceptance at greedy + BF16, plus probabilistic acceptance at temp>0 which prior mlx-lm experiments showed lifts acceptance from ~5% to 90% on quantized weights. Otherwise users would wait on the mlx-lm #990 chain to unblock probabilistic acceptance separately.

#223 is closed in favor of this. Happy to coordinate the merge whenever convenient.

janhilgard:
@Thump604 Superseded — MTP speculative decoding (`qwen3_5_mtp.py`, `--enable-mtp`) is already in `main` via #278. Closing.

PR #248, opened 2026-04-02
fix: replace manual decode loop with pipelined generation in SpecPrefill Phase 4

### LINKED ISSUE #247: SpecPrefill Phase 4 in SimpleEngine: very slow decode using `model()`
## Observed Problem

After SpecPrefill completes (scoring + sparse prefill), the Phase 4 decode loop uses manual `model()` + `mx.eval(y)` calls. On my setup, this produces **~0.3-0.4 tok/s** decode speed, while normal generation achieves 30+ tok/s.

### Resource observations

| Metric | Phase 4 decode | Normal generation |
|--------|-------------------|-------------------------------------|
| Wired memory | 30-40GB | 70-75GB |
| GPU idle | 50-65% | 0-3% |
| GPU power | <1W | 90-93W |
| Decode speed | 0.3-0.4 tok/s | 30+ tok/s |

Model weights are ~60GB (Qwen3.5-122B-A10B-4bit). During Phase 4, wired memory never reaches the full model size, and GPU is mostly idle.

## Replace with `stream_generate` fixes the decode speed

Replacing the manual decode loop with `mlx_lm.stream_generate`, passing the sparse-prefill cache via `prompt_cache`. After this change, decode speed returned to normal levels, with wired memory and GPU utilization matching normal generation behavior.

I verified with `temp=0` greedy decoding that both methods produce identical token sequences.

```python
sampler = make_sampler(temp=temperature, top_p=top_p)
first_token_id = sampler(logits[:, -1, :]).item()
first_text = tokenizer.decode([first_token_id])
eos_id = tokenizer.eos_token_id

results = [
    SimpleNamespace(
        text=first_text,
        finish_reason="stop" if first_token_id == eos_id else None,
    )
]

if first_token_id != eos_id:
    for chunk in mlx_stream_generate(
        model, tokenizer,
        prompt=mx.array([first_token_id]),
        max_tokens=max_tokens - 1,
        sampler=sampler,
        prompt_cache=cache,
    ):
        results.append(
            SimpleNamespace(
                text=chunk.text,
                finish_reason=getattr(chunk, "finish_reason", None),
            )
        )
```

Not sure if this is a proper fix. Is the difference in GPU utility expected between `model()` and `stream_generate()` ? Has anyone experienced similar issues?

## Environment

- Mac Studio M1 Ultra (64-core GPU, 128GB)
- mlx 0.31.1, mlx-lm 0.31.1
- vllm-mlx 0.2.7
- Target: Qwen3.5-122B-A10B-4bit, Draft: Qwen3.5-2B-4bit

Vigilans:
## Summary

Fixes #247. Try to make my first reviewed contribution here :)

SpecPrefill Phase 4 called `model()` + `mx.eval()` in a plain Python loop 
without `mx.async_eval` pipelining, causing the GPU to idle between steps 
(~0.3 tok/s vs 30+ tok/s normal decode).

- Extend `MLXLanguageModel.stream_generate()` to accept a `prompt_cache` 
  parameter and `mx.array`/`list[int]` prompt types, so SpecPrefill can 
  hand off the pre-populated sparse-prefill cache to the engine's standard 
  pipelined decode path
- Replace the manual decode loop in both SpecPrefill paths:
  - Non-MLLM: hand off to `self._model.stream_generate(prompt_cache=cache)`
  - MLLM text-only: hand off to `mlx_lm.stream_generate(prompt_cache=bc)` 
    (matching its normal path)
- First-token EOS check is preserved since `stream_generate` does not 
  validate whether the prompt itself is EOS

### Handoff target: Model layer `self._model.stream_generate`

There are three layers that can drive decode: Engine layer 
(`SimpleEngine.stream_generate`), Model layer 
(`MLXLanguageModel.stream_generate`), and `mlx_lm.stream_generate` directly.

The Engine layer is not suitable because `_stream_generate_specprefill` already 
holds `self._generation_lock` (asyncio.Lock, non-reentrant) — calling back 
into `SimpleEngine.stream_generate` would deadlock. It would also re-run 
SpecPrefill eligibility checks and re-tokenize the prompt unnecessarily 
(though no infinite recursion: the single-token prompt falls below the 
threshold).

The Model layer (`self._model.stream_generate`) is the right fit. The 
non-MLLM normal decode path already calls it (line 362), so routing 
SpecPrefill Phase 4 through the same method maintains symmetry and reuses 
its stop sequence handling. For the MLLM text-only path, `self._text_model` 
is a raw `mlx_lm TextModel` (nn.Module built by `build_text_model`) with no 
`stream_generate` method, so it calls `mlx_lm.stream_generate` directly — 
again symmetric with the MLLM normal path (line 1090).

### Extending `MLXLanguageModel.stream_generate` interface

The Model layer originally only accepted `prompt: str` and had no way to 
pass a pre-populated cache. Two changes were needed:

1. **`prompt_cache` parameter**: forwarded via kwargs to 
   `mlx_lm.stream_generate`, which passes it through to `generate_step`. 
   When `prompt_cache` is not None, `generate_step` skips creating a new 
   cache and uses the provided one directly.
2. **`prompt` type broadened to `str | mx.array | list[int]`**: Phase 4 
   passes `mx.array([first_token_id])` as prompt. `mlx_lm.stream_generate` 
   already accepts all three types natively; the only adaptation needed was 
   gating `tokenizer.encode()` behind an `isinstance(prompt, str)` check 
   for the prompt token count.

## Test plan

- [x] Existing `test_model_stream_generate` still passes
- [x] New tests: `test_model_stream_generate_with_prompt_cache`, 
      `test_model_stream_generate_with_list_prompt`
- [x] End-to-end: SpecPrefill activated on 12k-token prompt, Phase 4 
      decode runs at normal speed

### DIFF
--- a/tests/test_llm.py
+++ b/tests/test_llm.py
@@ -100,6 +100,58 @@ def test_model_stream_generate(small_model_name):
 
 
 @pytest.mark.slow
+def test_model_stream_generate_with_prompt_cache(small_model_name):
+    """Test streaming generation with pre-populated prompt_cache."""
+    pytest.importorskip("mlx_lm")
+    import mlx.core as mx
+    from mlx_lm.models.cache import make_prompt_cache
+
+    from vllm_mlx.models.llm import MLXLanguageModel
+
+    model = MLXLanguageModel(small_model_name)
+    model.load()
+
+    # Pre-populate cache by running a prefill
+    cache = make_prompt_cache(model.model)
+    tokens = model.tokenizer.encode("Hello")
+    model.model(mx.array([tokens]), cache=cache)
+    mx.eval([c.state for c in cache])
+
+    # Generate from a single token with the pre-populated cache
+    prompt_token = mx.array([tokens[-1]])
+    chunks = list(
+        model.stream_generate(
+            prompt=prompt_token,
+            max_tokens=10,
+            prompt_cache=cache,
+        )
+    )
+
+    assert len(chunks) > 0
+    assert any(chunk.finished for chunk in chunks)
+    # prompt_tokens should reflect the single-token prompt, not the full string
+    assert chunks[0].prompt_tokens == 1
+
+
+@pytest.mark.slow
+def test_model_stream_generate_with_list_prompt(small_model_name):
+    """Test streaming generation with list[int] prompt."""
+    pytest.importorskip("mlx_lm")
+
+    from vllm_mlx.models.llm import MLXLanguageModel
+
+    model = MLXLanguageModel(small_model_name)
+    model.load()
+
+    token_ids = model.tokenizer.encode("Hello")
+    chunks = list(model.stream_generate(prompt=token_ids, max_tokens=10))
+
+    assert len(chunks) > 0
+    assert any(chunk.finished for chunk in chunks)
+    assert chunks[0].prompt_tokens == len(token_ids)
+
+
+@pytest.mark.slow
 def test_model_chat(small_model_name):
     """Test chat interface."""
     pytest.importorskip("mlx_lm")
--- a/vllm_mlx/engine/simple.py
+++ b/vllm_mlx/engine/simple.py
@@ -1186,6 +1186,7 @@ class SimpleEngine(BaseEngine):
             from types import SimpleNamespace
 
             import mlx.core as mx
+            from mlx_lm import stream_generate as mlx_stream_generate
             from mlx_lm.models.cache import make_prompt_cache
             from mlx_lm.sample_utils import make_sampler
 
--- a/vllm_mlx/engine/simple.py
+++ b/vllm_mlx/engine/simple.py
@@ -787,41 +787,37 @@ class SimpleEngine(BaseEngine):
                     t_prefill,
                 )
 
-                # Phase 4: Generate (simple autoregressive, no MTP)
+                # Phase 4: Generate via engine's standard pipelined path
                 sampler = make_sampler(temp=temperature, top_p=top_p)
                 _cancel_check()
+                first_token_id = sampler(logits[:, -1, :]).item()
+                first_text = tokenizer.decode([first_token_id])
                 eos_id = tokenizer.eos_token_id
-                y = sampler(logits[:, -1, :])
-                mx.eval(y)
 
-                results = []
-                generated_ids = []
-                prev_decoded = ""
-
-                for _ in range(max_tokens):
-                    _cancel_check()
-                    tok_id = y.item()
-                    generated_ids.append(tok_id)
-
-                    decoded = tokenizer.decode(generated_ids)
-                    new_text = decoded[len(prev_decoded) :]
-                    prev_decoded = decoded
-
-                    is_eos = tok_id == eos_id
-                    results.append(
-                        SimpleNamespace(
-                            text=new_text,
-                            finish_reason="stop" if is_eos else None,
-                        )
+                results = [
+                    SimpleNamespace(
+                        text=first_text,
+                        finish_reason="stop" if first_token_id == eos_id else None,
                     )
+                ]
 
-                    if is_eos:
-                        break
-
-                    _cancel_check()
-                    logits = model(y.reshape(1, -1), cache=cache)
-                    y = sampler(logits[:, -1, :])
-                    mx.eval(y)
+                if first_token_id != eos_id:
+                    for chunk in self._model.stream_generate(
+                        prompt=mx.array([first_token_id]),
+                        max_tokens=max_tokens - 1,
+                        temperature=temperature,
+                        top_p=top_p,
+                        stop=stop,
+                        prompt_cache=cache,
+                    ):
+                        _cancel_check()
+                        new_text = chunk.text if hasattr(chunk, "text") else str(chunk)
+                        results.append(
+                            SimpleNamespace(
+                                text=new_text,
+                                finish_reason=getattr(chunk, "finish_reason", None),
+                            )
+                        )
 
                 return results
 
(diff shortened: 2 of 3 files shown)


Vigilans:
btw, I noticed SpecPrefill currently only triggers on the streaming path (`stream_chat` → `stream_generate`). The non-streaming `chat()` method calls `self._model.chat()` directly and bypasses `stream_generate` entirely, so SpecPrefill never activates for ·stream=False` requests. Is this intentional, or it will be implemented in the future?

Thump604:
@Vigilans, @waybarrios: independent technical review of this PR.

## Verification of the fix

Confirmed against current upstream main (b4fa030). The diff modifies two `_run_specprefill` call sites in `vllm_mlx/engine/simple.py` (the SimpleEngine streaming path and the BatchedEngine SpecPrefill path) to:

1. Sample the first token from SpecPrefill`s final logits (preserved behavior)
2. Hand off subsequent decode to `mlx_lm.stream_generate` with the SpecPrefill-prepared `prompt_cache`

The new `prompt_cache` parameter is added to `MLXLanguageModel.stream_generate` in `vllm_mlx/models/llm.py:177` and forwarded into mlx_lm`s `stream_generate` via the existing `mtp_kwargs` dict.

## Technical correctness

The root cause analysis in #247 holds. The manual `model()` + `mx.eval(y)` decode loop bypasses the pipelined kernel scheduling that `mlx_lm.stream_generate` provides, which is why the GPU stays idle (<1W vs 90-93W) and decode crawls at 0.3 tok/s vs 30+. The fix routes through the standard stream_generate path which gets full kernel pipelining and proper async eval batching.

Both `_run_specprefill` call sites get the same treatment, which is correct. Both BatchedEngine and SimpleEngine had the same bug.

## Backward compatibility

`prompt_cache=None` is the default, existing callers behave identically. The `prompt: Union[str, mx.array, list[int]]` signature widening is safe. Existing string callers still work, new token-list and array callers are needed for the SpecPrefill handoff.

## Minor observations (not blocking)

1. `mtp_kwargs` is now a slight misnomer since it also carries `prompt_cache` for non-MTP cases. Renaming to `extra_kwargs` would be clearer in a follow-up.
2. First-token decode is duplicated between the manual sample-and-decode block and what stream_generate would do internally. This is necessary because the first token must come from SpecPrefill`s final logits, not from a fresh forward pass. A code comment noting this would help future readers, but the logic is correct.
3. The two new tests in `tests/test_llm.py` exercise the new `prompt_cache` and `list[int] prompt` paths. An end-to-end SpecPrefill+stream_generate test that asserts the decode speed regression does not return would close the loop, but the production evidence in #247 is hard.

## On the related observation in your comment

The non-streaming `chat()` method bypassing SpecPrefill is a separate, related gap. It is on a different code path (the bypass happens before the streaming pipeline is reached) and would need its own change. The two issues are complementary and the fixes would compose well if both land.

## Recommendation

Merge candidate. Real fix to a real performance regression, hard production evidence, sound implementation, reasonable tests.

Thump604 (approved):
This is a correct fix and a big one. The manual `model() + mx.eval(y)` per-token loop foregoes MLX's kernel pipelining, which is exactly what gets SpecPrefill Phase 4 stuck at ~0.3 tok/s while the same model at the same context runs ~30 tok/s through a normal decode. Routing through `mlx_lm.stream_generate(prompt=mx.array([first_token]), prompt_cache=cache)` is the clean fix.

On your note about SpecPrefill currently only triggering on the streaming path: that's a separate accumulator gap I'm working on. Once both this PR and that refactor land, non-streaming /v1/chat/completions and /v1/completions will also reach Phase 4, and this fix becomes load-bearing there too. Worth mentioning in the PR description if you want to foreshadow it.

Post-merge I'd like to run this against a 16K/64K/128K needle harness on Qwen 3.5 122B + 2B draft as an independent regression signal. If that's useful I can post the run output here as a follow-up comment.

Approving.

Thump604:
Hey @Vigilans - the pipelined decode approach is correct and we've validated the same pattern independently on our local staging. Big speedup. Currently conflicting with main - can you rebase? Good to merge after that.

Vigilans:
> Hey @Vigilans - the pipelined decode approach is correct and we've validated the same pattern independently on our local staging. Big speedup. Currently conflicting with main - can you rebase? Good to merge after that.

Rebased on latest main and fixed the ruff lint errors from the [previous CI run](https://github.com/waybarrios/vllm-mlx/actions/runs/23910925523).

janhilgard (approved):
+1, agreeing with Thump604's approval. The fix is correct and well-scoped.

The manual `model() + mx.eval(y)` per-token loop was indeed the bottleneck -- it serializes every decode step and prevents MLX's kernel pipelining from overlapping compute and memory operations. Routing through `mlx_lm.stream_generate(prompt=mx.array([first_token_id]), prompt_cache=cache)` is the right fix.

A few minor observations (none blocking):

1. **`stream_generate` signature broadening in `llm.py`** -- accepting `Union[str, mx.array, list[int]]` is a nice generalization. The `TYPE_CHECKING` guard for `import mlx.core as mx` is correct since `mx` is only needed for the type annotation.

2. **First token is decoded standalone** (`self._text_tokenizer.decode([first_token_id])`) before handing off to `stream_generate`. This means the first token's text may differ slightly from what `stream_generate`'s internal detokenizer would produce (e.g., for tokens that need surrounding context for correct decode, like byte-fallback tokens). In practice this is unlikely to matter for the first token after SpecPrefill, but worth being aware of.

3. **The MLLM path** (second `_run_specprefill` at ~line 1204) uses `mlx_stream_generate` (the raw `mlx_lm` function) directly, while the LLM path uses `self._model.stream_generate` (the wrapper). This asymmetry is fine since the MLLM path needs the raw model + tokenizer, but it means MTP is not available in the MLLM SpecPrefill path (the wrapper handles MTP kwargs). Not a regression since MTP was not available there before either.

CI has no checks running -- would be good to trigger a run to confirm green before merge. Branch is mergeable with no conflicts.

Thump604:
This is approved by both reviewers and ready to go. Could you push a rebase onto current main to trigger CI? The branch hasn't had checks run.

PR #253, opened 2026-04-04
fix(server): integrate tool call parser into reasoning parser streaming path

mxl:
Running Qwen 3.5 model results in printing tool calls as raw xml and not applying actual tools.

## Summary
- run the streaming tool-call parser in the reasoning-parser path after reasoning content has been stripped from streamed output
- suppress tool-call markup from normal content chunks and emit structured `tool_calls` chunks when parsed tool calls are detected
- preserve the `tool_calls` finish reason on the final streamed chunk when generation ends with parsed tool calls

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 """Tests for the OpenAI-compatible API server."""
 
+import json
 import platform
 import sys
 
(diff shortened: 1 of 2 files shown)


waybarrios (commented):
Clean fix for a real bug — tool calls were leaking as raw XML when the reasoning parser was active. The approach is correct and tests are solid.

One minor bug:

### `request.model` vs `_model_name` inconsistency

The new tool_chunk in the reasoning path uses `model=request.model`, but every other chunk in the function (including the equivalent tool_calls chunk in the standard path) uses `model=_model_name`:

```python
# New code (reasoning path):
tool_chunk = ChatCompletionChunk(
    id=response_id,
    model=request.model,      # ← client-provided value
    ...
)

# Standard path (line ~2252) and all other chunks:
chunk = ChatCompletionChunk(
    id=response_id,
    model=_model_name,         # ← actual served model name
    ...
)
```

These can differ when `--served-model-name` is set. Should be `_model_name` for consistency.

waybarrios:
Pushed a small fix (1d16507) — was using `request.model` instead of `_model_name` in the reasoning tool chunk. Consistent with the rest of the function now.

PR #254, opened 2026-04-04
feat: add Gemma 4 tool call parser

keegoid:
## Summary

- Adds a new `gemma4_tool_parser.py` that handles Gemma 4's unique tool calling format (`<|tool_call>call:name{key:<|"|>value<|"|>}<tool_call|>`)
- Implements a state-machine converter for Gemma's custom argument syntax (unquoted keys, `<|"|>` string delimiters) to valid JSON
- Integrates Gemma 4 detection into the auto parser (both streaming and non-streaming paths)
- Adds `gemma4` to CLI `--tool-call-parser` choices
- 21 tests covering arg conversion, nested objects, arrays, hyphenated tool names, streaming, think tag stripping, and auto-detection

### Background

Gemma 4 models use a non-standard tool call format defined in their chat template. Unlike other model families that use JSON or XML, Gemma 4 uses special tokens (`<|tool_call>`, `<tool_call|>`, `<|"|>`) for structure and string quoting. This parser converts that format to OpenAI-compatible `tool_calls` responses.

Tested end-to-end with `mlx-community/gemma-4-31b-it-8bit` on Apple Silicon (M5 Max 128GB):
- Single and multi-tool calls
- Multi-turn with tool responses
- Vision + tool calling combined
- Streaming

### Usage

```bash
vllm-mlx serve mlx-community/gemma-4-31b-it-8bit --port 8000 --mllm   --enable-auto-tool-choice --tool-call-parser gemma4
```

## Test plan

- [x] All 21 new Gemma 4 parser tests pass
- [x] All 78 existing tool parser tests pass (99 total, 0 failures)
- [x] End-to-end tool calling verified against live Gemma 4 31B model
- [x] Vision + tool calling combo verified
- [x] Streaming tool call buffering verified
- [ ] Gemma 4 E4B model (not yet tested, same format expected)

### DIFF
--- a/tests/test_gemma4_parser.py
+++ b/tests/test_gemma4_parser.py
@@ -1,0 +1,237 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for the Gemma 4 tool call parser."""
+
+import json
+
+from vllm_mlx.tool_parsers.gemma4_tool_parser import (
+    Gemma4ToolParser,
+    _gemma_args_to_json,
+)
+
+# The Gemma 4 escape token for quotes
+Q = '<|"|>'
+
+
+def test_args_to_json_simple_strings():
+    raw = f"location:{Q}San Francisco{Q},unit:{Q}celsius{Q}"
+    result = "{" + _gemma_args_to_json(raw) + "}"
+    parsed = json.loads(result)
+    assert parsed == {"location": "San Francisco", "unit": "celsius"}
+
+
+def test_args_to_json_numbers():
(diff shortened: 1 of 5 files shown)


waybarrios:
The auto-detection logic you built for AutoToolParser was the right call. I pulled that approach into #269 so Gemma 4 tool calls work out of the box with --enable-auto-tool-choice. The standalone parser is also covered there. Thanks for flagging the RotatingKVCache issue on #256 too, that was a key catch. Closing in favor of #269.

PR #255, opened 2026-04-05
Preserve raw chat output for reasoning parsing

Thump604:
## Summary

This fixes a server/engine contract bug for reasoning-capable chat paths.

Before this change, chat output could be cleaned inside the engine before the server reasoning parser ran. For models like Gemma 4, that turns raw output such as:

`<|channel>thought\n...<channel|>Final answer`

into partially-sanitized visible text before parsing. In practice, hidden reasoning can leak into `message.content`, and the server no longer has the original text it needs to split reasoning from final content correctly.

## Fix

- add `raw_output=True` on parser-aware server chat paths
- preserve raw output in `SimpleEngine.chat()`
- preserve raw output in `BatchedEngine.generate()` / `stream_generate()`
- keep cleanup on the server side after reasoning parsing instead of before it

## Why this matters

Gemma 4 was the clearest failure case because its hidden reasoning starts in a channel token format. Once that prefix is sanitized too early, the reasoning parser cannot reliably identify it as hidden reasoning anymore.

This change keeps the engine/server boundary clean:
- engine can still return cleaned output by default
- server can explicitly request raw output when it needs to run reasoning/tool parsing on the original text

## Tests

Added targeted regressions for:
- raw-output preservation in batched MLLM scheduler paths
- `/v1/chat/completions` requesting raw output and returning split reasoning/content correctly

Tested with:
- `tests/test_batched_engine_text_scheduler.py`
- `tests/test_server.py -k ChatCompletionRawOutput`

### DIFF
--- a/tests/test_batched_engine_text_scheduler.py
+++ b/tests/test_batched_engine_text_scheduler.py
@@ -1,0 +1,76 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Focused regressions for BatchedEngine raw-output handling."""
+
+import asyncio
+from unittest.mock import AsyncMock, MagicMock
+
+from vllm_mlx.engine.batched import BatchedEngine
+from vllm_mlx.request import RequestOutput
+
+
+def _make_mllm_engine() -> BatchedEngine:
+    engine = BatchedEngine("dummy-model", force_mllm=True)
+    engine._loaded = True
+    return engine
+
+
+def test_generate_preserves_raw_output_for_mllm_scheduler():
+    async def _run():
+        engine = _make_mllm_engine()
+        engine._mllm_scheduler = AsyncMock()
+        engine._mllm_scheduler.generate.return_value = RequestOutput(
+            request_id="req-1",
+            output_text="<|channel>thought\nplan<channel|>Final answer",
+            prompt_tokens=9,
+            completion_tokens=3,
+            finished=True,
+            finish_reason="stop",
+        )
+
+        output = await engine.generate(prompt="hello", raw_output=True)
+
+        assert output.text == "<|channel>thought\nplan<channel|>Final answer"
+
+    asyncio.run(_run())
+
+
+def test_stream_generate_preserves_raw_output_for_mllm_scheduler():
+    async def _run():
+        engine = _make_mllm_engine()
+        engine._mllm_scheduler = MagicMock()
+        engine._mllm_scheduler.add_request_async = AsyncMock(return_value="req-1")
+
+        async def _stream_outputs(_request_id):
+            yield RequestOutput(
+                request_id="req-1",
+                output_text="<|channel>thought\n",
+                new_text="<|channel>thought\n",
+                prompt_tokens=4,
+                completion_tokens=1,
+                finished=False,
+            )
+            yield RequestOutput(
+                request_id="req-1",
+                output_text="<|channel>thought\nplan<channel|>Final answer",
+                new_text="plan<channel|>Final answer",
+                prompt_tokens=4,
+                completion_tokens=2,
+                finished=True,
+                finish_reason="stop",
+            )
+
+        engine._mllm_scheduler.stream_outputs = _stream_outputs
+
+        outputs = []
+        async for output in engine.stream_generate(prompt="hello", raw_output=True):
+            outputs.append((output.text, output.new_text))
+
+        assert outputs == [
+            ("<|channel>thought\n", "<|channel>thought\n"),
+            (
+                "<|channel>thought\nplan<channel|>Final answer",
+                "plan<channel|>Final answer",
+            ),
+        ]
+
+    asyncio.run(_run())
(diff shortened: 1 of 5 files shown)


Thump604:
Quick merge signal here: this one is clean now. The CI blockers were closed by the thread-safe test cleanup plus `raw_output` parity for `/v1/messages`, and the current branch is mergeable with the full matrix green.

This is the raw-output contract we need for reasoning/tool parsing, so if you're doing a fast sweep, this one is ready.

Thump604:
I rechecked this after syncing the current official Gemma 4 `chat_template.jinja` files into the live local model assets on April 11, 2026.

That template update changes the prompt shape and lowered prompt-token counts on the live Gemma canaries, but it does **not** remove the need for this raw-output contract. The evidence split is still the same on `:8080`: first-turn tool call works, second-turn OpenAI-style `role="tool"` replay works, and plain reasoning split still returns separate `content` + `reasoning_content`.

So my read is: this PR is still valid and still needed for parser-aware chat paths. The template update is an input-format improvement; this PR is about preserving unsanitized model output until the server-side reasoning/tool parsers have finished. That remains a separate contract issue.

Current status from my side: conceptually still green, but it now needs a restack on current `main` before it can move.

janhilgard (commented):
This PR addresses a real and important problem: reasoning parsers need the raw model output (with channel tokens, think tags, etc.) before `clean_output_text()` strips them. The fix is conceptually correct — add a `raw_output=True` flag that bypasses cleaning in the engine, and set it on all parser-aware server paths.

**Strengths:**
- The engine/server boundary is kept clean: engine defaults to cleaned output, server can opt into raw when needed
- Both `SimpleEngine.chat()` and `BatchedEngine.generate()`/`stream_generate()` are covered
- All four server entry points are covered: `/v1/chat/completions` (streaming + non-streaming), `/v1/messages` (streaming + non-streaming)
- Good regression tests for both the batched MLLM path and the server chat completion path

**Issues:**

1. **`raw_output=True` is now unconditional on all server paths, not just reasoning-parser paths.** Looking at `server.py`, `chat_kwargs["raw_output"] = True` is set in `create_chat_completion`, `create_anthropic_message`, `_stream_anthropic_messages`, and `stream_chat_completion` — always, regardless of whether `_reasoning_parser` or `_tool_call_parser` is configured. This means `clean_output_text()` is now *never* called for any server-initiated request. Is this intentional? If so, the PR title/description should clarify that `clean_output_text` is effectively dead code on the server path now. If not, the flag should be conditional: `chat_kwargs["raw_output"] = bool(_reasoning_parser or _tool_call_parser)`.

2. **`clean_output_text` on `new_text` in streaming**: In `BatchedEngine.stream_generate()`, `clean_output_text` is now applied (or not) to both `output.output_text` and `output.new_text`. When `raw_output=False`, applying `clean_output_text` to incremental `new_text` chunks could produce unexpected results if the special tokens span chunk boundaries. Was this already a problem before this PR, or is it new? Worth a comment either way.

3. **Test cleanup concern**: The test `TestChatCompletionRawOutput` manually swaps `srv._engine`, `srv._model_name`, and `srv._reasoning_parser` global state, then restores it in a `finally` block plus `asyncio.set_event_loop(asyncio.new_event_loop())`. This is fragile — if the module-level globals change names or new ones are added, the test silently breaks. Consider using `unittest.mock.patch` for cleaner isolation.

4. **Missing `raw_output` propagation in `SimpleEngine.stream_chat`**: `SimpleEngine.chat()` handles `raw_output` but I do not see the same treatment in `SimpleEngine.stream_chat()` / the streaming text path. If a streaming request goes through `SimpleEngine`, will it still get cleaned output?

Item 1 is the most important — please clarify whether the intent is to disable `clean_output_text` unconditionally or only when a parser is present.

Thump604:
Thanks for the detailed review. Addressing each point:

1. **Unconditional `raw_output=True`** — This is intentional. The server always needs the raw output now because reasoning parsing happens at the server layer, not the engine layer. `clean_output_text()` strips channel tokens that reasoning parsers need to see. When no parser is configured, the raw output is returned as-is, which is the correct behavior for models that don't emit special tokens. Making it conditional on parser presence would re-introduce the bug for models where we later add a parser. The cleaner contract is: engine returns raw, server decides what to clean.

2. **`clean_output_text` on incremental `new_text` chunks** — This was already a problem before this PR (chunk boundaries could split special tokens). The PR doesn't change the chunking behavior, just controls whether cleaning is applied. The real fix for chunk-boundary cleaning would be a stateful streaming cleaner, which is out of scope here.

3. **Test cleanup** — Fair point. I'll switch to `unittest.mock.patch` in a follow-up to reduce fragility.

4. **`SimpleEngine.stream_chat`** — Good catch. The streaming text path in SimpleEngine should also respect `raw_output`. I'll add that in a follow-up commit on this PR.

Thump604:
Superseded by #358 — narrower, single-commit implementation rebased on current main. Covers both SimpleEngine and BatchedEngine with raw_output parameter, plus server-side double-clean fix.

PR #213, opened 2026-03-23
feat: full sampling parameter support (top_k, min_p, presence_penalty, repetition_penalty)

Thump604:
## Summary

Pass all OpenAI-compatible sampling parameters through to mlx-lm's `make_sampler` and `make_logits_processors`. Previously only `temperature`, `top_p`, and `max_tokens` reached the engine — `top_k`, `min_p`, `presence_penalty`, and `repetition_penalty` were silently dropped.

## Changes

- **api/models.py**: Add `top_k`, `min_p`, `presence_penalty`, `repetition_penalty` to `ChatCompletionRequest` and `CompletionRequest`
- **request.py**: Add `presence_penalty` to `SamplingParams` dataclass
- **server.py**: Extract and pass all params in every code path (6 locations), log all params on request
- **models/llm.py**: Build sampler with `top_k`/`min_p`, build `logits_processors` for `presence_penalty`/`repetition_penalty`
- **engine/simple.py**: Fix `enable_thinking` to read `VLLM_MLX_ENABLE_THINKING` env var instead of hardcoding based on model name

## Motivation

Qwen 3.5 requires [specific sampling profiles](https://unsloth.ai/docs/models/qwen3.5) with all 7 parameters set. Without `presence_penalty`, accuracy drops to near-zero on evals. Without `top_k=20`, sampling doesn't match the model's training distribution.

## Test plan

- [x] All 4 Unsloth Qwen 3.5 profiles tested on 122B model
- [x] All 7 params logged in server stderr for every request
- [x] `presence_penalty=1.5` works without crash (requires mlx-lm MTP fix, separate PR)
- [x] `enable_thinking` toggle via env var verified (on/off)
- [x] Black + ruff clean

PR #262, opened 2026-04-06
Add per-request enable_thinking API parameter

janhilgard:
## Summary

- Adds `enable_thinking` field to `ChatCompletionRequest` for per-request control of thinking/reasoning mode
- When `false`: chat template rendered without `<think>` injection, reasoning parser bypassed
- When `true` or omitted: existing behavior preserved (thinking enabled by default)

Supported across all engine paths: SimpleEngine, BatchedEngine, MLXMultimodalLM.

## Motivation

Models like Qwen3/3.5 support `enable_thinking=False` in their chat templates to skip the thinking phase. This is useful when:
- Low latency is more important than reasoning quality
- Clients want direct answers without `<think>` overhead
- Applications need to toggle thinking per-request (e.g., simple vs complex questions)

Without this change, `enable_thinking` is hardcoded to `True` with no way to override from the API.

## Usage

```bash
# Disable thinking
curl /v1/chat/completions -d '{"enable_thinking": false, "messages": [...]}'

# OpenAI Python client
client.chat.completions.create(
    extra_body={"enable_thinking": False},
    messages=[...]
)
```

## Changes

| File | Change |
|------|--------|
| `api/models.py` | Add `enable_thinking: bool \| None = None` field |
| `server.py` | Pass to engine kwargs + bypass reasoning parser when `False` |
| `engine/simple.py` | Read from kwargs in `stream_chat` and `_stream_generate_text` |
| `engine/batched.py` | Add to `_apply_chat_template`, pass from `chat`/`stream_chat` |
| `models/mllm.py` | Pop from kwargs, pass to `get_chat_template` calls |

## Test plan

- [x] `enable_thinking=true` produces reasoning content (Qwen3.5)
- [x] `enable_thinking=false` produces direct answer without reasoning
- [x] Default (omitted) preserves existing behavior
- [x] Streaming mode works correctly for both values
- [x] Tested on SimpleEngine (port 1237) and BatchedEngine (port 1238)
- [x] Tested on MLLM path (port 1240)

### DIFF
--- a/vllm_mlx/api/models.py
+++ b/vllm_mlx/api/models.py
@@ -178,6 +178,8 @@ class ChatCompletionRequest(BaseModel):
     specprefill: bool | None = None
     # SpecPrefill: per-request keep percentage (0.0-1.0, None = use server default)
     specprefill_keep_pct: float | None = None
+    # Enable/disable thinking mode (None = server default, typically True)
+    enable_thinking: bool | None = None
 
 
 class AssistantMessage(BaseModel):
--- a/vllm_mlx/engine/batched.py
+++ b/vllm_mlx/engine/batched.py
@@ -12,6 +12,7 @@ LLM engine), so text-only requests must also be routed through it.
 """
 
 import logging
+import os
 from collections.abc import AsyncIterator
 from typing import Any
 
@@ -335,6 +336,7 @@ class BatchedEngine(BaseEngine):
         messages: list[dict[str, Any]],
         tools: list[dict] | None = None,
         num_images: int = 0,
+        enable_thinking: bool | None = None,
     ) -> str:
         """Apply chat template to messages.
 
@@ -363,9 +365,15 @@ class BatchedEngine(BaseEngine):
             if self._is_mllm and num_images > 0:
                 messages = self._prepare_mllm_messages(messages)
 
+            # Per-request enable_thinking override; fall back to env var / default True.
+            if enable_thinking is None:
+                enable_thinking = os.environ.get(
+                    "VLLM_MLX_ENABLE_THINKING", "true"
+                ).lower() in ("true", "1", "yes")
             template_kwargs = {
                 "tokenize": False,
                 "add_generation_prompt": True,
+                "enable_thinking": enable_thinking,
             }
             if tools:
                 template_kwargs["tools"] = tools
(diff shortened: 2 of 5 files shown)


Thump604:
@waybarrios, @janhilgard: coordination note with my #213.

#213 includes a (currently bundled) change to the `enable_thinking` heuristic in `vllm_mlx/engine/simple.py` that switches from a hardcoded `"coder" in model_name` string check to reading the `VLLM_MLX_ENABLE_THINKING` env var (server-level default).

This PR (#262) adds `enable_thinking` as a per-request field on `ChatCompletionRequest` (per-request override).

The two are complementary. Server-level env var sets the default, per-request field overrides it. The natural precedence order if both land is:

1. Per-request `enable_thinking` from this PR (highest priority)
2. Server-level `VLLM_MLX_ENABLE_THINKING` env var from #213
3. Default `True` (current behavior)

If you want, I can rebase #213 to land its env var change after this PR so the precedence is built in cleanly. Or this PR can land first and I will adjust the env var fallback to honor the per-request field.

No file-level conflict between the two PRs at the moment. Both mergeable.

janhilgard:
Thanks for the heads-up @Thump604!

I've just updated this PR to unify the fallback across **all** code paths — `SimpleEngine.stream_chat`, `BatchedEngine._apply_chat_template`, and both `MLLM.chat`/`MLLM.stream_chat` now consistently fall back to `VLLM_MLX_ENABLE_THINKING` env var (default `true`), replacing the old `"coder" in model_name` heuristic.

The three-level priority is now consistent everywhere:

1. **Per-request**: `extra_body.enable_thinking` (this PR)
2. **Environment**: `VLLM_MLX_ENABLE_THINKING=true|false|1|0|yes|no` (aligns with your #213)
3. **Default**: `true`

Both PRs can land in any order without conflict — if #213 lands first, it establishes the env var; this PR already respects it. If this one lands first, the env var support is already built in.

Thump604 (approved):
LGTM as the per-request thinking control and the fallback chain (request → env → default True) is the right shape. The plumbing through every simple-engine path is consistent with the #218 `chat_template_kwargs` work, and the runtime test I just ran against a runtime carrying the #218 cherry-pick confirms the chat_template_kwargs-based thinking override still works, so landing #262 just gives users a more discoverable top-level field on top of that.

Two non-blocking notes:

1. If #218 lands first, `_apply_chat_template` in `batched.py` will need a tiny rebase: #218's version sets `template_kwargs` via the `_merge_template_kwargs` helper and adds chat_template_kwargs keys that might collide with `enable_thinking`. Precedence question: if a caller sends both `enable_thinking: false` (top level) and `chat_template_kwargs: {enable_thinking: true}`, which wins? My read is that the dedicated top-level field should be the higher-precedence one since it's more specific, but I don't see that resolution codified. Happy with either order as long as it's documented.

2. The `os.environ` fallback inside `_apply_chat_template` reads the env var on every call, which is fine for runtimes that set it once at startup. If anything ever mutates it per-request, the current shape is racy. Non-issue for the current codebase.

Approving.

janhilgard:
@Thump604 Superseded — per-request `enable_thinking` parameter is already in `main` via #278. Closing.

PR #270, opened 2026-04-09
feat: add --compile flag for mx.compile model optimization

jackneil:
## Summary

- Adds `--compile` flag to `vllm-mlx serve` that wraps model forward pass with `mx.compile(shapeless=True)`
- Fuses elementwise Metal kernels, reducing kernel launch overhead
- Off by default, opt-in, zero impact when not used
- Adds `vllm-mlx bench-compile <model>` A/B benchmark command with warmup, cache clearing, and statistical reporting
- Benchmarked: +33% prefill speed, no decode change (decode is memory-bandwidth-bound on Apple Silicon)

## Usage

```bash
# Serve with compiled model
vllm-mlx serve mlx-community/Qwen3-8B-4bit --compile

# A/B benchmark
vllm-mlx bench-compile mlx-community/Qwen3-8B-4bit --prompts 5
```

## Test plan

- [x] 5 unit tests for compile wrapper
- [x] Integration test with real model
- [x] Benchmarked on Qwen3-0.6B and Qwen3.5-35B

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -44,6 +44,7 @@ vllm-mlx serve <model> [options]
 | `--embedding-model` | Pre-load an embedding model at startup | None |
 | `--enable-auto-tool-choice` | Enable automatic tool calling | False |
 | `--tool-call-parser` | Tool call parser (`auto`, `mistral`, `qwen`, `llama`, `hermes`, `deepseek`, `kimi`, `granite`, `nemotron`, `xlam`, `functionary`, `glm47`) | None |
+| `--compile` | Compile model forward pass with mx.compile for fused Metal kernels (experimental) | False |
 
 ### Examples
 
@@ -94,6 +95,16 @@ vllm-mlx serve mlx-community/Qwen3-4B-4bit \
   --rate-limit 60 \
   --timeout 120 \
   --continuous-batching
+```
+
+### Performance: mx.compile
+
+```bash
+# Serve with compiled model (may improve throughput 5-30%)
+vllm-mlx serve mlx-community/Qwen3-8B-4bit --compile
+
+# A/B benchmark to measure the real impact
+vllm-mlx bench-compile mlx-community/Qwen3-8B-4bit --prompts 5 --max-tokens 128
 ```
 
 ### Security
--- a/tests/test_bench_compile.py
+++ b/tests/test_bench_compile.py
@@ -1,0 +1,57 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Integration test for mx.compile wrapper with mlx_lm models."""
+
+import os
+
+import pytest
+
+try:
+    import mlx.core as mx
+    from mlx_lm import load
+
+    HAS_MLX_LM = True
+except ImportError:
+    HAS_MLX_LM = False
+
+from vllm_mlx.compile import apply_compile, is_compiled
+
+
+@pytest.mark.skipif(not HAS_MLX_LM, reason="mlx_lm not available")
+class TestCompileWithRealModel:
+    @pytest.fixture
+    def model_and_tokenizer(self):
+        model_name = os.environ.get(
+            "VLLM_MLX_TEST_MODEL", "mlx-community/Qwen3-0.6B-8bit"
+        )
+        try:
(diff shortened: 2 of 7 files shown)


Thump604:
Blocking issue: the current compile wrapper mutates `model.__call__` on the instance, but Python does not use an instance attribute for `obj(...)` special-method dispatch.

Minimal repro:
```python
class C:
    def __call__(self, x):
        return ('orig', x)

c = C()
def alt(x):
    return ('alt', x)

c.__call__ = alt
assert c.__call__(2) == ('alt', 2)
assert c(3) == ('orig', 3)  # still uses type(c).__call__
```

So `apply_compile()` can set the compiled flag and log success while the normal inference path still executes the original uncompiled `__call__`.

I think this needs a different implementation before merge: either wrap the model in a callable proxy used by the engine, or compile the actual function object the engine invokes at call sites rather than writing to the instance `__call__` attribute.

Thump604 (changes requested):
The current implementation is not safe to merge yet.

`apply_compile()` mutates `model.__call__` on the instance, but Python does not use an instance attribute for `obj(...)` special-method dispatch. In other words, `model.__call__ = compiled_call` does not make `model(...)` use the compiled function.

Minimal repro:
```python
class C:
    def __call__(self, x):
        return ('orig', x)

c = C()
def alt(x):
    return ('alt', x)

c.__call__ = alt
assert c.__call__(2) == ('alt', 2)
assert c(3) == ('orig', 3)
```

So this PR can set the compiled flag and log success while the normal inference path still executes the original uncompiled `__call__`.

I think this needs a different implementation before merge: either wrap the model in a callable proxy used by the engine, or compile the actual function object the engine invokes at call sites rather than writing to the instance `__call__` attribute.

janhilgard (commented):
Agreeing with Thump604's blocking review -- the `__call__` dispatch issue is fundamental. To elaborate:

**The core bug Thump604 identified is real and critical.** Python resolves `obj(args)` via `type(obj).__call__`, not `obj.__dict__["__call__"]`. So `model.__call__ = mx.compile(original_call, shapeless=True)` sets an instance attribute that is only reached via `model.__call__(x)` (explicit attribute access), NOT via `model(x)` (which is how MLX generation code invokes the model). The compiled function is never actually called during inference.

The unit tests pass because `apply_compile` returns the same model object, and `is_compiled` checks the flag (which IS set), but the actual forward pass is unchanged. The `test_compile_wraps_model` test calls `compiled_model(x)` which goes through the unmodified class `__call__`, so `mx.allclose` passes trivially (same function, same output).

To fix this properly, the standard pattern in MLX is:
```python
model.__class__.__call__ = mx.compile(model.__class__.__call__, shapeless=True)
```
But this mutates the class itself, affecting all instances. The safer approach would be to create a wrapper class or use `mx.compile` at the `nn.Module` level if MLX supports it.

Additional concerns beyond Thump604's review:

1. **`bench-compile` uses `mlx_lm.stream_generate` directly** rather than going through the vllm-mlx engine. This means the benchmark does not measure the actual serving path (batched engine, continuous batching, etc.), so the numbers may not reflect real-world impact.

2. **`server.py` simple engine path** does `apply_compile(inner_model)` where `inner_model = getattr(_engine._model, "model", None)`. This reaches into private internals of `MLXLanguageModel` and will break if the attribute name changes. Better to expose a public method on the engine.

3. **No interaction testing with MTP, KV cache quantization, or MLLM.** `mx.compile(shapeless=True)` may not be compatible with the dynamic shapes in BatchKVCache, MTP speculation, or vision preprocessing. This needs validation before marking as experimental.

4. **Branch has merge conflicts** with current main.

Thump604:
Hi @jackneil -- same question on this one. Review feedback is open and it has merge conflicts. Are you still planning to address the feedback and rebase? Will check back in two weeks.

waybarrios:
Closing this one. The wrapper sets `model.__call__` as an instance attribute, but Python dispatches through the class, so the compiled path never actually runs during inference. We also profiled mx.compile on this codebase and measured 0% gain, decode is GPU-bound and Python overhead is under 10%, so there's no headroom for this approach. Thanks for exploring it.

Issue #249, opened 2026-04-02
bug: load_model_with_fallback missing return on success path


dougborg:
## Summary

`vllm_mlx/utils/tokenizer.py:load_model_with_fallback()` is missing a `return` statement on the happy path, causing the server to crash with `TypeError: cannot unpack non-iterable NoneType object` for any model that loads cleanly via `mlx_lm.load()`.

## Regression

In v0.2.6 (line ~55):
```python
try:
    return load(model_name, tokenizer_config=tokenizer_config)
```

On current `main`:
```python
try:
    model, tokenizer = load(model_name, tokenizer_config=tokenizer_config)
# ... no return
```

The refactoring to add the `_load_strict_false` fallback path removed the `return` on the success path. Models that hit a `ValueError` (and go through a fallback) work fine; models that load cleanly return `None`.

## Fix

Add `return model, tokenizer` after the successful load:

```python
try:
    model, tokenizer = load(model_name, tokenizer_config=tokenizer_config)
    return model, tokenizer  # <-- missing
except ValueError as e:
```

## Reproduction

```bash
vllm-mlx serve mlx-community/Qwen3-Coder-30B-A3B-Instruct-4bit --port 4000
# TypeError: cannot unpack non-iterable NoneType object
```

## Environment

- macOS 26.3, M1 Max 64GB
- mlx-lm 0.31.1, mlx 0.31.1
- vllm-mlx from git main (b4fa030)

jtatum:
This is a dupe of #212

Thump604:
Same bug as #211, #212, #252. All four issues describe `load_model_with_fallback` in `vllm_mlx/utils/tokenizer.py` falling through after a successful `mlx_lm.load()` call and returning `None`, causing the caller to crash with `TypeError: cannot unpack non-iterable NoneType object`. Fixed by PR #243 (one-line `return model, tokenizer` added at line 55). PR is currently MERGEABLE on main.

jackneil:
Duplicate of #211. Fix in PR #243.

Thump604:
Duplicate of #212 — same root cause in `load_model_with_fallback`.

PR #272, opened 2026-04-10

### LINKED ISSUE #252: load_model_with_fallback missing return statement — model load succeeds but returns None
## Bug Description

`load_model_with_fallback()` in `vllm_mlx/utils/tokenizer.py` is missing a `return model, tokenizer` statement after the successful `try` block. When a model loads without raising a `ValueError`, the function falls through and implicitly returns `None`, causing `TypeError: cannot unpack non-iterable NoneType object` in the engine startup.

## Steps to Reproduce

```bash
vllm-mlx serve mlx-community/GLM-4.7-Flash-4bit --port 8000 --continuous-batching
```

## Error

```
File "vllm_mlx/engine/batched.py", line 257, in _start_llm
    self._model, self._tokenizer = load_model_with_fallback(
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: cannot unpack non-iterable NoneType object
```

## Root Cause

In `vllm_mlx/utils/tokenizer.py`, the `load_model_with_fallback` function:

```python
def load_model_with_fallback(model_name: str, tokenizer_config: dict = None):
    # ...
    try:
    except ValueError as e:
        # ... fallback handlers with return statements ...
        raise
    # <-- Missing: return model, tokenizer
```

The `return` statements only exist inside the `except` branches. When `load()` succeeds (no exception), execution falls off the end and returns `None`.

## Fix

Add `return model, tokenizer` after the `try/except` block:

```python
    try:
    except ValueError as e:
        ...
        raise

    return model, tokenizer  # <-- add this
```

## Environment

- vllm-mlx 0.2.7
- Python 3.14.3
- macOS (Apple Silicon M4 Pro)
- Model: mlx-community/GLM-4.7-Flash-4bit (loads fine with `mlx_lm.load()` directly)

mikepixelmagic-dev:
## Summary

`load_model_with_fallback()` in `utils/tokenizer.py` is missing a `return` statement on the happy path. When `mlx_lm.load()` succeeds without raising a `ValueError`, the function falls through the `try` block and implicitly returns `None`.

This causes a `TypeError` in `MLXLanguageModel.load()`:

```
self.model, self.tokenizer = load_model_with_fallback(...)
TypeError: cannot unpack non-iterable NoneType object
```

**The fix is a single line** — adding `return model, tokenizer` after the successful `load()` call.

## How to reproduce

```python
import asyncio
from vllm_mlx.models.llm import MLXLanguageModel

async def test():
    m = MLXLanguageModel('mlx-community/Qwen3.5-35B-A3B-4bit', trust_remote_code=True)
    m.load()  # TypeError: cannot unpack non-iterable NoneType object

asyncio.run(test())
```

Note: the bug only manifests when called inside an async context (e.g., `asyncio.run()`), which is how the server starts the engine via `SimpleEngine.start()`. Direct calls outside async may work due to Python caching behaviour.

## Related issues

Fixes #252, #249

## Test

Verified with Qwen3.5-35B-A3B (4-bit and 8-bit) on Apple M5 Max. Server starts and serves requests successfully after the fix.

### DIFF
--- a/vllm_mlx/utils/tokenizer.py
+++ b/vllm_mlx/utils/tokenizer.py
@@ -52,6 +52,7 @@ def load_model_with_fallback(model_name: str, tokenizer_config: dict = None):
 
     try:
         model, tokenizer = load(model_name, tokenizer_config=tokenizer_config)
+        return model, tokenizer
     except ValueError as e:
         # Fallback for models with non-standard tokenizers
         if "TokenizersBackend" in str(e) or "Tokenizer class" in str(e):


mikepixelmagic-dev:
Closing as duplicate of #243 which has the same fix. 👍 on getting it merged.

PR #268, opened 2026-04-09
feat: add Gemma 4 multimodal model support

jackneil:
## Summary

- Adds Gemma 4 multimodal model support (loading, inference, vision)
- Fixes BatchKVCache for Gemma 4's RotatingKVCache architecture
- Adds Gemma 4 reasoning parser for channel-based thinking
- Fixes RotatingKVCache isinstance check in MLLM batch generator
- Fixes missing return statement in tokenizer load_model_with_fallback

## Test plan

- [x] Model loads and generates text
- [x] Multimodal (image) input works
- [x] Reasoning parser extracts thinking blocks

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #250, opened 2026-04-03
feat: Gemma 4 tool call + reasoning parser

Thump604:
## Summary
- Gemma 4 tool call parser: parses `<|tool_call>call:name{params}<tool_call|>` with `<|"|>` string escaping
- Gemma 4 reasoning parser: extracts `<|channel>thought...<channel|>` blocks into `reasoning_content`
- Registered as `--tool-call-parser gemma4` and `--reasoning-parser gemma4`
- CLI choices updated

Addresses the server-side equivalent of ml-explore/mlx-lm#1096 (Gemma 4 native tool calls not parsed).

## Details

**Tool parser** (`tool_parsers/gemma4_tool_parser.py`):
- Regex extracts function name and parameter block from `<|tool_call>call:name{...}<tool_call|>`
- Handles `<|"|>` escaped string values, numeric/boolean coercion
- Arguments serialized to JSON string for OpenAI API compatibility
- Streaming support with start-marker detection

**Reasoning parser** (`reasoning/gemma4_parser.py`):
- Splits `<|channel>thought\n...reasoning...<channel|>` from final answer
- Non-streaming: returns `(reasoning, content)` tuple
- Streaming: tracks channel boundaries, emits reasoning deltas then content deltas

## Test plan
- [x] Non-streaming: `tool_calls` field populated with correct function name + JSON arguments
- [x] Non-streaming: `reasoning_content` separated from `content`, both present
- [x] Streaming: SSE chunks flow correctly
- [x] Concurrent: 3 simultaneous requests with BatchedEngine
- [x] Thinking + tools: model produces empty think channel then tool call (correct behavior)
- [x] Parser standalone unit tests pass

PR #273, opened 2026-04-10
fix: accept RotatingKVCache in MLLM prompt merge

Thump604:
Follow-up to the RotatingKVCache report in #256.

Context:
- `#256` adds trimming for `RotatingKVCache` before merge in `MLLMBatchGenerator._process_prompts`
- but the existing guard immediately above it still rejects anything that is not `KVCache`
- on Gemma 4 MLLM continuous batching, that makes the new trim path unreachable because `RotatingKVCache` is a sibling of `KVCache`, not a subclass

This patch does two things:
1. Relaxes the guard to accept either `KVCache` or `RotatingKVCache`
2. Adds a regression test that drives `_process_prompts()` with a real `RotatingKVCache` and asserts the merged cache is `BatchRotatingKVCache`

Validation:
- `python -m pytest tests/test_mllm_continuous_batching.py -q` -> `24 passed, 3 deselected`
- `python -m black --check --target-version py312 vllm_mlx/mllm_batch_generator.py tests/test_mllm_continuous_batching.py`

If you prefer, this can be cherry-picked directly onto `#256` rather than merged as a standalone PR.

### DIFF
--- a/tests/test_mllm_continuous_batching.py
+++ b/tests/test_mllm_continuous_batching.py
@@ -225,6 +225,53 @@ class TestMLLMBatchStats:
         assert stats.prompt_tps == 0
         assert stats.generation_tps == 0
 
+    def test_process_prompts_accepts_rotating_kv_cache(self, monkeypatch):
+        """MLLM prompt prefill should accept native sliding-window caches."""
+        import mlx_lm.models.cache as cache_mod
+        from mlx_lm.models.cache import BatchRotatingKVCache, RotatingKVCache
+        from vllm_mlx.mllm_batch_generator import MLLMBatchGenerator, MLLMBatchRequest
+
+        class DummyModel:
+            layers = [object()]
+
+        def fake_preprocess(request):
+            request.input_ids = mx.array([1, 2, 3], dtype=mx.int32)
+            request.pixel_values = None
+            request.attention_mask = None
+            request.image_grid_thw = None
+            request.extra_kwargs = {}
+
+        def fake_make_prompt_cache(_model):
+            cache = RotatingKVCache(max_size=16, keep=0)
+            cache.keys = mx.arange(8, dtype=mx.float32).reshape(1, 1, 4, 2)
+            cache.values = (mx.arange(8, dtype=mx.float32) + 1).reshape(1, 1, 4, 2)
+            cache.offset = 4
+            cache._idx = 4
+            return [cache]
+
+        generator = MLLMBatchGenerator(
+            DummyModel(),
+            MagicMock(),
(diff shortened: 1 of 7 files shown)


waybarrios:
Hey, thanks for pulling this out as a focused fix after keegoid flagged it on #256. The RotatingKVCache guard was a real blocker for Gemma 4 batching. It landed as part of #268 which we just merged. Between this, #250, and your review on #268, you've been all over the Gemma 4 effort, really appreciate it.

bludiesel:
Confirming this PR fixes a production-breaking bug on Apple Silicon with Gemma 4 26B A4B MLLM. Ran a full benchmark on M3 Ultra 512GB (2026-04-12) against `mlx-community/gemma-4-26B-A4B-it-heretic-4bit`.

## Reproduction on main

With `vllm-mlx serve <model> --mllm --use-paged-cache --continuous-batching`, any batch containing prompts of different sequence lengths crashes with:

```
ERROR:vllm_mlx.mllm_batch_generator:Failed to merge per-request caches (RotatingKVCache): ValueError: [broadcast_shapes] Shapes (1,8,48,256) and (1,8,177,256) cannot be broadcast.
```

On a 8-prompt benchmark suite (short, medium, long-gen, math, code, classification, Arabic, uncensored), main crashes on 2/8 single requests and the scheduler enters a broken state — all subsequent requests return `finish_reason: error` with empty content. `--kv-cache-quantization --kv-cache-quantization-bits 4` does NOT bypass the crash; it hits the same `_process_prompts` guard.

## After cherry-picking this PR

Same benchmark, patched branch:

```
=== SEQUENTIAL (8/8 success) ===
  short:          62w    1.3s
  medium:         391w   6.2s
  math:           266w   6.3s
  code:           380w   9.7s
  classification: 27w    1.3s
  arabic:         437w  11.6s
  uncensored:     477w   7.0s
  long_gen:       685w  11.6s

=== PARALLEL (8 concurrent, mixed-length prompts) ===
  8/8 successful
  Total wallclock:       20.0s
  Aggregate throughput:  237.7 tok/s
```

For context, on the same hardware and prompts:
- Patched vllm-mlx + continuous batching: 20.0s / 8 parallel / 237.7 tok/s
- mlx-vlm server (official):               24.7s / 8 parallel / ~130 tok/s
- Ollama (gguf backend):                   40.7s / 8 parallel / 151 tok/s
- vllm-mlx no --continuous-batching:       74.6s / 8 parallel / 82 tok/s

This patch is the difference between 25% failure and a fully working continuous-batching MLLM path. Would love to see this land — happy to re-run any additional test config if useful.

PR #275, opened 2026-04-10
fix: run MLLM prefill in thread executor to keep event loop responsive

janhilgard:
## Summary

- MLLM scheduler's `_process_loop()` was calling `self.step()` directly on the asyncio event loop, blocking all HTTP handlers (including `/health`) during prefill
- Apply the same hybrid executor pattern already used by `engine_core`: prefill-likely steps run via `run_in_executor`, decode-only steps remain inline
- This keeps `/health` and other endpoints responsive during long prefills

## Problem

When a large prompt arrives, prefill can block the event loop for several seconds. During this time, `/health` returns no response, causing monitoring dashboards to report the server as unavailable.

## Test plan

- [x] Start server with `--mllm --continuous-batching`
- [x] Send a long prompt request and simultaneously call `/health` — confirms 200 responses in ~65-73ms during active inference
- [x] Verify generation output is correct and streaming works

### DIFF
--- a/vllm_mlx/mllm_scheduler.py
+++ b/vllm_mlx/mllm_scheduler.py
@@ -19,6 +19,7 @@ Architecture:
 """
 
 import asyncio
+import concurrent.futures
 import logging
 import time
 import uuid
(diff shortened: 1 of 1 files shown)


janhilgard:
@Thump604 Superseded — MLLM prefill thread executor (`run_in_executor`) is already in `main` via #278. Closing.

PR #280, opened 2026-04-11
[codex] fix(simple): cooperatively cancel specprefill workers

Thump604:
## Summary
- add cooperative cancellation to the SimpleEngine SpecPrefill worker path
- thread a single `cancel_check` hook through `_prefill_draft`, `score_tokens`, and `sparse_prefill`
- add regression coverage for scoring-time and sparse-prefill cancellation without changing the current `main` decode shape

## Why
`_run_blocking_serialized()` now keeps the async generation lock held until the worker thread actually finishes, but long-running SpecPrefill phases still need a cooperative exit path after request cancellation. Without that, cancelled SpecPrefill requests can keep burning CPU inside blocking scoring/prefill loops even though the request is already gone.

## Validation
- `black --fast --check vllm_mlx/engine/simple.py vllm_mlx/specprefill.py tests/test_simple_engine.py`
- `pytest -q tests/test_simple_engine.py`

### DIFF
--- a/tests/test_simple_engine.py
+++ b/tests/test_simple_engine.py
@@ -4,6 +4,7 @@
 import asyncio
 from unittest.mock import MagicMock, patch
 
+import mlx.core as mx
 import pytest
 
 pytestmark = pytest.mark.anyio
--- a/tests/test_simple_engine.py
+++ b/tests/test_simple_engine.py
@@ -373,3 +373,316 @@ class TestSimpleEngineConcurrency:
 
         assert output.text == ""
         assert output.finish_reason == "stop"
+
+    @pytest.mark.anyio
+    async def test_cancellation_does_not_release_lock_before_worker_finishes(
+        self, mock_llm_model
+    ):
+        """A cancelled blocking chat call must not overlap the next worker."""
+        from threading import Event, Lock
+
+        from vllm_mlx.engine.simple import SimpleEngine
+
+        first_started = Event()
+        release_workers = Event()
+        call_count = 0
+        call_lock = Lock()
+
+        def chat_side_effect(**kwargs):
+            nonlocal call_count
+            with call_lock:
+                call_count += 1
+                current_call = call_count
+                mock_llm_model._concurrent_count += 1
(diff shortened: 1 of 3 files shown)


janhilgard (approved):
LGTM. Well-designed cooperative cancellation for SpecPrefill workers.

**Design:**
- Introduces `_SpecPrefillCancelled` as a cooperative cancellation sentinel
- A `threading.Event` + `_cancel_check()` closure is threaded through all blocking SpecPrefill phases: `score_tokens` -> `_prefill_draft` -> `_lookahead_decode` -> `sparse_prefill`
- `_run_blocking_serialized` gains an `on_cancel` callback that sets the event when the async task is cancelled
- The cancel check is polled at each step boundary (before/after MLX evals), giving cooperative exit points without interrupting mid-computation

**Why this works well:**
- The cancellation is cooperative, not preemptive — no risk of corrupting MLX state or leaving partial cache entries
- The same `cancel_check` callable is passed to both scoring and prefill phases, so cancellation is detected regardless of which phase is active
- `_SpecPrefillCancelled` is caught in `_run_all` and re-raised (not falling through to `_run_normal`), so a cancelled request does not silently start a full normal generation
- The `on_cancel` callback in `_run_blocking_serialized` is fire-and-forget with exception swallowing, which is correct since the event is the actual signal

**Test coverage is thorough:**
- `test_cancellation_does_not_release_lock_before_worker_finishes` — verifies no concurrent overlap
- `test_specprefill_path_does_not_prelock_serialized_runner` — verifies lock is not pre-acquired
- `test_specprefill_threads_same_cancel_check_to_helpers` — verifies the same callable reaches both phases
- `test_cancelling_specprefill_request_stops_during_scoring` — verifies cancellation during score phase
- `test_cancelling_specprefill_request_stops_during_sparse_prefill` — verifies cancellation during prefill phase

Clean implementation with no unnecessary complexity.

janhilgard:
This PR has merge conflicts with the current main branch. Could you rebase/resolve the conflicts? The code itself looks good and is ready to merge once conflicts are resolved.

PR #278, opened 2026-04-10
feat: backport production features — MTP, tool parsers, sampling, prefill

janhilgard:
## Summary

Backport battle-tested features from the production fork to upstream. These features have been running in production for days with multiple models (Qwen3.5-122B-A10B, Gemma 4 26B-A4B).

### New files (4)
- **`patches/qwen3_5_mllm.py`** — BatchKVCache offset fix for Qwen3.5 attention (same `mx.array.__iadd__` bug class as Gemma 4 fix in #256)
- **`patches/qwen3_5_mtp.py`** — Runtime MTP (Multi-Token Prediction) injection for Qwen3.5 models
- **`tool_parsers/minimax_tool_parser.py`** — MiniMax-M2 tool call parser
- **`scripts/add_mtp_weights_qwen35.py`** — Extract MTP weights from BF16 Qwen3.5 source models

### Modified files (17)

**MLLM batch generator** — Chunked prefill, mid-batch extend, MTP injection hooks, patch registration, repetition penalty, prefill abort on disconnect, per-request logits processors, think-suffix stripping for prefix cache, `last_query_index` template normalization for prefix-stable assistant turns

**MLLM scheduler** — Request status tracking, cache config forwarding, prefill abort support, fixed duplicate cache-clear block

**Server** — `enable_thinking` per-request, `tool_choice=none`, tool argument type coercion, `repetition_penalty`

**Engines** — MTP injection in batched engine, `enable_thinking` in simple engine, `gpu_memory_utilization` config

**CLI** — MiniMax parser choice, `--gpu-memory-utilization` flag

**API models** — `enable_thinking`, `repetition_penalty`

**Prefix cache** — Block LCP for hybrid models (SSM can't be rewound), think-suffix stripping for clean PREFIX matches, chat template normalization for Qwen3.5

**Other** — Qwen3.5/MiniMax MLLM patterns, TextModelArgs from VLM config dict, MTP validator in scheduler

## MTP speculative decoding results

| Model | Acceptance rate | Notes |
|-------|----------------|-------|
| Qwen3.5-122B-A10B (MoE) | 77-78% | 5.0 GB BF16 MTP weights |
| Qwen3.5-35B-A3B (MoE) | 79-85% | 1.7 GB BF16 MTP weights |
| Qwen3.5-27B (Dense) | 78-80% | 849 MB BF16 MTP weights |

**Critical**: MTP weights must be native BF16 extracted from source model — dequantized 4-bit weights give 0% acceptance.

[...]

PR #281, opened 2026-04-11
Add <function=name> format support to Qwen tool parser

janhilgard:
## Summary

- Qwen3.5 models generate tool calls in the native `<function=name><parameter=key>value</parameter></function>` format, but the `qwen` parser only handled `<tool_call>JSON</tool_call>` and `[Calling tool: name({...})]` — causing tool calls to be returned as raw text instead of structured `tool_calls`
- Add `FUNCTION_PATTERN` and `PARAM_PATTERN` to parse both `<parameter>` tags and inline JSON arguments
- Add streaming partial-marker buffering to prevent leaking incomplete markers (e.g. `<function` before `=` arrives) as text content to the client
- Buffer recovery: when a partial marker turns out to be a false positive, the buffered prefix is re-emitted with the next delta (no text lost)
- Add safety net in `server.py` Anthropic streaming paths to suppress any `<function` content that slips through the streaming parser

## Files changed

| File | Changes |
|------|---------|
| `qwen_tool_parser.py` | Add `FUNCTION_PATTERN`, `PARAM_PATTERN`, partial-marker streaming buffering |
| `server.py` | Safety net for `<function` in Anthropic streaming, `<function=` in fallback detection |

## Context

When using `--tool-call-parser qwen` with Qwen3.5 models, the model outputs:
```
<function=get_weather>
<parameter=city>
Prague
</parameter>
</function>
```

This format was not recognized, so responses had `tool_calls: null` and the raw XML leaked as text content. Clients (e.g. Claude Code) reported "Invalid tool parameters" errors.

### Streaming issue

Without partial-marker buffering, early tokens like `<function` (before `=` arrives) leaked as text content deltas. The fix buffers potential marker prefixes and re-emits them if they turn out not to be tool calls. An additional safety net in `server.py` catches any edge cases where markers slip through the streaming parser (e.g. due to multi-token MTP deltas or reasoning parser splits).

## Test plan

- [x] Non-streaming: `tool_calls` correctly populated with `name` and `arguments`
- [x] Streaming (OpenAI): tool call chunks emitted correctly
- [x] Streaming (Anthropic): `tool_use` content blocks with `input_json_delta`, no leaked text
- [x] Existing formats still work (XML `<tool_call>`, bracket `[Calling tool:]`)
- [x] JSON arguments inside `<function>` tags parsed correctly
- [x] Multiple `<parameter>` tags parsed into single arguments object
- [x] Parameter value type coercion (numbers, booleans, strings)
- [x] Partial marker buffering: `<function` suppressed until `=` confirms tool call
- [x] Buffer recovery: false positives (e.g. `<div>`) re-emit buffered prefix
- [x] Safety net: `<function` in content is suppressed before emission
- [x] End-of-stream fallback: `<function=` triggers tool call extraction

### DIFF
--- a/tests/test_tool_parsers.py
+++ b/tests/test_tool_parsers.py
@@ -1163,3 +1163,177 @@ class TestHermesStreamingFixes:
         assert len(emitted_calls) == 2
         assert emitted_calls[0]["function"]["name"] == "func1"
         assert emitted_calls[1]["function"]["name"] == "func2"
+
+
+class TestQwenFunctionFormat:
+    """Test Qwen parser's <function=name> format support."""
+
+    @pytest.fixture
+    def parser(self):
+        return QwenToolParser()
+
+    def test_function_format_with_parameters(self, parser):
+        """Test <function=name><parameter=key>value</parameter></function>."""
+        text = "<function=get_weather><parameter=city>Prague</parameter></function>"
+        result = parser.extract_tool_calls(text)
+        assert result.tools_called
+        assert result.tool_calls[0]["name"] == "get_weather"
+        args = json.loads(result.tool_calls[0]["arguments"])
+        assert args["city"] == "Prague"
+
+    def test_function_format_with_json(self, parser):
+        """Test <function=name>{"key": "val"}</function>."""
+        text = '<function=get_weather>{"city": "Prague"}</function>'
+        result = parser.extract_tool_calls(text)
+        assert result.tools_called
+        assert result.tool_calls[0]["name"] == "get_weather"
+        args = json.loads(result.tool_calls[0]["arguments"])
+        assert args["city"] == "Prague"
+
+    def test_function_format_multiple(self, parser):
+        """Test multiple <function=...> blocks."""
+        text = (
+            '<function=read_file>{"path": "/a.py"}</function>'
+            '<function=write_file>{"path": "/b.py", "content": "hello"}</function>'
+        )
+        result = parser.extract_tool_calls(text)
+        assert result.tools_called
+        assert len(result.tool_calls) == 2
+        assert result.tool_calls[0]["name"] == "read_file"
+        assert result.tool_calls[1]["name"] == "write_file"
+
+    def test_function_format_with_think_tags(self, parser):
+        """Test <function=...> with think tags."""
+        text = (
+            "<think>I need to check the weather.</think>\n"
+            '<function=get_weather>{"city": "Prague"}</function>'
+        )
+        result = parser.extract_tool_calls(text)
+        assert result.tools_called
+        assert result.tool_calls[0]["name"] == "get_weather"
+
+
+class TestQwenStreamingBuffering:
+    """Test Qwen parser streaming with partial-marker buffering."""
+
+    @pytest.fixture
+    def parser(self):
+        return QwenToolParser()
+
+    def test_streaming_function_format_complete(self, parser):
+        """Test streaming with <function=name>...</function> format."""
+        chunks = [
+            "<function=get_weather>",
+            "<parameter=city>Prague</parameter>",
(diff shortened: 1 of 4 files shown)


Thump604 (changes requested):
I found one blocker in the new streaming recovery path.

The parser now tries to recover false positives by re-emitting buffered text when a partial `<function` prefix turns out not to be a tool call. But the new server-side safety net swallows that recovered content again whenever it still contains `<function`. That means plain text like `<functional` or any literal content ending in `<function` at end-of-stream can be lost instead of re-emitted.

Minimal reproduction from the current logic:
- delta 1: `Look at <function` -> parser buffers and emits nothing
- delta 2: `al interface` -> parser correctly recovers `"<functional interface"`
- server safety net sees `<function` in recovered content and suppresses it again
- next delta only emits the later suffix, so the original recovered prefix is dropped

There is a second edge case at end-of-stream: a literal final `"<function"` with no `=` is buffered and then never flushed, because the fallback only checks for actual tool markers like `<function=`.

I would keep the parser-side buffering/recovery, but narrow or remove the server-side `<function` suppression so it does not fight the parser's recovery path. A focused regression test around a false-positive literal like `<functional` across a chunk boundary would catch this.

janhilgard:
Updated based on review feedback:

1. **Removed server-side `<function` safety net** — it was conflicting with the parser's recovery path. The parser correctly recovers false positives like `<functional interface` by re-emitting the buffered prefix, but the safety net was swallowing the recovered content again.

2. **Widened end-of-stream fallback** from `"<function="` to `"<function"` to handle truncated partial markers at stream end.

3. **Added regression tests** for:
   - `<functional` false positive across chunk boundary
   - Single `<` followed by non-marker text recovery
   - Streaming with multiple `<function=` blocks

The diff is now clean: 3 files, +311/-3 lines (rebased on latest `main` after PR #278 merge).

janhilgard:
@Thump604 Updated based on review feedback — removed the server-side safety net, added regression tests for false positives like `<functional`. Would appreciate another look when you get a chance. 🙏

janhilgard:
@Thump604 Review feedback addressed — ready for re-review.

Thump604:
I rechecked the updated streaming path and there is still one blocker in the false-positive recovery logic.

Minimal repro on the current branch:
- chunk 1: `"Look at <function"` -> returns `None`
- chunk 2: `"al interface"` -> returns `{"content": "<functional interface"}`

That preserves the buffered marker prefix, but it still drops the earlier buffered content from the same first chunk (`"Look at "`). So the user-visible stream still loses text; it is now just losing a smaller prefix than before.

The current regression test does not catch this because it only asserts that `"<function"` and `"al interface"` are present in the recovered content. It should also assert that the pre-marker content is preserved, e.g. the recovered content should be exactly `"Look at <functional interface"`.

I would keep the general direction, but the recovery branch needs to re-emit the full buffered suffix from the point buffering began, not only the matched partial marker fragment.

janhilgard:
Addressed the second review feedback — content before the partial marker is now preserved:

**What changed:**
- Added `_get_partial_marker_len()` to compute the exact length of a partial marker suffix
- Buffering now only holds back the marker suffix itself; content before it in the same delta (e.g. `"Look at "` from `"Look at <function"`) is emitted immediately via `delta_text[:safe_chars]`
- Added regression test asserting exact content: first chunk yields `{"content": "Look at "}`, recovery yields `{"content": "<functional interface"}`

All 23 Qwen tests pass. Ready for re-review @Thump604.

Thump604 (approved):
Rechecked the false-positive streaming case that blocked this earlier. The updated buffering logic now preserves the pre-marker content, the parser tests pass locally, and the server-side suppression conflict is gone. This looks good to merge.

PR #286, opened 2026-04-11
fix: handle 3D KV tensors in prefix cache

### LINKED ISSUE #142: Prefix cache tensor slicing fails for Qwen3.5 MoE models (3D KV cache)

Thump604:
Restack of #144 onto current `main`.

What changed:
- keep the existing 4D block-concat behavior intact
- add the missing 3D KV path for Qwen3.5-style `(heads, seq, dim)` caches
- store/use the correct sequence axis in block metadata
- fix reconstructed offset accounting for 3D states
- add a regression test that reconstructs a partial prefix from a 3D KV cache

Validation:
- `PYTHONPATH=/Users/David/code/vllm-mlx-pr144-restack /opt/ai-runtime/venv-live/bin/python -m pytest /Users/David/code/vllm-mlx-pr144-restack/tests/test_paged_cache.py -q`
- `PYTHONPATH=/Users/David/code/vllm-mlx-pr144-restack /opt/ai-runtime/venv-live/bin/python -m py_compile /Users/David/code/vllm-mlx-pr144-restack/vllm_mlx/prefix_cache.py /Users/David/code/vllm-mlx-pr144-restack/tests/test_paged_cache.py`
- `PYTHONPATH=/Users/David/code/vllm-mlx-pr144-restack /opt/ai-runtime/venv-live/bin/python -m black --check --fast /Users/David/code/vllm-mlx-pr144-restack/vllm_mlx/prefix_cache.py /Users/David/code/vllm-mlx-pr144-restack/tests/test_paged_cache.py`

Fixes #142. Supersedes #144.

### DIFF
--- a/tests/test_paged_cache.py
+++ b/tests/test_paged_cache.py
@@ -816,6 +816,42 @@ class TestBlockAwarePrefixCache:
 
         assert cache.reconstruct_cache(prefix_table) is None
 
+    def test_reconstructs_3d_kv_prefix_cache(self):
+        from mlx_lm.models.cache import KVCache
+        import mlx.core as mx
+
+        from vllm_mlx.paged_cache import BlockTable, PagedCacheManager
+        from vllm_mlx.prefix_cache import BlockAwarePrefixCache
+
+        paged_manager = PagedCacheManager(block_size=4, max_blocks=10)
+        cache = BlockAwarePrefixCache(model=None, paged_cache_manager=paged_manager)
+
+        tokens = list(range(8))
+        kv_keys = mx.arange(2 * 8 * 3).reshape(2, 8, 3)
+        kv_values = mx.arange(1000, 1000 + (2 * 8 * 3)).reshape(2, 8, 3)
+        extracted = [
+            {
+                "state": (kv_keys, kv_values),
+                "meta_state": "",
+                "class_ref": KVCache,
+                "class_name": "KVCache",
+            }
+        ]
+
+        block_table = cache.store_cache("req-1", tokens, extracted)
+        prefix_table = BlockTable(
+            request_id="req-prefix",
+            block_ids=[block_table.block_ids[0]],
+            num_tokens=4,
+        )
+
+        reconstructed = cache.reconstruct_cache(prefix_table)
+
+        assert reconstructed is not None
+        assert isinstance(reconstructed[0], KVCache)
+        assert reconstructed[0].state[0].tolist() == kv_keys[:, :4, :].tolist()
+        assert reconstructed[0].state[1].tolist() == kv_values[:, :4, :].tolist()
+
     def test_deduplicated_terminal_uses_correct_recurrent_snapshot(self):
         """Deduplication must not leak recurrent state across sequences."""
         from mlx_lm.models.cache import ArraysCache, KVCache
--- a/vllm_mlx/prefix_cache.py
+++ b/vllm_mlx/prefix_cache.py
@@ -658,7 +658,8 @@ class BlockAwarePrefixCache:
                 class_ref = layer_state.get("class_ref")
                 class_name = layer_state.get("class_name")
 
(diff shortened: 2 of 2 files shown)


janhilgard (approved):
LGTM. Clean fix for 3D KV tensor handling in prefix cache.

**What it does:**
- Replaces the hardcoded `_can_concatenate_cache_state` (4D-only) with `_cache_state_seq_axis` that returns the correct sequence axis for both 4D `(batch, heads, seq, dim)` and 3D `(heads, seq, dim)` tensors
- The 3D path (`ndim == 3, len(state) == 2`) correctly identifies Qwen3.5-style KV caches where the sequence axis is 1 instead of 2
- All downstream slicing (`_slice_concat_cache_state`, `_extract_block_tensor_slice`, `reconstruct_cache`) now uses the dynamically determined `seq_axis` instead of hardcoded `2`
- Offset accounting in `reconstruct_cache` is fixed for both the `_KVCache` and `KVCache` fallback paths

**Why it is correct:**
- The `ndims` set check ensures all tensors in state have the same dimensionality (rejects mixed shapes)
- The `len(state) == 2` guard for 3D tensors prevents false matches on non-KV state tuples
- The regression test creates a 3D `(2, 8, 3)` KV cache, stores 8 tokens across 2 blocks, reconstructs 4 tokens from the first block, and verifies the sliced output matches `[:, :4, :]`

One minor note: if a future model uses 3D tensors with `len(state) != 2`, it would fall through to `return None` and silently fail prefix caching rather than raising. That seems like acceptable defensive behavior for now.

PR #289, opened 2026-04-11
Upgrade torch and torchvision

perry2of5:
In PR 276 it was observed the lower bound of torch was lower than pip would resolve  it to. Since vllm-mlx advertises it should be run with python 3.10+, I upgraded the libraries as far as possible without compile errors. Attempting to upgrade torch to 2.9 and torchvision to 0.24, which are python 3.10+ led to dependency resolution errors so I went with the python 3.9 compatible versions listed on https://pypi.org/project/torchvision/ torch to 2.8 and torchvision to 0.21

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ dependencies = [
     # Video processing for VLM
     "opencv-python>=4.8.0",
     # Vision processor (required for transformers AutoProcessor)
-    "torchvision>=0.21.0",
+    "torchvision>=0.23.0",
     # Resource monitoring
     "psutil>=5.9.0",
     # Server
@@ -75,7 +75,7 @@ vllm = [
 ]
 vision = [
     "torch>=2.8.0",
-    "torchvision>=0.21.0",
+    "torchvision>=0.23.0",
 ]
 # Audio dependencies for TTS/STT (mlx-audio)
 audio = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -74,7 +74,7 @@ vllm = [
     "vllm>=0.4.0",
 ]
 vision = [
-    "torch>=2.3.0",
+    "torch>=2.8.0",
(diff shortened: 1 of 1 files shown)


Thump604:
`torch 2.8` + `torchvision>=0.21.0` isn't a valid combination. Official PyTorch docs map torch 2.8 to torchvision 0.23, and torchvision 0.21 explicitly requires `torch==2.6.0` in its wheel metadata.

If you want to raise the floor to torch 2.8, torchvision needs to move to `>=0.23.0` as well. Otherwise the existing bounds are safer.

perry2of5:
yes, you are correct. I was testing different combinations and committed the wrong one. Sorry for the sloppy PR. Fixed now.

janhilgard (approved):
LGTM. The version bump is straightforward and correct:

- `torch` minimum: `2.3.0` -> `2.8.0` (in `[vision]` extra)
- `torchvision` minimum: `0.21.0` -> `0.23.0` (in both `dependencies` and `[vision]` extra)

The `torch 2.8` / `torchvision 0.23` pairing matches the official PyTorch compatibility matrix. The earlier mismatch (`torch 2.8` + `torchvision 0.21`) that @Thump604 flagged has been fixed.

The PR description explains the rationale — torch 2.9 / torchvision 0.24 require Python 3.10+, so 2.8 / 0.23 is the right ceiling for the project's Python 3.9 compatibility claim.

One minor note: the `dependencies` section (line 47) only bumps `torchvision` to `>=0.23.0` but doesn't add an explicit `torch` lower bound there — `torch` is only constrained in `[vision]`. This is fine if torch is always pulled in transitively via torchvision, which it is (torchvision declares `torch==2.8.0` in its wheel metadata). Just noting for awareness.

CI passes across all Python versions. Good to merge.

Thump604:
Version fixed, CI green. Ready to merge.

PR #291, opened 2026-04-12
fix(specprefill): avoid dense tail expansion within cache window

Thump604:
## Summary
- avoid forcing full-tail inclusion for `RotatingKVCache` when the prompt still fits inside the cache window
- add a regression test that pins the no-eviction vs eviction boundary

## Why
On the Nemotron 120B live runtime path we cap unbounded attention caches with `max_kv_size=65536`, which means the SpecPrefill target cache is a `RotatingKVCache` even for prompts that still fit entirely inside the window. The previous sparse-prefill logic treated any `RotatingKVCache` as if eviction were already active and unioned in the full `max_size` tail. At 64K this effectively collapsed sparse prefill back into dense target work.

## Validation
- `pytest tests/test_specprefill_rotating_cache.py -q`
- `black --check --fast vllm_mlx/specprefill.py tests/test_specprefill_rotating_cache.py`
- `python -m py_compile vllm_mlx/specprefill.py tests/test_specprefill_rotating_cache.py`

Live runtime evidence from the promoted stack:
- Nemotron 64K clean cold rerun before fix: dense `271.98s`, sparse `395.37s`
- Nemotron 64K clean cold rerun after fix: dense `277.89s`, sparse `143.90s`
- Nemotron 16K clean cold rerun after fix: dense `68.48s`, sparse `36.26s`
- recall stayed correct in all fixed runs

### DIFF
--- a/tests/test_specprefill_rotating_cache.py
+++ b/tests/test_specprefill_rotating_cache.py
@@ -1,0 +1,84 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Regression tests for RotatingKVCache handling in sparse_prefill."""
+
+from __future__ import annotations
+
+import pytest
+
+try:
+    import mlx.core as mx
+
+    HAS_MLX = True
+except ImportError:
+    HAS_MLX = False
+
+pytestmark = pytest.mark.skipif(not HAS_MLX, reason="MLX not available")
+
+
+class _FakeAttention:
+    def __init__(self):
+        self.num_heads = 1
+        self.q_proj = lambda x: x
+
+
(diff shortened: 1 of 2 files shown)


janhilgard (approved):
LGTM. Clean one-line fix with clear regression tests.

The logic is correct: when `M <= max_rotating_size`, no eviction can happen yet, so forcing the full tail back in just collapses sparse prefill into dense work. The `M > max_rotating_size` guard is the right condition.

The Nemotron 64K benchmarks speak for themselves — 2x speedup over dense after the fix, vs being slower than dense before.

PR #295, opened 2026-04-12
Add GLM-4 reasoning parser and fix think tag / prefix cache bugs

janhilgard:
## Summary

- **Add `Glm4ReasoningParser`** (`--reasoning-parser glm4`) for GLM-4 models (GLM-4.5-Air, GLM-4.7) that use `<think>...</think>` tags but don't inject `<think>` in the prompt. Unlike Qwen3 where no tags = truncated reasoning, GLM-4 no tags = normal content. Streaming override emits pre-think deltas as content.
- **Fix duplicate `</think>` tag** in `BaseThinkingReasoningParser`: some models generate `<think></think></think>` with an extra closing tag that leaked into content.
- **Fix reasoning extraction with tool calls**: always parse reasoning from original `output.text` so `<think>` content is preserved even when tool calls are present.
- **Remove `<tool_call>` from `SPECIAL_TOKENS_PATTERN`**: these are structural tags used by GLM-4's tool calling format (`--tool-call-parser glm47`) and must not be stripped.
- **Fix prefix cache corruption on exact match**: trim stored cache by 1 in `_prompt_cache_save` so the last prompt token is reprocessed at the correct position on cache hit.
- **Fix prefix cache exact/prefix match with empty RotatingKVCache**: when stored cache has sliding-window layers with `keys=None` (all entries trimmed after long generation exceeding `sliding_window`), detect this and fall through to full prefill instead of producing garbage that causes client-visible hangs.
- **Fix prefix cache mutation corruption**: `fetch()` returns the same Python objects stored in the cache. `update_and_fetch()` mutates mutable attributes (`offset`, `_idx`) on these objects, corrupting the stored entry for future fetches. Fix: shallow-copy cache objects before use (`_copy_prefix_cache`), sharing immutable MLX arrays but with independent mutable state.
- **Extract `_trim_rotating_caches()` helper**: consolidates RotatingKVCache buffer trimming after prefix cache restore (previously inline in merge path).

## Test plan

- [x] Basic text generation (reasoning + content correctly separated)
- [x] Tool calling with reasoning (`tool_calls` + `reasoning_content` in response)
- [x] Streaming (reasoning chunks → content chunks transition)
- [x] Prefix cache exact hit (identical output on repeated prompts)
- [x] KV cache quantization Q8 (no degradation)
- [x] Verified Qwen3.5-122B (MLLM path) unaffected by scheduler.py fix
- [x] Gemma 4 json_schema repeat request: no longer hangs (empty RotatingKVCache detected, falls through to full prefill)
- [x] Prefix cache entries not corrupted after fetch (copy prevents mutation)

Tested with GLM-4.5-Air (106B/12B MoE, mixed 6/8-bit) and Gemma 4 26B-A4B (26B/4B MoE, mixed 6/8-bit) on Apple M3 Ultra.

### DIFF
--- a/tests/test_prefix_cache.py
+++ b/tests/test_prefix_cache.py
@@ -297,6 +297,136 @@ class TestSchedulerIntegration:
         assert default_config.prefix_cache_size == 100
 
 
+class TestTrimRotatingCaches:
+    """Regression tests for _trim_rotating_caches offset clamp."""
+
+    def test_offset_clamped_to_max_size(self):
+        """Restored cache with offset > max_size must be clamped.
+
+        Without clamping, RotatingKVCache._update_in_place computes
+        ``new_size = min(step, max_size - prev)`` which goes negative
+        when ``prev = offset % max_size`` exceeds max_size after trim.
+        """
+        mx = pytest.importorskip("mlx.core")
+        mlx_lm_cache = pytest.importorskip("mlx_lm.models.cache")
+        KVCache = mlx_lm_cache.KVCache
+        RotatingKVCache = mlx_lm_cache.RotatingKVCache
+
+        from vllm_mlx.mllm_batch_generator import MLLMBatchGenerator
+
+        # Simulate a RotatingKVCache with offset far beyond max_size
+        # (happens when prefix cache stores long-generation state)
+        rc = RotatingKVCache(max_size=128, keep=0)
+        rc.offset = 500  # Way beyond max_size
+        rc.keys = mx.zeros((1, 4, 128, 64))  # Full buffer at max_size
+        rc.values = mx.zeros((1, 4, 128, 64))
+        rc._idx = 128
+
+        # Also include a regular KVCache (should be unaffected)
+        kv = KVCache()
+        kv.offset = 200
+        kv.keys = mx.zeros((1, 4, 200, 64))
+        kv.values = mx.zeros((1, 4, 200, 64))
+
+        cache_list = [rc, kv]
+        MLLMBatchGenerator._trim_rotating_caches(cache_list)
+
+        # RotatingKVCache offset must be clamped to max_size
+        assert rc.offset <= rc.max_size
+        assert rc.offset == 128
+
+        # KVCache should be untouched
+        assert kv.offset == 200
+
+    def test_empty_rotating_cache_offset_reset(self):
+        """RotatingKVCache with keys=None should get offset reset to 0."""
+        mlx_lm_cache = pytest.importorskip("mlx_lm.models.cache")
+        RotatingKVCache = mlx_lm_cache.RotatingKVCache
+
+        from vllm_mlx.mllm_batch_generator import MLLMBatchGenerator
+
+        rc = RotatingKVCache(max_size=128, keep=0)
+        rc.offset = 500
+        rc.keys = None
+        rc.values = None
+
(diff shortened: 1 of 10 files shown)


Thump604:
Good work here - the prefix cache mutation fix (copy-on-fetch) and the empty RotatingKVCache detection are both real bugs worth fixing.

Two issues before this can merge:

1. **`_trim_rotating_caches()` doesn't clamp `offset`.** Same issue as noted on #296 - the helper trims `keys`/`values` and sets `_idx`, but `_update_in_place()` keys off `self.offset`. A restored cache with `offset > max_size` still computes negative `new_size` even after trim. Needs `layer_cache.offset = min(layer_cache.offset, layer_cache.max_size)` in the trim branch, plus a regression test covering this case.

2. **`SPECIAL_TOKENS_PATTERN` change is too broad.** Preserving `<tool_call>` makes sense for GLM tool parsing, but removing `<tool_call_reasoning>` from global cleanup changes response behavior on unrelated model paths.

Would you consider splitting this into two PRs?

- **Parser PR:** GLM-4 reasoning parser + duplicate `</think>` fix + the minimal tag-cleaning change needed for GLM
- **Cache PR:** exact-hit trim-by-1 / empty rotating cache detection / copy-on-fetch / rotating-cache normalization

That way the parser work can land independently while the cache fixes get the `offset` clamp sorted out with a regression test.

janhilgard:
Thanks for the thorough review! Both issues are now addressed in the force-pushed update:

**1. `_trim_rotating_caches()` offset clamp — Fixed**

Added `layer_cache.offset = min(layer_cache.offset, layer_cache.max_size)` at the end of the trim branch. This ensures that even when a prefix cache entry was stored with `offset >> max_size` (long generation), the restored cache won't compute negative `new_size` in `_update_in_place`.

Added regression test `TestTrimRotatingCaches::test_offset_clamped_to_max_size` in `test_prefix_cache.py` that creates a RotatingKVCache with `offset=500, max_size=128` and verifies the offset is clamped to 128 after trim.

**2. `SPECIAL_TOKENS_PATTERN` scope — Fixed**

Now only removes `</?tool_call>` (needed for GLM-4 tool format). `</?tool_call_reasoning>` is preserved in the pattern so unrelated model paths remain unaffected.

**Re: splitting into two PRs** — happy to split if you prefer. The parser and cache fixes are logically independent. Let me know and I'll create the separate branches.

Thump604 (approved):
Both issues are addressed correctly — approving.

**Offset clamp**: placing `layer_cache.offset = min(layer_cache.offset, layer_cache.max_size)` after the trim-if-block so it runs unconditionally for every RotatingKVCache with keys is the right call. The regression test exercises the exact scenario (offset=500, max_size=128, buffer already at max_size) and would fail without the clamp. Looks good.

**SPECIAL_TOKENS_PATTERN**: keeping `</?tool_call_reasoning>` in the stripping pattern while removing `</?tool_call>` is the correct scope. The `test_simple_engine.py` update confirms the intent — structural tool tags should flow through to the tool call parser rather than being silently dropped.

The rest of the changes (copy-on-fetch, empty RotatingKVCache detection, duplicate `</think>` stripping, GLM-4 parser, reasoning+tool_call ordering in server.py) all look sound. Nice consolidation of the inline trim logic into `_trim_rotating_caches`.

On the split question — both halves look ready, so I'm fine landing this as one PR. Up to you if you'd still prefer the separation for git history purposes.

PR #296, opened 2026-04-13
fix: guard tool parsing when no tools + trim RotatingKVCache after prefix restore

janhilgard:
## Summary

- **Skip tool call parsing when request has no tools**: When `--tool-call-parser` is globally configured but a request defines no tools (e.g. `response_format` JSON mode), the parser could misinterpret JSON model output as tool calls. Added guard `if request.tools:` in both OpenAI and Anthropic endpoints.

- **Trim RotatingKVCache after prefix cache restore**: Models with sliding-window attention (e.g. Gemma 4, `sliding_window=1024`) crash with "Negative dimensions not allowed" when prefix cache restores KV state with offset exceeding `max_size`. Extracted `_trim_rotating_caches()` helper and call it after prefix-hit, exact-hit, and merge paths.

## Test plan

- [x] Verify JSON `response_format` requests return content (not tool_calls) when no tools defined
- [x] Verify tool calling still works when tools are defined
- [x] Verify sliding-window model (Gemma 4) handles prefix cache restore for prompts > 1024 tokens
- [x] Run existing CI test suite

### DIFF
--- a/tests/test_specprefill_rotating_cache.py
+++ b/tests/test_specprefill_rotating_cache.py
@@ -82,3 +82,31 @@ def test_sparse_prefill_expands_tail_when_prompt_exceeds_window():
 
     flattened = [token for chunk in calls for row in chunk for token in row]
     assert flattened == [0, 2, 3, 4, 5, 6, 7, 8, 9]
+
+
+def test_trim_rotating_caches_clamps_offset():
+    """Regression: offset > max_size after prefix restore must be clamped.
+
+    RotatingKVCache._update_in_place computes
+    ``new_size = min(step, max_size - prev)`` where prev = offset.
+    If offset > max_size the result is negative → crash.
+    _trim_rotating_caches must clamp offset after trimming buffers.
+    """
+    from mlx_lm.models.cache import RotatingKVCache
+
+    from vllm_mlx.mllm_batch_generator import MLLMBatchGenerator
+
+    max_size = 4
+    cache = RotatingKVCache(max_size=max_size, keep=0)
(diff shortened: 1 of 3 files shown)


Thump604:
The tool parsing guard is clean - skipping the parser when no tools are requested is the right fix.

One blocking issue on the RotatingKVCache fix: `_trim_rotating_caches()` trims `keys`/`values` and sets `_idx`, but it never clamps `offset`. `_update_in_place()` keys off `prev = self.offset`, so a restored cache with `offset > max_size` still computes `new_size = min(step, max_size - prev)` as negative even after the trim. The `offset = 0` reset in the `keys is None` branch is correct, but the trim branch also needs `layer_cache.offset = min(layer_cache.offset, layer_cache.max_size)` (or just `= layer_cache.max_size` after trim).

A regression test covering a restored rotating cache where `offset > max_size` would also be valuable to lock this down.

janhilgard:
Good catch — updated `_trim_rotating_caches()` to clamp `offset` after trimming:

```python
# Clamp offset so _update_in_place doesn't compute negative new_size
if layer_cache.offset > layer_cache.max_size:
    layer_cache.offset = layer_cache.max_size
```

Also added a regression test `test_trim_rotating_caches_clamps_offset` that creates a RotatingKVCache with `offset=8` and `max_size=4` (simulating prefix restore) and verifies that after trimming, `offset <= max_size`.

Thump604:
The offset clamp looks correct — prevents the negative new_size path in _update_in_place after prefix restore. The regression test is exactly what's needed. CI is green. This is ready to merge.

PR #297, opened 2026-04-13
Fix test_cancellation_does_not_release_lock_before_worker_finishes

perry2of5:
Test failure: test_cancellation_does_not_release_lock_before_worker_finishes
AssertionError: CancelledError not raised

Fixed the test. It had not been updated when the `generate` method was changed to use `stream_generate`. Updated test ensures the single, previously tested path through the code is tested.

Also added lazy import cleanup and deferred MLX imports.

### DIFF
--- a/tests/test_simple_engine_cancel_serialization.py
+++ b/tests/test_simple_engine_cancel_serialization.py
@@ -3,78 +3,11 @@
 
 from __future__ import annotations
 
-import asyncio
-import threading
 import unittest
 from unittest.mock import MagicMock, patch
 
 
 class SimpleEngineCancelSerializationTests(unittest.IsolatedAsyncioTestCase):
-    async def test_cancellation_does_not_release_lock_before_worker_finishes(self):
-        """A cancelled request must not let a second MLX worker overlap."""
-        from vllm_mlx.engine.simple import SimpleEngine
-
-        model = MagicMock()
-        model.tokenizer = MagicMock()
-        model.tokenizer.encode = MagicMock(return_value=[1, 2, 3])
-        model._concurrent_count = 0
-        model._max_concurrent = 0
-
-        first_started = threading.Event()
-        release_workers = threading.Event()
-        call_count = 0
-        call_lock = threading.Lock()
-
-        def generate_side_effect(**kwargs):
-            nonlocal call_count
-            with call_lock:
-                call_count += 1
-                current_call = call_count
-                model._concurrent_count += 1
-                model._max_concurrent = max(
-                    model._max_concurrent, model._concurrent_count
-                )
-                if current_call == 1:
-                    first_started.set()
-
-            release_workers.wait(timeout=1.0)
-
-            with call_lock:
-                model._concurrent_count -= 1
-
-            result = MagicMock()
-            result.text = f"response-{current_call}"
-            result.tokens = [1, 2, 3]
-            result.finish_reason = "stop"
-            return result
-
-        model.generate = MagicMock(side_effect=generate_side_effect)
-
-        with patch("vllm_mlx.engine.simple.is_mllm_model", return_value=False):
-            engine = SimpleEngine("test-model")
-            engine._model = model
-            engine._loaded = True
-
-            task1 = asyncio.create_task(engine.generate(prompt="first", max_tokens=8))
-            await asyncio.to_thread(first_started.wait, 1.0)
-
-            task1.cancel()
-            task2 = asyncio.create_task(engine.generate(prompt="second", max_tokens=8))
-
-            await asyncio.sleep(0.05)
-            release_workers.set()
-
-            with self.assertRaises(asyncio.CancelledError):
-                await task1
-            result2 = await task2
-
-            self.assertEqual(result2.text, "response-2")
-            self.assertEqual(
-                model._max_concurrent,
-                1,
-                "cancellation released the generation lock before the first worker finished",
-            )
-
     async def test_specprefill_path_does_not_prelock_serialized_runner(self):
         """Specprefill streaming must let _run_blocking_serialized own the lock."""
         from vllm_mlx.engine.simple import SimpleEngine
--- a/vllm_mlx/engine/__init__.py
+++ b/vllm_mlx/engine/__init__.py
@@ -13,15 +13,24 @@ from .base import BaseEngine, GenerationOutput
 from .simple import SimpleEngine
 from .batched import BatchedEngine
 
-# Re-export from parent engine.py for backwards compatibility
-from ..engine_core import EngineCore, AsyncEngineCore, EngineConfig
+_ENGINE_CORE_NAMES = frozenset({"EngineCore", "AsyncEngineCore", "EngineConfig"})
+
+
+def __getattr__(name: str):
+    """Lazily re-export engine_core symbols to avoid importing mlx at package load time."""
+    if name in _ENGINE_CORE_NAMES:
+        from .. import engine_core
+
+        return getattr(engine_core, name)
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+
 
 __all__ = [
     "BaseEngine",
     "GenerationOutput",
     "SimpleEngine",
     "BatchedEngine",
-    # Core engine components
+    # Core engine components (lazy)
     "EngineCore",
     "AsyncEngineCore",
     "EngineConfig",
--- a/vllm_mlx/engine/simple.py
+++ b/vllm_mlx/engine/simple.py
@@ -692,10 +692,6 @@ class SimpleEngine(BaseEngine):
         model, then generates autoregressively. Falls back to normal generation
         on any error.
         """
-        import mlx.core as mx
-        from mlx_lm.models.cache import make_prompt_cache
-        from mlx_lm.sample_utils import make_sampler
-
         model = self._model.model
         tokenizer = self._model.tokenizer
         n_tokens = len(tokens)
@@ -711,6 +707,10 @@ class SimpleEngine(BaseEngine):
             """Score tokens, sparse prefill, generate autoregressively."""
             import time
             from types import SimpleNamespace
+
+            import mlx.core as mx
+            from mlx_lm.models.cache import make_prompt_cache
+            from mlx_lm.sample_utils import make_sampler
 
             from ..specprefill import (
                 cleanup_rope,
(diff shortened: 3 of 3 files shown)


Thump604:
The lazy import cleanup and deferred MLX imports look good. The cancellation-safe fast path is well-motivated.

Two issues with the new `_run_blocking_serialized` path for non-MLLM, no-specprefill:

1. **Usage accounting drops to zero.** `MLXLanguageModel.generate()` only returns `text`, `tokens`, and `finish_reason`. The `getattr(output, "prompt_tokens", 0)` fallback silently zeroes usage fields that the current `stream_generate()` accumulator path preserves.

2. **Stop sequences break.** `MLXLanguageModel.generate()` accepts `stop` but doesn't enforce it - stop matching lives in `stream_generate()`. The fast path would reintroduce broken stop handling.

Would you be able to either make the direct path compute usage and enforce stop, or keep `generate()` on the accumulator path and solve the cancellation safety without changing visible behavior? The lazy import and `__getattr__` changes could land independently if that helps.

waybarrios:
Took a close look at this PR. Found a few things worth addressing:

Usage accounting drops to zero on the fast path

The new `generate()` fast path calls `self._model.generate()` directly, but `MLXLanguageModel.generate()` returns a `GenerationOutput` that only has `text`, `tokens`, and `finish_reason` — no `prompt_tokens` or `completion_tokens` fields. So `getattr(output, "prompt_tokens", 0)` always returns `0`.

```python
# vllm_mlx/engine/simple.py - the fast path
output = await self._run_blocking_serialized(
    self._model.generate, ...
)
return GenerationOutput(
    text=text,
    tokens=output.tokens,
    prompt_tokens=getattr(output, "prompt_tokens", 0),      # always 0
    completion_tokens=getattr(output, "completion_tokens", 0), # always 0
    ...
)
```

This regresses [38479fe](https://github.com/waybarrios/vllm-mlx/commit/38479fe) which specifically fixed this by routing through `stream_generate()`.

https://github.com/waybarrios/vllm-mlx/blob/476a09bbefb980e92eb3f4b1470cb3f9fc0d243c/vllm_mlx/engine/simple.py#L305-L320

Stop sequences are silently ignored

`MLXLanguageModel.generate()` accepts `stop` in its signature but never uses it — stop matching lives in `stream_generate()`:

```python
# vllm_mlx/models/llm.py - stream_generate() has the logic
if stop:
    for stop_seq in stop:
        if stop_seq in accumulated_text:
            should_stop = True
            ...

# vllm_mlx/models/llm.py - generate() does NOT
def generate(self, prompt, ..., stop=None, ...):
    # stop is accepted but never referenced in the body
```

https://github.com/waybarrios/vllm-mlx/blob/476a09bbefb980e92eb3f4b1470cb3f9fc0d243c/vllm_mlx/models/llm.py#L279-L283

The new fast path has no test coverage

Both modified tests set `engine._draft_model = MagicMock()` which routes around the fast path:

```python
# tests/test_simple_engine.py
engine._draft_model = MagicMock()  # forces _draft_model is None → False
                                    # so the fast path is never hit
```

The `if not self._is_mllm and self._draft_model is None` branch is never exercised. A test with `_draft_model = None` would've caught both issues above.

https://github.com/waybarrios/vllm-mlx/blob/476a09bbefb980e92eb3f4b1470cb3f9fc0d243c/tests/test_simple_engine.py#L328-L330

Cache-HIT path still runs Metal ops outside the serialized worker

In `_stream_generate_text`, the cache-HIT block runs Metal operations on the async event-loop thread, before `_run_all`:

```python
# vllm_mlx/engine/simple.py - cache HIT block (outside _run_all)
backbone_cache = make_prompt_cache(self._text_model)  # Metal op
prompt_to_send = mx.array(suffix_tokens)               # Metal op

# ... later ...
await self._run_blocking_serialized(_run_all)  # only this is serialized
```

https://github.com/waybarrios/vllm-mlx/blob/476a09bbefb980e92eb3f4b1470cb3f9fc0d243c/vllm_mlx/engine/simple.py#L996-L1010

Suggestion: the simplest fix might be to compute `prompt_tokens` via the tokenizer and enforce stop sequences inline on the fast path. Or fall back to the accumulator approach and optimize later. Either way, a test hitting the fast path directly with `_draft_model = None` would help.

perry2of5:
I have a few thoughts:

1: we should add a regression test to prevent breaking the token counting again [476a09b](https://github.com/waybarrios/vllm-mlx/commit/476a09bbefb980e92eb3f4b1470cb3f9fc0d243c)

2: we should add cancel tests to address more cases

The cancel test right now only tests concurrent execution of the `generate` method of `SimpleEngine`. I think we need to expand this. My first question is: is testing at the generate method level the right granularity? Or should we be testing closer to the metal call level which in practice would mean testing calls into mlx, mlx-lm, etc. or groups of calls into those libraries. I'm researching to see if we could test just calls into those libraries or if some calls need to be grouped. For example, lines `[980](https://github.com/waybarrios/vllm-mlx/blob/main/vllm_mlx/engine/simple.py#L980)` to 984 of SimpleEngine

```
                        backbone_cache = make_prompt_cache(self._text_model)
                        for i, saved_state in enumerate(self._system_kv_snapshot):
                            backbone_cache[i].state = saved_state

                        prompt_to_send = mx.array(suffix_tokens)
```

I'll work on adding the token counting test and research more on the concurrency.

perry2of5:
In the end, it was a broken test, not an defect in `SimpleEngine`. I have fixed the test and verified that the second task does not access MLX through stream_generate before the cancel on the first task takes affect.

As noted by waybarrios, there are other paths through the code which still expose the possibilty of MLX being used from multiple coroutines or threads at one time. It may be worth moving the MLX class into a worker watching a queue to guarantee only a single action is sent to metal at on time to avoid corrouption.

Thump604:
Good work tracking this down. The test fix is the right call — the second task was racing into MLX before the first cancel completed.

On the broader Metal serialization question: the cache-HIT path running `make_prompt_cache()` and `mx.array()` outside the serialized worker is a real gap. We hit the same class of issue in the registry serving path — concurrent MLX evaluations from different models produced `[_MTLCommandBuffer addCompletedHandler:]` assertions. The fix there was serializing all cross-model execution, but same-model paths like the cache-HIT block have the same risk if two requests overlap.

The worker-queue approach would be the clean solution. Short of that, the `_run_blocking_serialized` wrapper needs to cover the cache-HIT Metal ops too — move `make_prompt_cache()` and the `mx.array()` call inside the serialized block.

For the usage accounting and stop sequence issues waybarrios flagged: those are real and block the fast path from landing. The accumulator approach (routing through `stream_generate()`) is the safe default until `generate()` gets proper usage counting and stop enforcement.

perry2of5:
### failing test
Thanks. After further review, I think the real issue is that the test was written for the old generate which didn't defer to stream_generate. 

Right now, in the test we don't have a race condition because task1 finishes before we even get to the cancel call. The test then fails because it couldn't cancel task1 which already completed. I rewrote to try to make it all async, but that actually broke the guarantees in a different way so I backed that out.

We should either remove the test or re-implement the old style generate, but with fixes for the issues you and waybarrios raised. My inclination is to assume that the performance hit for deferring to the stream_generate path is small and so we should drop the test and keep the current implementation until other, more important issues are dealt with.


### race guard
I can add the wrapper call for the cache-HIT Metal ops in this PR


### queued MLX ops
I can follow up with a PR to add a worker to centralize the MLX operations to avoid the race conditions. Do you have a good understanding of which operations need to be batched? For example these need to run as a batch based on waybarrios's comment:

```
backbone_cache = make_prompt_cache(self._text_model)  # Metal op
prompt_to_send = mx.array(suffix_tokens)               # Metal op
```

It seems like the calls into mlx-lm and mlx-vlm are all independent operations so they don't necessarily need to be batched with other calls in the area. I think I'll probably need your help on verifying the operations get grouped together, but hopefully not too much.

perry2of5:
Question: should we just drop the test for the old generate method? It isn't applicable to stream_generate and we should probably make race condition testing on that a separate PR.

Thump604:
Agree — drop the old generate-level test from this PR. The cancellation invariant that matters is at the stream_generate / _run_blocking_serialized boundary, not the top-level generate method. A separate PR for broader race condition coverage makes sense. This PR should stay focused on fixing the specific test that was broken.

janhilgard (commented):
Good progress on this PR. The discussion between @Thump604, @waybarrios, and @perry2of5 has been productive and reached a sensible consensus.

**Current state of the diff:** After the second commit (`1f1185c`), the PR now only contains two changes:

1. **Lazy `__getattr__` in `engine/__init__.py`** — Replaces eager `from ..engine_core import ...` with a `__getattr__` that defers the import until the symbols are actually accessed. This avoids pulling in MLX at package load time. Clean implementation, `__all__` is preserved for IDE/static analysis.

2. **Deferred MLX imports in `simple.py`** — Moves `import mlx.core as mx`, `make_prompt_cache`, `make_sampler`, and `mlx_stream_generate` from the outer async method scope into the inner `_run_specprefill()` and `_run_all()` functions that execute on the serialized worker thread. This is correct — these imports trigger MLX initialization and should happen on the thread that actually uses them.

**Observations:**

- The original test fix and the `generate()` fast path have been backed out, which aligns with the discussion consensus to drop the old generate-level cancellation test from this PR.
- However, the PR title still says "Fix test_cancellation_does_not_release_lock_before_worker_finishes" — the current diff no longer touches any test files. Consider updating the title and description to reflect the actual scope (e.g., "Defer MLX imports to serialized worker thread").
- The cache-HIT block (lines ~970-975 in the diff) still runs `make_prompt_cache()` and `mx.array()` on the async event-loop thread, outside `_run_blocking_serialized`. @Thump604 flagged this as a real gap — two overlapping requests could run Metal ops concurrently. Since the imports were moved into this block, it partially addresses the import issue, but the actual Metal execution is still unserialized. Consider moving these ops inside `_run_all()` as a follow-up or in this PR if scope permits.
- CI is all green, which is good.

**Summary:** The lazy import and deferred MLX import changes are clean and correct. The PR scope has narrowed to just these changes, so the metadata should be updated to match. The cache-HIT serialization gap is a known issue that can be tracked separately.

perry2of5:
This is ready for review. I did not add the concurrency guard for the mlx operations in this PR. I will send a new PR for that. This just removes the non-longer applicable test and improves the imports.

Thump604 (approved):
The lazy imports and test removal are correct.

Two minor observations:

1. The cache-HIT Metal ops (`make_prompt_cache`, `.state` assignment, `mx.array`) at lines 970-980 are still outside the serialized worker. Acknowledged as a follow-up — just confirming that's tracked.

2. `make_sampler()` moving into `_run_all` is a subtle correctness fix — previously it could touch MLX from the async event loop thread. Good catch.

The `__getattr__` lazy import in `engine/__init__.py` is clean.

perry2of5:
Opened https://github.com/waybarrios/vllm-mlx/issues/360 to track #1: metal ops. Thanks for the reviews.

perry2of5:
This fixes one of the failures in issue #364: tests/test_simple_engine_cancel_serialization.py::SimpleEngineCancelSerializationTests::test_cancellation_does_not_release_lock_before_worker_finishes

PR #298, opened 2026-04-14
Add Gemma 4 audio support to chat completions

Thump604:
## Summary
- add audio extraction to chat-completions preprocessing and batched multimodal scheduling
- teach the MLLM chat and streaming paths to accept OpenAI-style `audio_url` / base64 audio content parts
- plumb `audio` through BatchedEngine, the MLLM scheduler, and batch request preprocessing so Gemma 4 processors can emit `input_features` and `input_features_mask`
- add regression coverage for audio content extraction, audio input processing, and batched message preparation

## Validation
- `PYTHONPATH=/Users/David/code/vllm-mlx /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_api_utils.py tests/test_server.py tests/test_native_tool_format.py tests/test_mllm.py tests/test_mllm_mtp_routing.py -q`
- real `/v1/chat/completions` smoke against local `gemma-4-E2B-it-5bit` on both engine paths with a short WAV request
  - BatchedEngine response: `Hello world from Gemma audio test`
  - SimpleEngine response: `Hello world from Gemma audio test.`

## Note on local validation target
The local `gemma-4-E2B-it-5bit` MLX export has the audio tower weights, but its exported `processor_config.json` was missing the official `feature_extractor` block. For the endpoint smoke only, I used a temporary validation copy of the model directory with the official Gemma 4 E2B audio feature extractor config restored. The code in this PR does not depend on that workaround; it just made the local smoke exercise the real audio path end-to-end.

### DIFF
--- a/tests/test_api_utils.py
+++ b/tests/test_api_utils.py
@@ -201,17 +201,18 @@ class TestExtractMultimodalContent:
             Message(role="system", content="You are helpful."),
             Message(role="user", content="Hello"),
         ]
-        processed, images, videos = extract_multimodal_content(messages)
+        processed, images, videos, audios = extract_multimodal_content(messages)
 
         assert len(processed) == 2
         assert processed[0] == {"role": "system", "content": "You are helpful."}
         assert processed[1] == {"role": "user", "content": "Hello"}
         assert images == []
         assert videos == []
+        assert audios == []
 
     def test_none_content(self):
         messages = [Message(role="assistant", content=None)]
-        processed, images, videos = extract_multimodal_content(messages)
+        processed, images, videos, _ = extract_multimodal_content(messages)
         assert processed[0] == {"role": "assistant", "content": ""}
 
     def test_multimodal_with_image_url(self):
@@ -227,12 +228,13 @@ class TestExtractMultimodalContent:
                 ],
             )
         ]
-        processed, images, videos = extract_multimodal_content(messages)
+        processed, images, videos, audios = extract_multimodal_content(messages)
 
         assert len(processed) == 1
         assert processed[0]["content"] == "What is this?"
         assert images == ["https://example.com/img.png"]
         assert videos == []
+        assert audios == []
 
     def test_multimodal_with_dict_image_url(self):
         messages = [
@@ -247,7 +249,7 @@ class TestExtractMultimodalContent:
                 ],
             )
         ]
-        processed, images, videos = extract_multimodal_content(messages)
+        processed, images, videos, _ = extract_multimodal_content(messages)
         assert images == ["data:image/png;base64,abc"]
 
     def test_multimodal_with_string_image_url(self):
@@ -260,7 +262,7 @@ class TestExtractMultimodalContent:
                 ],
             )
         ]
-        processed, images, videos = extract_multimodal_content(messages)
+        processed, images, videos, _ = extract_multimodal_content(messages)
         assert images == ["https://example.com/img.png"]
 
     def test_multimodal_with_video(self):
@@ -273,7 +275,7 @@ class TestExtractMultimodalContent:
                 ],
             )
         ]
-        processed, images, videos = extract_multimodal_content(messages)
+        processed, images, videos, _ = extract_multimodal_content(messages)
         assert videos == ["/path/to/video.mp4"]
 
     def test_multimodal_with_video_url(self):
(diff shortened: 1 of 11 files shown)


janhilgard (commented):
This is a substantial and well-structured PR that threads audio support through the entire stack: API extraction, engine plumbing, scheduler, batch generator, MLLM model, and server. The approach mirrors the existing video/image patterns closely, which makes the code consistent and easy to follow. Good test coverage across multiple test files.

**Strengths:**
- Consistent pattern: audio extraction follows the same structure as video/image extraction throughout
- Security limits are sensible (100 MB audio, 140 MB base64)
- Cache is correctly disabled for audio inputs (since there is no audio pixel cache yet)
- Good regression tests for `extract_multimodal_content`, `_collect_audio_inputs`, `_prepare_mllm_messages`, and base64 detection
- `_prepare_mllm_messages` correctly converts `audio_url` to `{"type": "audio"}` for HuggingFace processors

**Issues / questions:**

1. **`apply_chat_template` — `num_audios` parameter**: The `_apply_chat_template` method now accepts `num_audios` but I do not see it being forwarded to the actual template applicator call. It is only used to decide whether to call `_prepare_mllm_messages`. Is the processor's `apply_chat_template` expected to handle a `num_audios` kwarg? If not, this is fine as-is; if yes, it needs to be threaded through like `num_images`.

2. **`download_audio` duplicates `download_video` almost verbatim**: The HEAD-then-GET pattern, streaming download with size check, content-type to extension mapping — all of this is copy-pasted from `download_video`. Consider extracting a shared `_download_media(url, max_size, ext_map, media_type)` helper. Not blocking, but it would reduce ~60 lines of duplication.

3. **`stream_generate` does not disable cache for audio**: In `generate()` and `chat()` you correctly add `if use_cache and all_audio: use_cache = False`, but `stream_generate()` does not have this guard. If `stream_generate` can receive audio inputs, it should also disable caching.

4. **`all_audios` vs `all_audio` naming inconsistency**: The variable is called `all_audios` in `chat()` / `stream_chat()` on `BatchedEngine` but `all_audio` in `MLXMultimodalLM`. Minor, but worth making consistent.

5. **Missing `audio` parameter in `stream_generate` on `BatchedEngine`**: `generate()` accepts `audio: list[str] | None = None` but I do not see the same parameter added to `stream_generate()` on `BatchedEngine`. If streaming audio is supported, this needs to be threaded through.

6. **No server-level `audio` forwarding**: In `server.py`, `audios` is extracted from `extract_multimodal_content` but I do not see it being passed into `chat_kwargs` for the non-MLLM path. The MLLM path extracts audio from messages directly, so this may be fine, but it is worth confirming that the LLM engine path does not silently drop audio inputs.

Overall this is solid work. The main concern is item 5 (streaming path completeness). The rest are minor or cosmetic.

Thump604:
Thanks for the review. Rebased on top of upstream main (which now includes #351) and addressed the feedback:

1. **`num_audios` forwarding** -- this is only used to decide whether to call `_prepare_mllm_messages`; the processor's `apply_chat_template` does not take a `num_audios` kwarg. No change needed.

2. **`download_audio` / `download_video` duplication** -- extracted a shared `_download_media(url, media_type, ext_map, default_ext, timeout, max_size)` helper. Both `download_video` and `download_audio` are now two-line wrappers. Also unified the audio path to use `_request_with_safe_redirects` (matching the video path's SSRF-safe request handling).

3. **`stream_generate` cache disable for audio** -- the `stream_generate` path on `MLXMultimodalLM` does not use `use_cache` / prompt cache at all, so there is nothing to disable. The cache guard in `generate()` and `chat()` covers the cached paths.

4. **Naming inconsistency (`all_audios` vs `all_audio`)** -- `all_audios` in `batched.py` follows the pattern of `all_images`/`all_videos` there; `all_audio` in `mllm.py` follows the pattern of the lower-level mlx_vlm API. Both are consistent within their respective files.

5. **Missing `audio` parameter in `stream_generate` on `BatchedEngine`** -- fixed. Added `audio: list[str] | None = None` parameter and forwarded it to the MLLM scheduler.

6. **Server-level `audio` forwarding for non-MLLM path** -- correct, the LLM engine path does not handle audio. Audio is only meaningful for MLLM models, and the MLLM path extracts audio from messages directly.

All tests compile-clean and black-clean. Rebase resolved the mllm.py conflict from #351 merge cleanly.

PR #299, opened 2026-04-14
Revert #287: self-merged in error

Thump604:
Reverts the test-only PR #287 which was merged by the author in error. The tests are correct and should be re-merged by a different contributor after review.

Apologies for the noise — this was a process mistake, not a code issue.


janhilgard (approved):
LGTM. Clean revert of the self-merged PR #287.

The PR description is transparent about the process mistake. The test file being removed (`test_tokenizer_utils.py`) contains valid tests that should indeed be re-merged after proper review by a different contributor.

Only question: will the tests be re-submitted as a new PR? Might be worth linking to it in the description when that happens.

PR #301, opened 2026-04-14
fix(mllm): increment total_prompt_tokens when scheduling requests

### LINKED ISSUE #260: MLLMScheduler total_prompt_tokens stays at 0 even after processing image requests
## Observed

In MLLM mode with continuous batching enabled, `mllm_scheduler.total_prompt_tokens` reports 0 in `/v1/status` even after the scheduler has successfully processed image requests. The companion field `total_completion_tokens` is correct.

## Reproduction

1. Start a vllm-mlx server with an MLLM model and continuous batching (Gemma 4 26B-A4B-it, 5-bit, MLLM mode in my run).
2. Issue at least one image-bearing chat completion request (so it routes through `MLLMScheduler` rather than the text scheduler).
3. Issue a few text-only requests (so the text scheduler shows non-zero counters for comparison).
4. `curl http://127.0.0.1:8080/v1/status` and look at `engine.mllm_scheduler`.

## Snapshot from my live run

```json
"mllm_scheduler": {
  "num_requests_processed": 1,
  "total_prompt_tokens": 0,
  "total_completion_tokens": 256,
  "batch_generator": {
    "prompt_tokens": 281,
    "prompt_time": 1.7035,
    "prompt_tps": 164.95,
    "generation_tokens": 256,
    "generation_time": 4.2541,
    "generation_tps": 60.18,
    "vision_encoding_time": 0.00907,
    "num_images_processed": 1
  }
}
```

The nested `batch_generator.prompt_tokens` is 281, so the prompt token count is being measured at the generator layer but never propagated up to the scheduler counter that `/v1/status` exposes.

## Why it matters

It looks small, but anything aggregating `mllm_scheduler.total_prompt_tokens` (dashboards, billing-style accounting, capacity planning) sees zero traffic for image requests even though the model is doing real work. The `text_scheduler.total_prompt_tokens` field is correct, so cross-scheduler comparisons drift over time.

## Suggested fix

Wire `MLLMScheduler.total_prompt_tokens` to accumulate `batch_generator.prompt_tokens` (or whatever upstream accounting field is canonical) when a batch finishes, mirroring the `total_completion_tokens` accounting that already works.

## Environment

- macOS 15.5, Mac Studio M2 Ultra (128GB)
- vllm-mlx 0.2.7 (live runtime, batched engine)
- Gemma 4 26B-A4B-it 5-bit, MLLM mode, continuous batching enabled

Thump604:
## Summary
- `MLLMScheduler._schedule_waiting()` was missing `self.total_prompt_tokens += request.num_prompt_tokens`
- After image/VLM requests, `total_prompt_tokens` stayed at 0 while `total_completion_tokens` was correctly tracked
- Matches the existing pattern in the LLM `TextBatchScheduler` at `scheduler.py:1968`

## One-line fix
```python
self.total_prompt_tokens += request.num_prompt_tokens
```
Added after `scheduled.append(request)` in `_schedule_waiting()`.

## Test plan
- [ ] Start server with a VLM model, send an image request
- [ ] Check `/v1/status` — `total_prompt_tokens` should be > 0
- [ ] Verify `total_completion_tokens` still works correctly

Closes #260

### DIFF
--- a/vllm_mlx/mllm_scheduler.py
+++ b/vllm_mlx/mllm_scheduler.py
@@ -498,6 +498,8 @@ class MLLMScheduler:
             self.running[request.request_id] = request
             scheduled.append(request)
 
+            self.total_prompt_tokens += request.num_prompt_tokens
+
         # Insert into batch generator
         if batch_requests and self.batch_generator is not None:
             uids = self.batch_generator.insert(batch_requests)


Thump604 (commented):
This one looks correct. `MLLMScheduler` was updating `total_completion_tokens` but never incremented `total_prompt_tokens` when requests moved from waiting -> running, so `/v1/status` under-reported prompt usage for multimodal requests. The added increment in `_schedule_waiting()` matches the existing accounting pattern in the text scheduler and fixes the obvious gap without widening scope.

janhilgard (approved):
LGTM. Clean one-liner fix.

`_schedule_waiting()` was correctly tracking `total_completion_tokens` but missing the `total_prompt_tokens` increment. The placement after `scheduled.append(request)` matches the existing pattern in `TextBatchScheduler` at `scheduler.py:1968`.

The fix is minimal, correct, and obvious. No edge cases to worry about since `request.num_prompt_tokens` is already computed before this point.

PR #302, opened 2026-04-14
fix(streaming): guarantee data: [DONE] SSE event on all paths

### LINKED ISSUE #101: Streaming SSE responses intermittently missing data: [DONE] termination event

Thump604:
## Summary
- `stream_completion`: wraps generation loop in try/except/finally so `data: [DONE]\n\n` is always emitted, even if the engine raises mid-stream
- `_disconnect_guard`: catches exceptions from the inner generator (previously only `StopAsyncIteration` was handled) and emits `[DONE]` before breaking — this covers `stream_chat_completion` and any other streaming generator routed through the guard

## Problem
OpenAI-compatible clients hang indefinitely when an exception occurs mid-stream because the `[DONE]` termination event is never sent. The happy path already emits `[DONE]`, but engine errors, reasoning parser crashes, or tool parser failures cause the generator to exit without it.

## Test plan
- [ ] Normal streaming chat/completion still ends with `data: [DONE]`
- [ ] Simulate engine error during streaming — client should receive `[DONE]` and close cleanly
- [ ] Client disconnect during streaming still handled correctly

Closes #101

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -1041,10 +1041,10 @@ class TestSseDoneTermination:
 
     @pytest.mark.anyio
     async def test_stream_completion_exception_still_emits_done(self, monkeypatch):
-        """When engine raises mid-stream, [DONE] is still emitted."""
+        """When engine raises mid-stream, [DONE] is still emitted via _ensure_sse_terminal."""
         from vllm_mlx.api.models import CompletionRequest
         from vllm_mlx.engine.base import GenerationOutput
-        from vllm_mlx.server import stream_completion
+        from vllm_mlx.server import _ensure_sse_terminal, stream_completion
 
         import vllm_mlx.server as server
 
@@ -1061,10 +1061,12 @@ class TestSseDoneTermination:
         monkeypatch.setattr(server, "_default_max_tokens", 100)
 
         request = CompletionRequest(model="test-model", prompt="Say hello")
+        # Wrap with _ensure_sse_terminal, matching server routing
         chunks = [
             chunk
-            async for chunk in stream_completion(
-                ExplodingEngine(), "Say hello", request
+            async for chunk in _ensure_sse_terminal(
+                stream_completion(ExplodingEngine(), "Say hello", request),
+                "data: [DONE]\n\n",
             )
         ]
 
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -1001,7 +1001,7 @@ class TestSseDoneTermination:
     data: [DONE] event, even when the engine raises mid-stream.
     """
 
-    @pytest.mark.asyncio
+    @pytest.mark.anyio
     async def test_stream_completion_normal_emits_done(self, monkeypatch):
         """Normal stream_completion yields exactly one [DONE] at the end."""
         from vllm_mlx.api.models import CompletionRequest
@@ -1039,7 +1039,7 @@ class TestSseDoneTermination:
         ), f"Expected exactly 1 [DONE], got {len(done_chunks)}"
         assert chunks[-1] == "data: [DONE]\n\n", "[DONE] must be the last chunk"
 
-    @pytest.mark.asyncio
+    @pytest.mark.anyio
     async def test_stream_completion_exception_still_emits_done(self, monkeypatch):
         """When engine raises mid-stream, [DONE] is still emitted."""
         from vllm_mlx.api.models import CompletionRequest
@@ -1074,7 +1074,7 @@ class TestSseDoneTermination:
         ), f"Expected exactly 1 [DONE], got {len(done_chunks)}"
         assert chunks[-1] == "data: [DONE]\n\n", "[DONE] must be the last chunk"
 
-    @pytest.mark.asyncio
+    @pytest.mark.anyio
     async def test_ensure_sse_terminal_normal_no_duplicate(self):
         """Wrapper passes through the generator's own [DONE] without duplicating."""
         from vllm_mlx.server import _ensure_sse_terminal
@@ -1095,7 +1095,7 @@ class TestSseDoneTermination:
             len(done_chunks) == 1
         ), f"Expected exactly 1 [DONE], got {len(done_chunks)}"
 
-    @pytest.mark.asyncio
+    @pytest.mark.anyio
     async def test_ensure_sse_terminal_exception_emits_done(self):
         """Wrapper emits [DONE] when inner generator raises before reaching it."""
         from vllm_mlx.server import _ensure_sse_terminal
@@ -1117,7 +1117,7 @@ class TestSseDoneTermination:
         ), f"Expected exactly 1 [DONE], got {len(done_chunks)}"
         assert chunks[-1] == "data: [DONE]\n\n", "[DONE] must be the last chunk"
 
-    @pytest.mark.asyncio
+    @pytest.mark.anyio
     async def test_ensure_sse_terminal_anthropic_protocol(self):
         """Wrapper emits Anthropic message_stop, not OpenAI [DONE], on exception."""
         import json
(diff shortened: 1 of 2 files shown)


Thump604 (commented):
Directionally this makes sense, but I do not think we should merge it without regression coverage. This changes both `_disconnect_guard()` and `stream_completion()` SSE termination behavior, so I want tests that prove exactly one terminal `data: [DONE]` is emitted on the normal path and on an exception raised mid-stream. Until those exist, I would treat this as an unverified protocol change rather than a finished fix for #101.

Thump604 (commented):
I found one blocker even with the new tests: `_disconnect_guard()` is shared infrastructure, and it also wraps the Anthropic streaming endpoint (`/v1/messages`). Emitting OpenAI-style `data: [DONE]\n\n` from the generic guard on exceptions will leak the wrong terminal frame into the Anthropic SSE protocol. The new tests only cover `stream_completion()` and a synthetic `_disconnect_guard()` case, so they do not catch this cross-endpoint regression. I think the fix needs to stay endpoint-specific: either keep the `[DONE]` guarantee inside the OpenAI streaming generators, or make `_disconnect_guard()` accept a protocol-specific terminal frame instead of hard-coding `[DONE]`.

Thump604:
Addressed the protocol blocker in \`887c7d5\`:

**Design change:** Terminal frame guarantee is now endpoint-specific, not hard-coded in the shared guard.

- New \`_ensure_sse_terminal(generator, terminal_frame)\` wrapper guarantees exactly one terminal frame even on mid-stream exceptions
- \`_disconnect_guard()\` no longer emits any SSE frames — it stays protocol-agnostic
- All three streaming endpoints wrapped at their call sites:
  - \`/v1/completions\` → \`data: [DONE]\n\n\`
  - \`/v1/chat/completions\` → \`data: [DONE]\n\n\`
  - \`/v1/messages\` → \`event: message_stop\` (Anthropic protocol)

**Tests added:**
- \`test_ensure_sse_terminal_normal_no_duplicate\` — happy path passes through without double emission
- \`test_ensure_sse_terminal_exception_emits_done\` — engine crash → OpenAI \`[DONE]\` emitted
- \`test_ensure_sse_terminal_anthropic_protocol\` — engine crash → Anthropic \`message_stop\` emitted, NOT OpenAI \`[DONE]\`

42/42 test_server.py pass locally.

Thump604 (commented):
This addresses my prior blocker. Making the terminal frame endpoint-specific via `_ensure_sse_terminal()` keeps `_disconnect_guard()` protocol-agnostic, and the new tests cover the key cases I was worried about: normal OpenAI completion, OpenAI exception path, no double-emission when the inner generator already yields the terminal frame, and Anthropic getting `message_stop` rather than OpenAI `[DONE]`. I do not see a remaining protocol issue from my side.

janhilgard (commented):
## Review: fix(streaming): guarantee data: [DONE] SSE event on all paths

**Overall: Clean, well-structured fix for a real reliability problem.**

### Strengths

1. **`_ensure_sse_terminal` is a great abstraction.** The wrapper cleanly separates the "guarantee terminal frame" concern from the generator logic. The `emitted` flag to avoid double-emission is correct.

2. **Protocol-aware terminal frames.** The Anthropic path correctly uses `event: message_stop` instead of `data: [DONE]`. The test explicitly asserts that OpenAI terminal frames do not leak into Anthropic streams.

3. **Comprehensive test coverage.** Four targeted test cases covering:
   - Normal completion emits exactly one `[DONE]`
   - Exception mid-stream still emits `[DONE]`
   - Inner generator already yielding `[DONE]` does not produce duplicates
   - Anthropic protocol gets the correct terminal frame

4. **Error handling in `_disconnect_guard`.** The new `except Exception` block prevents unhandled generator exceptions from killing the response without logging.

5. **`stream_completion` now has try/except/finally.** The `[DONE]` yield in the `finally` block is a belt-and-suspenders approach that pairs well with the outer `_ensure_sse_terminal` wrapper.

### Potential issues

1. **Silent error swallowing in `_ensure_sse_terminal`.** When the inner generator raises, the exception is caught and logged but not re-raised. This means the client gets a clean `[DONE]` termination but never learns the stream was incomplete. Consider whether emitting an error event (e.g., a JSON chunk with an `error` field) before the terminal frame would be more useful for debugging.

2. **Double error logging.** When `stream_completion` raises:
   - The `except` block inside `stream_completion` logs it
   - `_ensure_sse_terminal` catches and logs it again
   - `_disconnect_guard` may also log it
   
   Three log entries for one error could be noisy in production. Not a blocker, but worth noting.

3. **`stream_chat_completion` is not similarly wrapped with try/except/finally.** Only `stream_completion` got the inner try/except treatment. If `stream_chat_completion` raises, it relies entirely on `_ensure_sse_terminal`. This is fine functionally but the inconsistency is worth noting.

### Minor

- The `_ensure_sse_terminal` function signature uses `AsyncIterator[str]` for both param and return. This is correct but note that it is technically an `AsyncGenerator` -- not important for runtime, just a typing nit.

Overall this is a solid reliability improvement. The core `_ensure_sse_terminal` pattern is clean and composable.

Thump604:
Both review rounds are addressed — the endpoint-specific terminal frame via `_ensure_sse_terminal()` resolved the Anthropic cross-protocol concern. Tests cover the four key cases. Ready for maintainer merge.

On janhilgard's notes: the double logging is a fair observation — happy to deduplicate in a follow-up if it's noisy in production. The `stream_chat_completion` inconsistency is intentional; it relies on the outer wrapper, but I can add the inner try/except for symmetry if preferred.

PR #303, opened 2026-04-14
fix(scheduler): respect --chunked-prefill-tokens 0 when memory cache is active

### LINKED ISSUE #178: prefix-cache + large prompts: mid_prefill_cache re-enables chunked_prefill causing crash
Title: prefix-cache + large prompts: mid_prefill_cache re-enables chunked_prefill causing crash

---

**Environment**
- Hardware: Apple M2 Max, 96GB Unified Memory
- vllm-mlx version: latest (git, March 2026)
- Python: 3.14
- Model: mlx-community/Qwen3.5-35B-A3B-4bit

**Description**

When using `--enable-prefix-cache`, the server internally enables `mid_prefill_cache` which re-enables `chunked_prefill` — even when `--chunked-prefill-tokens 0` is explicitly set. This causes a crash on large prompts (>19k tokens), specifically when using Claude Code which sends a ~20k token system prompt.

**Steps to Reproduce**

1. Start server with prefix cache enabled:
```bash
vllm-mlx serve <model> \
  --continuous-batching \
  --enable-auto-tool-choice \
  --tool-call-parser qwen \
  --chunked-prefill-tokens 0 \
  --enable-prefix-cache
```

2. Send a request with >19k prompt tokens (e.g. via Claude Code):
```bash
export ANTHROPIC_BASE_URL=http://localhost:8000
export ANTHROPIC_API_KEY=not-needed
claude
```

**Server Log (crash)**
[mid_prefill_cache] enabled, interval=8192
[chunked_prefill] installed with budget=999999 tokens per step
[schedule] prompt_tokens=22615 tokens_to_prefill=22615
ERROR: too many values to unpack (expected 6)
File "scheduler.py", line 388, in _chunked_next
) = zip(*batch_prompts)
ValueError: too many values to unpack (expected 6)

**Workaround**
```bash
--chunked-prefill-tokens 0 --disable-prefix-cache
```

→ No crash, but 28s Time-to-First-Token on every request because the system prompt (~20k tokens) is never cached.

**Expected Behavior**

- `--chunked-prefill-tokens 0` should fully disable chunked_prefill, including when triggered internally by mid_prefill_cache
- OR: prefix-cache should work independently of chunked_prefill
- Result would be: fast TTFT on second request due to cached system prompt

**Impact**

Claude Code is practically unusable with prefix cache enabled:
- Without prefix cache: 28s TTFT every request 
- With prefix cache: instant crash 
- Expected with working prefix cache: ~1-2s TTFT after first request ✅

This is the main blocker for using Claude Code with vllm-mlx productively.

Thump604:
## Summary
`--chunked-prefill-tokens 0` was being silently overridden when `memory_aware_cache` was active (default-on). The condition at `scheduler.py:1275` OR'd the two, re-enabling chunked prefill with a 999,999 token budget even when the user explicitly disabled it.

## One-line fix
```python
# Before (broken):
need_chunked = chunked_budget > 0 or self.memory_aware_cache is not None

# After:
need_chunked = chunked_budget > 0
```

`memory_aware_cache` fetch/store works independently via the completion-time cache extraction path (`_process_responses` at line 2154). The mid-prefill save callback is a performance optimisation for long prompts, not a requirement for cache correctness.

## Test plan
- [x] `test_chunked_prefill_zero_means_disabled` — config with tokens=0 and memory cache=True does NOT enable chunked prefill
- [x] `test_chunked_prefill_positive_enables` — positive token count still enables chunked prefill
- [ ] Start server with `--chunked-prefill-tokens 0 --enable-prefix-cache`, send >20k token prompt — no crash

Closes #178

### DIFF
--- a/tests/test_batching.py
+++ b/tests/test_batching.py
@@ -834,5 +834,52 @@ class TestEngineAsync:
         assert not engine.engine.is_running()
 
 
+class TestChunkedPrefillConfig:
+    """Regression tests for chunked prefill configuration (#178)."""
+
+    def test_prompt_cache_save_installed_without_chunked_prefill(self):
+        """When chunked_prefill_tokens=0 but memory_aware_cache is active,
+        _install_prompt_cache_save should still patch _process_prompts."""
+        from vllm_mlx.scheduler import _install_prompt_cache_save
+
+        calls = []
+
+        class FakeBatchGen:
+            def _process_prompts(self, prompts):
+                class FakeBatch:
+                    uids = [42]
+                    num_tokens = [0]
+
+                    def extract_cache(self, idx):
+                        return f"cache-{idx}"
+
+                return FakeBatch()
+
+        bg = FakeBatchGen()
+        orig_fn = bg._process_prompts
+        _install_prompt_cache_save(bg, lambda uid, cache: calls.append((uid, cache)))
+
+        # _process_prompts was patched
+        assert bg._process_prompts is not orig_fn
+        bg._process_prompts([])
+        assert calls == [(42, "cache-0")], f"Expected callback to fire, got {calls}"
+
+    def test_chunked_prefill_zero_does_not_install_chunked_next(self):
+        """chunked_prefill_tokens=0 must not install the chunked _next patch,
+        even when use_memory_aware_cache=True."""
+        config = SchedulerConfig(chunked_prefill_tokens=0, use_memory_aware_cache=True)
+        need_chunked = config.chunked_prefill_tokens > 0
+        assert not need_chunked, (
+            "chunked_prefill_tokens=0 must not enable the chunked prefill "
+            "monkey-patch even when use_memory_aware_cache=True"
+        )
+
+    def test_chunked_prefill_positive_enables(self):
+        """Positive chunked_prefill_tokens should enable chunked prefill."""
+        config = SchedulerConfig(chunked_prefill_tokens=4096)
+        need_chunked = config.chunked_prefill_tokens > 0
+        assert need_chunked
+
+
 if __name__ == "__main__":
     pytest.main([__file__, "-v"])
--- a/vllm_mlx/scheduler.py
+++ b/vllm_mlx/scheduler.py
@@ -129,6 +129,33 @@ class SchedulerOutput:
     has_work: bool = False
 
 
+def _install_prompt_cache_save(batch_gen: "BatchGenerator", prompt_cache_save) -> None:
+    """Monkey-patch ``_process_prompts`` to capture prompt-only cache state.
+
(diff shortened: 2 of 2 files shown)


janhilgard (commented):
## Review: fix(scheduler): respect --chunked-prefill-tokens 0 when memory cache is active

**Overall: Clean, well-scoped fix with good separation of concerns.**

### Strengths

1. **Root cause is clearly addressed.** The previous logic `need_chunked = chunked_budget > 0 or self.memory_aware_cache is not None` silently re-enabled chunked prefill when the user explicitly set `--chunked-prefill-tokens 0`. The fix correctly decouples these two orthogonal features.

2. **`_install_prompt_cache_save` is a clean extraction.** Pulling the `_process_prompts` monkey-patch into its own function means the prompt cache capture works independently of chunked prefill. This is the right decomposition: memory-aware cache needs the save callback, but it does not need the chunked `_next` patch.

3. **The install order is correct.** The comment "call this **before** `_install_chunked_prefill`" is important because chunked prefill also patches `_process_prompts`. The code correctly installs the lightweight hook only when `not need_chunked`, so there is no conflict.

4. **Good test coverage.** Three tests:
   - `_install_prompt_cache_save` actually fires the callback
   - `chunked_prefill_tokens=0` does not enable chunked patching even with `use_memory_aware_cache=True`
   - Positive `chunked_prefill_tokens` still enables it

### Potential issues

1. **`_install_prompt_cache_save` does not check `batch.num_tokens[e] == 0` semantics.** The callback fires only when `batch.num_tokens[e] == 0`, which presumably indicates a "prompt-only" batch entry that has completed prefill. This condition should be documented more explicitly -- someone reading the code later may not understand why `num_tokens == 0` is the trigger.

2. **The `_batch_cls` import is unused.** The function imports `Batch` but never uses `_batch_cls` anywhere. It appears to be a leftover from an earlier iteration. Consider removing it.

3. **Silent exception swallowing.** The bare `except Exception: pass` in the patched `_process_prompts` means if `extract_cache` fails, the error is silently dropped. For a cache optimization this is probably acceptable, but a `logger.debug` would help with diagnosing cache misses.

4. **The `prompt_cache_cb` variable is hoisted above the `if need_chunked` block.** This is fine structurally, but the flow could be clearer -- `prompt_cache_cb` is created unconditionally when `memory_aware_cache is not None`, then passed into `_install_chunked_prefill` when chunked is enabled, or into `_install_prompt_cache_save` when it is not. The logic is correct but relies on reading both branches to understand the full picture.

### Minor

- The `_self=batch_gen` default argument in the inner function is unused (the function accesses `batch_gen` from the closure directly via `_orig_process_prompts`). It looks like a leftover from a refactor.

Good fix that respects user intent. The separation of prompt cache saving from chunked prefill is the right architectural choice.

Issue #107, opened 2026-02-23
Streaming mode returns tool calls as raw content text instead of structured tool_calls delta


Dubworx:
## Bug Report

**vllm-mlx version:** v0.2.5  
**Hardware:** Apple M1 Ultra, 128GB RAM  
**OS:** macOS Sequoia (Darwin 24.6.0)  
**Model:** `Qwen3-Coder-30B-A3B-Instruct-MLX-4bit`  
**Port:** 7780

## Summary

When streaming is enabled, tool calls are sent as raw `<tool_call>` XML markup in the `content` field with `tool_calls: null`, instead of being parsed into structured `tool_calls` delta objects per the OpenAI API spec. Non-streaming tool calls work correctly with the same server configuration.

## Server Start Command

```bash
vllm-mlx serve /path/to/Qwen3-Coder-30B-A3B-Instruct-MLX-4bit --port 7780
```

> **Note:** Server started WITHOUT `--enable-auto-tool-choice` or `--tool-call-parser` flags. The `--reasoning-parser qwen3` flag is known to cause Metal GPU assertion failures, and these flags were omitted to avoid similar crashes. The non-streaming path works correctly without these flags, demonstrating the server CAN parse tool calls — it just doesn't do so in streaming mode.

### Non-streaming (works correctly)

```bash
curl -s http://localhost:7780/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "Qwen3-Coder-30B-A3B-Instruct-MLX-4bit",
    "stream": false,
    "tools": [{"type": "function", "function": {"name": "list_directory", "description": "List files in a directory", "parameters": {"type": "object", "properties": {"path": {"type": "string"}}, "required": ["path"]}}}],
    "messages": [{"role": "user", "content": "List the files in my home directory /Users/testuser"}]
  }'
```

**Result (correct):**
```json
{
  "choices": [{
    "message": {
      "role": "assistant",
      "content": null,
      "tool_calls": [{"id": "call_62723f91", "type": "function", "function": {"name": "list_directory", "arguments": "{\"path\": \"/Users/testuser\"}"}}]
    },
    "finish_reason": "tool_calls"
  }]
}
```

### Streaming (broken)

```bash
curl -s http://localhost:7780/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "Qwen3-Coder-30B-A3B-Instruct-MLX-4bit",
    "stream": true,
[...]

PR #304, opened 2026-04-14
fix(server): stream structured tool calls without parser flags

Thump604:
## Summary
- add a generic streaming tool-parser fallback when tools are present but no explicit parser flags are configured
- keep configured parser behavior unchanged when  is set
- add regression coverage for streaming tool-call parsing without parser flags and for plain streamed text with tools present

## Why
Issue #107 shows a real mismatch between non-streaming and streaming behavior: non-streaming chat completions already fall back to generic tool parsing when no parser flags are configured, but streaming skipped tool parsing entirely and leaked raw tool markup as content.

This change makes the streaming path match the existing generic non-streaming behavior for the same request shape.

## Validation
- ============================= test session starts ==============================
platform darwin -- Python 3.12.12, pytest-9.0.2, pluggy-1.6.0
rootdir: /tmp/vllm-mlx-issue107
configfile: pytest.ini (WARNING: ignoring pytest config in pyproject.toml!)
plugins: asyncio-1.3.0, anyio-4.13.0
asyncio: mode=Mode.STRICT, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 42 items / 3 deselected / 39 selected

../../../tmp/vllm-mlx-issue107/tests/test_server.py .................... [ 51%]
...................                                                      [100%]

======================= 39 passed, 3 deselected in 2.51s =======================
-

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -713,6 +713,177 @@ class TestStreamChatCompletion:
     """Tests for streaming chat completion behavior."""
 
     @pytest.mark.anyio
+    async def test_stream_without_parser_flags_emits_structured_tool_calls(
+        self, monkeypatch
+    ):
+        """Streaming tools should still parse without explicit parser flags."""
+        from vllm_mlx.engine.base import GenerationOutput
+        from vllm_mlx.server import (
+            ChatCompletionRequest,
+            Message,
+            stream_chat_completion,
+        )
+        import vllm_mlx.server as server
+
+        class FakeEngine:
+            model_name = "fake-engine"
+
+            async def stream_chat(self, messages, **kwargs):
+                chunks = [
+                    GenerationOutput(text="", new_text="<tool_call>", finished=False),
+                    GenerationOutput(
+                        text="",
+                        new_text="<function=list_directory>",
+                        finished=False,
+                    ),
+                    GenerationOutput(
+                        text="",
+                        new_text="<parameter=path>/Users/testuser</parameter>",
+                        finished=False,
+                    ),
+                    GenerationOutput(
+                        text="",
+                        new_text="</function>",
+                        finished=False,
+                    ),
+                    GenerationOutput(
+                        text="",
+                        new_text="</tool_call>",
+                        finished=True,
+                        finish_reason="stop",
+                        prompt_tokens=5,
+                        completion_tokens=7,
+                    ),
+                ]
+                for chunk in chunks:
+                    yield chunk
+
+        monkeypatch.setattr(server, "_model_name", "served-model")
+        monkeypatch.setattr(server, "_reasoning_parser", None)
+        monkeypatch.setattr(server, "_enable_auto_tool_choice", False)
+        monkeypatch.setattr(server, "_tool_call_parser", None)
+        monkeypatch.setattr(server, "_tool_parser_instance", None)
+
+        request = ChatCompletionRequest(
+            model="served-model",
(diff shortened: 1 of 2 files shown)


janhilgard (commented):
## Review: fix(server): stream structured tool calls without parser flags

**Overall: Good feature that fills a gap, but the approach has some complexity concerns.**

### What this does

When the server has no explicit `--tool-call-parser` or `--enable-auto-tool-choice` flags, but the request includes `tools`, this PR auto-instantiates an `AutoToolParser` as a fallback. This means streaming responses get structured `tool_calls` instead of raw markup leaking as content -- matching the existing non-streaming behavior.

### Strengths

1. **Correct diagnosis of the asymmetry.** Non-streaming requests already fall through to generic `parse_tool_calls()` which handles tool markup. Streaming had no equivalent fallback.

2. **`_get_streaming_tool_parser` is well-structured.** The function checks `tool_choice == "none"`, tries the configured parser first, then falls back to `AutoToolParser`. This layering is clean.

3. **`_streaming_tool_markup_possible` consolidates the heuristic.** Replacing `"<" in content` with a proper marker tuple check is an improvement. The markers cover XML-based formats, Mistral, Qwen bracket format, MiniMax, and Anthropic invoke format.

4. **Two good tests.** One verifies structured tool calls appear in streaming output; the other verifies plain text is not interfered with.

5. **`_tool_choice_disabled` is a useful extraction.** Encapsulating the `tool_choice == "none"` check prevents duplicated logic.

### Potential issues

1. **Performance regression for all streaming requests with tools.** Previously, the fast path only checked `"<" in text`. Now `_streaming_tool_markup_possible` calls `any(marker in text for marker in _STREAMING_TOOL_MARKERS)` on `tool_accumulated_text + content` for every chunk. For long conversations with tools, `tool_accumulated_text` can grow to tens of thousands of characters. The `O(n * m)` substring search on every token could add measurable latency. Consider checking only on `content`/`delta_text` (the new portion) rather than the full accumulated text.

2. **Auto-parser instantiation happens on every request.** In `_get_streaming_tool_parser`, when there is no configured parser but tools are present, a new `AutoToolParser` is created per request. For a hot path under load this could be a concern. Consider caching or reusing the parser instance.

3. **The marker list is missing the bare bracket pattern from PR #305.** `_STREAMING_TOOL_MARKERS` includes `[Calling tool:` and `[TOOL_CALLS]` but not bare `[func(` patterns. If #305 merges first, this will need updating.

4. **Duplicated marker-check logic across two PRs.** PR #305 adds `_STREAMING_BARE_BRACKET_MARKER` and `_STREAMING_BARE_BRACKET_PARTIAL`, while this PR adds `_STREAMING_TOOL_MARKERS` and `_streaming_tool_markup_possible`. These two PRs will conflict. It would be good to coordinate which one lands first and have the other rebase.

### Minor

- The `global _tool_parser_instance` removal from both streaming paths is good cleanup, matching the move to `_get_streaming_tool_parser`.
- The fallback end-of-stream check now uses `_streaming_tool_markup_possible(tool_accumulated_text)` instead of a hardcoded list of 3 patterns -- this is a nice consolidation.

Solid improvement that closes a real gap in streaming tool call handling. The main concern is the coordination with PR #305 and the performance characteristics of marker scanning on accumulated text.

Thump604:
Follow-up from the two-week merge-readiness audit: I opened #484 to track the streaming tool parser marker/performance audit that spans this PR and #305.

PR #305, opened 2026-04-14
fix(auto-parser): support bare bracket tool calls

### LINKED ISSUE #146: auto tool parser: bare bracket format [func({...})] not recognized
## Description

Some models (observed with `Qwen3.5-35B-A3B-8bit`) occasionally generate tool calls in a bare bracket format:

```
[read({"file_path": "/some/path"})]
```

This is distinct from the three formats the `auto` parser already handles:
1. `<tool_call><function=name>...</function></tool_call>` (nemotron/XML)
2. `[Calling tool: func({...})]` (Qwen bracket)
3. `[{"name":"func","arguments":{...}}]` (raw JSON)

Format 4 has no `Calling tool:` prefix, so the Qwen bracket matcher doesn't catch it. It passes through as raw content text.

## Workaround

Added to `auto_tool_parser.py`:

```python
BARE_BRACKET_PATTERN = re.compile(r"\[(\w+)\((\{.*?\})\)\]", re.DOTALL)
```

With a matching block in `extract_tool_calls()` between the Qwen bracket and Nemotron checks, plus explicit tool name markers in the streaming path (`[read(`, `[write(`, `[exec(`, etc.) and a regex fallback `re.search(r'\[\w+\(\{', current_text)`.

## Environment

- vllm-mlx 0.2.6
- Model: `Qwen3.5-35B-A3B-8bit` (MLX 8-bit)
- `--enable-auto-tool-choice --tool-call-parser auto`

Thump604:
## Summary
- teach the auto tool parser to recognize bare bracket tool calls like `[read({"file_path": "/tmp/test.py"})]`
- wake the streaming server fast-path up on bare bracket tool markup so tool calls do not leak as raw text
- add regression coverage for parser extraction and streaming chat completions

## Test plan
- [x] `PYTHONPATH=/tmp/vllm-mlx-issue146 /opt/ai-runtime/venv-live/bin/python -m pytest /tmp/vllm-mlx-issue146/tests/test_tool_parsers.py -q`
- [x] `PYTHONPATH=/tmp/vllm-mlx-issue146 /opt/ai-runtime/venv-live/bin/python -m pytest /tmp/vllm-mlx-issue146/tests/test_server.py -q`
- [x] `/opt/ai-runtime/venv-live/bin/python -m black --check --fast /tmp/vllm-mlx-issue146/vllm_mlx/tool_parsers/auto_tool_parser.py /tmp/vllm-mlx-issue146/vllm_mlx/server.py /tmp/vllm-mlx-issue146/tests/test_tool_parsers.py /tmp/vllm-mlx-issue146/tests/test_server.py`

Closes #146.

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -1043,6 +1043,93 @@ class TestStreamChatCompletion:
         assert payloads[2]["choices"][0]["finish_reason"] == "stop"
 
     @pytest.mark.anyio
+    async def test_auto_parser_streams_bare_bracket_tool_calls(self, monkeypatch):
+        """Bare bracket tool calls should stream as structured tool_calls."""
+        from vllm_mlx.engine.base import GenerationOutput
+        from vllm_mlx.server import (
+            ChatCompletionRequest,
+            Message,
+            stream_chat_completion,
+        )
+        import vllm_mlx.server as server
+
+        class FakeEngine:
+            model_name = "fake-engine"
+
+            async def stream_chat(self, messages, **kwargs):
+                chunks = [
+                    GenerationOutput(text="", new_text="[read(", finished=False),
+                    GenerationOutput(
+                        text="",
+                        new_text='{"file_path": "/tmp/test.py"}',
+                        finished=False,
+                    ),
+                    GenerationOutput(
+                        text="",
+                        new_text=")]",
+                        finished=True,
+                        finish_reason="stop",
+                        prompt_tokens=4,
+                        completion_tokens=3,
+                    ),
+                ]
+                for chunk in chunks:
+                    yield chunk
+
+        monkeypatch.setattr(server, "_model_name", "served-model")
+        monkeypatch.setattr(server, "_reasoning_parser", None)
+        monkeypatch.setattr(server, "_enable_auto_tool_choice", True)
+        monkeypatch.setattr(server, "_tool_call_parser", "auto")
+        monkeypatch.setattr(server, "_tool_parser_instance", None)
+
+        request = ChatCompletionRequest(
+            model="served-model",
+            messages=[Message(role="user", content="hi")],
+            tools=[
+                {
+                    "type": "function",
+                    "function": {
+                        "name": "read",
+                        "description": "Read a file",
+                        "parameters": {
+                            "type": "object",
+                            "properties": {"file_path": {"type": "string"}},
+                            "required": ["file_path"],
+                        },
+                    },
+                }
+            ],
+            stream=True,
+        )
+
+        chunks = [
+            chunk
+            async for chunk in stream_chat_completion(
+                FakeEngine(), request.messages, request
+            )
+        ]
+
+        payloads = [
+            json.loads(chunk.removeprefix("data: ").strip())
+            for chunk in chunks
+            if chunk != "data: [DONE]\n\n"
+        ]
+        tool_payloads = [
+            payload
+            for payload in payloads
+            if payload["choices"] and payload["choices"][0]["delta"].get("tool_calls")
+        ]
+
+        assert len(tool_payloads) == 1
+        delta = tool_payloads[0]["choices"][0]["delta"]
+        assert delta["tool_calls"][0]["function"]["name"] == "read"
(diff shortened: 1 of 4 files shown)


janhilgard (commented):
## Review: fix(auto-parser): support bare bracket tool calls

**Overall: Correct feature addition with good test coverage, but the streaming detection logic has a concerning complexity issue.**

### What this does

Adds support for the `[func_name({...})]` bare bracket tool call format to the `AutoToolParser`. This format is used by some models that do not prefix with `[Calling tool:` or use XML/Mistral-style markup.

### Strengths

1. **Clean pattern addition in `extract_tool_calls`.** The `BARE_BRACKET_PATTERN` is inserted at the right position in the priority chain (after Qwen bracket, before Nemotron). The pattern `\[(\w+)\((\{.*?\})\)\]` correctly captures the function name and JSON arguments.

2. **Good test coverage.** Three new tests:
   - `test_detects_bare_bracket` for non-streaming extraction
   - `test_auto_streaming_bare_bracket` for incremental streaming
   - `test_auto_parser_streams_bare_bracket_tool_calls` for end-to-end server streaming

3. **Server-side streaming detection.** `_streaming_tool_markup_possible` is added to detect bare bracket patterns in accumulated text, replacing the previous `"<" in content` check.

### Issues

1. **The `has_marker` logic in `extract_tool_calls_streaming` is convoluted and likely buggy.**
   ```python
   has_marker = any(m in current_text for m in markers) and (
       self.BARE_BRACKET_PARTIAL_PATTERN.search(current_text) is not None
       or self.BARE_BRACKET_PATTERN.search(current_text) is not None
       or "[Calling tool:" in current_text
       or self.MISTRAL_TOKEN in current_text
       or "<" in current_text
   )
   ```
   The `markers` list now includes `"["` which matches nearly any text containing a square bracket (e.g. JSON arrays in content, markdown links). The second condition then acts as a guard, but the overall logic is hard to reason about.
   
   **Specific bug:** A model output like `"See [this link](url)"` would match `"["` in markers, then the second conjunction falls through to `"<" in current_text` which would be false (no angle brackets), but if any of the other conditions happen to match, it would incorrectly trigger tool parsing. The intent seems right but the expression is fragile.

2. **`BARE_BRACKET_PARTIAL_PATTERN` early return can suppress legitimate tool detections.**
   ```python
   if (
       self.BARE_BRACKET_PARTIAL_PATTERN.search(current_text) is not None
       and self.BARE_BRACKET_PATTERN.search(current_text) is None
   ):
       return None
   ```
   This returns `None` (meaning "not content, not tool call -- buffer") when a partial bare bracket is detected but the full pattern is not complete. This is correct for actual bare bracket tool calls, but could incorrectly suppress streaming output if the model produces text like `[read(` followed by something that is not a JSON object.

3. **Ordering concern with Qwen bracket pattern.** `QWEN_BRACKET_PATTERN` is `\[Calling tool:\s*(\w+)\((\{.*?\})\)\]` and `BARE_BRACKET_PATTERN` is `\[(\w+)\((\{.*?\})\)\]`. Since Qwen bracket is checked first (step 3) and bare bracket second (step 4), this is fine. But the bare bracket pattern would also match Qwen bracket format text, so if the ordering ever changes, there would be double-counting.

4. **The `_streaming_tool_markup_possible` in server.py overlaps with PR #304.** Both PRs modify the same streaming fast-path check in `server.py`. These will conflict on merge. Consider coordinating the merge order.

### Minor

- The renumbering of steps (4 -> 5, 5 -> 6, etc.) in `extract_tool_calls` is correct and helpful for readability.
- The `_STREAMING_BARE_BRACKET_PARTIAL` regex could be more precise -- `\[\w+\($` only anchors at end of string, which is correct for streaming but could miss patterns mid-text in non-streaming contexts.

The core feature is useful and the non-streaming path is clean. The streaming detection logic needs simplification -- the compound boolean with `"["` as a marker is the main concern.

Thump604:
Follow-up from the two-week merge-readiness audit: I opened #484 to track the streaming tool parser marker/performance audit that spans this PR and #304.

PR #307, opened 2026-04-14
feat: add registry-backed multi-model serving

Thump604:
## Summary

Registry-backed multi-model serving with memory-budget eviction for Apple Silicon unified memory.

- **`model_registry.py`**: Complete rewrite — `ModelManager` with acquire/release lease protocol, lazy loading, LRU memory-budget eviction, and configurable wait/fail/preempt contention policies
- **`server.py`**: Per-request engine routing via `RequestModelContext` and `_acquire_request_model`; lease-safe streaming via `_stream_with_model_context`; per-request reasoning/tool parser instantiation; registry-aware `/v1/models`, `/health`, `/v1/status`
- **`cli.py`**: `--models-config` flag with mutual exclusion against positional model arg and `--served-model-name`
- **Docs**: New Multi-Model Serving guide (`docs/guides/model-registry.md`), updated CLI reference and server docs
- **Tests**: 4 async ModelManager lifecycle tests + 4 server integration tests (all passing)

## Usage

```bash
# Single model (unchanged)
vllm-mlx serve mlx-community/Llama-3.2-3B-Instruct-4bit --port 8000

# Multi-model via registry
vllm-mlx serve --models-config models.yaml --port 8000
```

Registry YAML example:
```yaml
manager:
  memory_budget_gb: 100
  contention_policy:
    strategy: wait_then_preempt
    wait_timeout_s: 45
    preempt_after_s: 15

models:
  - name: fast
    path: /path/to/small-model
    preload: true
    estimated_memory_gb: 4
  - name: smart
    path: /path/to/large-model
    continuous_batching: true
    estimated_memory_gb: 36
```

Clients route requests via the standard OpenAI `model` field.

## v2 rebase notes

Rebased cleanly onto current `upstream/main` (b07e2a3). Key integration points with recent upstream changes:
- Metrics tracking (`_metrics.track_inference`, `tracker.finish`) preserved in all endpoint paths
- `_disconnect_guard()` and `_ensure_sse_terminal()` wrappers preserved
- `_normalize_messages()` and response_format handling preserved
- Streaming JSON fence stripper preserved in `stream_chat_completion`

## Test plan

- [x] `pytest tests/test_model_registry.py -x -q` — 4 passed
- [x] `pytest tests/test_server.py -x -q` — 41 passed (including 4 new registry tests)
- [x] `black --check` on all modified files
- [ ] Manual smoke test with real models and registry YAML
- [ ] Verify contention policy behavior under concurrent load

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,14 @@ vllm-mlx serve mlx-community/Llama-3.2-3B-Instruct-4bit --port 8000 --continuous
 vllm-mlx serve mlx-community/Llama-3.2-3B-Instruct-4bit --port 8000 --api-key your-secret-key
 ```
 
+### Multi-Model Serving
+
+```bash
+vllm-mlx serve --models-config /etc/vllm-mlx/models.yaml --port 8000
+```
+
+This mode uses a YAML registry of named models with lazy loading, memory-budget eviction, and configurable wait/fail/preempt behavior. See [Multi-Model Serving](docs/guides/model-registry.md) for the registry format and ops rollout guidance.
+
 ### Use with OpenAI SDK
 
 ```python
@@ -226,6 +234,7 @@ For full documentation, see the [docs](docs/) directory:
   - [Reasoning Models](docs/guides/reasoning.md)
   - [MCP & Tool Calling](docs/guides/mcp-tools.md)
   - [Continuous Batching](docs/guides/continuous-batching.md)
+  - [Multi-Model Serving](docs/guides/model-registry.md)
 
 - **Reference**
   - [CLI Commands](docs/reference/cli.md)
--- a/docs/guides/model-registry.md
+++ b/docs/guides/model-registry.md
@@ -1,0 +1,229 @@
+# Multi-Model Serving
+
+`vllm-mlx` can serve a registry of named models behind one process and one OpenAI-compatible API surface.
+
+This mode is designed for Apple Silicon machines where unified memory is the main constraint:
+
+- models load lazily on first use
+- idle models are evicted with an LRU policy under a memory budget
+- contention can be configured to wait, fail fast, or preempt active models
+- `/v1/models` reflects the configured registry instead of a single default model
+
+## When to Use It
+
+Use registry-backed serving when you want one server to expose multiple models such as:
+
+- a small low-latency chat model
+- a larger reasoning or coding model
+- a multimodal model for image or video requests
+
+Keep single-model serving when you want the smallest operational surface and the highest per-model simplicity.
+
+## Start the Server
+
+```bash
+vllm-mlx serve --models-config /etc/vllm-mlx/models.yaml --host 0.0.0.0 --port 8000
+```
+
+You can still use global serve flags such as:
+
+- `--api-key`
+- `--rate-limit`
+- `--timeout`
+- `--default-temperature`
+- `--default-top-p`
+- `--reasoning-parser`
+- `--enable-auto-tool-choice`
+- `--tool-call-parser`
+
+Do not combine `--models-config` with:
+
+- a positional model argument
+- `--served-model-name`
+
+## Registry File
+
+The registry is a YAML file with two top-level sections:
+
+- `manager`: global budget and contention behavior
+- `models`: named model entries that clients select via the OpenAI `model` field
+
+Example:
+
+```yaml
+manager:
+  memory_budget_gb: 100
+  contention_policy:
+    strategy: wait_then_preempt
+    wait_timeout_s: 45
+    preempt_after_s: 15
+
+models:
+  - name: fast
+    path: /Users/david/ai-models/mlx_models/gemma-4-E2B-it-5bit
+    preload: true
+    continuous_batching: false
+    estimated_memory_gb: 4
+
+  - name: smart
+    path: /Users/david/ai-models/mlx_models/Qwen3.5-27B-VLM-MTP-8bit
+    continuous_batching: true
+    enable_mtp: true
+    estimated_memory_gb: 36
+
+  - name: vision
+    path: /Users/david/ai-models/mlx_models/gemma-4-31B-it-6bit
+    mllm: true
+    continuous_batching: true
+    estimated_memory_gb: 44
+```
+
+## Manager Settings
+
+### `memory_budget_gb`
+
+Total resident-model budget for the registry manager.
+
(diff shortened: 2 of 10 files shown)


janhilgard (commented):
## Review: feat: add registry-backed multi-model serving

**Overall: Ambitious and well-architected feature. The core ModelManager is solid, but the scope is large and there are several areas worth discussing before merge.**

### Strengths

1. **Clean architectural separation.** `model_registry.py` owns all lifecycle/eviction logic, `server.py` routes requests through `_acquire_request_model`, and `cli.py` handles argument validation. The boundaries are clear.

2. **ModelManager lifecycle is well-designed.**
   - Shared in-flight load dedup (concurrent acquires for the same model share one load future)
   - LRU idle eviction under a memory budget
   - Five contention strategies covering real operational needs
   - `ModelLease` as an async context manager ensures release even on exceptions
   - `LoadedModel.active_tasks` tracking enables preemption via task cancellation

3. **Request-scoped parser instances.** `_build_tool_parser` and `_build_reasoning_parser` create fresh parser instances per request/stream instead of sharing mutable global state. This is the right approach for multi-model serving where different engines may have different tokenizers.

4. **Good test coverage.** The new `test_model_registry.py` covers:
   - Shared in-flight load deduplication
   - LRU eviction preserving budget
   - Preempt policy cancelling active requests
   - Non-local source requiring explicit memory estimate
   
   The server tests cover CLI parsing, `/v1/models` registry response, validation, and reasoning parser construction.

5. **Comprehensive documentation.** The `model-registry.md` guide covers the YAML format, contention strategies, sizing rules, operational checks, rollout guidance, and failure modes. This is production-quality ops documentation.

### Issues to discuss

1. **`_resolve_source` calls `ensure_model_downloaded` synchronously in an async context.** The `_instantiate_model` method is `async`, but `_resolve_source` does synchronous file I/O and potentially network downloads. This will block the event loop during model loading. Consider making `_resolve_source` async or running it in a thread executor.

2. **Memory estimation fallback is too coarse.**
   ```python
   available = _safe_available_memory_bytes()
   if available > 0:
       return max(available // 8, 1)
   ```
   Estimating a model's memory as 1/8 of available system memory when no weight files are found is likely to be wildly inaccurate. This code path triggers for local directories that exist but have no `.safetensors`/`.gguf` files (e.g., a model directory with only config files during a partial download). The `logger.warning` is good, but the estimate could cause incorrect eviction decisions.

3. **`_stream_with_model_context` nesting order.** The streaming path is now `_stream_with_model_context(model_ctx, _disconnect_guard(stream_fn, raw_request))`. This means the model lease is released in the `finally` block of `_stream_with_model_context`, which wraps the disconnect guard. If the client disconnects mid-stream, `_disconnect_guard` handles the disconnection, then `_stream_with_model_context` releases the lease. This ordering looks correct, but it means lease release happens after the disconnect guard has already broken out of its loop. Worth verifying that the lease release does not try to interact with the engine after disconnect cleanup.

4. **`yaml` is now a hard dependency.** `import yaml` at the top of `model_registry.py` means PyYAML must be installed even for single-model serving. Consider making it a conditional import inside `load_registry_config`, or adding it to optional dependencies.

5. **No hot-reload of the registry config.** Once loaded at startup, the YAML config is fixed. For operational flexibility, a `/v1/registry/reload` endpoint or SIGHUP handler would be valuable, but that can be a follow-up.

6. **`_committed_bytes_locked` counts unloading models.** Models in the `_unloading` dict are counted toward committed bytes. This is conservative (prevents over-commitment during slow unloads) but could cause unnecessary wait/fail decisions if unloading takes a long time. The behavior is correct but should be documented.

7. **`_build_reasoning_parser` fallback path.** When `_reasoning_parser_name` is None but `_reasoning_parser` is not None (pre-registry single-model mode), the function does `type(_reasoning_parser)(tokenizer)`. This reconstructs from the class of the existing instance. If the parser class has non-trivial constructor arguments beyond tokenizer, this could fail. The `try/except TypeError` handles it, but it is fragile.

8. **Non-streaming paths release lease in `finally` after `_wait_with_disconnect`.** If the output is `None` (client disconnected), the code returns `Response(status_code=499)` but the lease has already been released in the `finally` block. This is correct -- just noting that the release happens even on the 499 path.

### Minor

- The `_ModelOwnershipRegistry` is now much simpler (no weak refs, no threading locks). The simplification is welcome since the `ModelManager` handles the complex lifecycle.
- `_list_available_model_names` accesses `_model_manager._registry` directly. A property or method on `ModelManager` would be cleaner than reaching into internals.
- The `_self=batch_gen` parameter in `_patched_process_prompts` (from #303 overlap) appears to be unused in the function body.

### Merge readiness

This is a significant feature addition (~900 lines of new code in `model_registry.py` alone). The architecture is sound and the tests are good. The main items to address before merge:
- Event loop blocking in `_resolve_source`
- The `yaml` hard dependency question
- Coordination with PRs #304 and #305 which modify overlapping server.py code

Strong work overall. The contention policy design and lease-based lifecycle management are particularly well thought out.

Thump604:
Thanks for the thorough review. Pushed `bb16a79` addressing the main items:

1. **`_resolve_source` event loop blocking** -- now runs through `asyncio.to_thread` so model downloads and path resolution no longer block the event loop. Test updated to accommodate the thread yield.

2. **yaml hard dependency** -- moved to a lazy import inside `load_registry_config`. Single-model serving no longer requires PyYAML installed.

3. **`_build_reasoning_parser` fragility** -- the root issue was that `_reasoning_parser_name` was never set on startup, so the reliable name-based path was always skipped. Now set from `args.reasoning_parser` so reconstruction always goes through `get_reasoning_parser(name)` instead of `type(instance)(tokenizer)`.

4. **`_list_available_model_names` internals access** -- added `registered_model_names` property to `ModelManager`; `server.py` uses the public API now.

On the other points:
- **Memory estimation fallback** -- agree it is coarse, but the code path only triggers for local dirs without weight files (partial download), and the warning is explicit. Happy to tighten it if you have a preferred heuristic.
- **Hot-reload** -- good follow-up, keeping out of this PR scope.
- **`_committed_bytes_locked` counting unloading models** -- conservative by design to prevent over-commitment during slow unloads. Can add a docstring note.
- **`_self=batch_gen`** -- not present in the current code; likely resolved during the #303/#305 rebase.

All 4 registry tests pass.

PR #308, opened 2026-04-14
feat: add /v1/rerank endpoint for cross-encoder reranking

Thump604:
## Summary

- Add `/v1/rerank` endpoint following Jina/Cohere API convention
- Dedicated `RerankEngine` module (`vllm_mlx/rerank.py`) — separate from the chat engine
- Adapter contract (`RerankAdapter` ABC) for per-model-family tokenization, score extraction, and normalization
- `SigmoidAdapter` default for Jina Reranker v2, BGE Reranker v2, MS-MARCO MiniLM families
- MLX-native BERT/RoBERTa classifier forward pass (`vllm_mlx/rerank_forward.py`) — no transformers modeling stack at inference
- Token-budget batching (not fixed document count) with `asyncio.to_thread` for non-blocking scoring
- Concurrency semaphore prevents rerank bursts from monopolizing chat latency
- Preserves caller metadata when documents are `{text: ..., ...}` objects
- Reranker model listed in `/v1/models` with explicit `owned_by: vllm-mlx-reranker`
- `--rerank-model` CLI flag matching the existing `--embedding-model` pattern
- 404 when no reranker model configured (no unconstrained lazy loading)

## API

```
POST /v1/rerank
{
  "model": "mlx-community/jina-reranker-v2-base-multilingual",
  "query": "What is deep learning?",
  "documents": ["doc1", "doc2", {"text": "doc3", "id": "custom"}],
  "top_n": 2,
  "return_documents": true
}
```

## Test plan

- [x] `python -m pytest tests/test_rerank.py -v` — 32 tests pass
- [x] `python -m pytest tests/test_server.py tests/test_embeddings.py` — no regressions
- [x] `python -m black --check` — all files clean
- [ ] Manual smoke test with `mlx-community/jina-reranker-v2-base-multilingual` (optional)

### DIFF
--- a/tests/test_rerank.py
+++ b/tests/test_rerank.py
@@ -706,3 +706,137 @@ class TestRerankCLI:
         parser = build_parser()
         args = parser.parse_args(["serve", "test-model"])
         assert args.rerank_model is None
+
+
+# =============================================================================
+# Integration Tests - Full Round-Trip
+# =============================================================================
+
+
+class TestRerankIntegration:
+    """Full round-trip tests exercising the endpoint with mocked RerankEngine."""
+
+    @pytest.fixture()
+    def client(self):
+        from fastapi.testclient import TestClient
+
+        from vllm_mlx.server import app
+
+        return TestClient(app)
+
+    def test_full_roundtrip_string_documents(self, client):
+        """Test full request/response cycle with string documents."""
+        import vllm_mlx.server as srv
+
+        mock_engine = MagicMock()
+        mock_engine.model_name = "test-reranker"
+        mock_engine.score_pairs.return_value = [0.2, 0.8, 0.5]
+        mock_engine.count_tokens.return_value = 45
+
+        original = srv._rerank_engine
+        srv._rerank_engine = mock_engine
+        try:
+            resp = client.post(
+                "/v1/rerank",
+                json={
+                    "model": "test-reranker",
+                    "query": "machine learning",
+                    "documents": [
+                        "Cooking recipes for beginners",
+                        "Introduction to neural networks",
+                        "History of computer science",
+                    ],
+                    "top_n": 2,
+                    "return_documents": True,
+                },
+            )
+        finally:
+            srv._rerank_engine = original
+
+        assert resp.status_code == 200
+        body = resp.json()
+
+        # Verify structure
+        assert "model" in body
+        assert body["model"] == "test-reranker"
+        assert "results" in body
+        assert "usage" in body
+        assert body["usage"]["total_tokens"] == 45
+
+        # Verify top_n
+        assert len(body["results"]) == 2
+
+        # Verify sorting (descending by score)
+        assert (
+            body["results"][0]["relevance_score"]
+            >= body["results"][1]["relevance_score"]
+        )
+
+        # Verify index preservation
+        assert body["results"][0]["index"] == 1  # "Introduction to neural networks"
+        assert body["results"][1]["index"] == 2  # "History of computer science"
+
+        # Verify documents included
+        assert (
+            body["results"][0]["document"]["text"] == "Introduction to neural networks"
+        )
+        assert body["results"][1]["document"]["text"] == "History of computer science"
+
+    def test_full_roundtrip_object_documents(self, client):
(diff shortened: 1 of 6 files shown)


janhilgard (commented):
## Review: feat: add /v1/rerank endpoint for cross-encoder reranking (#308)

**Verdict: Well-structured feature with good API design, but the custom BERT forward pass is a maintenance concern and there are a few issues to address.**

### Architecture

The design follows the existing embedding endpoint pattern well:
- `RerankEngine` (model loading + scoring) in `rerank.py`
- `RerankAdapter` ABC for family-specific tokenization/scoring/normalization
- `SigmoidAdapter` default for Jina/BGE/MS-MARCO families
- Custom MLX BERT forward pass in `rerank_forward.py`
- Jina/Cohere-compatible API at `/v1/rerank`
- `--rerank-model` CLI flag mirroring `--embedding-model`
- 404 when no model configured (no lazy loading)

### Strengths

- **Adapter pattern** is the right abstraction for supporting multiple reranker families (Jina, BGE, Cohere). The ABC cleanly separates tokenization, score extraction, and normalization.
- **Token-budget batching** is a smart approach — better than fixed document count, adapts to varying document lengths.
- **Concurrency semaphore** prevents rerank bursts from monopolizing the event loop.
- **Preserves caller metadata** in object documents (`{text, id, source}` etc.), which is important for RAG pipelines.
- **Good API validation**: empty query, empty documents, top_n > documents, locked model mismatch — all covered with appropriate HTTP status codes.
- **32 thorough tests** covering models, adapters, engine, endpoint, CLI, and integration.

### Concerns

**1. Custom BERT forward pass in `rerank_forward.py` is a maintenance burden**

The 187-line hand-rolled BERT forward pass (embeddings, multi-head attention, FFN, pooler, classifier) works for standard BERT/RoBERTa but will need modification for every new architecture variant:
- DeBERTa (disentangled attention)
- ALBERT (cross-layer parameter sharing)
- Any model with GQA or different attention patterns

Consider using `mlx-lm` or `mlx-transformers` for model loading if available, and only falling back to the custom forward pass for unsupported architectures. Alternatively, document which model families are supported and fail explicitly for others.

**2. `snapshot_download` in `_load_mlx_model` contradicts CLAUDE.md guidelines**

The code calls `huggingface_hub.snapshot_download()` directly. While this is for reranker models (not LLM models), it still goes through the HF API. For this use case with small models (~300MB), this is probably fine, but worth noting.

**3. `build_parser` alias**

```python
build_parser = create_parser
```
This alias is added for "test compatibility" but pollutes the module's public API. Better to update the test to use `create_parser` directly.

**4. `/v1/models` side effect**

The PR also adds embedding engine to `/v1/models`:
```python
if _embedding_engine is not None:
    models.append(ModelInfo(id=_embedding_engine.model_name, owned_by="vllm-mlx-embedding"))
```
This is a good addition but is technically out of scope for a reranking PR. Consider splitting into a separate commit or PR.

**5. `_gelu` wrapper is unnecessary**

```python
def _gelu(x: mx.array) -> mx.array:
    """GELU activation (exact form)."""
    return nn.gelu(x)
```
This one-line wrapper adds no value. Just call `nn.gelu(x)` directly.

**6. No attention dropout**

The forward pass has no dropout, which is correct for inference. But the code should document this assumption since the same weights might be used for fine-tuning in the future.

**7. `hidden_act` from config is ignored**

The config has `"hidden_act": "gelu"` but the code always uses GELU regardless. Some cross-encoder models use `gelu_new` or `silu`. This should at minimum be checked and warned about.

### Test observations

Tests are well-structured with proper mock isolation. The `TestClassifierForward` tests actually run MLX forward passes with random weights, which is good for verifying shape correctness. The endpoint tests properly save/restore global state in try/finally blocks.

### Verdict

Good feature with clean API design and solid test coverage. The custom BERT forward pass works but is a maintenance concern. The adapter pattern is the right foundation for multi-family support. Address the `hidden_act` issue and consider reducing the scope of custom forward pass code before merge.

Thump604:
Follow-up from the two-week merge-readiness audit: I opened #483 to track the reranker forward/config support review debt from this PR, specifically the hard-coded activation / hidden_act contract.

PR #309, opened 2026-04-14
feat: SSD KV cache tiering with async promote

Thump604:
## Summary

- Add cold NVMe tier behind `MemoryAwarePrefixCache`
- Evicted entries spill to SSD via async writer thread instead of being discarded
- Cold-tier fetches use synchronous promote in `_schedule_waiting()` with RAM budget reservation (SSD I/O stays out of `fetch()` per spec; async engine-level promote is scaffolded for future use)
- SQLite WAL-mode index (replaces mutable JSON) for atomic metadata with thread-safe locking
- Per-layer serializer interface supporting KVCache, RotatingKVCache, and ArraysCache
- Crash consistency via atomic temp-file + rename writes, corrupt entry quarantine
- Disk LRU eviction and startup reconciliation
- All seven metrics from day one: `spill_count`, `spill_bytes`, `ssd_hits`, `ssd_misses`, `reload_latency`, `reload_bytes`, `promotion_failures`
- CLI flags: `--ssd-cache-dir`, `--ssd-cache-max-gb`

## Architectural requirements met

1. SSD I/O never enters synchronous `fetch()` — promote runs in `_schedule_waiting()` (sync disk read ~220ms vs 30-90s recompute; async engine-level handoff scaffolded as `promote_from_ssd()` for future integration)
2. SQLite index, not mutable JSON — thread-safe with `threading.Lock`
3. Prefix matching uses full token blob comparison; `check_ssd()` returns `match_type` and `matched_tokens` so prefix hits correctly set `remaining_tokens` for suffix prefill
4. RAM budget tentatively reserved before disk read, released on failure
5. Per-layer serializer interface with explicit support matrix
6. Atomic writes + rename, versioned schema, file permissions, quarantine
7. All seven required metrics exposed from day one

## Test plan

- [x] 53 new tests in `tests/test_ssd_cache.py` — all pass
- [x] 42 existing tests in `tests/test_memory_cache.py` — no regressions
- [x] `python -m black --check` — all files clean
- [x] Unit: SSDCacheConfig, SSDCacheStats, SSDIndex, LayerSerializer round-trips
- [x] Unit: Async writer thread (spill, atomicity, queue-full)
- [x] Integration: Eviction spill path from MemoryAwarePrefixCache
- [x] Integration: Async promote with budget reservation/denial
- [x] Integration: Full round-trip (RAM → SSD → RAM)
- [x] Integration: Hybrid ArraysCache round-trip
- [x] Integration: Capacity eviction end-to-end

### DIFF
--- a/tests/test_ssd_cache.py
+++ b/tests/test_ssd_cache.py
@@ -204,11 +204,9 @@ class TestSSDIndex:
         assert lru[0]["file_path"] == "b.safetensors"
 
 
-import tempfile
 import numpy as np
 
 from vllm_mlx.ssd_cache import (
-    LayerSerializer,
     KVCacheSerializer,
     ArraysCacheSerializer,
     get_serializer_for_layer,
--- a/tests/test_ssd_cache.py
+++ b/tests/test_ssd_cache.py
@@ -710,3 +710,24 @@ class TestCapacityManagement:
         # The orphaned index entry should have been removed
         assert tier2._index.lookup_exact(tokens) is None
         tier2.close()
+
+
+class TestCLIIntegration:
+    """Tests for CLI argument parsing of SSD cache flags."""
+
+    def test_scheduler_config_has_ssd_fields(self):
+        from vllm_mlx.scheduler import SchedulerConfig
+
+        config = SchedulerConfig()
+        assert config.ssd_cache_dir is None
+        assert config.ssd_cache_max_gb == 10.0
+
+    def test_scheduler_config_custom_ssd(self):
+        from vllm_mlx.scheduler import SchedulerConfig
+
+        config = SchedulerConfig(
+            ssd_cache_dir="/tmp/test-ssd",
+            ssd_cache_max_gb=5.0,
+        )
+        assert config.ssd_cache_dir == "/tmp/test-ssd"
+        assert config.ssd_cache_max_gb == 5.0
--- a/tests/test_ssd_cache.py
+++ b/tests/test_ssd_cache.py
@@ -309,3 +309,60 @@ class TestLayerSerializer:
     def test_get_serializer_unknown_raises(self):
         with pytest.raises(ValueError, match="Unsupported"):
             get_serializer_for_layer("not a cache layer")
+
+
+from vllm_mlx.ssd_cache import SSDCacheTier
+
+
+class TestSSDCacheTierCore:
+    """Tests for SSDCacheTier initialization and directory setup."""
+
+    def test_init_creates_directory(self, tmp_path):
+        cache_dir = str(tmp_path / "ssd_tier_test")
+        config = SSDCacheConfig(cache_dir=cache_dir)
+        tier = SSDCacheTier(config)
+        try:
+            assert os.path.isdir(cache_dir)
+            assert os.path.exists(os.path.join(cache_dir, "data"))
+            assert os.path.exists(os.path.join(cache_dir, "index.db"))
+        finally:
+            tier.close()
+
+    def test_dir_permissions(self, tmp_path):
+        cache_dir = str(tmp_path / "ssd_tier_perms")
+        config = SSDCacheConfig(cache_dir=cache_dir, dir_permissions=0o700)
+        tier = SSDCacheTier(config)
+        try:
+            stat = os.stat(os.path.join(cache_dir, "data"))
+            # Check owner permissions (at least rwx for owner)
+            assert stat.st_mode & 0o700 == 0o700
+        finally:
+            tier.close()
+
+    def test_entry_hash_deterministic(self):
+        tokens = (1, 2, 3, 4, 5)
+        h1 = SSDCacheTier._entry_hash(tokens)
+        h2 = SSDCacheTier._entry_hash(tokens)
+        assert h1 == h2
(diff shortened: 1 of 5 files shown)


janhilgard (commented):
## Review: feat: SSD KV cache tiering with async promote (#309)

**Verdict: Impressive scope and solid architecture, but several issues need attention before merge.**

### Architecture Overview

This is a substantial feature (1000+ line new module) that adds a cold NVMe tier behind MemoryAwarePrefixCache. The architecture is well thought out:

- **SQLite WAL-mode index** instead of mutable JSON — correct for crash consistency and concurrent access
- **Async writer thread** with bounded queue for non-blocking spills
- **Per-layer serializer interface** (KVCache, ArraysCache) with duck-typing dispatch
- **Atomic temp-file + rename** writes for crash consistency
- **RAM budget reservation before disk read** to prevent thrashing
- **Startup reconciliation** to clean orphaned entries
- **7 metrics from day one** — this is excellent engineering practice

The test suite is comprehensive (53 tests covering unit, integration, and end-to-end scenarios).

### Issues

**1. `lookup_prefix` is an O(N) table scan**

```python
cur = self._conn.execute(
    "SELECT ... FROM entries WHERE num_tokens <= ? ORDER BY num_tokens DESC",
    (query_len,),
)
```
This fetches ALL entries with `num_tokens <= query_len`, then does Python-side blob comparison. With 10,000 max entries, this is a full table scan on every prefix lookup. For production use with many cached entries, this will become a bottleneck.

Consider: storing a truncated token prefix (first K tokens) as an indexed column for fast filtering, then doing exact blob comparison only on candidates.

**2. `_try_promote_ssd_pending` accesses private members of `MemoryAwarePrefixCache`**

```python
self.memory_aware_cache._current_memory + memory_bytes > self.memory_aware_cache._max_memory
```
This breaks encapsulation. The cache should expose a `try_reserve(nbytes) -> bool` / `release(nbytes)` API instead of the scheduler reaching into private fields. This is fragile if MemoryAwarePrefixCache internals change.

**3. `_reconstruct_ssd_layers` creates ad-hoc `ArraysCacheLayer` objects**

```python
layer_obj = type("ArraysCacheLayer", (), {"state": state_arrays})()
```
This creates a dynamic type that may not be compatible with downstream code expecting actual `ArraysCache` instances (e.g., for `update_and_fetch` calls). If the promoted cache is used in a BatchKVCache context, this could silently produce wrong results.

**4. Thread safety concerns in `_try_promote_ssd_pending`**

The synchronous promote reads `self.memory_aware_cache._current_memory` and writes to it without holding the cache's lock. If another thread (the spill writer) triggers an eviction that also modifies `_current_memory`, there could be a race. The `_current_memory` updates should go through the cache's own methods.

**5. `hasattr(self, "_ssd_tier")` guards are defensive but fragile**

Multiple places use `hasattr(self, "_ssd_tier")` instead of initializing `_ssd_tier = None` in `__init__`. The `_ssd_tier` is only set inside a conditional block in `__init__`, so if `config.ssd_cache_dir is None`, the attribute may not exist at all. Initialize it unconditionally in `__init__` to `None`.

**6. `promote_from_ssd` double-releases budget**

```python
cache_layers = await self._ssd_tier.async_promote(tokens, reserve_budget, release_budget)
# ...
release_budget(candidate["memory_bytes"])  # <-- releases again after async_promote already handles it
```
`async_promote` calls `release_budget_fn` on failure internally. On success, it does NOT release. But then the caller releases again unconditionally after the await. If `async_promote` succeeds, the budget is released and then `store()` re-accounts — but this creates a window where the budget accounting is temporarily incorrect (briefly negative). This needs careful review.

**7. Missing `close()` in some error paths**

The `SSDIndex` connection should be closed if `SSDCacheTier.__init__` fails after opening the index. Consider a try/except in `__init__`.

### Strengths

- Excellent test coverage with real I/O (tmp_path fixtures, actual safetensors serialization)
- Atomic writes with quarantine for corrupt entries
- Budget reservation before disk read is the right pattern
- Queue-full graceful degradation (drop, don't block)
- Reconciliation on startup is essential for production use

### Verdict

The overall architecture is strong and the test coverage is excellent. The prefix lookup scalability, encapsulation violations, and budget accounting issues should be addressed before merge. The feature would be a significant improvement for long-context workloads.

sjswerdloff:
I'm surprised that of the concerns raised by @janhilgard that no issues were generated and referenced (never mind directly addressed) from the comments.

Thump604:
You're right. I should have converted Jan's review concerns into tracked follow-up issues when this merged, or explicitly documented which ones had been addressed and why the remaining risk was acceptable.

I went back through the review and opened a tracking issue here: https://github.com/waybarrios/vllm-mlx/issues/480

The issue lists each concern from the review and the current read against `main`. I do not want this to remain hidden in a merged PR thread, so the next step is to either fix the items in focused PRs, explicitly document any accepted tradeoffs, or consider a normal revert PR if the audit shows SSD tiering is unsafe as exposed.

Thump604:
Follow-up PR is open at #481. It addresses the SSD cache review-debt items tracked in #480 with focused source changes and regression coverage. I’m keeping it narrow so review can stay on the specific issues raised here.

PR #288, opened 2026-04-11
test: use anyio in async regression slice

Thump604:
Derived from the current-main restack work for #226, but limited to the pushable test-side cleanup.

Changes:
- convert the current async regression slice to `@pytest.mark.anyio`
- replace `asyncio.get_event_loop().run_until_complete(...)` with `asyncio.run(...)` in the API key tests

Why split this out:
- I also prepared the Python 3.13 CI matrix/workflow changes locally, but this auth context cannot push workflow-file updates (`.github/workflows/ci.yml`) because GitHub requires `workflow` scope for OAuth-app pushes.
- This PR lands the tested Python-side compatibility cleanup now, without waiting on that auth limitation.

Local validation:
- `PYTHONPATH=/Users/David/code/vllm-mlx-pr226-async-tests /opt/ai-runtime/venv-live/bin/python -m pytest /Users/David/code/vllm-mlx-pr226-async-tests/tests/test_server.py -q -k "verify_api_key or reasoning_stream"`
- `PYTHONPATH=/Users/David/code/vllm-mlx-pr226-async-tests /opt/ai-runtime/venv-live/bin/python -m pytest /Users/David/code/vllm-mlx-pr226-async-tests/tests/test_batching.py -q -k TestEngineAsync`
- `PYTHONPATH=/Users/David/code/vllm-mlx-pr226-async-tests /opt/ai-runtime/venv-live/bin/python -m pytest /Users/David/code/vllm-mlx-pr226-async-tests/tests/test_streaming_latency.py -q -k test_output_collector`
- `PYTHONPATH=/Users/David/code/vllm-mlx-pr226-async-tests /opt/ai-runtime/venv-live/bin/python -m black --check --fast /Users/David/code/vllm-mlx-pr226-async-tests/tests/test_batching.py /Users/David/code/vllm-mlx-pr226-async-tests/tests/test_batching_deterministic.py /Users/David/code/vllm-mlx-pr226-async-tests/tests/test_continuous_batching.py /Users/David/code/vllm-mlx-pr226-async-tests/tests/test_server.py /Users/David/code/vllm-mlx-pr226-async-tests/tests/test_streaming_latency.py`

Related to #226.

PR #226, opened 2026-03-25
test: make Python 3.13 async suite pass and cover it in CI

krystophny:
## Summary
Fix Python 3.13 test execution and add real 3.13 coverage to the project test jobs.

## Changes
- standardize async tests on `anyio` with the asyncio backend
- replace Python 3.13-fragile `asyncio.get_event_loop().run_until_complete(...)` with `asyncio.run(...)`
- remove dead `asyncio_mode = "auto"` config (pytest-asyncio reference after its removal)
- add `anyio>=4.0` as explicit dev dependency
- add Python 3.13 to the Ubuntu test matrix and the Apple Silicon MLX job

## Status
- refreshed onto current upstream `main` (`b4fa030`) on 2026-04-09
- no logic changes beyond the base refresh

## Validation
- `python -m pytest tests/ -q --timeout=120` -> `1020 passed, 2 failed, 9 skipped, 20 deselected`
- the two failures are throughput assertions in `tests/test_batching_deterministic.py::TestBatchingPerformance::test_batched_faster_than_sequential[asyncio]` and `tests/test_continuous_batching.py::TestContinuousBatchingIntegration::test_batching_improves_throughput[asyncio]`
- the async-harness changes themselves are exercising correctly; these failures are performance thresholds, not event-loop/plugin failures

PR #312, opened 2026-04-14
ci: restack Python 3.13 anyio coverage

krystophny:
## Summary
- add Python 3.13 to the Ubuntu unit-test matrix and Apple Silicon MLX job
- switch the dev/test async plugin wiring from pytest-asyncio to anyio
- simplify the shared anyio backend fixture to asyncio-only on current main

## Why
This is the minimal replacement for #226 after #288 landed the test-file conversions. It keeps only the remaining CI/dependency/backend wiring on top of current main.

## Local validation
- python -m pytest tests/test_server.py -q -k "verify_api_key or reasoning_stream"
- python -m pytest tests/test_batching.py -q -k TestEngineAsync
- python -m pytest tests/test_streaming_latency.py -q -k test_output_collector

Supersedes the remaining scope of #226.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.10", "3.11", "3.12"]
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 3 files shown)


krystophny:
Opened as the requested minimal replacement for #226 on top of current main and re-ran the focused local async validation slice before pushing.

Thump604 (approved):
Clean CI-only change — adds 3.13 to the matrix, swaps pytest-asyncio to anyio (consistent with #288), simplifies the conftest backend fixture. All 9 CI jobs green including the new 3.13 rows. Approving.

janhilgard (approved):
+1 from me. Clean, minimal CI change. Agreeing with Thump604's approval.

One small observation: the `anyio_backend` fixture in `conftest.py` is now `scope="session"` which is correct for performance, but worth noting that if trio support is ever added, this will need revisiting (not a concern today since trio is not a dependency).

All 9 CI jobs green, `pyproject.toml` dev deps are consistent with the CI install. LGTM.

PR #313, opened 2026-04-14
feat: add guided structured output decoding

Thump604:
## Summary
- add token-level guided decoding for `response_format` requests behind a new internal `vllm_mlx.guided_decoding` abstraction
- use Outlines as the first backend while keeping the public API contract as `response_format`
- preserve existing tool-calling behavior by keeping tool requests on the legacy prompt-instruction fallback
- carry guided logits processors through simple, batched, and MLLM scheduling/generation paths

## Details
- `json_object` and `json_schema` now use true token-level constrained decoding when tools are not active
- invalid `json_schema` requests return HTTP 400 before generation
- non-streaming responses still run post-generation JSON parse/validation so the external response contract stays stable
- tool requests continue using the legacy JSON instruction path so structured output guidance does not interfere with tool parsing

## Validation
- `PYTHONPATH=/tmp/vllm-mlx-guided-output /opt/ai-runtime/venv-live/bin/python -m pytest /tmp/vllm-mlx-guided-output/tests/test_structured_output.py /tmp/vllm-mlx-guided-output/tests/test_batching.py /tmp/vllm-mlx-guided-output/tests/test_mllm_continuous_batching.py /tmp/vllm-mlx-guided-output/tests/test_server.py -q`
  - `122 passed, 2 skipped, 8 deselected`
- `/opt/ai-runtime/venv-live/bin/python -m black --check --fast vllm_mlx/guided_decoding.py vllm_mlx/server.py vllm_mlx/scheduler.py vllm_mlx/mllm_scheduler.py vllm_mlx/mllm_batch_generator.py vllm_mlx/models/llm.py vllm_mlx/models/mllm.py vllm_mlx/engine/batched.py tests/test_structured_output.py tests/test_batching.py tests/test_mllm_continuous_batching.py tests/test_server.py`
- real local smoke with isolated Outlines install against `/Users/David/ai-models/mlx_models/Qwen3.5-2B-OptiQ-4bit`
  - `json_object -> {"status": "ok", "value": 7}`
  - `json_schema -> {"status": "ok", "value": 11}`

### DIFF
--- a/docs/guides/server.md
+++ b/docs/guides/server.md
@@ -605,6 +605,11 @@ See [Reasoning Models Guide](reasoning.md) for full details.
 
 Force the model to return valid JSON using `response_format`:
 
+For `json_object` and `json_schema`, vLLM-MLX uses token-level guided
+decoding when tools are not active. Tool-calling requests keep the legacy
+prompt-instruction fallback so structured output and tool parsing do not
+compete for the same generation surface.
+
 ### JSON Object Mode
 
 Returns any valid JSON:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,6 +54,8 @@ dependencies = [
     "mcp>=1.0.0",
     # JSON Schema validation for structured output
     "jsonschema>=4.0.0",
+    # Token-level guided decoding for structured output
+    "outlines>=1.2.12",
     # pytz is required by gradio but not declared as a dependency
     # See: https://github.com/waybarrios/vllm-mlx/issues/23
     "pytz>=2024.1",
--- a/tests/test_batching.py
+++ b/tests/test_batching.py
@@ -145,6 +145,11 @@ class TestSamplingParams:
         assert params.stop == ["END"]
         assert params.stop_token_ids == [1, 2]
 
+    def test_response_format_field(self):
+        """Structured-output requests must carry response_format through batching."""
+        params = SamplingParams(response_format={"type": "json_object"})
+        assert params.response_format == {"type": "json_object"}
+
 
 class TestRequestOutput:
     """Tests for RequestOutput."""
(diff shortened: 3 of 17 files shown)


janhilgard (commented):
## Review: feat: add guided structured output decoding (#313)

**Verdict: Well-architected feature, but raises concerns about the Outlines dependency and a few implementation details that need discussion.**

### Architecture

The design is sound — a `GuidedDecodingFactory` wraps Outlines' `from_mlxlm` + `get_json_schema_logits_processor` behind a clean internal interface. The key routing decision (`_should_use_guided_decoding`) correctly disables guided decoding when tools are active, falling back to the existing prompt-injection path. This avoids the fundamental conflict where guided JSON decoding and tool-call parsing would compete for the same generation surface.

The response_format is threaded through all four engine paths:
- `SimpleEngine` (LLM): via `generate()` / `stream_generate()` kwargs
- `SimpleEngine` (MLLM): via `_merge_guided_logits_processors()` 
- `BatchedEngine` (Scheduler): via `SamplingParams.response_format` -> `_build_request_logits_processors()`
- `BatchedEngine` (MLLM Scheduler): via `MLLMBatchRequest.logits_processors`

### Strengths

- Clean separation: `guided_decoding.py` is a standalone module with `normalize_response_format`, `response_format_to_schema`, `uses_guided_decoding`, and `GuidedDecodingFactory`.
- DRY: `normalize_response_format()` is extracted and reused in `tool_calling.py` (replacing two identical copies).
- Correct processor composition: guided processors are combined with penalty processors (repetition, presence) rather than replacing them.
- Lazy factory initialization avoids overhead when structured output is not used.
- Good test coverage across all paths (server, scheduler, MLLM scheduler, batching).

### Concerns

**1. Hard dependency on `outlines>=1.2.12`**

Adding `outlines` as a required dependency in `pyproject.toml` is a significant change. Outlines pulls in `torch` (or at least `interegular`, `airspeed-velocity`, `lark`, etc.) which adds substantial install size. Many vllm-mlx users run on Apple Silicon with minimal dependencies. Consider making this optional:
```python
# pyproject.toml
[project.optional-dependencies]
guided = ["outlines>=1.2.12"]
```
With a runtime check in `GuidedDecodingFactory.__init__` that raises a clear error if not installed.

**2. `SimpleEngine.stream_chat` MLLM text-only path bypass**

In `engine/simple.py`, when `response_format is not None`, the text-only fast path (which uses MTP) is skipped:
```python
and kwargs.get("response_format") is None
```
This means structured output requests always go through the slower MLLM path even for text-only prompts. This is likely intentional (Outlines processors may not work with MTP), but worth documenting.

**3. `json_object` schema is very permissive**

For `json_object` mode, the schema is:
```python
{"type": "object", "additionalProperties": True}
```
This constrains output to be a JSON object but allows any keys/values. This matches the OpenAI spec, but Outlines may compile a very large FSM for this. Worth testing whether this causes latency spikes on first request for each new `json_object` request.

**4. `GuidedDecodingFactory` is created per-scheduler but processors are per-request**

The factory calls `from_mlxlm(model, tokenizer)` once, then `build_processors()` creates a new `get_json_schema_logits_processor` for each request. Outlines compiles an FSM from the schema, which can be expensive. If many requests use the same schema (e.g., `json_object`), consider caching compiled processors by schema hash.

**5. MLLM path duplicates the merge pattern**

`_merge_guided_logits_processors` in `mllm.py` is called in three places (`stream_generate`, `chat`, `stream_chat`) with identical boilerplate. Consider extracting to a decorator or a single pre-processing step.

### Test notes

The test `test_invalid_guided_decoding_schema_returns_400` creates its own `pytest.MonkeyPatch()` instead of using the fixture parameter — this works but is unusual. The `monkeypatch.undo()` in a finally block is correct but the `@pytest.mark.anyio` test could use the fixture-based approach for consistency.

### Verdict

The feature is well-designed and the routing logic is correct. The main concern is the hard Outlines dependency — making it optional would be a significant improvement. The implementation is solid otherwise.

Thump604:
Superseded by #362 (lm-format-enforcer). Outlines FSM is incompatible with MLX lazy evaluation (mlx-lm#1157, outlines#1847). The lm-format-enforcer approach uses pure-Python logits masking that works cleanly with MLX's graph optimization.

PR #314, opened 2026-04-14
feat: add Prometheus server metrics

Thump604:
## Summary
- add opt-in Prometheus metrics with a new `/metrics` endpoint
- instrument HTTP and inference request paths, including TTFT and token counters
- expose scrape-time engine, scheduler, cache, model-registry, and MCP gauges
- wire `--enable-metrics` through both `vllm-mlx serve` and `python -m vllm_mlx.server`
- document the new metrics surface and add dedicated regression coverage

## Validation
- `python -m compileall vllm_mlx/server.py vllm_mlx/metrics.py vllm_mlx/cli.py tests/test_metrics.py`
- `PYTHONPATH=/tmp/vllm-mlx-prometheus:/tmp/vllm-mlx-metrics /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_metrics.py tests/test_server.py -q`

### DIFF
--- a/docs/guides/server.md
+++ b/docs/guides/server.md
@@ -37,6 +37,7 @@ vllm-mlx serve mlx-community/Llama-3.2-3B-Instruct-4bit --port 8000 --continuous
 | `--api-key` | API key for authentication | None |
 | `--rate-limit` | Requests per minute per client (0 = disabled) | 0 |
 | `--timeout` | Request timeout in seconds | 300 |
+| `--enable-metrics` | Expose Prometheus metrics on `/metrics` | False |
 | `--continuous-batching` | Enable batching for multi-user | False |
 | `--use-paged-cache` | Enable paged KV cache | False |
 | `--cache-memory-mb` | Cache memory limit in MB | Auto |
@@ -127,6 +128,23 @@ GET /health
 ```
 
 Returns server status.
+
+### Metrics
+
+```bash
+GET /metrics
+```
+
+Prometheus scrape endpoint for server, cache, scheduler, and request metrics.
+The endpoint is disabled by default and is enabled with `--enable-metrics`.
+
+```bash
+vllm-mlx serve mlx-community/Llama-3.2-3B-Instruct-4bit \
+  --enable-metrics
+```
+
+`/metrics` is intentionally unauthenticated. Expose it only on a trusted
+network or behind a reverse proxy / firewall that limits who can scrape it.
 
 ### Anthropic Messages API
 
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -27,6 +27,7 @@ vllm-mlx serve <model> [options]
 | `--api-key` | API key for authentication | None |
 | `--rate-limit` | Requests per minute per client (0 = disabled) | 0 |
 | `--timeout` | Request timeout in seconds | 300 |
+| `--enable-metrics` | Expose Prometheus metrics on `/metrics` | False |
 | `--continuous-batching` | Enable batching for multi-user | False |
 | `--cache-memory-mb` | Cache memory limit in MB | Auto |
 | `--cache-memory-percent` | Fraction of RAM for cache | 0.20 |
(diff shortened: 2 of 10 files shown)


Thump604 (commented):
Clean, well-scoped feature. The architecture choices are sound: lazy Prometheus init, dedicated `CollectorRegistry`, opt-in flag, `__unmatched__` cardinality collapse, and scrape-endpoint self-exclusion from the HTTP metrics middleware.

A few observations (none blocking):

**Gauge vs Counter for monotonic scraped values** — `cache_hits`, `cache_misses`, `cache_evictions`, `engine_steps_executed`, and `cache_tokens_saved` are monotonically increasing values exposed as Gauges. This is technically correct for scrape-time-sampled absolutes (you can't `.inc()` a Counter from an absolute value), but breaks the Prometheus convention that monotonic series should be typed as Counters. `rate()` still works on Gauges, but some dashboard tooling may not auto-detect monotonicity. A `# NOTE: monotonic but scraped as absolute` inline comment would help future readers.

**Metal memory unit conversion** — `metal_active_memory_gb * 1e9` uses SI giga (1,000,000,000). If the engine reports GiB (which is typical for memory), the multiplier should be `1024**3`. Worth confirming what the engine actually means by "gb."

**Scrape-time gauge refresh is sync** — `_update_engine_gauges()` runs synchronously inside the `/metrics` handler. Currently `engine.get_stats()` and `mcp_manager.get_server_status()` are lightweight dict returns so this is fine, but if they ever acquire I/O or lock contention, the scrape handler would block. Something to keep in mind.

**Test quality is good** — Tests assert exact metric line values including label combinations and counts, not just endpoint status codes. The streaming test verifies TTFT histogram count. The `FakeEngine` approach with monkeypatched `_metrics` gives proper isolation.

**Security posture is documented** — `/metrics` intentionally skips auth, which is standard for Prometheus exporters. The docs call this out with the right guidance (trusted network / reverse proxy).

LGTM. Nice work.

Thump604:
Reviewed — looks good. Clean integration, proper cardinality control, and the lazy init means zero overhead when disabled.

**Strengths:**
- Dedicated `CollectorRegistry` avoids global pollution
- `__unmatched__` path collapse prevents scraper/probe cardinality explosion
- `/metrics` self-exclusion from HTTP middleware prevents scrape feedback loops
- `InferenceTracker._finished` guard prevents double-counting
- Tests assert exact metric line values, not just endpoint liveness

**Three non-blocking observations:**

1. **Gauge vs Counter for monotonic values**: `cache_hits`, `cache_misses`, `engine_steps_executed` etc. are monotonically increasing but exposed as Gauges. This is correct for scrape-time absolutes, but some Prometheus tooling (e.g., `rate()`) expects Counters for monotonic series. Not a blocker — just something to revisit if users report alerting gaps.

2. **Metal memory unit conversion**: `* 1e9` is SI giga (10^9). If the engine reports GiB, the multiplier should be `1024**3` (~1.074e9). Worth verifying which unit `metal_memory` actually uses.

3. **Sync gauge refresh at scrape time**: `_update_engine_gauges()` is synchronous and currently fine since `get_stats()` is lightweight. Worth noting if stats collection ever gains I/O.

LGTM — ready to merge.

janhilgard (approved):
LGTM — clean, well-scoped feature. Nice work.

**Strengths:**
- Dedicated `CollectorRegistry` avoids global pollution — clean isolation
- `__unmatched__` path collapse prevents scraper/probe cardinality explosion
- `/metrics` self-exclusion from HTTP middleware prevents scrape feedback loops
- `InferenceTracker` with `_finished` / `_ttft_observed` guards prevents double-counting
- Lazy Prometheus init means zero overhead when `--enable-metrics` is not passed
- Tests assert exact metric line values including label combinations, not just endpoint liveness
- Streaming tests verify TTFT histogram count — good coverage
- Security posture documented (unauthenticated `/metrics`, trusted network guidance)

**Three non-blocking observations:**

1. **Gauge vs Counter for monotonic values**: `cache_hits`, `cache_misses`, `cache_evictions`, `engine_steps_executed`, `cache_tokens_saved` are monotonically increasing but exposed as Gauges. This is technically correct since you can't `.inc()` a Counter from an absolute scraped value, but `rate()` works better with Counters in some Prometheus tooling. A brief inline comment noting these are "monotonic but scraped as absolute" would help future readers.

2. **Metal memory unit conversion**: `metal_active_memory_gb * 1e9` uses SI giga (10^9). If the engine reports GiB (which is typical for memory), the multiplier should be `1024**3` (~1.074e9). Worth verifying which unit `get_stats()` actually returns.

3. **`prometheus-client` as hard dependency**: Added unconditionally in `pyproject.toml` but only used when `--enable-metrics` is passed. Consider moving it to an optional extras group (e.g., `[metrics]`) to keep the base install lighter — though it's a small package (~50KB), so this is a very minor point.

None of these block merge.

PR #315, opened 2026-04-14
fix: preserve reasoning when tool calls are present

### LINKED ISSUE #161: Reasoning content leaks into content field when tool calls are present (Nemotron)

Thump604:
## Summary
- extract non-streaming reasoning before running tool-call parsing
- reuse the same coexistence logic for both OpenAI and Anthropic response paths
- add endpoint-level regressions proving reasoning survives alongside structured tool calls / tool_use blocks

Closes #161.

## Validation
- `python -m compileall vllm_mlx/server.py tests/test_server.py`
- `PYTHONPATH=/tmp/vllm-mlx-issue161 /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_server.py -q`

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -1369,6 +1369,192 @@ class TestStreamChatCompletion:
         assert payloads[2]["choices"][0]["finish_reason"] == "stop"
 
 
+class TestReasoningAndToolCallsNonStreaming:
+    """Non-streaming coexistence of reasoning extraction and tool parsing."""
+
+    @pytest.fixture()
+    def client(self):
+        """Create a FastAPI test client."""
+        from fastapi.testclient import TestClient
+
+        from vllm_mlx.server import app
+
+        return TestClient(app)
+
+    def test_chat_completion_preserves_reasoning_with_tool_calls(
+        self, client, monkeypatch
+    ):
+        """Reasoning should survive when tool calls are present in final output."""
+        import vllm_mlx.server as server
+        from vllm_mlx.engine.base import GenerationOutput
+        from vllm_mlx.server import ToolCall, FunctionCall
+
+        parsed_inputs = []
+
+        class FakeEngine:
+            model_name = "fake-engine"
+            is_mllm = False
+            preserve_native_tool_format = False
+
+            async def chat(self, messages, **kwargs):
+                return GenerationOutput(
+                    text="<think>Need tool</think><tool_call>",
+                    prompt_tokens=7,
+                    completion_tokens=3,
+                    finish_reason="stop",
+                )
+
+        class FakeReasoningParser:
+            def extract_reasoning(self, model_output):
+                assert model_output == "<think>Need tool</think><tool_call>"
+                return "Need tool", "<tool_call>"
+
+        def fake_parse_tool_calls(text, request):
+            parsed_inputs.append(text)
+            if text == "<tool_call>":
+                return None, [
+                    ToolCall(
+                        id="call_1",
+                        type="function",
+                        function=FunctionCall(
+                            name="get_weather",
+                            arguments='{"city":"Paris"}',
+                        ),
+                    )
+                ]
+            return text, None
+
+        monkeypatch.setattr(server, "_engine", FakeEngine())
+        monkeypatch.setattr(server, "_model_name", "test-model")
+        monkeypatch.setattr(server, "_default_timeout", 30.0)
+        monkeypatch.setattr(server, "_default_max_tokens", 128)
+        monkeypatch.setattr(server, "_api_key", None)
+        monkeypatch.setattr(
+            server,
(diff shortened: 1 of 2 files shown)


janhilgard (commented):
## Review: fix: preserve reasoning when tool calls are present (#315)

**Verdict: Solid bug fix with correct ordering. LGTM with minor notes.**

### Problem & Solution

Previously, non-streaming responses ran tool parsing first. When tool calls were found, reasoning extraction was skipped entirely (the `if not tool_calls:` guard). This meant that model output like `<think>Need tool</think><tool_call>...` lost the reasoning segment.

The new `_extract_reasoning_and_tool_calls()` helper correctly orders the operations:
1. Extract reasoning first (strips think tags, returns cleaned text)
2. Parse tool calls from the **cleaned** text (not the raw output)

This ensures reasoning is preserved regardless of whether tool calls are present.

### Strengths

- Single helper `_extract_reasoning_and_tool_calls()` used by both OpenAI and Anthropic non-streaming paths, eliminating code duplication.
- The `allow_reasoning` parameter correctly gates on `request.enable_thinking is not False` for OpenAI, while Anthropic always allows it (matching existing behavior).
- Thorough test coverage: `TestReasoningAndToolCallsNonStreaming` has two tests covering both OpenAI (`/v1/chat/completions`) and Anthropic (`/v1/messages`) endpoints with full monkeypatching. The assertions verify that:
  - Tool parsing receives the cleaned text (not raw output)
  - `reasoning_content` is populated in the response
  - `tool_calls` / `tool_use` blocks are present alongside reasoning
  - `finish_reason` is correctly set to `tool_calls` / `tool_use`

### Edge case to consider

When `_reasoning_parser.extract_reasoning()` returns `(reasoning_text, None)` (meaning the entire output was reasoning with no cleaned text), the code sets `text_for_tool_parse = ""`. Tool parsing on an empty string should return `("", None)` or `(None, None)` in most parsers, but it might be worth confirming that all registered tool parsers handle empty input gracefully.

### One minor note

The Anthropic path no longer checks `if not tool_calls:` before extracting reasoning, which is the intended fix. However, the streaming path (not touched by this PR) may still have the same issue. If streaming also exhibits this bug, a follow-up PR would be needed.

### Verdict

Clean, correct fix. The ordering change is the right approach, and the shared helper reduces duplication. Good work.

janhilgard:
This PR has merge conflicts with the current main branch. Could you rebase/resolve the conflicts? The code itself looks good and is ready to merge once conflicts are resolved.

Thump604:
Thanks for the thorough review. Good catch on the streaming path — it has the same ordering issue. I'll file a follow-up PR for that once this one lands.

On the empty string edge case: all registered tool parsers handle empty input gracefully (return `(text, None)` or equivalent), so no issue there, but worth the callout.

PR #316, opened 2026-04-14
fix(api): emit a single reasoning_content field

### LINKED ISSUE #310: Streaming delta emits both `reasoning` and `reasoning_content` as sibling keys with identical values — breaks strict      JSON deserializers (Goose/serde)
### Summary

In streaming chat-completion responses, every `delta` chunk serializes **both** a `reasoning` field and a `reasoning_content` field as sibling keys of the same JSON object, with identical values. Strict JSON deserializers that alias both names to a single target field (e.g. [goose-rs](https://github.com/block/goose), via `#[serde(alias = "reasoning")]`) reject these payloads with `duplicate field "reasoning_content"` and abort the SSE stream.

This happens on **every** delta chunk, **regardless** of whether `--reasoning-parser` is set, and regardless of whether the model actually emits `<think>` tags. Non-reasoning models (e.g. `Qwen3-Coder-30B-A3B-Instruct-8bit`) also trigger it — their regular output ends up populating both fields.

### Reproduction

```bash
# Server:
vllm-mlx serve mlx-community/Qwen3-Coder-30B-A3B-Instruct-8bit \
  --continuous-batching --enable-auto-tool-choice --tool-call-parser qwen \
  --reasoning-parser qwen3 --port 8100 --host 127.0.0.1

# Request:
curl -sN -X POST http://127.0.0.1:8100/v1/chat/completions \
  -d '{"model":"mlx-community/Qwen3-Coder-30B-A3B-Instruct-8bit",
       "messages":[{"role":"user","content":"Say hi in one word"}],
       "stream":true,"max_tokens":10}'
```

Representative chunk (abridged):

```json
{"choices":[{"index":0,"delta":{
  "role":"assistant",
  "content":null,
  "reasoning":"Hello",
  "tool_calls":null,
  "reasoning_content":"Hello"
},"finish_reason":null}]}
```

Both `reasoning` and `reasoning_content` appear in the same JSON object with identical values. I also confirmed the same duplicate serialization *without* the `--reasoning-parser` flag.

### Expected

Either (a) emit only `reasoning_content` (the OpenAI-contemporary name), (b) emit only `reasoning` (the legacy vllm-mlx name), or (c) provide a server flag to pick one. Don't emit both.

### Root cause

`vllm_mlx/api/models.py:436-450` declares the streaming delta schema as:

```python
class ChatCompletionChunkDelta(BaseModel):
    role: str | None = None
    content: str | None = None
    reasoning: str | None = None      # regular field
    tool_calls: list[dict] | None = None

    @computed_field                    # computed property ALSO serialized
    @property
    def reasoning_content(self) -> str | None:
        """Alias for reasoning field. Serialized for backwards compatibility with clients expecting reasoning_content."""
        return self.reasoning
```

Pydantic's `@computed_field` serializes the property *in addition to* the regular `reasoning` field, not in place of it. Same pattern exists at `models.py:191-205` (`AssistantMessage`) for non-streaming responses, so the bug should reproduce for `stream=false` calls too (not verified here).

### Impact

Goose (the block/goose orchestrator) uses `serde(alias = "reasoning")` to accept either legacy or new naming. When both names appear in the same JSON object, serde errors `duplicate field "reasoning_content"` and drops the stream. This forces users to either:

1. Fork Goose with lenient parsing (not practical for most users), or
2. Route Goose to a different backend (Ollama in our case), *bypassing vllm-mlx for orchestration entirely* — we lose vllm-mlx's continuous batching for every Goose-initiated call.

The Goose project itself isn't vllm-mlx-specific, but any Rust/serde-based OpenAI client that aliases these two names will hit the same error. Any new client following the OpenAI spec (which only uses `reasoning_content` currently, I believe) and *also* tolerating the legacy `reasoning` name will break here.

### Suggested fixes (in order of preference)

1. **Pydantic field alias** — replace the `@computed_field` with a standard `Field(alias="reasoning_content")` on the `reasoning` field, serialize by alias, and drop the computed property. Produces exactly one field in the JSON.

2. **Server flag** — `--legacy-reasoning-field` (or `--reasoning-field-name={reasoning,reasoning_content,both}`) that selects which name(s) to emit. `both` stays as today's behavior for current callers; default flips to `reasoning_content` only.

3. **Client negotiation** — inspect the request `User-Agent` or a new `X-Reasoning-Field` request header to pick per-request.

Option 1 is cleanest, but will break clients that *only* look for the legacy `reasoning` name. Option 2 is the safer migration.

### Environment

- `vllm-mlx` 0.2.8 (PyPI)
- `mlx` 0.31.1, `mlx-lm` 0.31.1
- Python 3.14 on macOS 15 (Darwin 25.4.0)
- Model: `mlx-community/Qwen3-Coder-30B-A3B-Instruct-8bit`
- Client breaking: [`goose`](https://github.com/block/goose) v1.30.0 (serde-based Rust deserializer)

### Context / why I'm reporting this

We run an 11-agent local SDLC orchestrator built on Goose + vllm-mlx. When we tried to route Goose's orchestrator provider to vllm-mlx (via the OpenAI-compatible API), every call failed at the first streaming chunk. We worked around it by splitting provider: Ollama serves Goose's orchestrator turns (with a smaller 4-bit model), while agents continue to call vllm-mlx directly for batched workloads. The split works but costs a model slot in RAM and adds orchestration complexity. We'd love to collapse back to a single `GOOSE_PROVIDER=openai` path once this is fixed.

Happy to test any patch / try branches.

Thump604:
## Summary
- stop serializing both `reasoning` and `reasoning_content` in the same response object
- make `reasoning_content` the wire/schema field while preserving `reasoning` as an internal/property alias
- add regression coverage for model serialization and streaming SSE payloads

Closes #310.

## Validation
- `PYTHONPATH=/tmp/vllm-mlx-issue310 /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_api_models.py tests/test_server.py -q`
- `/opt/ai-runtime/venv-live/bin/python -m black --check --fast vllm_mlx/api/models.py tests/test_api_models.py tests/test_server.py`

### DIFF
--- a/tests/test_api_models.py
+++ b/tests/test_api_models.py
@@ -631,7 +631,7 @@ class TestModelSerialization:
         msg = AssistantMessage(content="Answer", reasoning="Thought")
         data = msg.model_dump()
         assert data["reasoning_content"] == "Thought"
-        assert data["reasoning"] == "Thought"
+        assert "reasoning" not in data
 
     def test_chat_completion_response_json(self):
         resp = ChatCompletionResponse(
@@ -650,6 +650,35 @@ class TestModelSerialization:
         delta = ChatCompletionChunkDelta(reasoning="thinking")
         data = delta.model_dump()
         assert data["reasoning_content"] == "thinking"
+        assert "reasoning" not in data
+
+    def test_chat_completion_response_serializes_reasoning_content_only(self):
+        resp = ChatCompletionResponse(
+            model="test-model",
+            choices=[
+                ChatCompletionChoice(
+                    message=AssistantMessage(content="Answer", reasoning="Thought")
+                )
+            ],
+        )
+        data = resp.model_dump()
+        message = data["choices"][0]["message"]
+        assert message["reasoning_content"] == "Thought"
+        assert "reasoning" not in message
+
+    def test_chat_completion_chunk_serializes_reasoning_content_only(self):
+        chunk = ChatCompletionChunk(
+            model="test-model",
+            choices=[
+                ChatCompletionChunkChoice(
+                    delta=ChatCompletionChunkDelta(reasoning="thinking"),
+                )
+            ],
+        )
+        data = chunk.model_dump()
+        delta = data["choices"][0]["delta"]
+        assert delta["reasoning_content"] == "thinking"
+        assert "reasoning" not in delta
 
     def test_response_format_json_schema_alias(self):
         schema = ResponseFormatJsonSchema(
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -820,7 +820,8 @@ class TestStreamChatCompletion:
         ]
 
         assert payloads[0]["choices"][0]["delta"]["role"] == "assistant"
-        assert payloads[1]["choices"][0]["delta"]["reasoning"] == "reasoning"
+        assert payloads[1]["choices"][0]["delta"]["reasoning_content"] == "reasoning"
+        assert "reasoning" not in payloads[1]["choices"][0]["delta"]
         assert len(tool_payloads) == 1
         assert (
             tool_payloads[0]["choices"][0]["delta"]["tool_calls"][0]["function"]["name"]
@@ -923,7 +924,8 @@ class TestStreamChatCompletion:
         ]
 
         assert tool_parser.calls == []
-        assert payloads[1]["choices"][0]["delta"]["reasoning"] == "reasoning"
+        assert payloads[1]["choices"][0]["delta"]["reasoning_content"] == "reasoning"
+        assert "reasoning" not in payloads[1]["choices"][0]["delta"]
         assert payloads[2]["choices"][0]["delta"]["content"] == "final answer"
         assert payloads[2]["choices"][0]["finish_reason"] == "stop"
 
--- a/vllm_mlx/api/models.py
+++ b/vllm_mlx/api/models.py
@@ -12,7 +12,7 @@ These models define the request and response schemas for:
 import time
 import uuid
 
-from pydantic import BaseModel, Field, computed_field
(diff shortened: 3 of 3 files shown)


janhilgard (commented):
## Review: fix(api): emit a single reasoning_content field (#316)

**Verdict: Clean, well-scoped fix. LGTM with one minor question.**

### Strengths

- Correctly inverts the serialization direction: `reasoning_content` is now the Pydantic field (the wire name), and `reasoning` becomes a read-only property alias. This eliminates the duplicate-field problem from `computed_field` cleanly.
- `AliasChoices("reasoning_content", "reasoning")` with `validation_alias` means internal callers can still construct with `reasoning=...` keyword, preserving backward compatibility on the **input** side while the **output** JSON only contains `reasoning_content`.
- Both `AssistantMessage` and `ChatCompletionChunkDelta` are updated symmetrically.
- Good test additions: the new `test_chat_completion_response_serializes_reasoning_content_only` and `test_chat_completion_chunk_serializes_reasoning_content_only` verify the full nested serialization path (response -> choice -> message/delta), not just the model in isolation. The `assert "reasoning" not in data` assertions cover the exact regression from #310.

### Minor observation

The `reasoning` property (now a plain `@property`) is no longer serialized but is still accessible at runtime as an attribute. Internal code that reads `.reasoning` (e.g., in `server.py` streaming logic) will silently go through the property. This is fine, but worth confirming that no codepath writes `.reasoning = "..."` after construction — since `reasoning` is now a property with no setter, that would raise `AttributeError`. A quick grep confirms the field is typically set at construction time via kwargs, so this should be safe.

### Verdict

The approach is correct and the test coverage is solid. Good fix.

PR #325, opened 2026-04-14
fix(mllm): block SSRF in remote media fetches

### LINKED ISSUE #318: security: SSRF via unrestricted image/video URL fetching
## Parent issue

Tier 1 sub-issue of #68 (finding #3).

## Problem

`download_image()` and `download_video()` in `mllm.py` fetch user-supplied URLs with `requests.get()` and no host validation. An attacker can reach:

- Cloud metadata endpoints (`http://169.254.169.254/...`)
- Internal network services (`http://10.0.0.1/admin`)
- Localhost services (`http://127.0.0.1:6379/`)

## Reproduction

```bash
  -d '{
    "model": "vlm",
    "messages": [{"role": "user", "content": [
      {"type": "text", "text": "describe"},
      {"type": "image_url", "image_url": {"url": "http://169.254.169.254/latest/meta-data/iam/security-credentials/"}}
    ]}]
  }'
```

The server fetches the URL, potentially leaking cloud credentials or internal service responses.

## Fix

Add a `_validate_url_safety(url)` helper that:

1. Parses the URL and extracts the hostname
2. Resolves the hostname to IP addresses
3. Rejects any IP in: `127.0.0.0/8`, `10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`, `169.254.0.0/16`, `::1`, `fe80::/10`
4. Rejects `file://` scheme

Call it at the top of `download_image()` and `download_video()` before any `requests.get()`.

## Locations

- `mllm.py:185-270` (`download_image`)
- `mllm.py:273-367` (`download_video`)

## Scope

One new validation function + two call sites. No behavioral change for legitimate HTTPS image URLs.

Thump604:
## Summary
- validate remote image/video URLs before any network fetch
- reject non-http(s), localhost, and any host that resolves to non-global addresses
- validate every redirect hop so redirected private targets are blocked too
- add regressions proving blocked URLs never reach requests and redirects cannot bypass the guard

## Testing
- PYTHONPATH=/tmp/vllm-mlx-issue318 /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_mllm.py -q -k 'validate_url_safety or safe_redirects or blocks_unsafe_url_before_request'
- /opt/ai-runtime/venv-live/bin/python -m black --check --fast vllm_mlx/models/mllm.py tests/test_mllm.py
- /opt/ai-runtime/venv-live/bin/python -m compileall vllm_mlx/models/mllm.py tests/test_mllm.py

Closes #318.

### DIFF
--- a/tests/test_mllm.py
+++ b/tests/test_mllm.py
@@ -123,6 +123,67 @@ class TestMLLMHelperFunctions:
         assert not is_url("/path/to/file.jpg")
         assert not is_url("data:image/png;base64,AAAA")
 
+    @pytest.mark.parametrize(
+        "url",
+        [
+            "file:///etc/passwd",
+            "http://127.0.0.1/private.jpg",
+            "http://169.254.169.254/latest/meta-data/",
+            "http://localhost/private.jpg",
+            "http://[::1]/private.jpg",
+        ],
+    )
+    def test_validate_url_safety_rejects_unsafe_targets(self, url):
+        """Test that local and private targets are rejected."""
+        from vllm_mlx.models.mllm import UnsafeRemoteURLError, _validate_url_safety
+
+        with pytest.raises(UnsafeRemoteURLError):
+            _validate_url_safety(url)
+
+    def test_validate_url_safety_allows_public_ip(self):
+        """Test that public literal IPs remain allowed."""
+        from vllm_mlx.models.mllm import _validate_url_safety
+
+        _validate_url_safety("https://8.8.8.8/image.jpg")
+
+    def test_request_with_safe_redirects_blocks_unsafe_redirect(self, monkeypatch):
+        """Test that redirect hops are validated before a second request."""
+        from vllm_mlx.models import mllm
+
+        calls = []
+
+        class FakeResponse:
+            def __init__(self, url, status_code, headers):
+                self.url = url
+                self.status_code = status_code
+                self.headers = headers
+                self.is_redirect = status_code in {301, 302, 303, 307, 308}
+                self.is_permanent_redirect = status_code in {301, 308}
+
+            def close(self):
+                return None
+
+        def fake_request(method, url, **kwargs):
+            calls.append((method, url, kwargs["allow_redirects"]))
+            return FakeResponse(
+                url,
+                302,
+                {"location": "http://127.0.0.1/internal.jpg"},
+            )
+
+        monkeypatch.setattr(mllm.requests, "request", fake_request)
+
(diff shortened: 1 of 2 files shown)


janhilgard (commented):
## Review: fix(mllm): block SSRF in remote media fetches (#325)

**Overall: Well-designed SSRF mitigation. A few items worth discussing.**

### Strengths

- `_validate_url_safety()` covers the important vectors: `file://` scheme, `localhost`, link-local (`169.254.x.x`), loopback (`127.0.0.1`, `[::1]`), and private RFC1918 ranges — all checked via `ipaddress.ip_address.is_global`.
- `_request_with_safe_redirects()` is the right pattern: disable `allow_redirects`, manually follow each hop, and validate before every request. This closes the classic redirect-to-internal-IP bypass.
- DNS resolution happens at validation time so hostname-based attacks that resolve to private IPs are caught.
- Tests verify that `requests.request` is never called for blocked URLs (the `fail_request` monkeypatch pattern is clean).
- The `response.url` fix for file extension detection is a nice bonus — the original code used the initial URL, not the final URL after redirects.

### Issues to discuss

1. **TOCTOU between DNS resolution and actual connection**: `_validate_url_safety()` resolves the hostname via `socket.getaddrinfo()`, but `requests.request()` will do its own independent resolution. A DNS rebinding attack could return a safe IP to the validator and then a private IP to the actual connection. This is a known hard problem — mitigating it fully would require hooking into the socket layer (e.g., a custom `requests` adapter that pins the resolved IP). Worth documenting as a known limitation even if not fixed in this PR.

2. **`is_global` rejects some legitimate IPs**: `ipaddress.is_global` returns `False` for `100.64.0.0/10` (Carrier-Grade NAT / shared address space, RFC 6598) and `192.0.0.0/24` (IANA special purpose). These are edge cases but worth noting. The current behavior (reject by default) is the safe choice.

3. **Double HEAD+GET means double validation**: Both `download_image` and `download_video` make a HEAD request followed by a GET, each going through `_request_with_safe_redirects`. That means up to `2 * (max_redirects + 1)` DNS lookups per media fetch. Not a correctness issue, but something to be aware of for latency-sensitive paths.

4. **Error message leaks internal IPs**: `UnsafeRemoteURLError` includes the resolved blocked addresses in the message (`resolves to blocked address(es): 127.0.0.1`). This could leak internal DNS topology to an attacker probing the server. Consider a generic message in the HTTP response while logging the details server-side.

### Verdict

The core SSRF protection is solid and the redirect-hop validation is well implemented. Items 1 and 4 above are worth considering as follow-ups.

PR #327, opened 2026-04-14
fix(mllm): block local path traversal in multimodal input

### LINKED ISSUE #320: security: path traversal via local filesystem paths in multimodal input
## Parent issue

Tier 2 sub-issue of #68 (finding #4).

## Problem

`process_image_input()` and `process_video_input()` in `mllm.py` check if user-supplied input exists as a local filesystem path and return it directly:

```python
# mllm.py:529
if len(image) < 4096 and Path(image).exists():
    return image

# mllm.py:446
if Path(video).exists():
    return video
```

An attacker can read arbitrary files by sending `{"image_url": {"url": "/etc/passwd"}}`.

## Fix

Remove the `Path.exists()` local file checks. Only accept:
- `http://` and `https://` URLs
- `data:` base64-encoded content

## Locations

- `mllm.py:446` (`process_video_input`)
- `mllm.py:529` (`process_image_input`)

## Scope

Remove two filesystem-path code paths. Replace with scheme validation.

Thump604:
## Summary
- remove local filesystem path handling from multimodal image/video inputs
- accept only http(s) URLs and data: base64 payloads for user-provided media
- update multimodal docs/tests to reflect the tightened contract

## Testing
- `PYTHONPATH=/private/tmp/vllm-mlx-issue320 /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_mllm.py tests/test_video.py -q`
- `/opt/ai-runtime/venv-live/bin/python -m black --check --fast vllm_mlx/models/mllm.py vllm_mlx/multimodal_processor.py tests/test_mllm.py tests/test_video.py`
- `/opt/ai-runtime/venv-live/bin/python -m compileall vllm_mlx/models/mllm.py vllm_mlx/multimodal_processor.py tests/test_mllm.py tests/test_video.py`

Closes #320.

### DIFF
--- a/tests/test_mllm.py
+++ b/tests/test_mllm.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 """Tests for MLX Multimodal Language Model (MLLM) wrapper."""
 
+import base64
 import platform
 import sys
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


janhilgard (commented):
## Review: fix(mllm): block local path traversal in multimodal input (#327)

**Overall: Good hardening. One concern about breaking legitimate local-model workflows.**

### Strengths

- Removing `Path(video).exists()` and `Path(image).exists()` checks from `process_video_input()` and `process_image_input()` closes the local file read / path traversal vector via the API. An attacker could previously pass `/etc/passwd` as an "image URL" and the server would happily read it.
- The error messages are clear and list exactly what IS supported (http(s) URLs and data: base64).
- Tests are updated correctly: old tests that validated local path passthrough now assert `ValueError`, and new tests use proper `data:` base64 payloads.
- The video test updates in `test_video.py` are clean — replacing `tempfile` local paths with base64 data URIs throughout.

### Concern: internal usage and `_prepare_images`/`_prepare_video`

The `_prepare_images()` method calls `process_image_input()` for each image. If any internal code path or script passes local file paths to `generate()` directly (not through the HTTP API), those will now break with `ValueError`. This is the correct security posture for an API server, but worth noting in the PR body or changelog that direct SDK usage with local paths is no longer supported — users must use base64 data URIs or URLs instead.

Similarly, `describe_video()` documents only URL usage now (the docstring example for local files is removed), which is consistent.

### Nit

This PR has 2 commits — consider squashing before merge per the repo contribution guidelines.

Overall, this is a clean and well-scoped change.

PR #328, opened 2026-04-14
fix(server): require explicit trust_remote_code opt-in

### LINKED ISSUE #321: security: trust_remote_code=True by default enables arbitrary code execution
## Parent issue

Tier 2 sub-issue of #68 (finding #5).

## Problem

All engine constructors default to `trust_remote_code=True`:
- `simple.py:33`
- `batched.py:140`
- `mllm.py:689`

This allows HuggingFace model repositories to execute arbitrary Python during model loading via custom `configuration_*.py` or `modeling_*.py` files.

## Fix

1. Change default to `trust_remote_code=False` in all engine constructors
2. Add `--trust-remote-code` CLI flag to explicitly opt in
3. Pass the flag through the engine creation path

## Scope

Default change + one new CLI flag wired to three constructors.

Thump604:
## Summary
- default `trust_remote_code` to false across engine and MLLM constructors
- add explicit `--trust-remote-code` opt-in to both `vllm-mlx serve` and `python -m vllm_mlx.server`
- thread the flag through `load_model()` into both simple and batched engine creation
- add regressions covering parser defaults and engine wiring

## Testing
- `PYTHONPATH=/tmp/vllm-mlx-issue321 /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_server.py -q`
- `/opt/ai-runtime/venv-live/bin/python -m black --check --fast vllm_mlx/engine/simple.py vllm_mlx/engine/batched.py vllm_mlx/models/mllm.py vllm_mlx/server.py vllm_mlx/cli.py tests/test_server.py`
- `/opt/ai-runtime/venv-live/bin/python -m compileall vllm_mlx/engine/simple.py vllm_mlx/engine/batched.py vllm_mlx/models/mllm.py vllm_mlx/server.py vllm_mlx/cli.py tests/test_server.py`

Closes #321.

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -4,6 +4,7 @@
 import json
 import platform
 import sys
+from unittest.mock import MagicMock, patch
 
 import pytest
 
@@ -171,6 +172,23 @@ class TestCompletionRequest:
 class TestServeCli:
     """Test serve CLI argument parsing."""
 
+    def test_trust_remote_code_flag_defaults_false(self):
+        """Serve CLI should require explicit opt-in for remote code loading."""
+        from vllm_mlx.cli import create_parser
+
+        parser = create_parser()
+        args = parser.parse_args(["serve", "mlx-community/Llama-3.2-3B-Instruct-4bit"])
+        assert args.trust_remote_code is False
+
+        args = parser.parse_args(
+            [
+                "serve",
+                "mlx-community/Llama-3.2-3B-Instruct-4bit",
+                "--trust-remote-code",
+            ]
+        )
+        assert args.trust_remote_code is True
+
     def test_tool_call_parser_accepts_harmony_aliases(self):
         """GPT-OSS/Harmony parsers should be selectable from the serve CLI."""
         from vllm_mlx.cli import create_parser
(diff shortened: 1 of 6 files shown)


janhilgard (commented):
## Review: fix(server): require explicit trust_remote_code opt-in (#328)

**Overall: Important security default change with thorough wiring. A few issues to discuss.**

### Strengths

- Changing `trust_remote_code` default from `True` to `False` in `SimpleEngine`, `BatchedEngine`, and `MLXMultimodalLM` is a meaningful security improvement. HuggingFace remote code execution is a known attack surface (malicious model repos can execute arbitrary Python).
- The flag is properly threaded through the full call chain: CLI parser -> `serve_command()` -> `load_model()` -> engine constructors. Both `vllm-mlx serve` (cli.py) and `python -m vllm_mlx.server` (server.py) get the flag.
- Good startup logging: the security summary block now includes the remote code loading status.
- Test coverage is solid: parser defaults, explicit opt-in, and engine wiring are all verified.

### Issues

1. **`main()` refactoring scope creep**: The standalone server `main()` function in `server.py` was significantly refactored — the argument parser was extracted into `create_parser()` and the body was restructured. While the refactoring is reasonable, it goes beyond the PR scope (adding `--trust-remote-code`) and makes the diff harder to review. The extracted `create_parser()` looks correct, but the `main()` function now only forwards a subset of args to `load_model()` — notably missing: `stream_interval`, `gpu_memory_utilization`, `served_model_name`, `prefill_step_size`, `specprefill`, and `scheduler_config`. This appears to be a pre-existing issue (these were likely already missing in the standalone server path), but the refactoring makes it more visible. Worth verifying that the standalone `python -m vllm_mlx.server` path still passes all needed args.

2. **Models that REQUIRE `trust_remote_code=True`**: Many HuggingFace models (e.g., Qwen family, ChatGLM, DeepSeek, Phi) require `trust_remote_code=True` to load their custom modeling code. Changing the default to `False` will break these out of the box. The CLI flag makes it easy to opt in, but this is a potentially breaking change that should be called out prominently in release notes. Consider also logging a helpful error message when model loading fails due to `trust_remote_code=False` — something like "If this model requires custom code, re-run with --trust-remote-code".

3. **`MLXMultimodalLM.__init__` default changed but not wired**: The default on `MLXMultimodalLM.__init__` is changed from `True` to `False`, but I don't see `trust_remote_code` being passed through from the engine constructors to `MLXMultimodalLM`. If `SimpleEngine` or `BatchedEngine` construct `MLXMultimodalLM` internally, they need to forward their own `trust_remote_code` value. Otherwise the MLLM constructor will silently use `False` regardless of the CLI flag. This should be verified.

### Verdict

The security intent is correct and the CLI/test changes are well done. Issue #3 (MLLM wiring gap) should be verified before merge — if the engines do not forward `trust_remote_code` to `MLXMultimodalLM`, the flag will have no effect on multimodal model loading.

PR #329, opened 2026-04-14
fix(server): enforce MCP sandbox on execute endpoint

### LINKED ISSUE #322: security: /v1/mcp/execute bypasses tool sandbox validation
## Parent issue

Tier 2 sub-issue of #68 (finding #7).

## Problem

The `POST /v1/mcp/execute` REST endpoint calls `_mcp_manager.execute_tool()` directly without going through `ToolSandbox.validate_tool_execution()`. Rate limits, blocklists, and argument pattern checks defined in the sandbox are never applied to REST-initiated tool executions.

## Fix

Call `get_sandbox().validate_tool_execution()` before `_mcp_manager.execute_tool()`, or route through the `ToolExecutor` which already applies sandbox validation.

## Locations

- `server.py:816-834`

## Scope

Add one validation call before the existing execution call.

Thump604:
## Summary
- route `/v1/mcp/execute` through `ToolExecutor` instead of calling the manager directly
- enforce existing sandbox and argument-validation policy on REST-initiated MCP executions
- lazily initialize the executor if MCP is configured but no executor instance exists yet
- add endpoint regressions proving the REST path no longer bypasses sandbox handling

## Testing
- `PYTHONPATH=/private/tmp/vllm-mlx-issue322 /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_server.py -q`
- `/opt/ai-runtime/venv-live/bin/python -m black --check --fast vllm_mlx/server.py tests/test_server.py`
- `/opt/ai-runtime/venv-live/bin/python -m compileall vllm_mlx/server.py tests/test_server.py`

Closes #322.

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -4,7 +4,7 @@
 import json
 import platform
 import sys
-from unittest.mock import MagicMock, patch
+from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
 from fastapi.testclient import TestClient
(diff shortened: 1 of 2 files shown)


janhilgard (commented):
## Review: fix(server): enforce MCP sandbox on execute endpoint

**Overall:** Good catch — routing `/v1/mcp/execute` through `ToolExecutor` instead of calling `_mcp_manager.execute_tool` directly closes a real bypass where REST callers could skip sandbox/argument validation entirely. The fix is clean and well-tested.

### Strengths

- Correctly reuses the existing `ToolExecutor` path (same sandbox policy as the LLM-driven tool-call flow).
- Lazy initialization of `_mcp_executor` is safe — the `_mcp_manager` existence check runs first.
- Both test cases are solid: one proves the executor is called instead of the raw manager, the other proves sandbox blocks surface through the REST response.

### Issues / suggestions

1. **`_mcp_executor` is not declared at module level in this diff.** The `global _mcp_executor` statement inside `execute_mcp_tool` works only if `_mcp_executor` is already declared elsewhere in `server.py`. I see from the upstream main that line 152 already has `_mcp_executor = None` and the lifespan initializes it at line ~2727 — so this works in practice. But the diff by itself doesn't show this, which could confuse reviewers. A brief comment in the function noting "initialized in lifespan, lazily created here as fallback" would help.

2. **Race condition on lazy init.** If two concurrent requests hit the endpoint when `_mcp_executor is None`, both could create separate `ToolExecutor` instances. In practice this is harmless (ToolExecutor is stateless around the manager), but an `asyncio.Lock` or a simple `if` guard would be cleaner.

3. **The synthetic tool_call dict is hand-assembled.** If `ToolExecutor.execute_tool_calls` ever changes its expected shape, this will break silently. Consider adding a brief comment referencing the expected schema, or extracting the dict construction into a small helper.

4. **Minor: uuid import.** The `uuid` module is already imported in server.py, so `uuid.uuid4()` works — just confirming.

LGTM with the minor nits above. Solid security fix.

PR #333, opened 2026-04-14
fix(mcp): harden newline and traversal validation

Thump604:
## Summary
- harden MCP validator checks against newline-delimited injection and URL-encoded traversal forms
- normalize likely path-like values instead of relying only on a raw `../` regex match
- add focused regressions for newline separators, encoded traversal, and config-model rejection paths

## Test plan
- `PYTHONPATH=/private/tmp/vllm-mlx-issue68-f8 /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_mcp_security.py -q`
- `/opt/ai-runtime/venv-live/bin/python -m black --check --fast vllm_mlx/mcp/security.py tests/test_mcp_security.py`
- `/opt/ai-runtime/venv-live/bin/python -m compileall vllm_mlx/mcp/security.py tests/test_mcp_security.py`

Addresses finding #8 in #68.

### DIFF
--- a/tests/test_mcp_security.py
+++ b/tests/test_mcp_security.py
@@ -98,7 +98,16 @@ class TestMCPCommandValidator:
         with pytest.raises(MCPSecurityError) as exc_info:
             validator.validate_command("../../../bin/bash", "test-server")
 
-        assert "dangerous pattern" in str(exc_info.value)
+        assert "path traversal" in str(exc_info.value)
+
+    def test_command_newline_blocked(self):
+        """Test that newline separators in commands are rejected."""
+        validator = MCPCommandValidator(check_path_exists=False)
+
+        with pytest.raises(MCPSecurityError) as exc_info:
+            validator.validate_command("npx\ncat /etc/passwd", "test-server")
+
+        assert "newline characters" in str(exc_info.value)
 
 
 class TestArgumentValidation:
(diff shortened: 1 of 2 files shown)


janhilgard (commented):
## Review: fix(mcp): harden newline and traversal validation

**Overall:** Solid defense-in-depth. The two new checks (`_check_control_chars` and `_check_path_traversal`) close real bypass vectors: newline injection to smuggle commands past single-line regex checks, and percent-encoded `%2e%2e/` to bypass the literal `../` pattern. The implementation is thorough.

### Strengths

- `_check_control_chars` is applied consistently across all validation surfaces: command, args, env values, URLs, and URL query strings.
- `_check_path_traversal` uses `posixpath.normpath` for canonicalization AND explicit `..` segment detection — belt and suspenders approach that handles edge cases like `foo/../../etc/passwd`.
- URL validation decomposes the URL with `urlparse` before checking the path component, which is the correct approach (avoids false positives on query parameters containing `../`).
- Good test coverage: newline tests for commands, args, env, and URLs; encoded traversal tests for args, URLs, and config.
- The existing `test_path_traversal_blocked` assertion change from `"dangerous pattern"` to `"path traversal"` is correct — the new `_check_path_traversal` now fires before the generic `DANGEROUS_PATTERNS` loop.

### Issues

1. **The early-exit heuristic in `_check_path_traversal` may skip checks incorrectly.** The condition:
   ```python
   if "/" not in candidate and "\\" not in candidate and "%2e" not in value.lower():
       continue
   ```
   uses `value.lower()` (the original value) instead of `candidate.lower()` for the `%2e` check. When `candidate` is the decoded form, this is checking the wrong string. In practice this doesn't cause a bug because if `%2e` was in `value`, `decoded != value` would be true and both candidates would be checked. But the logic is confusing — consider using `candidate` consistently or just removing the early-exit (the normpath check is cheap).

2. **Double traversal on Windows-style paths.** The `candidate.replace("\\", "/")` handles backslash paths, but `posixpath.normpath` does not resolve `\` as a separator. Since `replace` runs first, this should be fine, but it's worth a comment explaining the conversion order.

3. **`_check_path_traversal` is not applied to env var names**, only values. An env var named `../../etc` would be unusual but technically possible. Low risk since env var names are checked against `dangerous_env_vars` separately.

4. **`\r` (carriage return) is included in `CONTROL_CHARS` but not `\0` (null byte).** Null bytes can cause truncation in some C-based tools. Consider adding `\0` to the list.

5. **Minor: the `CONTROL_CHARS` tuple could be a `frozenset` for O(1) lookup** instead of O(n) tuple iteration. Negligible for 2 elements but more idiomatic.

Good PR. The main suggestion is point 1 (consistent variable usage in the heuristic) and point 4 (null byte).

PR #337, opened 2026-04-14
fix(server): bind localhost by default

Thump604:
## Summary
- change both serve entrypoints to bind `127.0.0.1` by default instead of `0.0.0.0`
- keep external exposure explicit via `--host 0.0.0.0` and update the docs/help text accordingly
- add a regression test covering the CLI and standalone server parser defaults

## Test plan
- `PYTHONPATH=/private/tmp/vllm-mlx-issue68-bind-local /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_server.py -q`
- `/opt/ai-runtime/venv-live/bin/python -m black --check --fast vllm_mlx/server.py vllm_mlx/cli.py tests/test_server.py`
- `/opt/ai-runtime/venv-live/bin/python -m compileall vllm_mlx/server.py vllm_mlx/cli.py tests/test_server.py`

### DIFF
--- a/docs/guides/server.md
+++ b/docs/guides/server.md
@@ -2,6 +2,8 @@
 
 vllm-mlx provides a FastAPI server with full OpenAI API compatibility.
 
+By default the server binds only to `127.0.0.1`. Use `--host 0.0.0.0` only when you intentionally want to expose it beyond the local machine.
+
 ## Starting the Server
 
 ### Simple Mode (Default)
@@ -33,7 +35,7 @@ vllm-mlx serve mlx-community/Llama-3.2-3B-Instruct-4bit --port 8000 --continuous
 | Option | Description | Default |
 |--------|-------------|---------|
 | `--port` | Server port | 8000 |
-| `--host` | Server host | 0.0.0.0 |
+| `--host` | Server host | 127.0.0.1 |
 | `--api-key` | API key for authentication | None |
 | `--rate-limit` | Requests per minute per client (0 = disabled) | 0 |
 | `--timeout` | Request timeout in seconds | 300 |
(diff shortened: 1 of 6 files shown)


janhilgard (commented):
## Review: fix(server): bind localhost by default

**Addresses:** Issue #68, finding #13

**Summary:** Changes the default `--host` from `0.0.0.0` to `127.0.0.1` in both the `cli.py` serve parser and the standalone `server.py` parser. Docs and help text updated accordingly.

### Strengths
- Correct security default. Binding to all interfaces with no auth is an easy footgun, and localhost-by-default follows the principle of least privilege.
- Good refactor extracting `create_parser()` from `main()` in `server.py` so the test can inspect parser defaults directly.
- Test coverage is clean and covers both entrypoints.
- Docs consistently updated across all three reference pages.

### Issues / Suggestions

1. **Breaking change for existing users.** Anyone currently relying on the implicit `0.0.0.0` binding (e.g. Docker containers, LAN-accessible setups, systemd services) will silently stop accepting remote connections after upgrading. This deserves a mention in the PR description and ideally a note in the changelog or migration guide. Consider whether a deprecation warning (log a warning when `--host` is not explicitly set) would smooth the transition, though I acknowledge a clean default change is also defensible.

2. **Minor: `create_parser()` docstring.** The new function says `"Create the standalone server argument parser."` but it is also now importable and used from tests. A slightly broader docstring like `"Create the argument parser for the standalone server entrypoint."` would be more accurate since the word "standalone" might confuse readers into thinking it cannot be reused.

Overall this is a clean, well-scoped change. The only real concern is the silent behavioral change for existing deployments.

**Verdict:** LGTM with the caveat that the breaking change should be documented.

Issue #16, opened 2026-01-21
Issue: Simple mode issue when run with kilo-code


longcuxit:
I run "GLM-4.7-Flash" with simple cmd: `vllm-mlx serve "GLM-4.7-Flash"`
Sometime serve exit with error message: `-[_MTLCommandBuffer addCompletedHandler:]:1011: failed assertion `Completed handler provided after commit call'`

I think it throw error when kilo-code trigger double request (optimize promt, read file ...)
No errors were reported when running with the `--continuous-batching` option, but the program ran out of memory very quickly.

My device is M3 Max 128gb RAM

waybarrios:
Thanks for reporting this! I was able to reproduce and fix the Metal crash issue.

Tested with `mlx-community/GLM-4.7-Flash-4bit` - ran concurrent requests and the server now stays stable:

```
Testing 5 concurrent requests...
All requests completed!
Server is still running - NO CRASH!

Testing 10 concurrent requests...
Successful responses: 10/10
Server is still running!
```

The fix adds a lock to serialize MLX operations in SimpleEngine mode, which prevents the Metal command buffer conflicts. I'll merge this into main soon.

One note: GLM-4.7-Flash needs mlx-lm >= 0.30.2 since it uses the `glm4_moe_lite` architecture (older versions don't support it).

Regarding the memory growth with `--continuous-batching` - still looking into that one. Any details you can share about how fast memory grows would be helpful.

PR #339, opened 2026-04-14
fix(server): cap request max_tokens

Thump604:
## Summary
- add a separate `--max-request-tokens` ceiling for API requests instead of treating the default generation length as the only limit
- enforce positive `max_tokens` in the OpenAI and Anthropic request models and reject oversized requests before any engine work
- apply the same ceiling to completions, chat completions, and Anthropic messages, and document the new server option

## Test plan
- `PYTHONPATH=/private/tmp/vllm-mlx-issue68-max-tokens /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_server.py -q`
- `/opt/ai-runtime/venv-live/bin/python -m black --check --fast vllm_mlx/api/models.py vllm_mlx/api/anthropic_models.py vllm_mlx/cli.py vllm_mlx/server.py tests/test_server.py`
- `/opt/ai-runtime/venv-live/bin/python -m compileall vllm_mlx/api/models.py vllm_mlx/api/anthropic_models.py vllm_mlx/cli.py vllm_mlx/server.py tests/test_server.py`

Part of #68 (finding #16).

### DIFF
--- a/docs/guides/server.md
+++ b/docs/guides/server.md
@@ -45,6 +45,7 @@ vllm-mlx serve mlx-community/Llama-3.2-3B-Instruct-4bit --port 8000 --continuous
 | `--cache-memory-mb` | Cache memory limit in MB | Auto |
 | `--cache-memory-percent` | Fraction of RAM for cache | 0.20 |
 | `--max-tokens` | Default max tokens | 32768 |
+| `--max-request-tokens` | Maximum `max_tokens` accepted from API clients | 32768 |
 | `--default-temperature` | Default temperature when not specified | None |
 | `--default-top-p` | Default top_p when not specified | None |
 | `--stream-interval` | Tokens per stream chunk | 1 |
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -35,6 +35,7 @@ vllm-mlx serve <model> [options]
 | `--no-memory-aware-cache` | Use legacy entry-count cache | False |
 | `--use-paged-cache` | Enable paged KV cache | False |
 | `--max-tokens` | Default max tokens | 32768 |
+| `--max-request-tokens` | Maximum `max_tokens` accepted from API clients | 32768 |
 | `--stream-interval` | Tokens per stream chunk | 1 |
 | `--mcp-config` | Path to MCP config file | None |
 | `--paged-cache-block-size` | Tokens per cache block | 64 |
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -9,6 +9,7 @@
 | `--host` | Server host address | `127.0.0.1` |
 | `--port` | Server port | `8000` |
 | `--max-tokens` | Default max tokens | `32768` |
+| `--max-request-tokens` | Maximum `max_tokens` accepted from API clients | `32768` |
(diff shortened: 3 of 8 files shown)


janhilgard (commented):
## Review: fix(server): cap request max_tokens

**Addresses:** Issue #68, finding #16

**Summary:** Adds a `--max-request-tokens` server-side ceiling that rejects API requests whose `max_tokens` exceeds the configured limit. Also enforces `gt=0` on request model fields via Pydantic.

### Strengths
- Clear separation between `--max-tokens` (default for omitted values) and `--max-request-tokens` (hard ceiling). This is a useful operational knob.
- Validation happens early (`_resolve_request_max_tokens`) before engine allocation, which is the right place.
- Pydantic `Field(gt=0)` on all three request models (OpenAI chat, completions, Anthropic) is a good defensive layer.
- Thorough test coverage: CLI parsing, resolution logic, and per-endpoint rejection tests.
- CLI cross-validates that `--max-tokens <= --max-request-tokens`.
- Docs updated consistently across all three reference pages.

### Issues / Suggestions

1. **Default ceiling of 32768 may be too restrictive.** The current `--max-tokens` default is 32768, and now `--max-request-tokens` also defaults to 32768. Many users run models with 128K context (as this project does in production). Anyone sending `max_tokens=65536` will get a 400 error after upgrading unless they also pass `--max-request-tokens`. This is a silent breaking change. Consider either:
   - Defaulting `--max-request-tokens` to something higher (e.g. 131072 or even no limit by default), or
   - Making the feature opt-in (no ceiling unless explicitly configured)

2. **The constraint `--max-tokens <= --max-request-tokens` seems backwards.** The *default* generation length must be <= the *maximum allowed* request length. That check is correct. But it means you cannot set a low default (e.g. 2048) with a high ceiling (e.g. 131072) unless you explicitly pass both flags. This is fine, but the error message in `cli.py` (`"--max-tokens cannot exceed --max-request-tokens"`) could be clearer: `"--max-tokens (default generation length) cannot exceed --max-request-tokens (request ceiling)"`.

3. **Anthropic `max_tokens` is required (no default), but Pydantic `Field(gt=0)` removes the "required" signal.** Before this PR, `max_tokens: int` meant "required, no default". After: `max_tokens: int = Field(gt=0)` -- this still works because `Field(gt=0)` without a `default=` argument keeps the field required in Pydantic v2. But it is worth double-checking that Pydantic v1 (if still supported) handles this the same way.

4. **`_resolve_request_max_tokens` returns `_default_max_tokens` for `None`, but does not clamp it.** If someone configures `--max-tokens 65536 --max-request-tokens 65536`, this works. But if the globals get out of sync (e.g. programmatic `load_model()` call with mismatched values), `_default_max_tokens` could exceed `_max_request_tokens`. The function should arguably also validate `_default_max_tokens <= _max_request_tokens` or clamp to `min(_default_max_tokens, _max_request_tokens)`.

5. **Minor: duplicate validation.** Both `cli.py` `serve_command()` and `server.py` `load_model()` validate `max_tokens >= 1`, `max_request_tokens >= 1`, and `max_tokens <= max_request_tokens`. This is fine for defense-in-depth but could be extracted into a shared helper.

**Verdict:** Solid change overall. The main concern is the default ceiling being potentially too low for existing users with large-context models.

Issue #19, opened 2026-01-23
pip ERROR: ResolutionImpossible


bjer:
Issue installing latest tag and also from main branch

In a clean python 3.12.12 environment

```bash
git clone https://github.com/waybarrios/vllm-mlx.git
git checkout v0.2.3
pip install -e .
```
&rarr;

```
ERROR: Cannot install vllm-mlx and vllm-mlx==0.2.3 because these package versions have conflicting dependencies.

The conflict is caused by:
    vllm-mlx 0.2.3 depends on mlx-lm>=0.30.2
    mlx-audio 0.2.10 depends on mlx-lm<0.30.0 and >=0.18.0

Additionally, some packages in these conflicts have no matching distributions available for your environment:
    mlx-lm

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip to attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```

Same result for main branch

PR #343, opened 2026-04-14
fix(mcp): block high-risk tools by default

Thump604:
## Summary
- change MCP high-risk tool detection from warn-only to block-by-default
- add a root-level `allowed_high_risk_tools` MCP config allowlist for trusted exceptions
- thread the allowlist into the MCP executor sandbox and add regression coverage for default blocking and explicit allowlisting

## Test plan
- `PYTHONPATH=/private/tmp/vllm-mlx-issue68-high-risk-tools /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_mcp_security.py -q`
- `/opt/ai-runtime/venv-live/bin/python -m black --check --fast vllm_mlx/mcp/types.py vllm_mlx/mcp/config.py vllm_mlx/mcp/security.py vllm_mlx/server.py tests/test_mcp_security.py`
- `/opt/ai-runtime/venv-live/bin/python -m compileall vllm_mlx/mcp/types.py vllm_mlx/mcp/config.py vllm_mlx/mcp/security.py vllm_mlx/server.py tests/test_mcp_security.py`

Part of #68 (finding #19).

### DIFF
--- a/tests/test_mcp_security.py
+++ b/tests/test_mcp_security.py
@@ -342,6 +342,34 @@ class TestMCPServerConfigSecurity:
         )
         assert config.url == "https://api.example.com/mcp"
 
+    def test_config_parses_allowed_high_risk_tools(self):
+        """Root MCP config should accept explicit high-risk tool allowlists."""
+        from vllm_mlx.mcp.config import validate_config
+
+        config = validate_config(
+            {
+                "servers": {},
+                "allowed_high_risk_tools": [
+                    "trusted__execute_command",
+                    "run_shell",
+                ],
+            }
+        )
+
+        assert config.allowed_high_risk_tools == {
+            "trusted__execute_command",
+            "run_shell",
+        }
+
+    def test_config_rejects_invalid_allowed_high_risk_tools(self):
+        """High-risk allowlists must be a list of non-empty strings."""
+        from vllm_mlx.mcp.config import validate_config
+
+        with pytest.raises(ValueError) as exc_info:
+            validate_config({"servers": {}, "allowed_high_risk_tools": ["ok", ""]})
+
+        assert "allowed_high_risk_tools" in str(exc_info.value)
+
     def test_skip_security_validation_field_rejected(self):
         """Test that config-file security bypass is rejected explicitly."""
         with pytest.raises(ValueError) as exc_info:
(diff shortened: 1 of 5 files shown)


janhilgard (commented):
## Review: fix(mcp): block high-risk tools by default

**Addresses:** Issue #68, finding #19

**Summary:** Changes high-risk MCP tool detection from warn-and-continue to block-by-default. Adds an `allowed_high_risk_tools` config field so operators can explicitly allowlist trusted tools. The allowlist flows from config -> ToolSandbox -> runtime enforcement.

### Strengths
- **Correct security posture.** Warn-only for dangerous tools is security theater. Block-by-default with explicit opt-in is the right pattern.
- **Flexible allowlisting.** Supports both short names (`"execute_command"`) and fully-qualified names (`"server__execute_command"`), which covers both targeted and broad allowlisting use cases.
- **Case-insensitive matching** via `.lower()` normalization avoids trivial bypasses.
- **Good error message.** The `MCPSecurityError` message tells the operator exactly what to add to their config to unblock a tool.
- **Clean integration.** The allowlist is threaded from `MCPConfig` -> `load_mcp_config` -> `init_mcp` -> `ToolSandbox` -> `ToolExecutor`. No shortcuts or global state.
- **Config validation** rejects empty strings in the allowlist.
- **Tests cover** default blocking, short-name allowlisting, full-name allowlisting, and config parsing (valid + invalid).

### Issues / Suggestions

1. **`from_dict` on MCPConfig does not validate `allowed_high_risk_tools`.** The `MCPConfig.from_dict()` classmethod in `types.py` does `set(data.get("allowed_high_risk_tools", []))` without any validation. The validation only happens in `validate_config()` in `config.py`. If someone constructs an `MCPConfig` via `from_dict` directly (e.g. in tests or programmatic use), they bypass the empty-string check. Consider adding a `__post_init__` validation or documenting that `from_dict` is raw/unvalidated.

2. **The `_check_high_risk_tool` method now raises on the first matching pattern but the old code did `break`.** This means if a tool name matches multiple high-risk patterns (e.g. both `"exec"` and `"shell"`), it will raise on the first match. The old code would only warn once. This is actually fine (raise is more correct), just noting the behavior change.

3. **Pattern matching is substring-based.** The `HIGH_RISK_TOOL_PATTERNS` list uses `if pattern in tool_lower`. A tool named `executor_status` would be flagged because it contains `"exec"`. This is pre-existing behavior, not introduced by this PR, but worth noting that the allowlist mechanism is now the mitigation for false positives.

4. **`allowed_high_risk_tools` in the example config is an empty list `[]`.** This is correct -- it documents the field exists. But the example config also doesn't include any high-risk tools in its `servers`, so an operator might not understand the interaction. A comment in the example would help.

5. **Minor: `set()` vs `frozenset()`.** Since `allowed_high_risk_tools` is never mutated after initialization, `frozenset` would be a clearer signal of immutability in `MCPConfig`. Not a blocker.

**Verdict:** Well-designed change with good defense-in-depth. The only real gap is the unvalidated `from_dict` path.

PR #345, opened 2026-04-14
fix(mcp): stop loading config from cwd

Thump604:
## Summary
- remove implicit MCP config discovery from the current working directory
- keep explicit `--mcp-config` / explicit path loading working, along with the per-user `~/.config/vllm-mlx` fallback
- add regression coverage proving ambient `./mcp.json` files are ignored while explicit and user-config paths still load

## Test plan
- `PYTHONPATH=/private/tmp/vllm-mlx-issue68-mcp-config-cwd /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_mcp_security.py -q`
- `/opt/ai-runtime/venv-live/bin/python -m black --check --fast vllm_mlx/mcp/config.py tests/test_mcp_security.py`
- `/opt/ai-runtime/venv-live/bin/python -m compileall vllm_mlx/mcp/config.py tests/test_mcp_security.py`

Part of #68 (finding #18).

### DIFF
--- a/tests/test_mcp_security.py
+++ b/tests/test_mcp_security.py
@@ -6,7 +6,9 @@ These tests verify that the MCP command validation properly prevents
 command injection attacks and other security vulnerabilities.
 """
 
+import json
 import re
+
 import pytest
 from vllm_mlx.mcp.security import (
     MCPCommandValidator,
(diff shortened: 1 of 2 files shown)


janhilgard (commented):
## Review: fix(mcp): stop loading config from cwd

**Addresses:** Issue #68, finding #18

**Summary:** Removes `./mcp.json` and `./mcp.yaml` from the implicit config discovery search paths. Per-user config (`~/.config/vllm-mlx/`) and explicit paths (`--mcp-config` / env var) still work.

### Strengths
- **Correct fix for the threat.** Ambient CWD config loading is a well-known attack vector (cf. `.npmrc`, `.env`, `.git/config` poisoning). If a user runs `vllm-mlx serve` in a directory containing a malicious `mcp.json`, it would silently load attacker-controlled MCP server definitions. Removing CWD from the implicit search path eliminates this.
- **Minimal, surgical change.** Only 2 lines removed from `CONFIG_SEARCH_PATHS` and a docstring update. Very low risk of regression.
- **Explicit paths still work.** `load_mcp_config("./mcp.json")` with a relative path argument is still honored, which is the right behavior -- explicit is fine, implicit is dangerous.
- **Thorough test coverage.** Three new tests:
  - Verifies CWD `mcp.json` is ignored in auto-discovery
  - Verifies explicit `"./mcp.json"` argument still loads
  - Verifies `~/.config/vllm-mlx/mcp.json` discovery still works
- **Tests use `monkeypatch` properly** to isolate CWD, HOME, and env vars without side effects.

### Issues / Suggestions

1. **No issues found.** This is a clean, well-tested, correctly-scoped security fix.

2. **Optional improvement:** If you want belt-and-suspenders, `_find_config_file()` could log a debug message when it skips a CWD config, so operators who previously relied on CWD loading can diagnose why their config stopped being picked up. But this is minor -- the explicit `--mcp-config` path and the env var are well-documented alternatives.

**Verdict:** LGTM. Clean and correct.

PR #347, opened 2026-04-14
Fix MLLM tool call arguments not converted from JSON string to dict

janhilgard:
## Summary

- The MLLM code path in `server.py` bypasses `extract_multimodal_content()`, which is responsible for converting `tool_call.arguments` from JSON strings to dicts. Chat templates that iterate over arguments (e.g. GLM-4.6V: `{% for k, v in _args.items() %}`) crash with `jinja2.exceptions.UndefinedError: 'str object' has no attribute 'items'`.
- Add the same JSON string-to-dict conversion in the MLLM code path.
- Enable `SUPPORTS_NATIVE_TOOL_FORMAT` for the GLM-4.7 tool parser, since its chat template requires dict-form arguments.

## Test plan

- [x] Existing `test_native_tool_format.py` updated — GLM-4.7 parser added to native support assertions
- [x] All 100 tests pass (`pytest tests/test_native_tool_format.py tests/test_tool_parsers.py`)
- [x] Verified end-to-end on GLM-4.6V (107B MoE) with `--mllm` flag: tool call response cycle returns 200 instead of 500

### DIFF
--- a/tests/test_native_tool_format.py
+++ b/tests/test_native_tool_format.py
@@ -13,6 +13,7 @@ from vllm_mlx.tool_parsers import (
     DeepSeekToolParser,
     FunctionaryToolParser,
     Gemma4ToolParser,
+    Glm47ToolParser,
     GraniteToolParser,
     HermesToolParser,
     KimiToolParser,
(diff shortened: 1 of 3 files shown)


Thump604:
Good catch. The MLLM path bypasses extract_multimodal_content() which handles this conversion for the LLM path. The fix is clean and the test correctly exercises the GLM4 native tool format path. CI green.

PR #348, opened 2026-04-15
fix: stop stripping <tool_call> tags and add GLM-4 reasoning parser

janhilgard:
## Summary

- **Remove `<tool_call>` from `SPECIAL_TOKENS_PATTERN`** — `clean_output_text()` was stripping `<tool_call>` / `</tool_call>` tags from model output *before* the tool call parser had a chance to detect them. This broke tool call detection for all models using XML-style tool call markup (GLM-4.7, GLM-4.6V, etc.).
- **Add `Glm4ReasoningParser`** — GLM-4 models do *not* inject `<think>` into the prompt, so output without think tags is normal content (not truncated reasoning). The new parser inherits from `BaseThinkingReasoningParser` (which already handles this correctly) and adds GLM-4.6V `<|begin_of_box|>`/`<|end_of_box|>` token stripping plus a streaming override that emits content (not reasoning) in the pre-think phase.

## Root cause

`SPECIAL_TOKENS_PATTERN` included `</?tool_call>|</?tool_call_reasoning>`. The `clean_output_text()` function applied this regex to model output in `batched.py` before returning the text to the server, which then passed the *already-stripped* text to the tool call parser. Since the `<tool_call>` tags were gone, the parser could never match them → `tool_calls: null` on every request.

## Test plan

- [x] Tool call detection: `get_weather({"city": "Prague"})` correctly parsed
- [x] Text generation: `content` field populated (not null)
- [x] Reasoning extraction: `<think>` blocks correctly separated
- [x] Gemma 4 unaffected (different parser, no regressions)

### DIFF
--- a/tests/test_simple_engine.py
+++ b/tests/test_simple_engine.py
@@ -184,7 +184,10 @@ class TestSimpleEngineConcurrency:
                 ],
             )
 
-            assert output.text == '{"name":"bash","arguments":{"command":"pwd"}}'
+            assert (
+                output.text
+                == '<tool_call>{"name":"bash","arguments":{"command":"pwd"}}</tool_call>'
+            )
             assert output.tokens == [7, 8, 9]
             assert output.prompt_tokens == 11
             assert output.completion_tokens == 4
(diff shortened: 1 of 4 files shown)


janhilgard:
Superseded by #295 which includes both fixes (SPECIAL_TOKENS_PATTERN + Glm4ReasoningParser) along with additional prefix cache improvements.

PR #349, opened 2026-04-15
fix: patch GLM-4.6V (glm4v_moe) for BatchKVCache compatibility

janhilgard:
## Summary

- Add runtime patch for `GLM4VModel.__call__` to handle `BatchKVCache` offset correctly
- `GLM-4.6V` (`glm4v_moe`) computes `position_ids` from `cache[0].offset` at the model level using `mx.arange()`. With `BatchKVCache`, offset is an `mx.array` (per-batch offsets), but `mx.arange` does not support `mx.array` start/stop — producing wrong position_ids that corrupt RoPE for all decoder layers
- Symptoms: garbled output, `[gMASK]` token spam, or incoherent generation in continuous batching mode
- Fix converts `mx.array` offset to `int` before `mx.arange`, similar to the existing Qwen3.5 patch approach

## Test plan

- [x] Verify GLM-4.6V text generation produces coherent output with `--continuous-batching` (tested: "Paris", poems, math)
- [x] Verify tool calling works correctly (proper `<tool_call>` tags parsed: `get_weather({"location": "Berlin"})`)
- [x] Verify no `[gMASK]` token spam in output (confirmed across 15+ sequential requests)
- [x] Verify Gemma4 model still works (patches are independent: "Hello." with coherent reasoning)
- [x] CI tests pass (7/7: lint, type-check, test-matrix 3.10/3.11/3.12, test-apple-silicon, tests)

### DIFF
--- a/vllm_mlx/mllm_batch_generator.py
+++ b/vllm_mlx/mllm_batch_generator.py
@@ -366,9 +366,11 @@ class MLLMBatchGenerator:
         # Patch attention for BatchKVCache compatibility
         from .patches.qwen3_5_mllm import patch_qwen35_attention_for_batching
         from .patches.gemma4_mllm import patch_gemma4_attention_for_batching
+        from .patches.glm4v_moe_mllm import patch_glm4v_moe_for_batching
 
         patch_qwen35_attention_for_batching()
         patch_gemma4_attention_for_batching()
+        patch_glm4v_moe_for_batching()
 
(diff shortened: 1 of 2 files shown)


Thump604:
Clean patch, follows the same pattern as the Qwen 3.5 and Gemma 4 BatchKVCache patches. The offset int conversion is the right fix for mx.arange. CI green. Good to merge.

PR #350, opened 2026-04-15
fix(guided): Outlines tokenizer compat for structured output decoding

Thump604:
## Summary

Fixes Outlines integration issues for guided structured output decoding (#313) on Apple Silicon:

- **Tokenizer backend patching:** Outlines `TransformerTokenizer` reads `eos_token_id`, `eos_token`, `all_special_tokens` from the raw `tokenizers.Tokenizer` backend, but modern transformers only sets these on the `PreTrainedTokenizer` wrapper. `GuidedDecodingFactory` now patches them onto the raw backend.
- **Prompt token stripping:** `BatchGenerator` passes full token history to logits processors, but Outlines FSM expects only generated tokens. Wrapper tracks prompt length and slices.
- **Token type adaptation:** `BatchGenerator` passes `list[int]`, Outlines expects `mx.array`. Wrapper converts.
- **`build_raw_processors()`:** for serial `stream_generate` path which handles token tracking natively.

## Known limitation

The Outlines FSM compiles and runs, but hits state transition errors on the first generated token (`No next state found for state 256 with token ID 198`). This appears to be a tokenizer-to-vocabulary mapping issue between how Outlines builds the FSM and the model's actual token space. Investigation ongoing.

## Test plan

- [x] Tokenizer patching verified: `eos_token_id`, `eos_token`, `all_special_tokens` set on raw backend
- [x] Factory initialization succeeds on Qwen 3.5 122B
- [x] Processor builds and runs (no import/attribute errors)
- [ ] FSM state transitions need vocabulary alignment fix

### DIFF
--- a/docs/guides/server.md
+++ b/docs/guides/server.md
@@ -605,6 +605,11 @@ See [Reasoning Models Guide](reasoning.md) for full details.
 
 Force the model to return valid JSON using `response_format`:
 
+For `json_object` and `json_schema`, vLLM-MLX uses token-level guided
+decoding when tools are not active. Tool-calling requests keep the legacy
+prompt-instruction fallback so structured output and tool parsing do not
+compete for the same generation surface.
+
 ### JSON Object Mode
 
 Returns any valid JSON:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,6 +54,8 @@ dependencies = [
     "mcp>=1.0.0",
     # JSON Schema validation for structured output
     "jsonschema>=4.0.0",
+    # Token-level guided decoding for structured output
+    "outlines>=1.2.12",
     # pytz is required by gradio but not declared as a dependency
     # See: https://github.com/waybarrios/vllm-mlx/issues/23
     "pytz>=2024.1",
--- a/tests/test_batching.py
+++ b/tests/test_batching.py
@@ -145,6 +145,11 @@ class TestSamplingParams:
         assert params.stop == ["END"]
         assert params.stop_token_ids == [1, 2]
 
+    def test_response_format_field(self):
+        """Structured-output requests must carry response_format through batching."""
+        params = SamplingParams(response_format={"type": "json_object"})
+        assert params.response_format == {"type": "json_object"}
+
 
 class TestRequestOutput:
     """Tests for RequestOutput."""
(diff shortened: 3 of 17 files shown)


janhilgard (commented):
Good follow-up to #313 — the tokenizer backend patching and prompt-token stripping are the right fixes for Outlines integration on Apple Silicon. The architecture of `GuidedDecodingFactory` with lazy init and the wrapper pattern for adapting BatchGenerator's interface is clean.

**Blocker:**

The PR description acknowledges FSM state transition errors on the first generated token (`No next state found for state 256 with token ID 198`). This means guided decoding **doesn't actually work yet**. I'd suggest keeping this as a draft until the vocabulary alignment issue is resolved — merging broken guided decoding could confuse users who try `--enable-metrics` + `response_format`.

**Issues:**

1. **`outlines>=1.2.12` as hard dependency** — same concern as on #313. This pulls in a significant dependency tree (torch-adjacent) even when structured output is never used. Should be an optional extra (`pip install vllm-mlx[guided]`) with a clear import error at runtime.

2. **Per-token `_log.info()` in production wrapper** — `_wrap_outlines_processor` calls `_log.info()` on every single generated token, including printing token IDs:
   ```python
   _log.info("Guided wrapper: generated=%d tokens, first_few=%s", len(generated), generated[:5])
   ```
   This will flood logs in production. Should be `_log.debug()` at most, or removed entirely since the DEBUG block below already handles diagnostic output.

3. **`build_raw_processors()` appears unused** — it's defined in `GuidedDecodingFactory` but I don't see any callsite in the diff. Is this for a future PR? If so, consider deferring it.

4. **Triplicated merge pattern in `mllm.py`** — The `response_format` / `existing_logits_processors` extraction + `_merge_guided_logits_processors()` block is copy-pasted identically in `stream_generate`, `chat`, and `stream_chat`. Extract into a small private method to reduce drift risk.

5. **`numpy` import in debug path** — `import numpy as np` inside the wrapper's debug branch is heavy. Consider `result.tolist()` + pure Python check, or guard with `if _log.isEnabledFor(logging.DEBUG)` (which you already do — but the import still happens at function definition scope).

6. **No CI checks** — branch has no CI runs. Needs a push to trigger the test matrix.

**Good parts:**
- `normalize_response_format()` DRYs up the duplicated conversion in `tool_calling.py`
- `_should_use_guided_decoding()` correctly routes tool-calling requests to the legacy prompt-injection path
- `SimpleEngine.stream_chat` correctly skips the text-only MTP fast path when `response_format` is set
- Test coverage is solid — scheduler wiring, server integration, and structured output helpers all tested
- Tokenizer backend patching is a real fix for a real Outlines compatibility issue

Overall: good architecture, needs the FSM fix before merge.

Thump604:
Closing — the vllm-mlx-side fixes belong on #313 and the remaining FSM issue is an Outlines tokenizer compatibility bug, not a vllm-mlx issue.

PR #351, opened 2026-04-15
feat(mllm): add audio_url support in chat()

miguel-flowstate:
## Summary

- `MLLM.chat()` currently handles `image_url` and `video_url` content parts but silently drops `audio_url`, so audio data never reaches the model even though `mlx_vlm.generate()` fully supports it
- Extract `audio_url` content parts from OpenAI-compatible messages and pass them through to `generate(audio=...)`
- Insert `{"type": "audio"}` tokens into the chat template content list so the prompt is correctly formatted

## Changes

1. Collect `audio_url` paths alongside `image_url` during message content parsing
2. Include `{"type": "audio"}` entries in the structured content list (handled by `get_chat_template` via `_flatten_content`)
3. Pass `audio=all_audio_urls` to `mlx_vlm.generate()`
4. Updated log line to include audio count

## Test plan

- [x] Verified with Gemma 4 E4B + WAV audio via `mlx_vlm.generate()` directly — model produces correct transcription ("Ask not what your country can do for you...")
- [x] Confirmed processor returns `input_features` (387x128 mel spectrogram) and `input_features_mask` when audio is provided
- [x] Confirmed `audio_url` content parts are parsed correctly through the `/v1/chat/completions` endpoint
- [ ] Test with image + audio combined in a single request
- [ ] Test with image + audio + video combined

**Note:** Models need a `preprocessor_config.json` with the audio feature extractor definition for audio to work. Without it, the processor's `feature_extractor` is `None` and audio features are not produced.

### DIFF
--- a/vllm_mlx/models/mllm.py
+++ b/vllm_mlx/models/mllm.py
@@ -1319,9 +1319,10 @@ class MLXMultimodalLM:
         from mlx_vlm import generate
         from mlx_vlm.prompt_utils import get_chat_template
 
-        # Extract text and images from messages
-        # Build chat_messages for multi-turn support WITH proper image tokens per message
+        # Extract text, images and audio from messages
+        # Build chat_messages for multi-turn support WITH proper image/audio tokens per message
         all_image_urls = []  # Raw URLs/paths to process later
+        all_audio_urls = []  # Raw audio URLs/paths to process later
         chat_messages = []  # List of properly formatted messages for chat template
 
         logger.info(f"MLLM.chat() called with {len(messages)} messages")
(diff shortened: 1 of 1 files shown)


Thump604:
The audio_url extraction pattern matches the existing image_url/video_url handling. A few things to check before merge:

1. Does mlx-vlm's generate path actually accept audio inputs? The audio_urls are passed through but I don't see where they're processed into model-consumable format.
2. The audio URLs need the same SSRF protection as image/video URLs (#325). Any user-supplied URL should be validated before fetching.
3. CI needs to run — currently showing no checks.

miguel-flowstate:
Thanks for the review!

1. **mlx-vlm generate audio support** — Yes, `mlx_vlm.generate()` accepts an `audio` kwarg (see [generate.py L580](https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/generate.py#L580)), which flows through `prepare_inputs()` → `process_inputs()` into the processor. We verified end-to-end with Gemma 4 (both E2B and E4B): the `Gemma4Processor` produces `input_features` (mel spectrogram) and `input_features_mask` when audio is provided, and the model correctly transcribes speech. The key prerequisite is that the model repo includes a `preprocessor_config.json` with the `feature_extractor` section so the processor loads the `Gemma4AudioFeatureExtractor`.

2. **SSRF protection** — Good callout. Looking at the current codebase, the SSRF validation from #325 hasn't landed on main yet — image and video URLs go through the same unvalidated path today. Once #325 merges, audio URLs should be routed through the same `validate_url_safety` check. Happy to add that in this PR if #325 lands first, or we can address it as a follow-up that covers all three media types uniformly.

3. **CI** — Understood, will keep an eye on it.

janhilgard (commented):
## Review: feat(mllm): add audio_url support in chat()

**Overall: Good, minimal addition that follows existing patterns. One logic bug to fix.**

### Strengths

1. **Follows the established pattern** for `image_url` handling exactly -- clean, predictable code.
2. **Handles both string and object forms** of `audio_url` (matching OpenAI spec flexibility).
3. **Passes `audio=` kwarg to `mlx_vlm.generate()`** -- correctly enables end-to-end audio support for models like Gemma 4.

### Bug: `msg_audio_count` uses accumulated total, not per-message count

```python
msg_audio_count = len(all_audio_urls)  # ← BUG: this is the TOTAL across ALL messages
```

If message 1 has 1 audio and message 2 has 1 audio:
- Message 1: `msg_audio_count = 1` ✓
- Message 2: `msg_audio_count = 2` ✗ (should be 1, inserts 2 `{"type": "audio"}` tokens)

This will produce incorrect prompt formatting for multi-message audio conversations.

**Fix:** Track per-message audio count the same way `msg_image_count` is tracked:

```python
msg_audio_count = 0  # Initialize at start of message loop
# ... inside the content loop:
elif item_type == "audio_url":
    aud_url = item.get("audio_url", {})
    if isinstance(aud_url, str):
        all_audio_urls.append(aud_url)
    else:
        all_audio_urls.append(aud_url.get("url", ""))
    msg_audio_count += 1
```

### Minor

- No unit tests added. Given the small scope this is acceptable for now, but at least one test for multi-message audio parsing would catch the bug above.

PR #355, opened 2026-04-16
Fix Gemma 4 streaming reasoning parser: buffer partial markers

janhilgard:
## Summary

The streaming reasoning parser for Gemma 4 had two edge-case bugs that caused marker leakage when channel transition markers spanned token boundaries.

### Problems

1. **Partial `<|channel>` marker leak**: when a delta ended with a prefix of `<|channel>` (e.g. `<|` or `<|chan`), the prefix was emitted as reasoning text instead of being buffered for the next delta.

2. **Leading newline after transition**: the tokenizer often emits `\n` as its own token right after `<|channel>response`. This leading newline was being included in the first content delta.

### Fix

- **`_trailing_partial_marker_len()`** detects any trailing proper prefix of the three transition markers (`<|channel>response`, `<channel|>`, `<|channel>`) and buffers it internally in `self._pending`. On subsequent deltas, buffered text is re-combined with new text; only the "safe" prefix (without the new trailing partial) is parsed.
- **Leading newline stripping** for the first content delta after the `<|channel>response` marker (`self._content_seen` flag).
- **New `finalize_stream()` hook** on `ReasoningParser` base class lets the server flush any still-buffered text when the stream ends mid-marker (e.g. model hits `max_tokens` during `<|channel>` and the closing never arrives). Default implementation is a no-op. Gemma 4 override flushes pending text as reasoning (or content if already past the transition). Integrating this into `server.py` streaming endpoints is a small follow-up that can be done in a separate PR.

## Test plan

- [x] Added `tests/test_gemma4_streaming_edge.py` with 7 passing unit tests:
  - Marker split across multiple tokens
  - Leading newline stripping
  - Realistic token-by-token stream (matches production output)
  - Standard `<channel|>` format with marker splits
  - Absence of transition (all reasoning)
  - `finalize_stream()` hook behavior
  - Buffered partial marker flushed on stream end
- [x] Verified in production: identical reasoning/content split between streaming and non-streaming paths for deterministic prompts (`temperature=0, seed=N`) with Gemma 4 26B-A4B.

### DIFF
--- a/tests/test_gemma4_streaming_edge.py
+++ b/tests/test_gemma4_streaming_edge.py
@@ -1,0 +1,192 @@
+#!/usr/bin/env python3
+"""Unit tests for Gemma4 streaming parser edge cases."""
+
+import sys
+
+sys.path.insert(0, "/Users/janhilgard/vllm-mlx-upstream")
+
+from vllm_mlx.reasoning.gemma4_parser import Gemma4ReasoningParser
+
+
+def stream_parse(parser, text, token_list):
+    """Feed token_list to parser one-by-one, accumulating reasoning/content."""
+    parser.reset_state()
+    accumulated = ""
+    reasoning_parts = []
+    content_parts = []
+    for tok in token_list:
+        prev = accumulated
+        accumulated += tok
+        msg = parser.extract_reasoning_streaming(prev, accumulated, tok)
+        if msg is None:
+            continue
+        if msg.reasoning:
+            reasoning_parts.append(msg.reasoning)
+        if msg.content:
+            content_parts.append(msg.content)
+    return "".join(reasoning_parts), "".join(content_parts)
+
+
+def test_channel_marker_split_across_tokens():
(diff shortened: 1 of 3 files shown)


Thump604 (approved):
Code looks good — the partial-marker buffering logic in `_trailing_partial_marker_len` is correct and handles the real split-token issue with `<|channel>` markers that we've also observed.

Two items:

1. **Hardcoded test path:** `sys.path.insert(0, "/Users/janhilgard/vllm-mlx-upstream")` in the test file should be removed or made relative for CI.

2. **`finalize_stream()` is currently unwired** — as noted in the description, it needs a server.py follow-up to actually flush the buffer at stream end. Until then, if `max_tokens` truncation happens mid-marker, the buffered text is silently lost. Not a blocker for merging the parser fix, but worth tracking.

The `_content_seen` flag for leading newline stripping after transition is a nice touch — the tokenizer commonly emits `\n` as its own token right after `<|channel>response`.

PR #356, opened 2026-04-16
fix(api): make response_format actually work on chatty tool-calling models

janhilgard:
## Summary

Two related bugs made `response_format={"type": "json_object" | "json_schema"}` silently return empty content on chatty reasoning models like MiniMax-M2, GLM-4.6, and friends.

## Bug 1 — raw-JSON tool-call fallback hijacks user data

When no specific tool parser matched, `parse_tool_calls` fell through to a raw-JSON fallback that treated ANY object with a `"name"` field as a tool call. Schemas with a `"name"` data field (person name, user name, product name — one of the most common JSON Schema fields) were reclassified as tool calls: `{"name": "John", "age": 25}` became `function.name="John"` with empty arguments, `content=null`, and `finish_reason="tool_calls"`.

Two complementary guards:

1. New `_looks_like_tool_call(obj)` heuristic requires BOTH `"name"` AND `"arguments"` to match the OpenAI tool-call wire format.
2. `parse_tool_calls` only runs the raw-JSON fallback when the caller actually passed `tools`. Without `tools`, any JSON is user data.

Genuine tool calls keep working — XML wrappers (`<invoke>`, `<tool_call>`, `<function=>`) are unaffected, and raw-JSON tool calls still parse when the request declared `tools`.

## Bug 2 — extract_json_from_text loses chatty + truncated output

Even when the tool-call hijack was avoided, two failure modes silently returned `None` from `extract_json_from_text` and fell through to "Failed to extract valid JSON":

1. **Chatty preamble + truncation.** `` Let me format this as JSON:\n```json\n{... `` and hit `max_tokens` before the closing fence. The old matcher only recognised *complete* fences.
2. **Mid-object truncation.** `{"name": "John", "age": 25, "city": "Pra` — the greedy `\{[\s\S]*\}` regex required a closing `}`.

`extract_json_from_text` now tries five ordered strategies:

1. Parse entire text as JSON.
2. Extract from complete markdown code blocks.
3. *(new)* Extract from an *unterminated* markdown fence.
4. Balanced-brace scan respecting strings/escapes (replaces the greedy regex, which could match garbage between the first `{` and the last `}`).
5. *(new)* Repair a truncated prefix by closing unterminated strings and brackets, with fallbacks that strip dangling separators, keys, and partial scalars.

## Stricter system prompt

`build_json_system_prompt` now emits explicit STRICT output rules (first char `{`/`[`, no markdown fences, no preamble, no chain-of-thought) to reduce the incidence of the above failures in the first place.

## Verified against live MiniMax-M2.7

Before (same request, same seed):
```
content: null
tool_calls: [{"function": {"name": "John", "arguments": "{}"}}]
finish_reason: tool_calls
```

After:
```
content: '{"name": "John", "age": 25}'
tool_calls: null
finish_reason: stop
```

## Test plan

- [x] `pytest tests/test_structured_output.py tests/test_minimax_tool_calling.py` — 54 passed, 2 skipped
- [x] Live MiniMax-M2.7 — `response_format=json_schema` returns valid JSON in `content`, `tool_calls=null`, `finish_reason=stop`
- [x] Live MiniMax-M2.7 — genuine tool call (`get_weather`) still emitted with correct `name=get_weather`, `arguments={"city":"Prague"}`
- [x] Live MiniMax-M2.7 — plain chat (no tools, no response_format) unaffected
- [x] New unit tests: `TestRawJsonToolCallHijackPrevention` (6 tests), chatty/truncation coverage in `TestExtractJsonFromText` (9 tests), STRICT-rules coverage in `TestBuildJsonSystemPrompt` (2 tests)

### DIFF
--- a/tests/test_structured_output.py
+++ b/tests/test_structured_output.py
@@ -126,6 +126,172 @@ class TestExtractJsonFromText:
         result = extract_json_from_text(text)
         assert result == {"outer": {"inner": {"deep": "value"}}}
 
+    def test_chatty_preamble_with_json(self):
+        """
+        Test the classic Minimax / chain-of-thought failure mode:
+        the model explains itself before emitting JSON.
+        """
+        text = (
+            "Let me format this as JSON:\n" '```json\n{"name": "John", "age": 25}\n```'
+        )
+        result = extract_json_from_text(text)
+        assert result == {"name": "John", "age": 25}
+
+    def test_unterminated_markdown_fence(self):
+        """
+        Test truncation: model started ```json fence but ran out of
+        tokens before closing it. Previously returned None.
+        """
+        text = 'Let me format this as JSON:\n```json\n{"name": "John", "age": 25}'
+        result = extract_json_from_text(text)
+        assert result == {"name": "John", "age": 25}
+
+    def test_truncated_json_unclosed_object(self):
+        """Test repair of JSON truncated mid-object (max_tokens hit)."""
+        text = '{"name": "John", "age": 25, "city": "Prague"'
+        result = extract_json_from_text(text)
+        assert result == {"name": "John", "age": 25, "city": "Prague"}
+
+    def test_truncated_json_unclosed_string(self):
+        """Test repair of JSON truncated mid-string."""
+        text = '{"name": "John", "city": "Pra'
+        result = extract_json_from_text(text)
+        assert result == {"name": "John", "city": "Pra"}
+
+    def test_truncated_json_unclosed_nested(self):
+        """Test repair of JSON truncated deep inside nested structures."""
+        text = '{"user": {"name": "John", "items": [1, 2, 3'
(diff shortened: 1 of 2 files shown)


Thump604 (approved):
Both fixes are correct and high-value:

1. **`_looks_like_tool_call()` guard** — checking for both `name` (non-empty string) and `arguments` (dict or string) matches the OpenAI wire format precisely. The architectural fix of only running the raw-JSON fallback when `tools` were actually requested is the right approach.

2. **Balanced-brace JSON scanner** — `_scan_balanced_json` properly handles escaped quotes and braces inside strings. The old greedy `(\{[\s\S]*\})` regex was matching from first `{` to last `}` across garbage. The ordered repair strategies in `_repair_truncated_json` are pragmatic and safe (first `json.loads` wins).

The stricter system prompt with explicit STRICT rules is a good complementary change.

PR #357, opened 2026-04-16
Strip markdown code fences from streaming response_format content

janhilgard:
## Summary

When a client requests structured output via `response_format={"type": "json_object" | "json_schema", ...}` **and** uses streaming, models that don't have guided-decoding enforcement (e.g. Gemma 4) often wrap the JSON in a markdown fence, so the client sees:

```
```json
{"pain_points": [...]}
```
```

instead of `{"pain_points": [...]}`. The non-streaming path already strips these via `parse_json_output` / `extract_json_from_text`, but the streaming path emitted the raw deltas — breaking `json.loads` on the client.

## Repro

Before this PR, on a Gemma 4 server with `response_format={"type": "json_schema", ...}`:

| Mode          | Output                                                 |
|---------------|--------------------------------------------------------|
| Non-streaming | `{"pain_points": [...]}` ✓ valid JSON                 |
| Streaming     | ``` ```json\n{"pain_points":[...]}\n``` ``` ❌ fenced |

After this PR:

| Mode          | Output                                                 |
|---------------|--------------------------------------------------------|
| Non-streaming | `{"pain_points": [...]}` ✓ valid JSON                 |
| Streaming     | `{"pain_points":[...]}` ✓ valid JSON                  |

## Design — `StreamingJsonFenceStripper`

A small stateful filter (`vllm_mlx/api/tool_calling.py`) that:

- Strips a leading fence (`` ``` ``, `` ```json ``, `` ```\n ``, `` ```json\n ``, with optional leading whitespace), detecting and consuming fences **split across delta boundaries** (e.g. one delta sends `` ``` `` and the next sends `json\n{...}`).
- Holds back a small tail (~5 chars + any trailing whitespace/backticks) so a closing fence cannot straddle the emit boundary.
- On `finalize()`, drops any trailing `` ``` `` / `` \n``` ``, and also drops a strict-prefix fence that never completed (e.g. stream ended on `` ```js ``).

## Wiring

`stream_chat_completion` in `vllm_mlx/server.py`:

1. Instantiates `fence_stripper` when `request.response_format.type in {"json_object", "json_schema"}`, handling both the pydantic `ResponseFormat` model and a plain `dict`.
2. Routes `content` through `fence_stripper.feed(...)` in both content emit sites (reasoning-parser path and standard path) — but **only** when no tool calls have been detected (tool outputs are never wrapped in JSON fences).
3. On the final delta (`output.finished == True`), appends `fence_stripper.finalize()` to the content of that chunk so buffered tail text is flushed in-place.

## Tests

28 unit tests in `tests/test_streaming_fence_stripper.py` cover:

- No-fence passthrough, incremental emits, inner backticks preserved
- Leading fences in every form, including split across every single character
- Trailing fences (including split across deltas, extra whitespace)
- Full wraps and empty objects/arrays
- Edge cases: single trailing backtick in string, partial fence prefix at EOS, bare-backtick content

## Verified

Tested against a live Gemma 4 service (`mlx-community/gemma-4-26b-a4b-it` mixed 6/8-bit) with streaming + `json_schema`:

- Before: `'```json\n{"pain_points": [...]}\n```'` — not parseable
- After:  `'{"pain_points": [...]}'` — valid JSON ✓

Non-JSON streaming and Minimax streaming (which already worked via a stricter system prompt) both unchanged — regression-verified.

## Test plan

- [x] `pytest tests/test_streaming_fence_stripper.py -v` — 28 passed
- [x] `uvx black vllm_mlx/ tests/test_streaming_fence_stripper.py` — clean
- [x] Live test: Gemma 4 streaming + `json_schema` → valid JSON
- [x] Regression: plain streaming (no `response_format`) unaffected

### DIFF
--- a/tests/test_streaming_fence_stripper.py
+++ b/tests/test_streaming_fence_stripper.py
@@ -1,0 +1,136 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for ``StreamingJsonFenceStripper`` — strips markdown code fences
+from streamed ``response_format`` content.
+"""
+
+from vllm_mlx.api.tool_calling import StreamingJsonFenceStripper
+
+
+def _stream(chunks):
+    """Feed ``chunks`` through a fresh stripper and collect emitted text."""
+    s = StreamingJsonFenceStripper()
+    out = []
+    for c in chunks:
+        out.append(s.feed(c))
+    out.append(s.finalize())
+    return "".join(out)
+
+
+class TestStreamingFenceStripperNoFence:
+    def test_single_delta_no_fence(self):
+        assert _stream(['{"answer": true}']) == '{"answer": true}'
+
+    def test_multiple_deltas_no_fence(self):
+        assert _stream(["{", '"a"', ":", "1", "}"]) == '{"a":1}'
+
+    def test_empty_input(self):
+        assert _stream([]) == ""
+
+    def test_only_empty_deltas(self):
+        assert _stream(["", "", ""]) == ""
+
+    def test_content_with_inner_backticks_preserved(self):
+        # Backticks inside a string value must not be interpreted as a fence.
+        text = '{"code": "x `y` z"}'
+        assert _stream([text]) == text
+
+
+class TestStreamingFenceStripperLeadingFence:
+    def test_leading_backticks_json_newline(self):
+        assert _stream(['```json\n{"a": 1}']) == '{"a": 1}'
+
+    def test_leading_backticks_json(self):
+        assert _stream(['```json{"a": 1}']) == '{"a": 1}'
+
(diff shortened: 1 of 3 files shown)


Thump604 (approved):
Clean implementation. The `StreamingJsonFenceStripper` design with small holdback buffer (`_TAIL_HOLDBACK = 5` for `\n` + triple-backtick + `\n`) is correct and handles split fences across delta boundaries well.

Good that it only activates for `json_object` / `json_schema` response_format and skips when tool calls are detected. The `finalize()` method properly handles buffered tail text at stream end.

28 unit tests cover the matrix well. Composes cleanly with #355 (reasoning parser buffering runs first, fence stripping runs on clean content).

PR #362, opened 2026-04-16
Add JSON Schema constrained decoding via lm-format-enforcer

janhilgard:
## Summary

- Add **grammar-guided constrained decoding** for `response_format` (both `json_schema` and `json_object`) using [`lm-format-enforcer`](https://github.com/noamgat/lm-format-enforcer)
- Token-level logits masking ensures the model can **only emit tokens that form valid JSON** matching the requested schema
- Works with all engine paths: SimpleEngine, BatchedEngine (non-MLLM scheduler), and MLLM BatchGenerator (continuous batching)
- Falls back gracefully to prompt-only guidance when `lm-format-enforcer` is not installed

### New module: `vllm_mlx/constrained/`
- `json_schema_processor.py` — `JSONSchemaLogitsProcessor` that wraps `lm-format-enforcer`'s `TokenEnforcer`
  - **`_simplify_schema()`** — resolves `$ref` references (with cycle detection for recursive definitions), removes unsupported `not` keywords, and converts `type` arrays to `anyOf` — fixes crash on complex schemas like OpenAI's `OpenAiAnyType` pattern
  - Schema-aware key filtering to work around known enforcer bugs with multi-char boundary tokens
  - EOS guard prevents premature stop before JSON is complete
  - **Graceful fallback** — when the enforcer gets stuck on an unsupported schema (empty allowed-set), disables constrained decoding for that request instead of force-terminating with EOS after 3 tokens
  - Dynamic vocab sizing from logits tensor (handles models where embedding dim > tokenizer vocab, e.g. MiniMax-M2.7 200,064 vs 200,000)
- `cache.py` — LRU cache for `TokenEnforcerTokenizerData` (built once per model, reused across requests)

### Wiring
- `server.py` — builds processor when `response_format` is present, injects into all engine paths
- `mllm_batch_generator.py` — applies logits processors in all 3 prefill paths (cache miss, prefix hit, exact hit) and in `_step()` with correct token context
- `scheduler.py` — merges JSON processor with repetition penalty processors per request; **always passes `logits_processors` (empty list) to mlx_lm BatchGenerator** to prevent `TypeError: 'NoneType' object is not iterable` when a non-constrained request follows a constrained one
- Reasoning parsers are automatically suppressed when constrained decoding is active (incompatible — model cannot emit `<think>` tags when constrained to JSON)

### Anthropic endpoint
- `response_format` field added to `AnthropicRequest` (extension) and propagated through `anthropic_to_openai()` adapter

## Test plan

- [x] 16 unit tests pass (`tests/test_constrained_decoding.py`)
- [x] Gemma 4 26B-A4B: single request, 3 concurrent batch, streaming — all valid JSON
- [x] MiniMax-M2.7 (230B): single request, 3 concurrent batch, streaming — all valid JSON
- [x] Complex `$ref`/`not` schemas (e.g. `fact_batch` with recursive `OpenAiAnyType`) — valid JSON with constrained decoding
- [x] Non-constrained request after constrained request — no crash (logits_processors fix)
- [x] `json_object` mode (no schema) works on both models
- [x] Regression: normal requests without `response_format` unaffected
- [x] Black/ruff clean
- [x] All CI checks pass (9/9)

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,6 +55,8 @@ dependencies = [
     "mcp>=1.0.0",
     # JSON Schema validation for structured output
     "jsonschema>=4.0.0",
+    # Constrained decoding (grammar-guided generation for response_format)
+    "lm-format-enforcer>=0.10.9",
     # pytz is required by gradio but not declared as a dependency
     # See: https://github.com/waybarrios/vllm-mlx/issues/23
     "pytz>=2024.1",
--- a/tests/test_constrained_decoding.py
+++ b/tests/test_constrained_decoding.py
@@ -1,0 +1,305 @@
+# SPDX-License-Identifier: Apache-2.0
+"""
+Unit tests for JSON-schema constrained decoding.
+
+Covers:
+
+* ``vllm_mlx.constrained.cache`` — tokenizer-data construction and caching.
+* ``vllm_mlx.constrained.json_schema_processor.JSONSchemaLogitsProcessor``
+  — mask shape, token filtering, schema enforcement.
+* ``vllm_mlx.api.tool_calling.build_json_logits_processor`` — builder glue.
+* ``vllm_mlx.api.anthropic_adapter.anthropic_to_openai`` — propagation of
+  the OpenAI-compatible ``response_format`` extension field from
+  ``/v1/messages`` requests.
+
+These tests are pure-logic: they use a minimal fake tokenizer (ASCII
+vocabulary) to avoid pulling in a 200k-entry HF tokenizer at test time.
+They skip gracefully when ``lm-format-enforcer`` is not installed, so CI
+running with a minimal dependency set still passes.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from vllm_mlx.api.anthropic_adapter import anthropic_to_openai
+from vllm_mlx.api.anthropic_models import AnthropicMessage, AnthropicRequest
+from vllm_mlx.api.models import ResponseFormat, ResponseFormatJsonSchema
+from vllm_mlx.api.tool_calling import build_json_logits_processor
+from vllm_mlx.constrained import is_available
+from vllm_mlx.constrained.cache import clear_cache, get_tokenizer_data
+
+# Skip all processor-level tests if the optional dependency is missing.
+pytestmark_lmfe = pytest.mark.skipif(
+    not is_available(),
+    reason="lm-format-enforcer not installed",
+)
+
+
+# ---------------------------------------------------------------------------
+# Fake tokenizer — just enough API surface for the cache + processor.
+# ---------------------------------------------------------------------------
+
+
+class _FakeTokenizer:
+    """
+    Deterministic, tiny tokenizer for unit tests.
+
+    Vocabulary is a small ASCII set (digits, brackets, whitespace, quotes,
+    a handful of letters).  IDs start at 0 and are stable across instances.
+    Special tokens (EOS/BOS) occupy the end of the range.
+    """
+
+    def __init__(self) -> None:
+        chars = list('0123456789{}[]:," \n\ttrueflasnul')
+        # Deduplicate while preserving order.
+        seen: set[str] = set()
+        unique: list[str] = []
+        for c in chars:
+            if c not in seen:
+                seen.add(c)
+                unique.append(c)
+        self._id_to_tok = unique + ["<eos>", "<bos>"]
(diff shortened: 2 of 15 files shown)


Thump604 (commented):
This is excellent timing — we disabled guided decoding (#313, Outlines-based) because Outlines' FSM requires eager token materialization (`mx.eval`) inside the streaming generation loop, which is incompatible with MLX's lazy evaluation model (tracked in ml-explore/mlx-lm#1156, #1157, and dottxt-ai/outlines#1847).

`lm-format-enforcer`'s logits-masking approach avoids that entirely — the processor only needs the logits tensor and token IDs, not materialized decode state. This is a cleaner fit for MLX.

Key questions for review:

1. **Lazy eval compatibility:** Does `JSONSchemaLogitsProcessor.__call__` trigger any `mx.eval()` internally, or does it operate purely on the numpy/python side after the logits are materialized by the caller? If the caller has already called `mx.eval(logits)` before passing to the processor, this is fine. If the processor triggers eval internally, it could interfere with MLX's graph optimization.

2. **MTP interaction:** When MTP (multi-token prediction) is active, the backbone logits AND draft logits both pass through `logits_processors`. Does the `TokenEnforcer` state correctly handle the backbone→draft→verify sequence, or does the draft step confuse the FSM state? This is the exact issue we found in ml-explore/mlx-lm#1156 — stateful processors see stale tokens during MTP draft steps.

3. **Hard dependency vs optional:** The `try/except ImportError` fallback is good, but `pyproject.toml` lists it as a hard dep. Consider making it optional (extras_require) — not every user needs constrained decoding and the dependency tree adds weight.

The reasoning parser suppression when constrained decoding is active is the right call.

Once the merge conflicts are resolved, I'd like to test this on our Qwen 3.5 27B + Gemma 4 stack. If it passes, we can close #313 in favor of this approach.

janhilgard:
Thanks for the detailed review. Addressing each point:

**1. Lazy eval compatibility — no interference.**

`JSONSchemaLogitsProcessor.__call__` does `tokens.tolist()` which materializes the token array, but by that point the tokens are already evaluated (mlx-lm's sampler calls `.item()` on the previous step's token). The enforcer itself (`get_allowed_tokens()`) is pure Python — no MLX operations. The output `logits + mask` is lazy. So the processor slots into the mlx-lm pipeline without breaking graph optimization. This is the key difference from Outlines — Outlines' FSM needed `mx.eval(logits)` inside the decode loop to resolve the current token before advancing state, which stalled the pipeline.

**2. MTP interaction — safe by design.**

Logits processors are applied to **backbone logits only** (line 1887-1898 in `_mtp_step`). Draft tokens are sampled from MTP head logits **without** constraint enforcement. If a draft token violates the JSON schema, two things happen:
- The verify step may reject it (distribution mismatch → rollback to backbone token)
- Even if accepted, the **next** backbone step receives the full context including the draft token, and the enforcer's `get_allowed_tokens()` incrementally processes it — if it broke the schema, the allowed set will be restricted to corrective tokens

In practice, MTP only fires for single-user batches (`len(active_batch) > 1` skips MTP), so the impact window is narrow. I haven't tested constrained decoding + MTP together yet (Gemma 4 doesn't have MTP, MiniMax doesn't either), but the architecture is safe — worst case is a wasted draft token, not invalid output.

**3. Hard dependency — will move to optional.**

Good call. I'll move `lm-format-enforcer` to `[constrained]` extras. The `try/except ImportError` fallback already handles the missing case gracefully.

**Testing:** Already validated on our production stack:
- **Gemma 4 26B-A4B** (port 1236, mixed 6/8-bit): single request, 3-concurrent batch, streaming — all valid JSON, no regressions on unconstrained requests
- **MiniMax-M2.7 230B** (port 1242, mixed 6/8-bit): same test matrix, all passing
- Both `json_schema` (strict schema enforcement) and `json_object` (any valid JSON) modes tested

Rebased onto current main, merge conflicts resolved.

Thump604 (approved):
All three concerns addressed cleanly:

1. **Lazy eval** — enforcer is pure Python (`get_allowed_tokens()`), no `mx.eval` in the decode loop. This is the fundamental advantage over Outlines.
2. **MTP** — processors apply to backbone logits only, draft tokens are unconstrained. Worst case is a wasted draft, not invalid output.
3. **Hard dep** — moving to optional extras.

CI is 9/9 green. This unblocks structured output for us — we disabled Outlines-based guided decoding (#313) specifically because of the FSM + lazy eval incompatibility that this approach avoids entirely.

PR #365, opened 2026-04-17
Fix garbled output from stale disk-persisted prefix cache

janhilgard:
## Summary

Adds version validation and model fingerprint checking to disk-persisted prefix cache entries, preventing garbled output when loading stale or incompatible cache data.

### Problem

When a server restarts and loads prefix cache entries from disk that were saved by a previous session with different code, configuration, or model state, the restored KV cache tensors can be incompatible with the current model. This causes the model to generate garbled output (e.g., random code snippets, PHP tags, repeated tokens) instead of coherent text.

The issue is particularly insidious because:
- The cache reports a successful HIT, so it looks like everything is working
- Partial prefix matches from corrupted entries can cascade to poison other requests
- The corruption is silent — no errors, just wrong output

### Root Cause

`MemoryAwarePrefixCache.load_from_disk()` had no mechanism to detect whether saved entries are compatible with the current server state. Cache entries saved with one model configuration (quantization settings, code version, etc.) were blindly loaded and used with a different configuration.

### Fix

1. **Cache version constant** (`_CACHE_PERSIST_VERSION = 3`): Bumped from implicit v2. Any code change that affects cache format or semantics requires a version bump. Old caches are automatically rejected on load.

2. **Model fingerprint** (`_compute_model_fingerprint()`): Hashes key model architecture parameters (`num_hidden_layers`, `hidden_size`, `vocab_size`, `num_key_value_heads`, `head_dim`, `intermediate_size`, `model_type`) into a 16-char hex digest. Stored in the cache index at save time and validated at load time. Prevents loading cache entries from a different model.

3. **Validation on load**: `load_from_disk()` now checks both version (exact match) and fingerprint before loading any entries. On mismatch, logs a warning and returns 0 entries (the server continues normally, just without pre-warmed cache).

### Testing

- Verified on production MiniMax-M2.7 (230B MoE, port 1242):
  - Fresh server without disk cache → correct output ✓
  - In-session cache hit → correct output ✓  
  - Save to disk → restart → load from disk → correct output ✓
  - Tool calling with cache → correct function calls ✓
  - Structured output with cache → valid JSON ✓
- Verified Gemma 4 26B-A4B (port 1236) unaffected by editable install changes ✓
- Old stale cache (29 entries from previous session) correctly rejected after version bump ✓

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/vllm_mlx/memory_cache.py
+++ b/vllm_mlx/memory_cache.py
@@ -38,6 +38,9 @@ _BYTES_PER_MB = 1024 * 1024
 _DEFAULT_MEMORY_PERCENT = 0.20  # 20% of available RAM
 _MIN_MEMORY_BYTES = 100 * _BYTES_PER_MB  # Minimum 100MB
 _MAX_ENTRIES_FALLBACK = 50  # Fallback if memory detection fails
+# Bump this when the cache on-disk format or KV semantics change.
+# Loading a cache with a different version is rejected automatically.
+_CACHE_PERSIST_VERSION = 3
 
 
 def _get_available_memory() -> int:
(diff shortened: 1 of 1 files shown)


Thump604 (approved):
Clean fix for a real bug. The model fingerprint hash + version bump prevents stale disk-persisted prefix cache from poisoning a model swap or requant restart.

The fallback chain for config attribute access (model.config → model.args → model.model_config → model) covers all wrapper variants. Graceful degradation on mismatch (log + return 0) is the right behavior.

PR #367, opened 2026-04-17
Fix constrained decoding enforcer stuck + MLLM gpu-memory-utilization

janhilgard:
## Summary

- **`_simplify_schema()` now strips metadata keywords** (`default`, `examples`, `title`, `description`) that `lm-format-enforcer` doesn't understand, and **flattens nested `anyOf`/`oneOf`** wrappers (e.g. `anyOf: [{anyOf: [A, B]}, C]` → `anyOf: [A, B, C]`). This prevents the enforcer from getting stuck on schemas like the `fact_batch` pattern (nested anyOf with `$ref`, `not`, and `default` producing empty allowed-sets mid-generation).
- **`_start_mllm()` now sets `mx.set_memory_limit()`**, matching the existing LLM path. Previously `--gpu-memory-utilization` was silently ignored for MLLM models (Gemma 4, GLM-4.6V, Qwen3.5).

## Test plan

- [x] New unit tests: `test_strips_default_and_metadata`, `test_flattens_nested_anyof`, `test_fact_batch_schema_simplifies_cleanly`
- [x] All 19 existing + new tests pass (`pytest tests/test_constrained_decoding.py`)
- [x] `black` + `ruff` clean
- [ ] Production smoke test on MiniMax-M2.7 with `fact_batch` schema
- [ ] Verify no `enforcer stuck` warnings in logs after deploy
- [ ] After MLLM server restart, verify `Metal memory limits set` in logs

### DIFF
--- a/tests/test_constrained_decoding.py
+++ b/tests/test_constrained_decoding.py
@@ -301,5 +301,133 @@ class TestAnthropicAdapterResponseFormat:
         assert openai_req.response_format is None
 
 
+# ---------------------------------------------------------------------------
+# _simplify_schema — metadata stripping and anyOf flattening.
+# ---------------------------------------------------------------------------
+
+
+class TestSimplifySchema:
+    """Test ``_simplify_schema`` handles metadata and nested anyOf correctly."""
+
+    def test_strips_default_and_metadata(self):
+        from vllm_mlx.constrained.json_schema_processor import _simplify_schema
+
+        schema = {
+            "type": "object",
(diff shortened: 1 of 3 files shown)


Thump604 (approved):
Reviewed the constrained decoding and MLLM memory-limit changes. The schema simplification is scoped correctly, the new regression coverage matches the failure mode, and setting Metal memory limits on the MLLM batched path closes a real runtime gap. LGTM.

PR #368, opened 2026-04-17
Fix 3 async tests.

### LINKED ISSUE #364: Build failures on main
@perry2of5 I will take a look on this. Could you share more details like what is your, env python versions, etc? 

_Originally posted by @waybarrios in https://github.com/waybarrios/vllm-mlx/issues/363#issuecomment-4264794367_

perry2of5:
Switch from @pytest.mark.asyncio to @pytest.mark.anyio. The Apple Silicon CI runner only has anyio, not pytest-asyncio. Switching to @pytest.mark.anyio ensures tests can be run.

This partially fixes #364.

Thanks @Thump604 for the guidance.

### DIFF
--- a/tests/test_batched_engine.py
+++ b/tests/test_batched_engine.py
@@ -39,7 +39,7 @@ class TestBatchedEngineGenerate:
         mock.finish_reason = finish_reason
         return mock
 
-    @pytest.mark.asyncio
+    @pytest.mark.anyio
     async def test_tokens_field_is_populated(self):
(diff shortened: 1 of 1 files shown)


Thump604 (approved):
Correct fix — same pattern we applied in #288 and #302. The Apple Silicon CI runner has `anyio` but not `pytest-asyncio`, so `@pytest.mark.anyio` is the right marker. All 9 checks green.

PR #369, opened 2026-04-17
Change QuantizedKVCache to _QuantizedCacheWrapper to fix tests

perry2of5:
Fix the KV cache quantization isinstance failures (_QuantizedCacheWrapper vs QuantizedKVCache)

Partially fixes issue #364 (additional fixes in PR #368)

Thank you to @Thump604 for pointing me in the correct direction.

### DIFF
--- a/tests/test_kv_cache_quantization.py
+++ b/tests/test_kv_cache_quantization.py
@@ -2,11 +2,12 @@
 """Tests for KV cache quantization in prefix cache."""
 
 import mlx.core as mx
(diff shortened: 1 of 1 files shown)


Thump604 (approved):
Reviewed. This matches the current quantized cache implementation, where the quantized layers are wrapped in _QuantizedCacheWrapper rather than exposed as bare QuantizedKVCache instances. Narrow test-only fix, and it aligns the assertions with runtime reality. LGTM.

PR #371, opened 2026-04-18
Fix async audio tests @pytest.mark.asyncio -> @pytest.mark.anyio

perry2of5:


Thump604 (approved):
Same asyncio->anyio fix as #368. Apple Silicon CI runner has anyio, not pytest-asyncio. 9/9 green.

PR #372, opened 2026-04-18
Fix event loop errors in tests.

perry2of5:
Was receiving:

 RuntimeError: There is no current event loop in thread 'MainThread'.

Changed
    result = asyncio.get_event_loop().run_until_complete(
to
        result = asyncio.run(

and tests run locally in python 3.10, 3.11, 3.12, and 3.13

**what is wrong with CI that these are making it into main?**

### DIFF
--- a/tests/test_ssd_cache.py
+++ b/tests/test_ssd_cache.py
@@ -563,9 +563,7 @@ class TestAsyncFetchPath:
         def release_fn(nbytes):
             budget_released.append(nbytes)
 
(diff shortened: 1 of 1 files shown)


Thump604 (approved):
Clean fix. The deprecated get_event_loop().run_until_complete() pattern warns on Python 3.12+ and will break on 3.14 when get_event_loop() stops auto-creating loops. asyncio.run() is the right replacement. All 9 checks green.

PR #373, opened 2026-04-18
[NEW FEATURE] --warm-prompts for agent cold-start TTFT + bench-serve multi-msg

waybarrios:
## Motivation

Agent workloads (proxies to coding/reasoning assistants, MCP servers,
multi-agent orchestrators) all share the same pattern: every request
carries a multi-kilobyte system prompt that the operator knows in
advance. Today the first request pays the full prefill cost for that
system — even though the prefix cache could trivially reuse it if it
had been warmed.

First-hit TTFT grows with model size and concurrency, reaching several
seconds per request on multi-billion-parameter models and compounding
when multiple agents start at once. That latency lands exactly when a
user is waiting for their new agent to respond for the first time.

## Limitations in current main

1. Cold start always pays full prefill. The prefix cache is reactive:
   it only caches what it has already processed. The first time any
   client sends a large system prompt, the server prefills the whole
   thing.
2. bench-serve prompt files only support flat single-user lists. You
   cannot benchmark realistic agent traffic (system + user pairs).
3. The `count_prompt_tokens` preflight pollutes the cache. It sends a
   `max_tokens=1` request per prompt set before measurement,
   populating the prefix cache with the full prompt and defeating any
   cold-vs-warm comparison.
4. Hybrid SSM+attention models cannot benefit from partial prefix
   matches. Qwen3.5-MoE / Qwen3.6-MoE cache's SSM state cannot be
   trimmed, so `memory_cache.py` explicitly blocks LCP matching for
   them. A naive "warm the system prompt" strategy has no effect on
   those models.

### 1. `--warm-prompts <path>` on `vllm-mlx serve`

A JSON file of message arrays. Each entry is a full chat history, same
shape as `/v1/chat/completions` messages. At startup — after disk
cache load and before accepting traffic — the server runs a
`max_tokens=1` chat completion for each entry. The resulting KV state
lands in the prefix cache, so the first real request that shares a
prefix hits warm.

```bash
vllm-mlx serve mlx-community/Qwen3-4B-4bit \
  --continuous-batching \
  --warm-prompts ~/.config/vllm-mlx/agents.json
```

```json
[
  [{"role": "system", "content": "You are a code assistant..."}],
  [{"role": "system", "content": "You are a reviewer..."}]
]
```

Lifespan hook in `server.py`:

```python
# After engine start and disk cache load, before accepting traffic
if _warm_prompts_path and _engine is not None and hasattr(_engine, "stream_chat"):
    from vllm_mlx.prompt_warmup import load_warmup_file, warm_prefix_cache
    prompts = load_warmup_file(_warm_prompts_path)
    result = await warm_prefix_cache(_engine, prompts)
    logger.info(
        "[lifespan] Warm-up done (%s): %d completed, %d skipped, "
        "%d prompt tokens in %.1fs",
        result.get("mode", "?"), result["count"], result["skipped"],
        result["total_prompt_tokens"], result["elapsed_ms"] / 1000,
    )
```

### 2. Strict-prefix warmup for hybrid SSM+attention models

A naive warm-up (render the chat template with a placeholder user
message) stores an entry whose tokens do not form a strict prefix of a
real request: the placeholder user content diverges from real user
content, and everything after that point differs. Dense models can
recover via LCP matching, but hybrid models cannot — `memory_cache.py`
blocks LCP when any cache layer is non-trimmable.

The new warmer renders the template twice with different user
contents, finds the character position where the two strings diverge,
and stores a warm entry for the string up to that divergence point.
Real requests, which always use `add_generation_prompt=True`, begin
with exactly that prefix and match via strict prefix — no LCP needed.

```python
def _build_strict_prefix_string(tokenizer, messages, enable_thinking=True):
    """Render the chat template twice with two different user contents
    and truncate at the first divergent character — that position is
    exactly where user content gets inserted. Returns a string that is a
    strict prefix of any real request's rendered template for this system.
    """
    apply = getattr(tokenizer, "apply_chat_template", None)
    if apply is None:
        return None

    def _with_user(content):
        msgs = [dict(m) for m in messages]
        if msgs and msgs[-1].get("role") == "user":
            msgs[-1] = {**msgs[-1], "content": content}
        else:
            msgs = [*msgs, {"role": "user", "content": content}]
        return msgs

    kwargs = {"tokenize": False, "add_generation_prompt": True,
              "enable_thinking": enable_thinking}
    try:
        a = apply(_with_user("A"), **kwargs)
        b = apply(_with_user("B"), **kwargs)
    except Exception:
        # Template may reject enable_thinking on non-Qwen models — retry without
        kwargs.pop("enable_thinking", None)
        a = apply(_with_user("A"), **kwargs)
        b = apply(_with_user("B"), **kwargs)

    boundary = 0
    diverged = False
    for i in range(min(len(a), len(b))):
        if a[i] != b[i]:
            diverged = True
            break
        boundary = i + 1
    if not diverged or boundary < 16:
        return None
    return a[:boundary]
```

This is the difference between Qwen3.6-35B-A3B regressing 0.86–0.99x
and improving 1.44–1.56x with warm-prompts.

### 3. Three robustness optimizations

Parallel warm-up:

```python
results = await asyncio.gather(
    *(_one(i, msgs) for i, msgs in enumerate(prompts)),
    return_exceptions=False,
)
```

N prompts warm in roughly the time of one. Startup cost stays flat as
the warm-prompts file grows.

Auto user-terminator for templates that reject system-only messages
(Qwen3.6, DeepSeek-VL):

```python
def _ensure_user_terminator(messages):
    if messages and messages[-1].get("role") in ("user", "assistant"):
        return messages
    return [*messages, {"role": "user", "content": " "}]
```

Auto re-warm on `DELETE /v1/cache/prefix`:

```python
@app.delete("/v1/cache/prefix", dependencies=[Depends(verify_api_key)])
async def clear_prefix_cache():
    if _engine is None:
        return {"status": "no_engine"}
    cleared = False
    if hasattr(_engine, "clear_prefix_cache"):
        _engine.clear_prefix_cache()
        cleared = True

    rewarm_scheduled = False
    if cleared and _warm_prompts_path and hasattr(_engine, "stream_chat"):
        async def _rewarm():
            from vllm_mlx.prompt_warmup import load_warmup_file, warm_prefix_cache
            result = await warm_prefix_cache(_engine, load_warmup_file(_warm_prompts_path))
            logger.info("[clear_prefix_cache] re-warm done: %d completed", result["count"])
        asyncio.create_task(_rewarm())
        rewarm_scheduled = True
    return {"status": "cleared" if cleared else "not_supported",
            "rewarm_scheduled": rewarm_scheduled}
```

### 4. bench-serve upgrades

Prompt files now accept multi-message arrays — backwards compatible
with the flat single-user format, auto-detected on load:

```python
# load_prompt_set detects format and normalises to list[list[dict]]
first = raw[0]
if isinstance(first, dict):
    return [[msg] for msg in raw]   # flat → single-element lists
if isinstance(first, list):
    return raw                      # already list-of-lists
```

New flag `--system-prompt-file` prepends a system message to every
prompt:

```python
if system_prompt_file:
    sys_content = Path(system_prompt_file).expanduser().read_text()
    system_msg = {"role": "system", "content": sys_content}
    for ps, prompts in all_prompts.items():
        all_prompts[ps] = [
            msgs if msgs and msgs[0].get("role") == "system"
            else [system_msg] + msgs
            for msgs in prompts
        ]
```

New flag `--skip-preflight-token-count` skips the cache-polluting
`count_prompt_tokens` request. Auto-enabled when
`--system-prompt-file` is set so cold-vs-warm comparisons give honest
numbers (manual pass of the flag forces the skip regardless):

```python
skip_preflight_token_count=(
    args.skip_preflight_token_count or bool(args.system_prompt_file)
),
```

### 5. `DELETE /v1/cache/prefix` (admin)

New endpoint plus `clear_prefix_cache()` method on engine and
scheduler for operators who want to flush the in-memory cache without
bouncing the server. Disk cache is untouched.

## Benchmarks — main vs this PR

M4 Max / 128 GB, continuous batching, `long` prompt set (~2.5k user
tokens) + ~1.7k-token system prompt, `max_tokens=128`. Two separate
servers, one per condition, isolated cold start.

| Model | conc | cold TTFT | warm TTFT | Speedup |
|-------|-----:|----------:|----------:|--------:|
| Qwen3-0.6B-8bit | 1 | 563 ms | 419 ms | 1.34x |
| Qwen3-0.6B-8bit | 4 | 1 723 ms | 1 282 ms | 1.34x |
| Qwen3-0.6B-8bit | 8 | 3 708 ms | 2 661 ms | 1.39x |
| Llama-3.2-3B-4bit | 1 | 1 754 ms | 1 060 ms | 1.65x |
| Llama-3.2-3B-4bit | 4 | 5 926 ms | 3 945 ms | 1.50x |
| Llama-3.2-3B-4bit | 8 | 15 161 ms | 9 820 ms | 1.54x |
| Qwen3-4B-4bit | 1 | 4 937 ms | 2 191 ms | 2.25x |
| Qwen3-4B-4bit | 4 | 12 535 ms | 9 623 ms | 1.30x |
| Qwen3-4B-4bit | 8 | 38 148 ms | 23 878 ms | 1.60x |
| Qwen3.6-35B-A3B-4bit (MoE/hybrid) | 1 | 2 400 ms | 1 603 ms | 1.50x |
| Qwen3.6-35B-A3B-4bit | 4 | 8 735 ms | 6 054 ms | 1.44x |
| Qwen3.6-35B-A3B-4bit | 8 | 22 419 ms | 14 409 ms | 1.56x |

All 12 (model x concurrency) configurations improve. Generation tok/s
is neutral (+/-5%) for dense models; Qwen3.6-35B-A3B shows a 20-35%
decode drop at conc >= 4 that appears to be MoE routing interaction
with batched scheduling — investigating separately, not blocking this
PR since TTFT savings dominate end-to-end latency on agent workloads.

## Test plan

- `tests/test_prompt_warmup.py` — 40 tests covering file validation,
  multi-agent warm-up, strict-prefix string builder, user-terminator
  auto-append, parallel execution, error paths, unicode,
  BatchEncoding/Encoding normalisation.
- `tests/test_bench_serve.py` — updated for multi-message format
  (flat backwards compat, new list-of-lists, invalid shapes).
- `pytest tests/test_prompt_warmup.py tests/test_bench_serve.py` ->
  105 passed, 2 skipped.
- Manual on 4 models x 3 concurrencies x 2 conditions above.

## Usage

Put this in `~/.config/vllm-mlx/agents.json`:

```json
[
  [{"role": "system", "content": "You are a code assistant..."}]
]
```

Start the server:

```bash
vllm-mlx serve mlx-community/Qwen3-4B-4bit \
  --continuous-batching \
  --warm-prompts ~/.config/vllm-mlx/agents.json
```

Logs will show:

```
[lifespan] Warm-up done (strict-prefix): 1 completed, 0 skipped,
           1431 prompt tokens in 0.2s
```

The first real request that shares the warmed system prompt hits the
cache with `tokens_saved` around the warm-up prompt length and TTFT
roughly 1.3-2.3x lower than on a cold server.

## Breaking changes

None. All additions are opt-in flags and endpoints. Existing
bench-serve prompt files (flat `[{role, content}, ...]`) are
auto-detected and normalised internally.


janhilgard:
## Review

Overall: well-designed feature with solid test coverage. A few notes:

### Strengths

- **Strict-prefix warmup** is a clever solution for hybrid SSM+attention models where LCP matching is disabled. The two-probe divergence method to find the user-content boundary without parsing the template is elegant.
- **`DELETE /v1/cache/prefix` + auto re-warm** is a nice operator UX touch.
- **bench-serve multi-message format** with backwards-compatible auto-detection is clean.
- **40 tests** with good edge-case coverage (unicode, template failures, engine failures, concurrency timing).
- **`--skip-preflight-token-count`** fixes a real problem where the preflight polluted the cache and skewed warm-vs-cold benchmarks.

### Questions / minor issues

1. **`_normalise_token_ids`** (prompt_warmup.py L243-L274) — this function is defined but never called in this diff. Is it dead code left over from an earlier token-based warmup path that was replaced by the string-based approach? If unused, consider removing it to keep the module lean.

2. **Parallel warmup memory pressure** — `asyncio.gather` fires all N warm-up prompts concurrently. On a large model (e.g. 195 GB MoE) with 10+ warm prompts, that's N concurrent prefills at startup, each allocating KV cache. Might be worth a brief note in the docstring/help text that operators should keep the warm-prompts file small (1-3 entries covers most agent deployments), or consider an optional `--warm-prompts-concurrency` cap in the future.

3. **`warm_prompts_example.json`** — the 2700-char system prompt baked into the repo is a useful reference, but it's quite large for an example file. A shorter example with a comment pointing to the full version might be friendlier.

4. **`_build_strict_prefix_string` probe content** — using single-char `"A"` / `"B"` as probe content works for all known templates, but using slightly longer distinct strings (e.g. `"__PROBE_A__"` / `"__PROBE_B__"`) would add a tiny safety margin against hypothetical templates that treat single-char content specially. Very minor.

None of these are blockers. LGTM — nice work on the benchmarks across 4 models x 3 concurrencies.

Thump604 (approved):
Thorough and well-designed. The strict-prefix builder for hybrid SSM+attention models is the key insight — naive warm-up is useless without it on Qwen 3.5/3.6 MoE.

Reviewed:
- `prompt_warmup.py`: clean module, good validation, `_build_strict_prefix_string` handles the template-divergence probe correctly, parallel warm-up via `asyncio.gather` keeps startup cost flat
- `server.py`: warm-up runs after disk cache load (correct ordering), `DELETE /v1/cache/prefix` auto-re-warms in background
- `bench_serve.py`: backwards-compatible multi-message format, auto-detection of flat vs list-of-lists
- `cli.py`: `--warm-prompts` and `--skip-preflight-token-count` are clean opt-in
- 40 warmup tests + updated bench-serve tests, 105 passed

The bench-serve changes build cleanly on #366. `--system-prompt-file` + `--skip-preflight-token-count` give honest cold-vs-warm measurement — the original `count_prompt_tokens` preflight was indeed a cache-polluting issue.

One minor note: `_normalise_token_ids` handles 6 different tokenizer return shapes, which is defensive but warranted given the HF tokenizer zoo.

Benchmarks show real TTFT gains across all tested configs including the hybrid MoE path. Nice work.

waybarrios:
Thanks for the reviews. Follow-up commit will include these four changes:

1. **Remove `_normalise_token_ids`** — dead code from an abandoned
   token-ID-based warm path.
2. **Document warm-prompts sizing** — CLI help + module docstring note
   recommending 1–3 entries for typical agent deployments.
3. **Trim `warm_prompts_example.json`** to a concise reference system
   prompt; format spec stays in the module docstring.
4. **Rename probe strings** `"A"` / `"B"` → `"__PROBE_A__"` /
   `"__PROBE_B__"` for extra safety against hypothetical templates
   that treat single-char content specially.

**Context on (1).** It was the normaliser for the earlier token-ID
warm path, replaced once we discovered the server renders the chat
template with `tokenize=False` before handing a string to the engine.
Going through the same string→engine pipeline gives strict-prefix
tokens without any shape-normalisation, so the helper never got wired
up. @Thump604 — your read of it as defensive tokenizer-zoo handling
was the original intent, but since it's off the active path removing
it is cleaner. It can come back with tests if we ever accept
pre-tokenized input again.

**Context on (2).** `--warm-prompts-concurrency=N` is a trivial add
when a concrete user hits a memory limit. Deferring rather than
guessing a default now.

**Acknowledgment.** The bench-serve additions stack cleanly on
@Thump604's #366 because the prompt loading and request dispatch were
already well-isolated — the backwards-compat auto-detect for the flat
format was a 3-line addition at the top of `load_prompt_set`. Thanks
for shipping that with clean seams.

waybarrios:
@janhilgard I updated this pr based on you resquest.

PR #377, opened 2026-04-19
feat: server-side sampling defaults for top_k, min_p, presence/repetition_penalty

Thump604:
## Summary

This extends the server-side sampling defaults pattern from PR #41 (which added `--default-temperature` and `--default-top-p`) to cover the remaining sampling parameters:

- `--default-top-k` (int) -- server-wide top_k default
- `--default-min-p` (float) -- server-wide min_p default
- `--default-presence-penalty` (float) -- server-wide presence_penalty default
- `--default-repetition-penalty` (float) -- server-wide repetition_penalty default

### Why this matters

When deploying models that need specific sampling configurations (e.g., Unsloth-recommended presets for Qwen, Gemma, Nemotron families), operators currently have no way to set server-wide defaults for these parameters. Clients that omit them get hardcoded neutral fallbacks (`top_k=0`, `min_p=0.0`, `presence_penalty=0.0`, `repetition_penalty=1.0`), which may not be optimal for the loaded model.

### Design

Each new parameter follows the same three-tier resolution as temperature/top_p:

```
request value (if not None) > CLI default > fallback constant
```

The resolve functions use **`is not None` checks, not truthiness**. This is critical because `top_k=0` and `presence_penalty=0.0` are valid explicit values that a client may send to override a non-neutral server default. Using truthiness (`or`) would silently discard these.

### Changes

- **server.py**: 4 new globals, 4 new `_resolve_*` functions, 4 new CLI args, wired in `main()`, all `X or fallback` patterns replaced with resolve calls, sampling defaults logged at startup
- **cli.py**: 4 new `serve_parser` arguments, wired in `serve_command()`

### Cleanup

This also removes a small redundancy in the completions and chat handlers where `repetition_penalty` was set via `or 1.0` and then immediately overwritten by a conditional `if rep_penalty is not None` block. The resolve function handles both cases cleanly.

## Test plan

- [x] `pytest tests/test_server.py -q` -- 72 passed, 3 deselected
- [x] `black --check` -- no formatting changes needed
- [ ] Manual: start server with `--default-repetition-penalty 1.1 --default-min-p 0.1`, confirm startup log shows overrides, confirm requests without explicit params use server defaults, confirm requests with explicit `repetition_penalty=1.0` override the server default

### DIFF
--- a/vllm_mlx/cli.py
+++ b/vllm_mlx/cli.py
@@ -77,6 +77,14 @@ def serve_command(args):
         server._default_temperature = args.default_temperature
     if args.default_top_p is not None:
         server._default_top_p = args.default_top_p
+    if args.default_top_k is not None:
+        server._default_top_k = args.default_top_k
+    if args.default_min_p is not None:
+        server._default_min_p = args.default_min_p
+    if args.default_presence_penalty is not None:
+        server._default_presence_penalty = args.default_presence_penalty
+    if args.default_repetition_penalty is not None:
+        server._default_repetition_penalty = args.default_repetition_penalty
     server._max_audio_upload_bytes = args.max_audio_upload_mb * 1024 * 1024
     server._max_tts_input_chars = args.max_tts_input_chars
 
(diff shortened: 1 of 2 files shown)


Thump604:
I opened #418 as the current refreshed sampling-defaults and short-prefix-cache follow-up on top of current main. This older branch is dirty and superseded, so I am closing it to keep the review queue focused.

PR #382, opened 2026-04-20
feat(mllm): interleaved prefill/decode for continuous batching

janhilgard:
## Summary

- Add `install_chunked_prefill_mllm()` that replaces the MLLM batch generator's `_next()` with a state-machine interleaving one prefill chunk with one generation step, so active requests are no longer blocked during large prompt prefills
- Forward existing `--chunked-prefill-tokens` CLI argument to `MLLMSchedulerConfig` → batch generator (0 = disabled by default)
- Vision/multimodal requests bypass interleaving and use the original blocking prefill path

## Problem

When a new request with a large prompt (30-60K tokens) arrives while other requests are being served, the MLLM batch generator blocks ALL generation for 20-30 seconds during prefill. Active requests see throughput drop from ~50 tok/s to 1-5 tok/s.

## Solution

Process ONE chunk of the new prompt, then generate a token for the existing active batch, repeat. This keeps active request throughput at ~29 tok/s during a 30K-token prefill (vs 1-5 tok/s without interleaving).

### Design decisions

- **Synchronous `mx.eval()`** in generation step (not async) for safe context switching between prefill and batch caches
- **State machine** with `_partial` dict tracking in-progress partial prefill
- **Installation order**: chunked prefill FIRST, then MTP — MTP wraps `_next`, chunked replaces it, so MTP automatically wraps the chunked version
- **Abort-safe**: patched `remove()` clears partial state on disconnect
- **Prefix cache integration**: think-suffix stripping, empty RotatingKVCache detection, copy-on-fetch

### Files changed

| File | Change |
|------|--------|
| `mllm_batch_generator.py` | `install_chunked_prefill_mllm()` + idempotent `_preprocess_request` guard |
| `mllm_scheduler.py` | `chunked_prefill_tokens` config field + install before MTP |
| `engine/batched.py` | Forward `chunked_prefill_tokens` to MLLM config |

## Test plan

- [x] Concurrent load test: 2 short requests + 1 long (30K tokens) — short requests maintain ~29 tok/s
- [x] Abort test: disconnect client during long prefill, verify active batch continues
- [x] Vision request during interleaved prefill (should bypass chunking)
- [x] MTP + chunked prefill interaction

## Usage

```bash
vllm-mlx serve model --mllm --chunked-prefill-tokens 1024
```

### DIFF
--- a/vllm_mlx/engine/batched.py
+++ b/vllm_mlx/engine/batched.py
@@ -314,9 +314,12 @@ class BatchedEngine(BaseEngine):
             prefill_step_size = getattr(
                 self._scheduler_config, "prefill_step_size", None
             )
+        chunked_prefill = getattr(self._scheduler_config, "chunked_prefill_tokens", 0)
         mllm_extra = {}
         if prefill_step_size is not None:
             mllm_extra["prefill_step_size"] = prefill_step_size
+        if chunked_prefill:
+            mllm_extra["chunked_prefill_tokens"] = chunked_prefill
         mllm_config = MLLMSchedulerConfig(
             max_num_seqs=max_num_seqs,
             prefill_batch_size=prefill_batch_size,
--- a/vllm_mlx/mllm_batch_generator.py
+++ b/vllm_mlx/mllm_batch_generator.py
@@ -772,14 +772,17 @@ class MLLMBatchGenerator:
         3. Running vision encoder to get features
 
         Uses vision cache to skip processing for repeated images.
-        Idempotent: if input_ids is already set, returns immediately.
+        Idempotent: if input_ids is already set for text-only requests,
+        returns immediately (avoids redundant tokenization when called
+        from both chunked prefill interleaving and _process_prompts).
 
         Args:
             request: Request to preprocess
         """
         # Already preprocessed (e.g. by early executor offloading in
-        # _process_loop).  Only skip for text-only requests; vision
-        # requests need pixel cache lookup even if input_ids was set.
+        # _process_loop or chunked prefill interleaving).  Only skip for
+        # text-only requests; vision requests need pixel cache lookup even
+        # if input_ids was set.
         if request.input_ids is not None and not request.images and not request.videos:
             return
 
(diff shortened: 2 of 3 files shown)


waybarrios:
Found one bug that should block merge, plus two small notes.

**Bug: off-by-one in prefix cache store**

```python
# vllm_mlx/mllm_batch_generator.py
if batch_gen.prefix_cache is not None and req.input_ids is not None:
    try:
        input_ids_list = req.input_ids.reshape(-1).tolist()
        S = batch_gen._think_suffix_len
        cache_key = input_ids_list[:-S] if S > 0 else input_ids_list
        # Trim output: we stored 0 output tokens
        trim_amount = 1 + S  # last prompt token + think suffix
        store_cache = _trim_cache_offset(partial["cache"], trim_amount)
        batch_gen.prefix_cache.store(cache_key, store_cache)
```

Canonical `_maybe_store_prefix_cache` does:

```python
output_count = batch.num_tokens[i]
S = self._think_suffix_len
total_trim = output_count + S
```

Invariant is `output_count + S`. In chunked prefill `output_count = 0` at store time, so trim should be `S`, not `1 + S`. The exact-match fetch path already trims by 1, so this causes a double trim: stored cache is one position short, and any future request that exact-matches will produce corrupted output from KV/RoPE misalignment.

Fix: `trim_amount = S`.

- Chunked store: https://github.com/waybarrios/vllm-mlx/blob/3fbf27ae58339d2501867309799eaadc2f3741ae/vllm_mlx/mllm_batch_generator.py#L2418-L2428
- Canonical: https://github.com/waybarrios/vllm-mlx/blob/3fbf27ae58339d2501867309799eaadc2f3741ae/vllm_mlx/mllm_batch_generator.py#L1755-L1765

Also worth adding a round-trip test (chunked prefill, store, exact-match fetch, decode, compare vs non-chunked) so this doesn't regress.

**Minor**

1. Sync `mx.eval()` in `_generation_step()` vs `async_eval` elsewhere is fine per the PR body, but the "~29 tok/s" claim has no baseline. A small table (TTFT + decode tok/s, main vs PR, same hardware) would make it reviewable.

2. `merge([self])` works but is unusual:

```python
merged_cache = [
    request_cache[layer_idx].merge([request_cache[layer_idx]])
    for layer_idx in range(len(request_cache))
]
```

One-line comment saying "convert single-request cache to B=1 batch format" would help the next reader.

https://github.com/waybarrios/vllm-mlx/blob/3fbf27ae58339d2501867309799eaadc2f3741ae/vllm_mlx/mllm_batch_generator.py#L2395-L2404

waybarrios (changes requested):
Check my last comment.

janhilgard:
Thanks for the thorough review!

**Bug fix (off-by-one in prefix cache store)**: Fixed in fad71e9. Changed `trim_amount = 1 + S` to `trim_amount = S`, matching the canonical path's `output_count + S` invariant where `output_count = 0` at chunked store time. Good catch — the extra `+1` would have caused a one-position mismatch on exact-match fetch.

> Also worth adding a round-trip test

Agreed, will add a round-trip test once the fix lands.

**`merge([self])` comment**: Added clarifying comments on both sites: "Convert single-request cache to B=1 batch format so it can be extended into the active batch."

**Benchmark table**: I'll run a proper A/B comparison and post the numbers as a follow-up comment.

janhilgard:
## Benchmark: Interleaved Prefill/Decode

**Hardware:** Apple M3 Ultra, 256 GB unified memory  
**Model:** Qwen3.6-35B-A3B (mixed 6/8-bit, ~23 GB, MoE 3B active)  
**Budget:** `--chunked-prefill-tokens 1024`  
**Long prompt:** ~44K tokens (unique per test, no prefix cache hits)

### Latency — short request arriving during long prefill

Scenario: 1 background request generating → long prefill starts → short request arrives during prefill.

| Metric | main (blocking) | PR (interleaved) | Improvement |
|--------|-----------------|-------------------|-------------|
| Short request alone | 0.206s | 0.209s | — |
| **Short @ t=5 during 44K prefill** | **29.79s** | **7.27s** | **4.1× faster** |
| **Short @ t=15 during 44K prefill** | **19.73s** | **8.79s** | **2.2× faster** |
| Long prefill alone | 33.89s | 33.88s | — |
| Long prefill concurrent | 34.52s | 39.11s | +15% overhead |

### Decode throughput — existing requests during long prefill

| Metric | main (blocking) | PR (interleaved) |
|--------|-----------------|-------------------|
| Baseline single request | 81–89 tok/s | 80–85 tok/s |
| Short during 44K prefill (2 concurrent) | 5.3–6.0 tok/s¹ | 6.0–6.1 tok/s¹ |

¹ Average throughput is similar because most tokens are generated before/after the prefill; the key difference is latency for NEW requests arriving mid-prefill.

### Key finding

The main bottleneck in v1 was that new requests arriving during a partial prefill were stuck in `unprocessed_requests` until the prefill completed. Fixed in 9110d96 by processing short pending requests inline after each chunk, so they join the active batch immediately.

### Overhead

Long prefill takes ~15% longer when there's an active batch (39.1s vs 33.9s) due to interleaved generation steps. No overhead when no active batch.

sjswerdloff:
mx.eval([c.state for c in cache]) only works for ArraysCache and will OOM on hybrid models. Please see #413  as
  a possible fix pattern. You might need the same hasattr(c, "keys") check in this code.  Or I can re-do #413 once this merges.

janhilgard:
Fixed in ef3bffd — replaced `mx.eval([c.state for c in cache])` with cache-type-aware eval that handles both `ArraysCache` (`.state`) and `KVCache` (`.keys`/`.values`).

Matches the pattern from #413. Both `mx.eval` calls in `install_chunked_prefill_mllm()` (partial continuation and first-chunk processing) are updated.

Thanks @sjswerdloff for catching this!

janhilgard:
@waybarrios Both issues from your review and @sjswerdloff's comment are now fixed in the rebased branch:

1. **Off-by-one in prefix cache store**: `trim_amount = S` (was `1 + S`)
2. **`mx.eval` hybrid cache OOM**: replaced with `_eval_cache_tensors()` that handles both `KVCache` (`.keys/.values`) and `ArraysCache` (`.state`)
3. **`merge([self])` comments**: added on both call sites

Branch is cleanly rebased on current `main`, all 9 CI checks pass. Could you re-review when you get a chance?

janhilgard:
Superseded by #414, which was rebased on current main and merged. Closing.

Issue #380, opened 2026-04-20
bug: gemma 4 multiple parser and template bugs


0xAlcibiades:
`uv run vllm-mlx serve unsloth/gemma-4-31b-it-MLX-8bit --host 127.0.0.1 --port 1234 continuous-batching --embedding-model mlx-community/Qwen3-Embedding-4B-4bit-DWQ --enable-auto-tool-choice --tool-call-parser gemma4 --reasoning-parser gemma4 --max-tokens 262144 --max-request-tokens 262144`

Trying to run gemma 4 yield nonsense output. Same quantization works with lmstudio. Seems to be a parser/tokenizer issue.

If I drop continuous batching then i.e.

`uv run vllm-mlx serve unsloth/gemma-4-31b-it-MLX-8bit --host 127.0.0.1 --port 1234 --enable-auto-tool-choice --tool-call-parser gemma4 --reasoning-parser gemma4`

I get a better result but the reasoning parser is still broken:

```
Thinking: The user is greeting me ("Hello, how are you?").
According to the system instructions:
- I should be concise, direct, and to the point.
- I should minimize output tokens.
- I should avoid unnecessary preamble or postamble.
- I should avoid emojis unless requested.
- I should answer concisely with fewer than 4 lines.
A simple, polite, and concise response is appropriate.<channel|>I


am doing well, thank you. How can I help you today?
```

And have to wait for a full prefill of the entire context window for each turn.

There seem to be at least 2 issues:

1) Parser incorrect for multiple thinking/tool calls in a turn
2) Continuous batching broken with gemma 4 (or maybe more generally)

0xAlcibiades:
Some related issues and fixes:

- https://github.com/ggml-org/llama.cpp/pull/21418
- https://github.com/ggml-org/llama.cpp/pull/21390
- https://github.com/ggml-org/llama.cpp/pull/21406
- https://github.com/ggml-org/llama.cpp/pull/21327
- https://github.com/agentscope-ai/QwenPaw/issues/2947
- https://github.com/vllm-project/vllm/issues/40080

waybarrios:
Thanks for the detailed report and the upstream references. They pointed straight at the right fix classes.

PR #383 lands three of the issues reported here.

[...]

PR #383, opened 2026-04-21
[BUG] gemma4: tool-call stop token, nullable arg parsing, streaming reasoning leak (#380)

waybarrios:
## Purpose

Fixes three correctness bugs in the Gemma 4 stack reported in
[#380](https://github.com/waybarrios/vllm-mlx/issues/380):

1. The model keeps generating (or loops) after emitting a tool call —
   `<|tool_response>` is not treated as end-of-generation.
2. Tool calls with nullable (`"type": ["string", "null"]`) or
   enum-without-explicit-type schemas are dropped because the parser cannot
   understand bare, unwrapped values.
3. The streaming reasoning parser leaks the partial label `"th"` / `"tho"`
   into `reasoning_content` when the token `<|channel>thought` is split
   across delta boundaries.

None of these required touching the engine or KV-cache path. The PR is
strictly a parser/server correctness patch and has no impact on the
active v1 engine work.

---

### Symptom

When a Gemma 4 model issues a tool call, it often continues generating
after `<tool_call|>`, producing extra content or looping (confirmed by
downstream reports like QwenPaw #2947 and vLLM #40080).

### Root cause

Gemma 4's chat template renders `<|tool_response>` (token `50`, special)
right after the assistant emits a tool call — it is the signal that the
runtime should now execute the tool and return a response. In other
words, `<|tool_response>` is the turn boundary from the model's side and
must stop generation. Our tokenizer does not include it in the default
`eos_token_id` set, so the model never halts on it. This is the same fix
landed in
[ggml-org/llama.cpp#21418](https://github.com/ggml-org/llama.cpp/pull/21418).

### Fix

Add an `extra_stop_tokens` class attribute to the abstract `ToolParser`,
declare it on `Gemma4ToolParser`, and merge it into the chat request's
`stop` list at the server layer.

```python
# vllm_mlx/tool_parsers/abstract_tool_parser.py
class ToolParser(ABC):
    ...
    # Extra stop tokens specific to this parser's model format. The server
    # merges these into the request's `stop` list when the parser is active.
    extra_stop_tokens: list[str] = []
```

```python
# vllm_mlx/tool_parsers/gemma4_tool_parser.py
class Gemma4ToolParser(ToolParser):
    # <|tool_response> is the model-side turn boundary after a tool call.
    extra_stop_tokens = ["<|tool_response>"]
```

```python
# vllm_mlx/tool_parsers/__init__.py — new helper
def get_parser_stop_tokens(
    parser_name: str | None,
    user_stops: list[str] | None,
) -> list[str]:
    """Merge user-supplied stops with parser-declared extras (deduped)."""
    ...
```

```python
# vllm_mlx/server.py — inside create_chat_completion, just after tools are set
parser_name = _tool_call_parser if _enable_auto_tool_choice else None
merged_stop = get_parser_stop_tokens(parser_name, request.stop)
if merged_stop:
    chat_kwargs["stop"] = merged_stop
```

### Verification

End-to-end test hitting a live server with `--continuous-batching
--enable-auto-tool-choice --tool-call-parser gemma4` and the following
schema:

```python
{"type": "function", "function": {"name": "get_weather",
  "parameters": {"type": "object",
    "properties": {"city": {"type": "string"}},
    "required": ["city"]}}}
```

Request: `"What's the weather in Tokyo?"` — response:

```
finish_reason:     tool_calls
tool_calls count:  1
content:           None
completion tokens: 15
```

15 tokens, clean stop, no trailing rambling.

---

### Symptom

Tool calls using OpenAPI-style schemas with nullable types or enum fields
without an explicit `"type"` silently fail. The parser logs:

```
WARNING Gemma 4 tool parser: failed to parse args for call:set_state:
        Expecting value: line 1 column 11 (char 10)
```

### Root cause

The Jinja chat template decides whether to wrap a value in `<|"|> ... <|"|>`
by checking `value['type'] | upper == 'STRING'`. For a schema like
`{"type": ["string", "null"]}` or `{"enum": [...]}` with no `"type"`, that
branch is not taken, so the model's expected output is a **bare** value:

```
<|tool_call>call:set_state{domain:light}<tool_call|>
```

Our parser ran three passes — capture `<|"|>` strings, quote bare keys,
restore strings — but never quoted bare **values**. `json.loads` rejected
the result as invalid JSON, and the tool call was reported as not called.

This matches the failure class fixed in
[ggml-org/llama.cpp#21327](https://github.com/ggml-org/llama.cpp/pull/21327).

### Fix

Add a fourth pass that quotes identifiers appearing in value position
(after `:`, `[`, or `,`), while leaving JSON literals (`true`, `false`,
`null`) and numeric values untouched.

```python
# vllm_mlx/tool_parsers/gemma4_tool_parser.py
_BARE_VALUE = re.compile(r"(?<=[:\[,])(\s*)([A-Za-z_][\w\-]*)(?=\s*[,}\]])")
_JSON_LITERALS = frozenset({"true", "false", "null"})

def _quote_bare_value(m: re.Match) -> str:
    ws, word = m.group(1), m.group(2)
    if word in _JSON_LITERALS:
        return m.group(0)
    return f'{ws}"{word}"'

def _gemma4_args_to_json(text: str) -> str:
    text = _STRING_DELIM_RE.sub(_capture, text)        # 1. extract delimited strings
    text = _BARE_KEY.sub(r'"\1":', text)                # 2. quote bare keys
    text = _BARE_VALUE.sub(_quote_bare_value, text)     # 3. NEW: quote bare values
    text = _PLACEHOLDER_RE.sub(_restore, text)          # 4. restore strings
    return text
```

Placeholders (`\x00N\x00`) and negative numbers (starting with `-`) are
not captured because the pattern requires a leading letter/underscore,
so the existing paths are untouched.

### Verification

Live request with a realistic schema (nullable enum + enum-without-type):

```python
"parameters": {"type": "object", "properties": {
    "area":   {"type": "string"},
    "state":  {"type": ["string", "null"], "enum": ["on", "off"]},
    "brightness": {"enum": ["low", "medium", "high"]},
}}
```

Request: `"Set the kitchen light to on with medium brightness."` — response:

```
args_raw:  {"area":"kitchen","brightness":"medium","state":"on"}
args:      {'area': 'kitchen', 'brightness': 'medium', 'state': 'on'}
no <|"|> artifact in args_raw: True
```

All three fields are correctly typed strings and `json.loads` succeeds.

---

### Symptom

When the parser streams reasoning and the token sequence lands
`<|channel>` + `"tho"` + `"ught"` across delta boundaries, the first
delta emits `"tho"` into `reasoning_content` before the strip logic
recognises the `"thought"` label. Observed directly via a targeted
probe:

```
streamed reasoning = 'thoreal reasoning'
                     ^^^
```

### Root cause

The buffering function `_trailing_partial_marker_len` only considered the
bare markers `<|channel>response`, `<channel|>`, and `<|channel>`. Proper
prefixes of the longer label `<|channel>thought` were not buffered, so
partial characters of `"thought"` slipped through to the parent
streaming parser before the label-strip logic saw the full word.

### Fix

Register `<|channel>thought` as a buffering target too. The parser then
withholds any trailing proper prefix (e.g. `"<|channel>tho"`) until the
full label lands, at which point the existing strip path consumes it.

```python
# vllm_mlx/reasoning/gemma4_parser.py
_THOUGHT_MARKER = "<|channel>thought"
...
markers = (_RESPONSE_MARKER, _THOUGHT_MARKER, self.end_token, self.start_token)
```

### Verification

Streaming request with `enable_thinking=True` against the live server:

```
reasoning head:  'Thinking Process:\n1.  **Analyze the request:** ...'
content head:    '4'
no 'th'-ish leak at start:  True
no markers leaked:          True
```

---

### Unit tests added

- `tests/test_gemma4_tool_parser.py`:
  - `test_extra_stop_tokens_declares_tool_response`
  - `test_abstract_parser_default_empty_stop_tokens`
  - `test_merge_helper_adds_parser_extras`
  - `test_merge_helper_dedupes`
  - `test_merge_helper_unknown_parser_is_passthrough`
  - `test_merge_helper_none_parser_is_passthrough`
  - `test_bare_string_value_without_delimiters`
  - `test_bare_string_mixed_with_number_and_bool`
  - `test_bare_string_preserves_null_and_bool_literals`
  - `test_bare_string_in_array`
- `tests/test_gemma4_streaming_edge.py`:
  - `test_thought_label_split_across_deltas`

### Full test suite

```
$ pytest tests/ --ignore=tests/test_moe_top_k.py
========= 1561 passed, 11 skipped, 22 deselected in 25.35s =========
```

(One pre-existing flaky perf test, `test_batched_faster_than_sequential`,
fails under parallel load and passes in isolation — unrelated to this PR.)

### End-to-end

All three fixes validated against a running server with the flags from
the issue on M4 Max / 128 GB:

```
== Fix A — tool_response stop token ==                 PASS
== Fix B — nullable/enum bare string args ==           PASS
== Fix C — streaming reasoning, no label leak ==       PASS
```

---

## Out of scope — follow-up

The fourth symptom from #380 (**continuous-batching produces nonsense
output**) was **reproduced** while investigating this PR: under 12
concurrent requests with `enable_thinking=True` on
`mlx-community/gemma-4-e2b-it-mxfp4`, the content from one request
bleeds into the response of unrelated concurrent requests. Example:

```
[11] user: 'Briefly, what is dark matter?'
     content: ''   (response redirected)
[ 3] user: 'Explain what a hash table is in 30 words or less.'
     content: 'Dark matter is a hypothetical form of matter ...'
[ 4] user: 'Who painted the Mona Lisa?'
     content: 'Dark matter is a hypothetical form of matter ...'
[ 6] user: "Translate 'thank you' into German."
     content: 'Dunkle Materie.'   (literally "dark matter" in German)
```

This is a pre-existing engine-level bug (KV-cache / scheduler state
leak in the MLLM continuous-batching path). It is orthogonal to the
parser and server changes in this PR and will be filed separately with
the repro script.

---

Closes (partially) #380. Refs:
- [ggml-org/llama.cpp#21418](https://github.com/ggml-org/llama.cpp/pull/21418)
- [ggml-org/llama.cpp#21327](https://github.com/ggml-org/llama.cpp/pull/21327)

### DIFF
--- a/tests/test_gemma4_streaming_edge.py
+++ b/tests/test_gemma4_streaming_edge.py
@@ -27,6 +27,27 @@ def stream_parse(parser, text, token_list):
     return "".join(reasoning_parts), "".join(content_parts)
 
 
+def test_thought_label_split_across_deltas():
+    """The channel-name word 'thought' can arrive split across deltas in
+    production streams. The parser must buffer the partial prefix rather than
+    leak 'th' / 'tho' / 'thou' / 'thoug' into the reasoning output.
+    """
+    parser = Gemma4ReasoningParser()
+    tokens = [
+        "<|channel>",
+        "tho",        # partial "thought" label
+        "ught",       # completes the label
+        "\n",
+        "real reasoning",
+        "<channel|>",
+        "content",
+    ]
+    reasoning, content = stream_parse(parser, None, tokens)
+    assert reasoning == "real reasoning", f"Label leaked: {reasoning!r}"
+    assert content == "content", f"Got content: {content!r}"
+    print("[PASS] test_thought_label_split_across_deltas")
+
+
 def test_channel_marker_split_across_tokens():
     """<|channel> alone should not leak into reasoning if response follows."""
     parser = Gemma4ReasoningParser()
--- a/tests/test_gemma4_streaming_edge.py
+++ b/tests/test_gemma4_streaming_edge.py
@@ -35,8 +35,8 @@ def test_thought_label_split_across_deltas():
     parser = Gemma4ReasoningParser()
     tokens = [
         "<|channel>",
-        "tho",        # partial "thought" label
-        "ught",       # completes the label
+        "tho",  # partial "thought" label
+        "ught",  # completes the label
         "\n",
         "real reasoning",
         "<channel|>",
--- a/tests/test_gemma4_tool_parser.py
+++ b/tests/test_gemma4_tool_parser.py
@@ -167,7 +167,6 @@ class TestGemma4ToolParserExtract:
         args = json.loads(result.tool_calls[0]["arguments"])
         assert args == {"text": 'line1\nline2 said "hello"'}
 
-
 class TestGemma4ToolParserStreaming:
     """Test streaming tool call extraction."""
 
@@ -237,4 +236,47 @@ class TestGemma4Registration:
 
     def test_native_format_false(self):
         assert Gemma4ToolParser.SUPPORTS_NATIVE_TOOL_FORMAT is False
+
+    def test_extra_stop_tokens_declares_tool_response(self):
+        """Gemma 4 treats <|tool_response> (id 50) as end-of-generation
+        after a tool call. The parser exposes it so the server can merge it
+        into the request's stop sequences.
+        Reference: llama.cpp PR #21418.
+        """
+        parser = Gemma4ToolParser()
+        assert "<|tool_response>" in parser.extra_stop_tokens
+
+    def test_abstract_parser_default_empty_stop_tokens(self):
+        """Other parsers that don't override keep an empty default."""
+        from vllm_mlx.tool_parsers.abstract_tool_parser import ToolParser
+
+        assert ToolParser.extra_stop_tokens == []
+
+    def test_merge_helper_adds_parser_extras(self):
+        """get_parser_stop_tokens adds the parser's EOG tokens on top of user stops."""
+        from vllm_mlx.tool_parsers import get_parser_stop_tokens
+
+        merged = get_parser_stop_tokens("gemma4", ["END"])
+        assert "END" in merged
+        assert "<|tool_response>" in merged
+
+    def test_merge_helper_dedupes(self):
+        """Parser extra tokens already present in user stops aren't duplicated."""
+        from vllm_mlx.tool_parsers import get_parser_stop_tokens
+
+        merged = get_parser_stop_tokens("gemma4", ["<|tool_response>"])
+        assert merged.count("<|tool_response>") == 1
+
+    def test_merge_helper_unknown_parser_is_passthrough(self):
+        """Unknown parser name leaves user stops untouched."""
+        from vllm_mlx.tool_parsers import get_parser_stop_tokens
+
+        assert get_parser_stop_tokens("nonexistent_parser_xyz", ["A"]) == ["A"]
+
+    def test_merge_helper_none_parser_is_passthrough(self):
+        """No parser name returns user stops as-is."""
+        from vllm_mlx.tool_parsers import get_parser_stop_tokens
+
+        assert get_parser_stop_tokens(None, ["A"]) == ["A"]
+        assert get_parser_stop_tokens(None, None) == []
         assert Gemma4ToolParser.supports_native_format() is False
--- a/tests/test_gemma4_tool_parser.py
+++ b/tests/test_gemma4_tool_parser.py
@@ -189,7 +189,9 @@ class TestGemma4ToolParserExtract:
 
     def test_bare_string_preserves_null_and_bool_literals(self):
         """null/true/false must NOT be treated as bare strings."""
-        output = "<|tool_call>call:cfg{flag:null,ready:true,done:false,name:bob}<tool_call|>"
+        output = (
+            "<|tool_call>call:cfg{flag:null,ready:true,done:false,name:bob}<tool_call|>"
+        )
         result = self.parser.extract_tool_calls(output)
         args = json.loads(result.tool_calls[0]["arguments"])
         assert args == {"flag": None, "ready": True, "done": False, "name": "bob"}
--- a/tests/test_gemma4_tool_parser.py
+++ b/tests/test_gemma4_tool_parser.py
@@ -167,6 +167,42 @@ class TestGemma4ToolParserExtract:
         args = json.loads(result.tool_calls[0]["arguments"])
         assert args == {"text": 'line1\nline2 said "hello"'}
 
+    def test_bare_string_value_without_delimiters(self):
+        """Nullable type (e.g. ["string", "null"]) makes the template skip the
+        <|"|> wrap around string values. The parser must still produce valid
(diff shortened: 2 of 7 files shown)


Thump604 (approved):
Three clean fixes, well-scoped and well-documented.

**Fix A (stop token):** The `extra_stop_tokens` extension point on `ToolParser` is the right pattern -- it keeps the stop-token policy with the parser that knows the model format, and `get_parser_stop_tokens` gives the server a single merge point. The dedup in the helper is correct.

**Fix B (bare values):** The four-step ordering is correct -- bare-value quoting must happen after key quoting (step 2) to avoid double-processing, and before placeholder restoration (step 4) so placeholders are not accidentally matched. `_JSON_LITERALS` filter correctly preserves `true`/`false`/`null`. The `[A-Za-z_][\w\-]*` pattern is safe: digits-first values pass through, and hyphenated identifiers like `my-value` are handled.

**Fix C (streaming label):** One-line fix, directly addresses the buffering gap. Adding `_THOUGHT_MARKER` to the marker tuple is the minimal correct change.

Tests are thorough and cover the edge cases (mixed types, arrays, dedup, passthrough for unknown parsers).

The out-of-scope continuous-batching cross-request content leak is a real bug worth filing separately.

janhilgard (approved):
Clean, well-scoped PR. All three fixes are correct and well-tested.

**Fix A — stop token:** The `extra_stop_tokens` extension point on `ToolParser` is the right abstraction. `get_parser_stop_tokens` gracefully handles unknown parsers (try/except → passthrough), None parser names, and deduplicates. The server integration point in `create_chat_completion` is positioned correctly — after tools are wired but before constrained decoding.

**Fix B — bare values:** The four-step pipeline ordering is correct:

1. Extract `<|"|>`-delimited strings → placeholders protect contents from steps 2–3
2. Quote bare keys (won't touch values since lookbehind is `{` or `,`, not `:`)
3. Quote bare values (lookbehind on `:`, `[`, `,` ensures value position only)
4. Restore placeholders

`_BARE_VALUE` regex is safe against false positives:
- Placeholders (`\x00N\x00`) start with `\x00`, not `[A-Za-z_]` → no match
- Negative numbers start with `-` → no match
- Already-quoted keys were handled in step 2, their values follow `:` → correctly caught
- `_JSON_LITERALS` filter preserves `true`/`false`/`null`

Hyphenated identifiers like `my-value` are handled by `[\w\-]*` — good.

**Fix C — streaming label:** Minimal and correct. Adding `_THOUGHT_MARKER` to the marker tuple in `_trailing_partial_marker_len` makes the existing buffering machinery hold back partial prefixes of `<|channel>thought` until the full label lands. The `_strip_channel_name` logic then consumes it as before.

**Tests:** 11 new tests cover the key edge cases comprehensively — mixed types, arrays, dedup, passthrough for unknown/None parsers, and the split-across-deltas streaming scenario.

One observation for future reference: the out-of-scope cross-request content leak under concurrent `enable_thinking=True` requests (described at the bottom of the PR body) is a real issue I've also encountered — it's an engine-level bug in the MLLM batching path, not related to this PR.

LGTM.

PR #385, opened 2026-04-21
[BUG] gemma4 CB leaks content across requests via prefix cache LCP (#384)

### LINKED ISSUE #384: [BUG] Gemma 4 continuous batching leaks content across concurrent requests
## Origin

This bug was originally reported as one of the symptoms in
[#380](https://github.com/waybarrios/vllm-mlx/issues/380)
(`continuous-batching yields nonsense output`). While landing PR
[#383](https://github.com/waybarrios/vllm-mlx/pull/383) for the other
symptoms in that issue (parser and server fixes), I reproduced this
symptom cleanly and confirmed it is an engine-level bug that is
orthogonal to the parser work. Splitting it out into its own issue so it
can be tracked and fixed independently.

## Summary

When serving a Gemma 4 model with `--continuous-batching` and
`enable_thinking=True`, the content of one in-flight request bleeds into
the response of unrelated concurrent requests. The parser and server
layers are innocent. The corruption happens at the engine / KV-cache
level.

## Environment

- Hardware: Apple M4 Max, 128 GB unified memory
- OS: macOS 15.1
- Model tested: `mlx-community/gemma-4-e2b-it-mxfp4`
- Server flags:
  ```
  vllm-mlx serve mlx-community/gemma-4-e2b-it-mxfp4 \
    --host 127.0.0.1 --port 18082 \
    --continuous-batching \
    --enable-auto-tool-choice \
    --tool-call-parser gemma4 \
    --reasoning-parser gemma4 \
    --max-tokens 4096
  ```

## Reproduction

Fire 12 concurrent chat requests to the server. Each request has a
different question, `temperature=0.0`, `enable_thinking=true`, and no
tools. Repro script (Python, `aiohttp`):

```python
import asyncio, aiohttp, json

URL = "http://127.0.0.1:18082/v1/chat/completions"
MODEL = "mlx-community/gemma-4-e2b-it-mxfp4"

PROMPTS = [
    "What's 17 times 23?",
    "Name the tallest mountain in South America.",
    "Give me three programming languages that compile to native code.",
    "Explain what a hash table is in 30 words or less.",
    "Who painted the Mona Lisa?",
    "What gas do plants absorb during photosynthesis?",
    "Translate 'thank you' into German.",
    "What is the boiling point of water at sea level in Celsius?",
    "List two benefits of exercise.",
    "Who wrote Hamlet?",
    "What's the capital of Australia?",
    "Briefly, what is dark matter?",
]

async def fire(session, idx, prompt):
    body = {
        "model": MODEL,
        "messages": [
            {"role": "system", "content": "You are a concise assistant."},
            {"role": "user", "content": prompt},
        ],
        "max_tokens": 200,
        "temperature": 0.0,
        "stream": False,
        "enable_thinking": True,
    }
    async with session.post(URL, json=body) as resp:
        d = await resp.json()
    m = d["choices"][0]["message"]
    print(f"[{idx:2}] {prompt}")
    print(f"     content: {(m.get('content') or '')[:120]!r}")

async def main():
    async with aiohttp.ClientSession() as s:
        await asyncio.gather(*(fire(s, i, p) for i, p in enumerate(PROMPTS)))

asyncio.run(main())
```

## Observed behavior

Request 11 ("what is dark matter?") produces an empty `content` in its
own response, and its actual answer appears inside five unrelated
requests:

```
[11] user: 'Briefly, what is dark matter?'
     content: ''
[ 3] user: 'Explain what a hash table is in 30 words or less.'
     content: 'Dark matter is a hypothetical form of matter ...'
[ 4] user: 'Who painted the Mona Lisa?'
     content: 'Dark matter is a hypothetical form of matter ...'
[ 5] user: 'What gas do plants absorb during photosynthesis?'
     content: 'Dark matter is a hypothetical form of matter ...'
[ 6] user: "Translate 'thank you' into German."
     content: 'Dunkle Materie.'   (literally "dark matter" in German)
[ 9] user: 'Who wrote Hamlet?'
     content: 'Dark matter is a hypothetical form of matter ...'
```

The "Dunkle Materie" output is especially telling. The request prompt was
"translate thank you into German", but the model produced the German
translation of "dark matter". The sequence decoded the semantic content
of a different request while applying the tokenizer / language context of
its own prompt. That fingerprints a shared-state bug deep in the decode
path, not a prompt mix-up at dispatch.

## Expected behavior

Each concurrent request must produce an answer that corresponds to its
own prompt. Requests should not see each other's token stream.

## Single-request and low-concurrency behavior

- A single chat request with `enable_thinking=True` produces a clean,
  correctly-attributed response.
- 8 concurrent requests with `enable_thinking=False` also produce clean
  outputs (no corruption observed).
- The failure shows up when concurrency climbs to about 12 and
  `enable_thinking=True`.

## Known guardrails that are not enough

`vllm_mlx/patches/gemma4_mllm.py` patches
`mlx_vlm.models.gemma4.language.Attention.__call__` to snapshot
`cache.offset` before `update_and_fetch`, which prevents a RoPE offset
corruption on `BatchKVCache`. That fix is in place and activated from
`mllm_batch_generator.py`, but does not cover this bug. There is a
second path, likely one of:

- Sliding-window / hybrid attention layers keep a second offset that is
  not snapshotted by the current patch.
- The `is_kv_shared_layer=True` shortcut in the patched `__call__` reads
  `cache.state` without snapshotting a per-sequence offset.
- Prefix-cache key computation does not account for the `<|think|>`
  system-prompt token introduced when `enable_thinking=True`, producing
  false prefix hits across sequences and corrupting the decode state.

## Related

- Originally reported in [#380](https://github.com/waybarrios/vllm-mlx/issues/380).
- [#383](https://github.com/waybarrios/vllm-mlx/pull/383) fixes the
  parser and server symptoms from that issue, not this one.
- Likely connected to `mlx-vlm`'s batched attention path
  (`mlx_vlm.models.gemma4.language.Attention`).

## Acceptance criteria

- 12 concurrent requests with `enable_thinking=True` on any Gemma 4
  model produce correct, per-request answers.
- A regression test exercises the concurrent-thinking case so the bug
  does not return.

waybarrios:
## Purpose

Fixes [#384](https://github.com/waybarrios/vllm-mlx/issues/384). Under
12 to 32 concurrent chat requests with `enable_thinking=True`, the
content of one request bleeds into unrelated requests. A "who wrote
Hamlet?" request comes back with "Alexander Fleming discovered
penicillin"; a "translate thank you into German" request returns "Dunkle
Materie" (the German for "dark matter"). The contamination is
deterministic: the same slots keep picking up the same foreign content
across rounds, which lets it reproduce reliably inside a hammer script.
Disabling the prefix cache with `--disable-prefix-cache` eliminates the
contamination entirely, so the fault is inside the cache path, not the
engine or the scheduler.

This PR identifies the exact mechanism and applies a one-file fix with
targeted regression coverage.

## Root cause

The `MemoryAwarePrefixCache.fetch()` path in `vllm_mlx/memory_cache.py`
has a Longest-Common-Prefix (LCP) fallback: when the new request shares
only part of its token sequence with a stored entry, the cache trims
the stored entry back to the shared length and returns it. Trimming is
done by `_trim_cache_offset(cache, trim_by)`.

The rotating-cache branch already trimmed the circular buffer
correctly. The plain-`KVCache` branch did not. It created a shallow
copy that adjusted the offset but kept `keys` and `values` pointing at
the original full-size array:

```python
# before
tc = orig_cls.__new__(orig_cls)
tc.keys = layer_cache.keys
tc.values = layer_cache.values
tc.offset = max(layer_cache.offset - trim_by, 0)
```

Stored entries routinely hold 500+ tokens worth of data: the system
prompt, then the previous request's user message, then that request's
thinking, then its response. Only the first N tokens (the LCP length,
often 40 to 80 for the system prompt) are semantically valid for a new
request. The remaining positions still hold another request's private
content.

That stale content was fine for any reader that honoured `self.offset`,
because `update_and_fetch` returns only `self.keys[..., :self.offset, :]`.
But Gemma 4 has KV-shared attention layers that read `cache.state`
directly (see `vllm_mlx/patches/gemma4_mllm.py`, lines 84 to 86):

```python
if self.is_kv_shared_layer and cache is not None:
    state = cache.state
    keys, values = state[0], state[1]
```

`cache.state` returns the raw `(self.keys, self.values)` arrays and
ignores `self.offset`. The new request's attention therefore
conditioned on positions 60 through 500, still holding another
request's tokens. The model produced a coherent answer to the other
request, routed back to the current request's slot.

Because the shared system prompt is deterministic and so is the
scheduler order for a fixed set of concurrent requests, the same slot
in each round picked up the same foreign content. That is why the
observed pattern is reproducible: "Hamlet" always got "penicillin";
"largest ocean" always got "Japanese Yen"; "Mona Lisa" always got
"Edison / light bulb".

## Fix

`_trim_cache_offset` now slices `keys` and `values` down to
`new_offset` on the plain-`KVCache` branch:

```python
# vllm_mlx/memory_cache.py  —  _trim_cache_offset, plain-KVCache branch
new_offset = max(layer_cache.offset - trim_by, 0)
keys = layer_cache.keys
values = layer_cache.values
if keys is not None and hasattr(keys, "shape") and len(keys.shape) >= 3:
    if new_offset < keys.shape[-2]:
        # Slice the arrays down.  KVCache.update_and_fetch will
        # detect shape[-2] == offset on the first write, take the
        # grow branch, and mx.concatenate into a fresh array, so
        # in-place writes by the new request cannot leak into the
        # stored prefix entry even though MLX slicing is a view.
        tc.keys = keys[..., :new_offset, :]
        tc.values = values[..., :new_offset, :]
    else:
        tc.keys = keys
        tc.values = values
else:
    tc.keys = keys
    tc.values = values
tc.offset = new_offset
```

Why the slice is safe even though MLX slicing is a view: on the very
first call to `update_and_fetch` through the trimmed cache,
`prev + keys.shape[2]` exceeds `self.keys.shape[2]` (since `shape[-2]
== new_offset == prev` right after trim), so the grow branch runs,
`mx.concatenate` allocates a fresh buffer, and the subsequent in-place
assignment lands in the fresh buffer rather than the original stored
array. The rotating-cache branch is untouched because it was already
materialising new arrays via `_temporal_order` and `mx.concatenate`.

Earlier iterations of the fix added an explicit `mx.eval` to force
materialisation of the slice, but that produced a `Fatal Python error:
Aborted` under the scheduler's worker thread while another thread was
inside `mlx_lm.generate._next`. The bug was timing-related: calling
`mx.eval` on a slice concurrently with another thread eval-ing a parent
graph caused a race. The slice-only version is both simpler and free
of the race, because update_and_fetch's grow branch is the thing that
actually provides isolation (the eval was belt-and-suspenders).

## How to verify

Server command (same flags as the user's report in #380 and #384):

```
vllm-mlx serve mlx-community/gemma-4-e2b-it-mxfp4 \
  --host 127.0.0.1 --port 18082 \
  --continuous-batching \
  --enable-auto-tool-choice --tool-call-parser gemma4 \
  --reasoning-parser gemma4 \
  --max-tokens 4096
```

Hammer script (attached in the issue): 32 distinct prompts, 8 rounds of
varying concurrency (32, 16, 32, 16, 32, 24, 32, 20) = 204 requests
total, all with `temperature=0.0`, `stream=false`, `enable_thinking=true`.

| Condition | Contaminated responses |
|---|---|
| `main` | **22 / 204** |
| `main` + `--disable-prefix-cache` | 0 / 204 |
| this PR | **0 / 204** |

Example output on `main` before the fix:

```
[ 1] 'Name the tallest mountain in South America.'
     content: 'Japanese Yen.'         [!! foreign from 'Name the currency of Japan.' !!]
[ 4] 'Who painted the Mona Lisa?'
     content: 'Thomas Edison is credited with inventing the practical light bulb.'
[ 9] 'Who wrote Hamlet?'
     content: 'Alexander Fleming discovered penicillin.'
[23] 'Name the largest ocean on Earth.'
     content: 'Japanese Yen (JPY).'
```

After the fix, every slot returns the answer to its own prompt.

## Tests

Ten new unit tests in
`tests/test_memory_cache.py::TestTrimCacheOffset`, all using real
`mx.array` objects on MLX, covering the failure mode and adjacent
edge cases:

- `test_plain_kv_cache_array_sliced_to_new_offset` — after trim,
  `tc.keys.shape[-2] == new_offset`, not the original shape.
- `test_plain_kv_cache_no_stale_tokens_visible_via_state` — stores
  `1.0` in the shared prefix and `7.0` past it, confirms that the
  view returned by `cache.state` contains only `1.0`s. This is the
  direct regression of the #384 symptom.
- `test_plain_kv_cache_no_trim_preserves_array` — `trim_by == 0`
  leaves the array untouched.
- `test_plain_kv_cache_trim_by_exceeds_offset_clamps_to_zero` —
  trimming past the current offset yields an empty-but-valid trimmed
  cache (`offset == 0`, `shape[-2] == 0`).
- `test_plain_kv_cache_stored_entry_unaffected_after_trim` — running
  `_trim_cache_offset` on a layer must not mutate the source layer's
  `keys` / `values` / `offset`.
- `test_plain_kv_cache_in_place_write_does_not_corrupt_source` —
  after trim, calling `update_and_fetch(new_keys, new_values)` on the
  returned cache must not bleed into the source entry. Values in the
  source stay at their original `5.0` across the full range.
- `test_plain_kv_cache_multiple_layers_all_sliced` — caches with
  several `KVCache` layers get every layer sliced, not just the
  first.
- `test_plain_kv_cache_slice_works_for_float16_and_bfloat16` —
  fix is dtype-agnostic; fp16 and bf16 caches receive the same
  treatment and dtype is preserved.
- `test_plain_kv_cache_rotating_layers_unchanged_behavior` —
  rotating-cache branch still works after the change, dtype and
  `max_size` are preserved, output arrays are non-None.
- `test_fetch_returns_sliced_cache_on_lcp_match` — end-to-end
  regression: `MemoryAwarePrefixCache.fetch()` on a request that
  shares only a prefix with a longer stored entry returns a cache
  whose arrays are already sliced down. Asserts that the `7.0`
  private content from the stored entry is not visible anywhere in
  the returned cache.

## Test results

All 419 relevant tests pass locally on M4 Max / 128 GB:

```
$ pytest tests/test_memory_cache.py tests/test_prefix_cache.py \
         tests/test_gemma4_tool_parser.py tests/test_gemma4_streaming_edge.py \
         tests/test_reasoning_parser.py tests/test_tool_parsers.py \
         tests/test_batching_deterministic.py tests/test_native_tool_format.py \
         tests/test_chat_template_kwargs.py tests/test_gemma4_openai_format.py \
         tests/test_api_utils.py -q
419 passed in 5.40s
```

The 10 new regression tests pass in 2 seconds on their own:

```
$ pytest tests/test_memory_cache.py::TestTrimCacheOffset -v
10 passed in 1.92s
```

## Out of scope

- The parser-level fixes for #380 remain on their own branch
  ([#383](https://github.com/waybarrios/vllm-mlx/pull/383)) and merge
  independently. The two PRs have no file overlap.
- Only the plain-`KVCache` branch of `_trim_cache_offset` changes.
  The rotating-cache branch and the quantized-wrapper branch are
  untouched and covered by existing tests.
- No scheduler or engine changes. No change to how requests are
  dispatched, batched, or merged.

Closes #384. Refs #380.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -134,6 +134,7 @@ jobs:
             tests/test_chat_template_kwargs.py \
             tests/test_batching.py \
             tests/test_continuous_batching.py \
+            tests/test_memory_cache_mlx.py \
             -v --tb=short \
             -m "not slow" \
             -k "not Integration"
--- a/tests/test_memory_cache_mlx.py
+++ b/tests/test_memory_cache_mlx.py
@@ -1,0 +1,281 @@
+# SPDX-License-Identifier: Apache-2.0
+"""MLX-dependent regression tests for the LCP trim contamination fix (#384).
+
+These tests use real ``mlx_lm.models.cache.KVCache`` / ``RotatingKVCache``
+objects backed by ``mlx.core`` arrays.  They run only on the apple-silicon
+CI matrix (and locally on M-series hardware); the Linux ``test-matrix``
+job excludes this file because MLX has no Linux distribution.
+"""
+
+from unittest.mock import MagicMock
+
+
+class TestTrimCacheOffset:
+    """Tests for ``_trim_cache_offset``, focused on the LCP contamination fix.
+
+    Regression: when the LCP fetch path trimmed a cache entry by shrinking
+    the offset while still sharing the underlying (oversized) key/value
+    arrays, downstream attention layers that read ``cache.state`` directly
+    (e.g. Gemma 4 KV-shared layers) could see stale tokens from the previous
+    owner of the entry.  See issue #384.  The fix slices the arrays down to
+    new_offset so no memory beyond the new boundary remains accessible.
+    """
+
+    def test_plain_kv_cache_array_sliced_to_new_offset(self):
+        """Plain-KVCache-like layer: after trim, keys.shape[-2] == new_offset."""
+        import mlx.core as mx
+        from mlx_lm.models.cache import KVCache
+
+        from vllm_mlx.memory_cache import _trim_cache_offset
+
+        layer = KVCache()
+        # Pretend a previous request wrote 500 tokens worth of data
+        layer.keys = mx.arange(1 * 4 * 500 * 8, dtype=mx.float32).reshape(1, 4, 500, 8)
+        layer.values = mx.arange(1 * 4 * 500 * 8, dtype=mx.float32).reshape(
+            1, 4, 500, 8
+        )
+        layer.offset = 500
+
+        # New request shares only the first 60 tokens as prefix
+        trim_by = 500 - 60
+        trimmed = _trim_cache_offset([layer], trim_by)
+        tc = trimmed[0]
+
+        assert tc.offset == 60
+        # The underlying array MUST be shrunk, not just the offset pointer.
+        # Otherwise Gemma 4's cache.state-reading layers would see positions
+        # 60..500 filled with the previous request's tokens.
+        assert tc.keys.shape[-2] == 60
+        assert tc.values.shape[-2] == 60
+
+    def test_plain_kv_cache_no_stale_tokens_visible_via_state(self):
+        """A layer that reads the full cache.state must not see tokens past new_offset."""
+        import mlx.core as mx
+        from mlx_lm.models.cache import KVCache
+
+        from vllm_mlx.memory_cache import _trim_cache_offset
+
+        layer = KVCache()
+        # Positions 0..60: shared prefix (same for everyone).  Positions 60..500:
+        # private content from a previous request that must NOT leak.
+        shared = mx.ones((1, 4, 60, 8), dtype=mx.float32)
+        private = mx.full((1, 4, 440, 8), 7.0, dtype=mx.float32)
+        layer.keys = mx.concatenate([shared, private], axis=2)
+        layer.values = layer.keys
+        layer.offset = 500
+
+        tc = _trim_cache_offset([layer], 500 - 60)[0]
+
+        # cache.state is what KV-shared layers read directly.
+        keys_view, _ = tc.state
+        assert keys_view.shape[-2] == 60
+        # No "7.0" tokens anywhere — private content was excluded.
+        assert float(mx.max(keys_view).item()) == 1.0
+
+    def test_plain_kv_cache_no_trim_preserves_array(self):
+        """If trim_by == 0 or offset already equals shape, array is untouched."""
+        import mlx.core as mx
+        from mlx_lm.models.cache import KVCache
+
+        from vllm_mlx.memory_cache import _trim_cache_offset
+
+        layer = KVCache()
+        layer.keys = mx.ones((1, 4, 100, 8), dtype=mx.float32)
+        layer.values = mx.ones((1, 4, 100, 8), dtype=mx.float32)
+        layer.offset = 100
+
+        tc = _trim_cache_offset([layer], 0)[0]
+
+        assert tc.offset == 100
+        assert tc.keys.shape[-2] == 100
+
+    def test_plain_kv_cache_trim_by_exceeds_offset_clamps_to_zero(self):
+        """trim_by larger than offset yields an empty-but-valid trimmed cache."""
+        import mlx.core as mx
+        from mlx_lm.models.cache import KVCache
+
+        from vllm_mlx.memory_cache import _trim_cache_offset
+
+        layer = KVCache()
+        layer.keys = mx.ones((1, 4, 80, 8), dtype=mx.float32)
+        layer.values = mx.ones((1, 4, 80, 8), dtype=mx.float32)
+        layer.offset = 80
+
+        tc = _trim_cache_offset([layer], 1000)[0]
+
+        assert tc.offset == 0
+        assert tc.keys.shape[-2] == 0
+        assert tc.values.shape[-2] == 0
+
+    def test_plain_kv_cache_stored_entry_unaffected_after_trim(self):
+        """Calling _trim_cache_offset must not mutate the source layer in place.
+
+        The stored prefix-cache entry is the source here; a later lookup for
+        a different request should get the same pristine data.
+        """
+        import mlx.core as mx
+        from mlx_lm.models.cache import KVCache
+
+        from vllm_mlx.memory_cache import _trim_cache_offset
+
+        layer = KVCache()
+        full = mx.arange(1 * 2 * 200 * 4, dtype=mx.float32).reshape(1, 2, 200, 4)
+        layer.keys = full
+        layer.values = full
+        layer.offset = 200
+        original_shape = layer.keys.shape
+
+        _trim_cache_offset([layer], 150)
+
+        # Source entry keeps its full shape and offset.
+        assert layer.keys.shape == original_shape
+        assert layer.values.shape == original_shape
+        assert layer.offset == 200
+
+    def test_plain_kv_cache_in_place_write_does_not_corrupt_source(self):
+        """After trim, writing through the returned cache must not leak into
+        the stored entry.  This is the direct semantics of the fix: the stored
+        prefix-cache entry has to survive concurrent use by other requests.
+        """
+        import mlx.core as mx
+        from mlx_lm.models.cache import KVCache
+
+        from vllm_mlx.memory_cache import _trim_cache_offset
+
+        # Source stored entry: positions 0..300 holding 5.0.
+        layer = KVCache()
+        layer.keys = mx.full((1, 2, 300, 4), 5.0, dtype=mx.float32)
+        layer.values = mx.full((1, 2, 300, 4), 5.0, dtype=mx.float32)
+        layer.offset = 300
+
+        # New request shares first 50 tokens.
+        tc = _trim_cache_offset([layer], 300 - 50)[0]
+
+        # The trimmed cache now only has 50 tokens.  Writing new tokens via
+        # update_and_fetch allocates a new array (because prev + N > current
+        # shape) and does not touch the source.
+        new_keys = mx.zeros((1, 2, 10, 4), dtype=mx.float32)
+        new_values = mx.zeros((1, 2, 10, 4), dtype=mx.float32)
+        tc.update_and_fetch(new_keys, new_values)
+
+        # Source remains untouched (all 5.0 values preserved across full range).
+        assert layer.keys.shape[-2] == 300
+        assert float(mx.min(layer.keys).item()) == 5.0
+        assert float(mx.max(layer.keys).item()) == 5.0
+
+    def test_plain_kv_cache_multiple_layers_all_sliced(self):
+        """Caches with several KVCache layers: every layer gets sliced."""
+        import mlx.core as mx
+        from mlx_lm.models.cache import KVCache
+
+        from vllm_mlx.memory_cache import _trim_cache_offset
+
+        layers = []
+        for _ in range(5):
+            layer = KVCache()
+            layer.keys = mx.ones((1, 4, 200, 8), dtype=mx.float32)
+            layer.values = mx.ones((1, 4, 200, 8), dtype=mx.float32)
+            layer.offset = 200
+            layers.append(layer)
+
+        trimmed = _trim_cache_offset(layers, 150)
+
+        assert len(trimmed) == 5
+        for tc in trimmed:
+            assert tc.offset == 50
+            assert tc.keys.shape[-2] == 50
+            assert tc.values.shape[-2] == 50
+
+    def test_plain_kv_cache_slice_works_for_float16_and_bfloat16(self):
+        """Fix must be dtype-agnostic so quantized / mixed-precision KV caches
+        receive the same treatment as fp32.
+        """
+        import mlx.core as mx
+        from mlx_lm.models.cache import KVCache
+
+        from vllm_mlx.memory_cache import _trim_cache_offset
+
+        for dtype in (mx.float16, mx.bfloat16):
+            layer = KVCache()
+            layer.keys = mx.ones((1, 2, 120, 4), dtype=dtype)
+            layer.values = mx.ones((1, 2, 120, 4), dtype=dtype)
+            layer.offset = 120
+
+            tc = _trim_cache_offset([layer], 80)[0]
+
+            assert tc.offset == 40, f"dtype={dtype}"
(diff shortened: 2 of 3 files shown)


janhilgard (approved):
Excellent fix. The root cause analysis is thorough and the fix is clean.

## Review summary

**Root cause**: `_trim_cache_offset`'s plain-KVCache branch created a shallow copy that adjusted `.offset` but left `.keys` / `.values` pointing at the original full-size array. Readers that access the raw arrays (like Gemma 4's KV-shared layers via `cache.state`) bypass the offset boundary and see stale tokens from a previous request.

**Fix**: Slice `keys[..., :new_offset, :]` and `values[..., :new_offset, :]` so no memory beyond the shared prefix is accessible. On the first `update_and_fetch`, the grow branch runs `mx.concatenate` which allocates a fresh buffer, so the stored entry is never mutated.

## What I verified

- The diff is minimal and surgical — only the plain-KVCache branch of `_trim_cache_offset` changes (3 lines removed, 21 added in `memory_cache.py`).
- The rotating-cache branch is untouched and was already correct.
- The slice-based approach avoids the `mx.eval` race condition that crashed earlier iterations (documented in the PR body).
- MLX slicing returns a view, but `update_and_fetch`'s grow path allocates a new array on first write, so the stored entry remains isolated. This is correct.
- Guard conditions (`keys is not None`, `hasattr(keys, "shape")`, `len(keys.shape) >= 3`, `new_offset < keys.shape[-2]`) are defensive and handle edge cases cleanly.
- Tests cover the exact failure mode (7.0 private tokens not visible after trim), source immutability, dtype preservation, multiple layers, and an end-to-end `fetch()` regression test.
- All CI checks pass.

## Missing: `_QuantizedCacheWrapper` branch has the same bug

The PR body states the quantized-wrapper branch is "out of scope and covered by existing tests", but **it has the exact same stale-token leak** when `--kv-cache-quantization` is enabled (which our Gemma 4 production server uses).

**Bug flow with quantized cache:**

1. Request A stores `KVCache(offset=500, keys=[500 tokens])` → `store()` quantizes into `_QuantizedCacheWrapper(offset=500, keys=quantized_500_tokens)`
2. Request B matches LCP (shared=60 tokens)
3. `_trim_cache_offset`: `_QuantizedCacheWrapper` branch only reduces offset `500→60`, **keys/values tuples stay at full 500-token size** (lines 281-291 — unchanged by this PR)
4. `_dequantize_cache`: dequantizes ALL 500 tokens into plain KVCache, copies offset=60 → `KVCache(offset=60, keys=[500 dequantized tokens])`
5. Gemma 4 KV-shared layer reads `cache.state` → sees all 500 tokens including 440 stale ones from request A
6. Model attention conditions on stale content → garbage output (confirmed: `<|tool_call>call:web_search{query:<|"|>...de de de de de de...}`)

**Reproduction**: Long Czech system prompt (~1300+ tokens) + `temp=0.4` + 2 tools + `--kv-cache-quantization` → garbage output ~20-100% of the time. With `temp=0` the model is robust enough to override the noise.

**Fix**: Slice dequantized arrays in `_dequantize_cache()` after line 556 (`kv.offset = layer.offset`):

```python
# Slice the dequantized arrays down to offset so that readers
# which bypass offset (e.g. Gemma 4 KV-shared layers reading
# cache.state directly) cannot see stale tokens from a previous
# request.  Mirrors the plain-KVCache slice in
# _trim_cache_offset — see issue #384.
if (
    kv.keys is not None
    and hasattr(kv.keys, "shape")
    and len(kv.keys.shape) >= 3
    and kv.offset < kv.keys.shape[-2]
):
    kv.keys = kv.keys[..., : kv.offset, :]
    kv.values = kv.values[..., : kv.offset, :]
```

The fix goes in `_dequantize_cache()` (not `_trim_cache_offset`) because quantized data is stored as tuples `(data, scales, biases)` which can't be trivially sliced — slicing the dequantized fp32/bf16 arrays is both simpler and correct.

I have the fix + 5 regression tests ready locally (all 15 MLX cache tests pass). Happy to push a commit to this branch or open a follow-up PR.

janhilgard (changes requested):
Changing to request-changes: the `_QuantizedCacheWrapper` branch in `_dequantize_cache()` has the same stale-token leak described in the review comment above.

When `--kv-cache-quantization` is enabled (which our Gemma 4 production server uses), dequantized arrays are not sliced to `offset`, so KV-shared layers reading `cache.state` can still see stale tokens from previous requests.

The fix is straightforward — slice dequantized arrays after `kv.offset = layer.offset` in `_dequantize_cache()`. See the review comment for the exact code. This should be included in this PR before merge.

janhilgard (dismissed):
Changing from APPROVED to REQUEST_CHANGES — the `_QuantizedCacheWrapper` branch of `_dequantize_cache()` has the same stale-token leak as the plain-KVCache branch that this PR fixes. See my updated review above for full details and a ready-to-use fix.

The fix is a 9-line addition to `_dequantize_cache()` that slices the dequantized arrays down to `offset` before returning. Without it, servers using `--kv-cache-quantization` (like our Gemma 4 production instance) still see cross-request contamination via `cache.state`.

Thump604 (approved):
Strong root cause analysis and clean fix. The LCP trim path was the right place to look -- the offset-only shallow copy with Gemma 4's KV-shared `cache.state` bypass is an easy-to-miss interaction.

The slice-based approach is correct: MLX slicing returns a view but `update_and_fetch`'s grow branch materializes via `mx.concatenate`, so the stored entry stays isolated. The guard conditions are defensive and handle all edge cases. Tests directly exercise the contamination path.

I agree with janhilgard's review: the `_QuantizedCacheWrapper` branch in `_dequantize_cache()` has the same leak and should be fixed in this PR rather than as a follow-up. The failure mode is identical (stale tokens visible via `cache.state` after dequantization), and shipping the plain fix alone would leave quantized-cache users exposed to the same cross-request contamination. Since janhilgard has the fix and tests ready, including it here keeps the fix atomic.

janhilgard (approved):
The quantized-cache gap is now fixed. The new commit slices dequantized arrays to `offset` in `_dequantize_cache()`, mirroring the plain-KVCache fix from the first commit. Five new tests cover the quantized path. All CI checks pass.

LGTM — both the plain and quantized branches are now safe against stale-token leaks via `cache.state`.

waybarrios:
all set for this?

janhilgard:
Yes, all set! CI is green (all 9 checks pass), review is approved, and the fix has been tested on production with quantized KV cache enabled. Ready to merge.

PR #387, opened 2026-04-21
fix(gemma4): Comprehensive UAT and fixes

0xAlcibiades:
Follow-up to #380, #383, #385. The bugs enumerated below were found by running `unsloth/gemma-4-31b-it-MLX-8bit` under `--continuous-batching --enable-auto-tool-choice --tool-call-parser gemma4 --reasoning-parser gemma4` against a realistic workload — Honcho as a concurrent load generator with dialectic, deriver, summary, and embeddings all firing per turn; opencode and Hermes Agent for tool-call and reasoning streaming; 14-16K-token agent system prompts with multi-turn history — and fixing each failure until Gemma 4 was stable under that workload.

The branch was developed locally while #383 and #385 were in review, so it carries re-implementations of a few of those commits. Rebasing onto current `main` to drop the duplicates is fine; the net diff in this PR is what's new beyond them.

### Gemma 4 reasoning + tool parser

1. `<|tool_response>` was not declared as end-of-generation — the model kept generating past tool calls.
2. Bare string tool-call values produced by nullable / enum schemas (`["string","null"]`, enum-without-type) failed JSON parse because the template skipped the `<|"|>` wrap.
3. Streaming "thought" channel label leaked partial prefixes (`th`, `tho`, `thoug`) into `reasoning_content` when the label arrived split across deltas.
4. Reasoning parser leaked `<channel|>` marker into `content`.
5. Reasoning parser overwrote already-emitted content when the thought label completed mid-stream.
6. `BaseThinkingReasoningParser`'s containment check (`end_tok in current_text and end_tok not in previous_text`) missed a second end-token in streams with back-to-back reasoning blocks.
7. `BaseThinkingReasoningParser` didn't re-enter thinking from content on a fresh start token, so consecutive `<think>` blocks collapsed into one.
8. Multi-block tool-call responses only parsed the first `<|tool_call>...<tool_call|>` block; subsequent blocks were ignored.
9. Streaming tool-call emission lacked stable IDs and re-sent the full tool-call list on every delta that contained a closing marker.
10. `Gemma4ToolParser.SUPPORTS_NATIVE_TOOL_FORMAT` was False, so JSON-string `arguments` leaked through the template as `{{"k":"v"}}`.
11. OpenAI `role:tool` messages were not coalesced into the preceding assistant's `tool_responses` field, which is what the Gemma 4 template's `is mapping` branch actually expects.
12. String tool-response content was not wrapped as `{content: ...}` for that mapping branch.

### Prefix cache and KV correctness

13. `_trim_cache_offset` on plain `KVCache` shrank the offset pointer but left `keys`/`values` pointing at the full-size array. Gemma 4's KV-shared attention layers read `cache.state` directly and saw stale tokens from the previous owner past the new boundary.
14. `MemoryAwarePrefixCache` had no thread lock protecting `_entries` / `_sorted_keys` / `_current_memory` / `_stats`.
15. LCP scan only checked the two neighbors around the bisect insertion point. It missed valid LCP candidates when many entries shared a first token.
16. Non-quantized fetch returned references to the stored entry. The model's in-place `update_and_fetch` on `RotatingKVCache` buffers then corrupted that stored entry.
17. `_QuantizedCacheWrapper` dequantize did not slice the output to `layer.offset`. A post-trim wrapper whose quantized arrays still covered the pre-trim length leaked trailing slots.
18. `MLLMBatch.extract_cache` returned slice views into the shared batch KV tensor. Extracted caches mutated under their new owner as the batch kept generating for other slots.
19. `RotatingKVCache.offset` was being clamped to `max_size` after a sliding-window trim. The offset tracks the global token position and is consumed by RoPE; clamping it told RoPE a 12K-token prefill landed at position 1024, producing language mixing and repetition loops.
20. RotatingKVCache buffers that exceeded `max_size` after an LCP trim crashed `BatchRotatingKVCache.merge()` with "Negative dimensions not allowed".
21. Chunked prefill with `step > sliding_window` evicted the prior chunk wholesale at each boundary. Early queries in the next chunk saw no in-window context, breaking SWA layers.

### HF fast tokenizer concurrency

22. Concurrent async handlers calling `apply_chat_template` on the Rust-backed fast tokenizer raised `RuntimeError: Already borrowed`.
23. Same race on `add_request_async`'s prompt-token `encode`.
24. Same race on `_compute_think_suffix_len`.
25. Same race on `_preprocess_request`'s `prepare_inputs` call.
26. `prepare_inputs` persistently mutated `tokenizer.pad_token` as a side effect. The mutation leaked across requests.

### Sampling params and server defaults

27. Per-request samplers were built only when `top_k != 0 or min_p != 0`. Every other request silently dropped its per-request `temperature` / `top_p` / `top_k` / `min_p`.
28. The scheduler installed a hardcoded module-level `make_sampler(temp=0.7, top_p=0.9)` that overrode request values on the path that wasn't using per-request samplers.
29. `--default-top-k`, `--default-min-p`, `--default-presence-penalty`, and `--default-repetition-penalty` CLI flags didn't exist. Only `--default-temperature` and `--default-top-p` did.
30. Sampling defaults did not resolve from the loaded model's `generation_config.json`.
31. `_apply_generation_config_defaults()` ran before `BatchedEngine.start()` populated the processor. For MLLM the resolution was effectively a no-op.
32. `BatchedEngine.tokenizer` returned the wrapped tokenizer instead of the processor, so callers that needed `name_or_path` to find `generation_config.json` couldn't reach it for MLLM.
33. No per-request log line showed the resolved sampling kwargs. There was no way to verify what the server actually applied.

### Metal driver asserts under concurrent HTTP

34. The scheduler split `step()` between the event-loop thread (decode-only inline) and the `mllm-step` executor thread (prefill). `mx.async_eval` left encoder state on `mx.Stream` in a transitional state across the handoff, so the next thread's `with mx.stream(_stream):` created an encoder on the shared command buffer before the prior thread's submit had closed it — `'A command encoder is already encoding to this command buffer'`.
35. After fixing 34, the embedding / rerank / audio handlers still submitted MLX work from the event loop, which raced the scheduler's executor thread and surfaced as `'Completed handler provided after commit call'`.
36. `/v1/embeddings` called `_embedding_engine.count_tokens` and `.embed` synchronously on the event loop.
37. `/v1/rerank` used `asyncio.to_thread`, which picks arbitrary threadpool threads rather than the scheduler's executor.
38. `/v1/audio/transcriptions` ran `STTEngine.load` and `.transcribe` on the event loop.
39. `/v1/audio/speech` ran `TTSEngine.load` and `.generate` on the event loop.

### Scheduler and batch generator plumbing

40. The batch generator was built lazily on the first request. Concurrent first requests raced against the init path (attention patches, `_compute_think_suffix_len`, prefix cache setup) while it was still running.
41. `_preprocess_request` re-tokenized text-only requests when both chunked prefill interleaving and `_process_prompts` called it for the same request.
42. `_copy_prefix_cache` became a dead shallow-copy call site after `fetch()` was changed to deep-clone. Left the branch in a confused state.

### Multimodal / video

43. Video-native models could not flow through the batched scheduler. The processor's `<|image>...<|video|>...<image|>` timestamp interleave can't be reconstructed from a rendered chat-template string.
44. `_run_vision_encoding` called `self.model(input_ids, ..., attention_mask=...)` with the wrong kwarg name. Gemma 4's forward takes `mask=`; `attention_mask` fell into `**kwargs` and was ignored.
45. `_run_vision_encoding` dropped the embedding-time state returned by `get_input_embeddings` (M-RoPE deltas, position IDs, etc.). The language model didn't receive the state it needed for decode.
46. `process_image_input` rejected local filesystem paths. The video-frame extraction pipeline couldn't feed frames into the MLLM.

### Server and streaming lifecycle

47. `_ensure_sse_terminal` yielded the terminal frame during `GeneratorExit`. Yielding inside `finally` while unwinding from `GeneratorExit` produces Python's "async generator ignored GeneratorExit" warning.
48. `_disconnect_guard` cancelled helper tasks but never awaited them. Just cancelling leaves them suspended mid-await, which triggers the same warning when the consumer calls `aclose()` on us.
49. `MLXMultimodalLM.chat` and `stream_chat` each reimplemented a multi-turn message loop that dropped `tool_calls`, `tool_call_id`, and `name` fields and collapsed dict content.
50. `stream_chat` didn't forward `audio=all_audio_urls` to the mlx_vlm generator, so audio inputs were silently dropped from the streaming path.

## New features

`--chunked-prefill-tokens N` enables interleaved prefill/decode on the continuous-batching path. When a long text-only request arrives at an active batch, one chunk of its prefill runs per scheduler step, interleaved with generation for the active batch. In-flight throughput stays at 30-50 tok/s instead of collapsing to 1-5 tok/s for the full prefill duration. Handles abort-during-partial, post-prefill merge into the batch cache, and prefix-cache store on completion. Capped at `prefill_step_size` so the SWA safety cap applies.

Native video through the batched scheduler. `BatchedEngine._native_video_preprocess` runs `MLXMultimodalLM._prepare_native_video_inputs` under the tokenizer lock to produce `input_ids`, `pixel_values`, `mask`, and `video_grid_thw`. New `MLLMRequest.preprocessed_*` fields and a `MLLMBatchRequest.is_preprocessed` flag propagate the pre-tokenized inputs to the scheduler, which short-circuits its own preprocess pass. `_run_vision_encoding` mirrors upstream `mlx_vlm.generate.generate_step`: `model.get_input_embeddings(input_ids, pixel_values, mask=mask, **extras)` then `language_model(inputs_embeds=..., cache=..., **state)`.

## Tests

- `tests/test_memory_cache_mlx.py` (281 lines, from #385's cherry-pick; included here) — MLX-backed regression for `_trim_cache_offset` and end-to-end `MemoryAwarePrefixCache.fetch` on an LCP match. Covers dtype agnosticism (fp16 / bf16 / fp32), multiple layers, trim-exceeds-offset, source-entry preservation after trim, and quantized-wrapper slicing. Added to the apple-silicon CI matrix; excluded from the Linux `test-matrix` job because MLX has no Linux distribution.
- `tests/test_gemma4_tool_parser.py` — bare-value quoting, JSON-literal filtering, `<|tool_response>` in `extra_stop_tokens`, `get_parser_stop_tokens` merge / dedup / unknown-parser / None-parser passthrough.
- `tests/test_gemma4_streaming_edge.py` — streaming "thought" label split across deltas.

## Upstream references

- ggml-org/llama.cpp#21327 — Gemma 4 bare string tool-call values
- ggml-org/llama.cpp#21365 — Gemma 4 repetition bias
- ggml-org/llama.cpp#21418 — `<|tool_response>` as end-of-generation

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -134,6 +134,7 @@ jobs:
             tests/test_chat_template_kwargs.py \
             tests/test_batching.py \
             tests/test_continuous_batching.py \
+            tests/test_memory_cache_mlx.py \
             -v --tb=short \
             -m "not slow" \
             -k "not Integration"
--- a/tests/test_gemma4_streaming_edge.py
+++ b/tests/test_gemma4_streaming_edge.py
@@ -27,6 +27,27 @@ def stream_parse(parser, text, token_list):
     return "".join(reasoning_parts), "".join(content_parts)
 
 
+def test_thought_label_split_across_deltas():
+    """The channel-name word 'thought' can arrive split across deltas in
+    production streams. The parser must buffer the partial prefix rather than
+    leak 'th' / 'tho' / 'thou' / 'thoug' into the reasoning output.
+    """
+    parser = Gemma4ReasoningParser()
+    tokens = [
+        "<|channel>",
+        "tho",        # partial "thought" label
+        "ught",       # completes the label
+        "\n",
+        "real reasoning",
+        "<channel|>",
+        "content",
+    ]
+    reasoning, content = stream_parse(parser, None, tokens)
+    assert reasoning == "real reasoning", f"Label leaked: {reasoning!r}"
+    assert content == "content", f"Got content: {content!r}"
+    print("[PASS] test_thought_label_split_across_deltas")
+
+
 def test_channel_marker_split_across_tokens():
     """<|channel> alone should not leak into reasoning if response follows."""
     parser = Gemma4ReasoningParser()
--- a/tests/test_gemma4_streaming_edge.py
+++ b/tests/test_gemma4_streaming_edge.py
@@ -35,8 +35,8 @@ def test_thought_label_split_across_deltas():
     parser = Gemma4ReasoningParser()
     tokens = [
         "<|channel>",
-        "tho",        # partial "thought" label
-        "ught",       # completes the label
+        "tho",  # partial "thought" label
+        "ught",  # completes the label
         "\n",
         "real reasoning",
         "<channel|>",
--- a/tests/test_gemma4_tool_parser.py
+++ b/tests/test_gemma4_tool_parser.py
@@ -167,7 +167,6 @@ class TestGemma4ToolParserExtract:
         args = json.loads(result.tool_calls[0]["arguments"])
         assert args == {"text": 'line1\nline2 said "hello"'}
 
-
 class TestGemma4ToolParserStreaming:
     """Test streaming tool call extraction."""
 
@@ -237,4 +236,47 @@ class TestGemma4Registration:
 
     def test_native_format_false(self):
         assert Gemma4ToolParser.SUPPORTS_NATIVE_TOOL_FORMAT is False
+
+    def test_extra_stop_tokens_declares_tool_response(self):
+        """Gemma 4 treats <|tool_response> (id 50) as end-of-generation
+        after a tool call. The parser exposes it so the server can merge it
+        into the request's stop sequences.
+        Reference: llama.cpp PR #21418.
+        """
+        parser = Gemma4ToolParser()
+        assert "<|tool_response>" in parser.extra_stop_tokens
+
+    def test_abstract_parser_default_empty_stop_tokens(self):
+        """Other parsers that don't override keep an empty default."""
+        from vllm_mlx.tool_parsers.abstract_tool_parser import ToolParser
+
+        assert ToolParser.extra_stop_tokens == []
+
+    def test_merge_helper_adds_parser_extras(self):
+        """get_parser_stop_tokens adds the parser's EOG tokens on top of user stops."""
+        from vllm_mlx.tool_parsers import get_parser_stop_tokens
+
+        merged = get_parser_stop_tokens("gemma4", ["END"])
+        assert "END" in merged
+        assert "<|tool_response>" in merged
+
+    def test_merge_helper_dedupes(self):
+        """Parser extra tokens already present in user stops aren't duplicated."""
+        from vllm_mlx.tool_parsers import get_parser_stop_tokens
+
+        merged = get_parser_stop_tokens("gemma4", ["<|tool_response>"])
+        assert merged.count("<|tool_response>") == 1
+
+    def test_merge_helper_unknown_parser_is_passthrough(self):
+        """Unknown parser name leaves user stops untouched."""
+        from vllm_mlx.tool_parsers import get_parser_stop_tokens
+
+        assert get_parser_stop_tokens("nonexistent_parser_xyz", ["A"]) == ["A"]
+
+    def test_merge_helper_none_parser_is_passthrough(self):
+        """No parser name returns user stops as-is."""
+        from vllm_mlx.tool_parsers import get_parser_stop_tokens
+
+        assert get_parser_stop_tokens(None, ["A"]) == ["A"]
+        assert get_parser_stop_tokens(None, None) == []
         assert Gemma4ToolParser.supports_native_format() is False
--- a/tests/test_gemma4_tool_parser.py
+++ b/tests/test_gemma4_tool_parser.py
@@ -189,7 +189,9 @@ class TestGemma4ToolParserExtract:
 
     def test_bare_string_preserves_null_and_bool_literals(self):
         """null/true/false must NOT be treated as bare strings."""
-        output = "<|tool_call>call:cfg{flag:null,ready:true,done:false,name:bob}<tool_call|>"
+        output = (
+            "<|tool_call>call:cfg{flag:null,ready:true,done:false,name:bob}<tool_call|>"
+        )
         result = self.parser.extract_tool_calls(output)
         args = json.loads(result.tool_calls[0]["arguments"])
         assert args == {"flag": None, "ready": True, "done": False, "name": "bob"}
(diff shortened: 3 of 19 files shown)


waybarrios:
@Thump604 what can we do here? This is pretty huge change and there are some conflicts to resolve

Thump604:
I think there is valuable work here, but I would not try to merge this branch as one PR. It is currently conflicting, has no checks reported, and it mixes several independent risk surfaces in one diff: Gemma 4 parser fixes, MLLM message preservation, tool response adaptation, cache correctness, SSD cache, video routing, scheduler/threading, and cleanup. Even if each part is good, reviewing and conflict-resolving all of that together is too risky.

What I would do:

1. Rebase only as a salvage step, not as a merge target. Current `main` already absorbed some nearby work, including cache/threading/Qwen-serving fixes, so the first pass should identify what is still net-new versus already superseded.

2. Split into narrow PRs in dependency order:

- Gemma 4 parser correctness: multi-block reasoning re-entry, marker cleanup, parallel tool-call parsing, stable streaming tool-call IDs. Keep this to parser files and parser tests.
- Gemma 4 native tool-response adaptation: `prepare_messages`, role:tool coalescing, native tool format support, and the smallest server hooks needed to apply it. This needs rendered-template tests plus at least one tool-loop regression.
- MLLM message preservation: preserve assistant `tool_calls`, `tool_call_id`, dict/list content, and media ordering through `chat()` and `stream_chat()`. This should be model-agnostic and tested as such.
- Cache correctness: only the remaining cache ownership/trim/deep-copy fixes not already covered by current `main`. This needs invariant tests around LCP trim, source-entry immutability, quantized wrappers, rotating cache metadata, and exact-match reuse.
- Native video batching: separate PR. This is a different feature path with different failure modes.
- SSD cache tier for MLLM: separate PR. Storage policy, crash recovery, quantized serialization, and promotion behavior need their own review.
- Scheduler/thread ownership: separate only if there is still net-new behavior after current `main`. Some of this appears related to fixes that already landed, so I would compare carefully before keeping it.

3. Drop or fold pure logging/comment cleanup into the owning slice. I would not keep cleanup commits as part of the salvage branch unless they directly support a fix.

4. For each slice, I would expect:

- one clear failure mode in the PR body
- a minimal repro or failing regression test
- exact files touched and why
- CI green
- no broad performance claim without a small table showing hardware, model, prompt shape, cache state, TTFT, decode tok/s, and peak memory

My recommendation is: do not resolve conflicts on this PR and merge it. Mine it for the parts that are still valid, split those into reviewable PRs, and close this branch once the salvage stack exists. I am happy to review the split PRs or help identify what current `main` already superseded.

waybarrios:
@0xAlcibiades could you check this ⬆️  ?

0xAlcibiades:
Yeah, we can kill this PR, I more just did a "smash until it works" approach here. Feel free to lift the changes and fixes you see fit to take in.

Issue #394, opened 2026-04-22
No way to disable Qwen3 thinking mode server-wide


makinapower:
Running vllm-mlx with `mlx-community/Qwen3.6-35B-A3B-4bit-DWQ` and 
can't find a way to disable thinking mode server-wide.

vllm has `--default-chat-template-kwargs` for this but vllm-mlx doesn't 
seem to expose it.

Passing `chat_template_kwargs` per-request already works though:

  "chat_template_kwargs": {"enable_thinking": false}

Any chance this could be added as a server-level default?

Thump604:
This is now covered by merged PR #396. The server accepts `--default-chat-template-kwargs` and passes those kwargs into chat-template rendering, so Qwen thinking can be disabled server-wide with `--default-chat-template-kwargs '{"enable_thinking": false}'`.

Merged commit: ff49160a1074b7d1d7802d92dd919ad29d6688f9
Post-merge main CI: https://github.com/waybarrios/vllm-mlx/actions/runs/24889027884

Closing this as implemented.

PR #396, opened 2026-04-23
Support --default-chat-template-kwargs '{"enable_thinking": false}' 

perry2of5:
Address issue #394


perry2of5:
rebased and addressed merge conflicts. Will follow up and fix tests.

On a side note: I copied the approach from vllm and implemented it. However, I feel like it might be better to have a more generic strategy for setting defaults on requests. Possibly there could be a way to set the defaults by endpoint like --default-<endpoint>-template or something. I'm not sure it was wise of me to apply this to all methods instead of just the chat completions. Open to some discussion on this.

perry2of5:
I'm going to have to rethink this. This PR is too big for what it provides.

Thump604 (commented):
I like the direction. Server-wide `chat_template_kwargs` is the right control surface for issue #394, and the request-over-default merge behavior is useful.

I would clean up one thing before merge: the batching deterministic tolerance change looks unrelated to this PR.

```python
assert batch_throughput > seq_throughput * 0.8
```

This PR is about chat template defaults, not batching performance. Relaxing a throughput assertion in the same diff makes the review harder and weakens a test while the project is actively tightening performance qualification. I would revert that hunk and keep this PR scoped to the server default plumbing, endpoint propagation, docs, and tests for the new flag.

The refactor into `PreparedChatInvocation` is a bigger change than the issue strictly requires, but it does remove duplicated preparation logic across streaming/non-streaming chat and Anthropic paths. I am okay with that shape if the unrelated benchmark tolerance hunk is dropped and CI stays green.

perry2of5:
Thanks, @Thump604.

I reverted that change to the test, squished it down to one commit, rebased on main, and verified tests work.

Is this change too big to bundle in one PR? I think the overall changes are reasonable, but it might make all the outstanding PRs break.... Thoughts @waybarrios ?

Thump604 (approved):
I rechecked this after the squash, rebase, and removal of the unrelated batching tolerance change. I think this is acceptable as one PR now.\n\nIt is large by line count, but the current diff is cohesive: server-wide chat template kwargs defaults, request-over-default merge behavior, propagation through OpenAI chat, Anthropic, and Responses paths, plus tests/docs for that surface. Splitting it further would reduce file count per PR, but it would also make the shared preparation refactor harder to reason about across endpoints.\n\nI also ran the focused server tests locally through my runtime venv:\n\n\n\nCI is green and the PR is mergeable. I would merge this if Wayner is comfortable with the endpoint-wide scope.

Thump604:
Small correction to my approval body: the command block was swallowed by my shell when I submitted the review. The focused local check I ran was:

```bash
PYTHONPATH=/tmp/vllm-mlx-pr396 /opt/ai-runtime/venv-live/bin/python -m pytest -q \
  tests/test_chat_template_kwargs.py tests/test_responses_api.py tests/test_server.py
# 126 passed, 3 deselected
```

PR #397, opened 2026-04-23
fix(simple): preserve mtp after successful specprefill

### LINKED ISSUE #398: RuntimeError: There is no Stream(gpu, N) in current thread — engine dispatches scheduler.step() to a thread that doesn't own MLX streams
This was generated with Claude Code Opus 4.7 - I'm not a coder by trade but hopefully this helps someone else

## Summary

On Apple Silicon, every request to `vllm-mlx serve` fails at the first `mx.eval(...)` call inside `mlx_lm/generate.py` with:

```
```

(or `Stream(gpu, 0)` depending on which stream is active). The cause is that MLX streams are **thread-local** — the handle is only valid on the thread that created it — but `vllm_mlx/engine_core.py` dispatches `scheduler.step()` to a dedicated worker thread (`mlx-step`) via `loop.run_in_executor(_executor, ...)`. The streams used by `mlx_lm.generate` were created on whichever thread first imported the module (the main event-loop thread, during model load), so the executor thread cannot use them.

Forcing `scheduler.step()` to run inline on the event-loop thread fixes every request. The executor path appears to be fundamentally broken in combination with MLX's thread-local streams.

## Environment

- macOS (Apple Silicon)
- Python 3.12
- `vllm-mlx == 0.2.9`
- `mlx-lm == 0.31.3`
- `mlx == 0.31.2`

## Reproduction

```bash
pip install vllm-mlx
vllm-mlx serve mlx-community/Llama-3.2-3B-Instruct-4bit --port 8000 --continuous-batching
```

In another shell:

```bash
curl -s -X POST http://localhost:8000/v1/chat/completions \
  -d '{"model":"mlx-community/Llama-3.2-3B-Instruct-4bit","messages":[{"role":"user","content":"hi"}],"max_tokens":10}'
```

Response (200 OK but content is null, `finish_reason: "error"`):

```json
{"choices":[{"message":{"content":null,...},"finish_reason":"error"}],
 "usage":{"prompt_tokens":0,"completion_tokens":0,"total_tokens":0}}
```

Server log:

```
ERROR:vllm_mlx.scheduler:Error in batch generation step: There is no Stream(gpu, 1) in current thread.
  File ".../vllm_mlx/scheduler.py", line 2455, in step
  File ".../mlx_lm/generate.py", line 1855, in next
    return self._next()
  File ".../mlx_lm/generate.py", line 1841, in _next
    self._prompt_batch.prompt(prompts)
  File ".../mlx_lm/generate.py", line 1161, in prompt
    mx.eval([c.state for c in self.prompt_cache])
```

Same failure occurs with or without `--continuous-batching`, with `--max-num-seqs 1`, and with various models (Llama-3.1-8B-4bit, Llama-3.2-3B-4bit).

## Root cause

`mlx_lm/generate.py:226` creates a module-level stream that is explicitly thread-bound:

```python
generation_stream = mx.new_thread_local_stream(mx.default_device())
```

This runs once, on whichever thread first imports the module — the main thread during vllm-mlx startup/model load.

`vllm_mlx/engine_core.py:149` creates a dedicated worker thread:

```python
_executor = concurrent.futures.ThreadPoolExecutor(
    max_workers=1, thread_name_prefix="mlx-step"
)
```

and dispatches `scheduler.step()` to it:

```python
needs_executor = has_waiting or has_partial
if needs_executor:
    output = await loop.run_in_executor(_executor, self.scheduler.step)
else:
    output = self.scheduler.step()
```

When `needs_executor=True`, step() runs on `mlx-step` — a thread that has no access to the streams created on the main thread. Every MLX op inside `mlx_lm.generate` (which uses `with mx.stream(generation_stream): ...`) then raises `RuntimeError: There is no Stream(gpu, N) in current thread`.

Swapping `new_thread_local_stream` for `default_stream` in `mlx_lm/generate.py` does **not** fix it — in this MLX version even `Stream(gpu, 0)` is thread-local (the error just changes from index 1 to index 0). The problem is cross-thread handle usage, regardless of which stream.

## Workaround

Forcing step to run inline on the event-loop thread (the thread that owns the streams) makes every request succeed:

```diff
# vllm_mlx/engine_core.py (around line 186)
- needs_executor = has_waiting or has_partial
+ needs_executor = False
```

With this one-line change, identical `curl` → `200 OK, 8 tokens in 0.81s, 9.9 tok/s`. No other changes required (no `--max-num-seqs 1`, no launcher wrapper, no mlx-lm patch).

## Suggested fix

The executor-thread dispatch appears incompatible with MLX's thread-local streams as currently used by `mlx_lm.generate`. A few options:

1. **Drop the executor and run step() inline** (what the workaround does). Simplest. Downside: blocks the event loop during each step — probably fine since MLX is already the bottleneck and `max_num_seqs` is usually small.
2. **Load the model and import `mlx_lm.generate` on the executor thread** at engine startup, so the streams it creates live on `mlx-step`. Then all MLX work has to stay on that thread (the event loop must not touch MLX).
3. **Re-create streams on the executor thread** on first step, and ensure `mlx_lm.generate.generation_stream` is rebound to that stream before any inference runs. Brittle.

Option 1 is a single-line fix and what I'd suggest starting with.

## Happy to send a PR

If you'd like, I can open a PR that flips the dispatch to inline by default, with a flag to re-enable the executor for anyone whose setup works with it.

Thump604:
## Summary
- preserve text-route decoding controls after successful SimpleEngine SpecPrefill
- keep native MTP only on paths where request-local processors are speculation-safe or retired
- resume MTP after thinking-budget processor retirement through a shared continuation helper
- bind MLX generation streams inside the scheduler worker thread to fix cross-thread stream ownership failures

## Problem
Successful SimpleEngine SpecPrefill was not composing with the normal decode contract. The continuation could silently drop native MTP and request-local logits processors, making SpecPrefill+MTP measurements misleading and risking constrained-decoding violations.

Separately, continuous batching could run prefill on an executor thread and decode on the event-loop thread. MLX generation streams are thread-local, so that split can fail with `RuntimeError: There is no Stream(gpu, N) in current thread`.

## Fix
- seed request-local logits processors with the original prompt context before continuation
- sample the first post-prefill token through active processors
- continue through `mlx_lm.stream_generate()` with the backbone cache and MTP cache only when allowed
- fail closed for native MTP with active unsafe processors
- share processor-retirement resume logic between normal and SpecPrefill simple paths
- run continuous-batching and MLLM scheduler steps on one MLX worker thread and bind mlx-lm/mlx-vlm generation streams in that thread

## Validation
- `python -m pytest -q tests/test_engine_core_thread_streams.py tests/test_simple_engine.py tests/test_mllm_continuous_batching.py tests/test_batched_engine.py tests/test_llm.py tests/test_server_cache_controls.py`
- result: `66 passed, 9 deselected`
- `python -m black --check ...`
- result: `16 files would be left unchanged`
- GitHub Actions run `24860026718`: lint, type-check, test-matrix 3.10/3.11/3.12/3.13, Apple Silicon 3.11/3.13, and aggregate tests all passed

Fixes #398.

### DIFF
--- a/tests/test_batched_engine.py
+++ b/tests/test_batched_engine.py
@@ -92,3 +92,35 @@ class TestBatchedEngineGenerate:
         assert result.prompt_tokens == 7
         assert result.completion_tokens == 1
         assert result.finish_reason == "stop"
+
+
+class TestBatchedEngineCacheRestore:
+    def _make_mllm_engine(self):
+        from vllm_mlx.engine.batched import BatchedEngine
+
+        with patch("vllm_mlx.engine.batched.is_mllm_model", return_value=True):
+            engine = BatchedEngine("test-mllm")
+
+        engine._loaded = True
+        engine._is_mllm = True
+        return engine
+
+    def test_load_cache_from_disk_bootstraps_mllm_batch_generator(self):
+        engine = self._make_mllm_engine()
+
+        prefix_cache = MagicMock()
+        prefix_cache.load_from_disk.return_value = 2
+        scheduler = MagicMock()
+        scheduler.batch_generator = None
+
+        def ensure_batch_generator():
+            scheduler.batch_generator = MagicMock(prefix_cache=prefix_cache)
+
+        scheduler._ensure_batch_generator.side_effect = ensure_batch_generator
+        engine._mllm_scheduler = scheduler
+
+        loaded = engine.load_cache_from_disk("/tmp/cache")
+
+        assert loaded == 2
+        scheduler._ensure_batch_generator.assert_called_once_with()
+        prefix_cache.load_from_disk.assert_called_once_with("/tmp/cache")
--- a/tests/test_engine_core_thread_streams.py
+++ b/tests/test_engine_core_thread_streams.py
@@ -1,0 +1,112 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Regression coverage for MLX stream/thread ownership in engine loops."""
+
+import asyncio
+import threading
+from types import SimpleNamespace
+
+import pytest
+
+
+class _SchedulerOutput:
+    outputs = []
+    finished_request_ids = []
+
+
+@pytest.mark.anyio
+async def test_engine_core_runs_all_scheduler_steps_on_one_worker_thread(monkeypatch):
+    """Continuous batching must not bounce MLX steps across threads."""
+    from vllm_mlx.engine_core import EngineConfig, EngineCore
+
+    engine = object.__new__(EngineCore)
+    engine.config = EngineConfig(step_interval=0, stream_interval=1)
+    engine._running = True
+    engine._steps_executed = 0
+    engine._output_collectors = {}
+    engine._stream_states = {}
+    engine._finished_events = {}
+
+    main_thread = threading.get_ident()
+    bind_threads: list[int] = []
+
+    class FakeScheduler:
+        batch_generator = SimpleNamespace(_partial=None)
+
+        def __init__(self):
+            self.calls = 0
+            self.step_threads: list[int] = []
+            self.close_threads: list[int] = []
+
+        def has_requests(self):
+            return self.calls < 3
+
+        def step(self):
+            self.step_threads.append(threading.get_ident())
+            self.calls += 1
+            if self.calls == 3:
+                engine._running = False
+            return _SchedulerOutput()
+
+        def _close_batch_generator(self):
+            self.close_threads.append(threading.get_ident())
+
+    scheduler = FakeScheduler()
+    engine.scheduler = scheduler
+
+    def bind_streams():
+        bind_threads.append(threading.get_ident())
+
+    monkeypatch.setattr("vllm_mlx.engine_core.bind_generation_streams", bind_streams)
+
+    await asyncio.wait_for(engine._engine_loop(), timeout=2)
+
+    assert scheduler.step_threads
+    assert len(set(scheduler.step_threads)) == 1
+    assert scheduler.step_threads[0] != main_thread
+    assert bind_threads == [scheduler.step_threads[0]]
+    assert scheduler.close_threads == [scheduler.step_threads[0]]
+
+
+@pytest.mark.anyio
+async def test_mllm_scheduler_runs_all_steps_on_one_worker_thread(monkeypatch):
+    """MLLM scheduler loop follows the same thread-local stream invariant."""
+    from vllm_mlx.mllm_scheduler import MLLMScheduler
+
+    scheduler = object.__new__(MLLMScheduler)
+    scheduler._running = True
+
+    main_thread = threading.get_ident()
+    bind_threads: list[int] = []
+    step_threads: list[int] = []
+    close_threads: list[int] = []
+
+    class FakeBatchGenerator:
+        _partial = None
+
+        def close(self):
+            close_threads.append(threading.get_ident())
+
+    scheduler.batch_generator = FakeBatchGenerator()
+
+    def has_requests():
+        return len(step_threads) < 3
+
+    def step():
+        step_threads.append(threading.get_ident())
+        if len(step_threads) == 3:
+            scheduler._running = False
+
+    def bind_streams():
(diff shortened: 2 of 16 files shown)


janhilgard (commented):
## Review: fix(simple): preserve mtp after successful specprefill

Solid PR — the core fix (preserving MTP after successful SpecPrefill) is clearly correct and well-motivated. Text-only routing for MLLM even without MTP is a nice bonus. Test coverage is excellent (557 lines in test_simple_engine.py alone).

### Architecture concerns

**1. Non-greedy sampling guard in `_mtp_step` may be too aggressive**

```python
has_non_greedy_sampling = any(
    getattr(req, "temperature", 0.0) not in (0, 0.0)
    or getattr(req, "top_p", 1.0) < 1.0
    ...
    for req in active_requests
)
```

This disables MTP for **all** temperature-sampled requests in the batched path. In practice, many requests use `temperature=0.6` / `top_p=0.95` (thinking models). The always-advance verifier is _correctness-safe_ for non-greedy — if the draft doesn't match the sampled token, we just use the sampled token (advance by 1 instead of 2). Acceptance rate drops, but the result is correct.

The guard makes MTP a greedy-only feature in the batched path, which is a **significant throughput regression** for production workloads using non-zero temperature. Consider limiting the guard to cases where acceptance would be provably zero (e.g., very high temperature), or document it as a known regression.

**2. Duplicated retirement flow (~100 lines)**

The processor retirement logic (break out of stream_generate → trim cache → start MTP-enabled continuation) is copy-pasted between `_run_all()` and `_run_specprefill()`. Both paths share identical:
- `_processors_retired()` check → break
- `trim_prompt_cache(bc, 1)` → seed construction
- `resume_kwargs` with MTP re-enable
- Second `mlx_stream_generate()` call

Extracting this into a `_resume_with_mtp_after_retirement(model, bc, ...)` helper would reduce ~50 lines of duplication and make both paths more maintainable.

**3. `_processors_can_retire` / `_processors_retired` duplicated**

These identical helpers exist in both `simple.py` and `mllm_batch_generator.py`. Consider a shared module (e.g., `vllm_mlx/utils/processors.py`).

### Correctness questions

**4. `logits_processors` guard in `_mtp_step` — interaction with `is_retired`**

```python
or (logits_processors is not None and any(logits_processors))
```

After `_drop_retired_processors()` sets a slot to `None`, this `any()` check would still trigger if other non-retired processors are present. But when ALL processors are retired and `logits_processors == [None]`, `any([None])` is False — so MTP re-enables on the next step. Is this the intended behavior? The slot becomes `None` but the list still has length > 0. Worth a comment clarifying the None-slot convention.

**5. `trim_prompt_cache(bc, 1)` in retirement flow**

When breaking out of `stream_generate` after the processor retires, the cache contains KV up through the last yielded token. The trim removes that last position, then `stream_generate(prompt=seed)` re-prefills it. This is correct for backbone cache, but what about MTP cache state from the first stream_generate call? The first call may have populated MTP state that now has stale entries. Creating a fresh `make_mtp_cache()` should handle this, but worth a comment confirming that fresh MTP cache is intentional (not just "MTP was off in phase 1").

### Minor items

**6. `prefill_step_size` fallback in `_start_mllm()`**

```python
if prefill_step_size is None:
    prefill_step_size = getattr(self._scheduler_config, "prefill_step_size", None)
```

Previously `prefill_step_size` was only forwarded when explicitly set via `mllm_prefill_step_size`. Now it falls back to `prefill_step_size`, which could change default MLLM prefill behavior for users who set `prefill_step_size` for the LLM scheduler but didn't intend it to affect MLLM. Worth documenting this fallback behavior.

**7. `_bind_worker_generation_streams()` per-call overhead**

Creating a new `mx.Stream` on every `_run_blocking_serialized()` call adds per-request overhead. Since `asyncio.to_thread()` may reuse threads from the pool, could this leak streams over time? Not a blocker, but worth investigating if we see memory growth.

### Summary

The fix is correct and well-tested. Main concerns are the non-greedy MTP guard (item 1) and code duplication (items 2-3). Items 4-7 are minor or informational.

PR #400, opened 2026-04-23
fix: apply RMSNorm offset fixup to MTP weights

janhilgard:
## Summary

- Fix 0% MTP acceptance rate on dense Qwen3.5/3.6 models by applying the HuggingFace RMSNorm weight offset convention (`actual_gamma = 1 + stored_weight`) to MTP weights
- Guard `scales_key` lookup to prevent false positive dequantization on non-quantized (BF16) MTP weights
- Convert fused expert format (`experts.gate_up_proj`) to split `switch_mlp` format for MoE MTP layers (Qwen3.6 compatibility)

## Root Cause

Qwen3.5/3.6 models store RMSNorm weights as **offsets** (`actual_gamma = 1 + stored_weight`). The main model's `sanitize()` in mlx_vlm adds +1.0 automatically, but MTP weights loaded via `inject_mtp_support()` bypass `sanitize()` entirely.

This caused MTP norm layers to multiply by near-zero or negative values, producing near-uniform logits (std ~1.0) and 0% acceptance rate.

**MoE models (35B-A3B, 122B-A10B) were unaffected** because their MTP checkpoints don't include norm weights (they default to 1.0). **Dense models (27B) are affected** — all 7 norm weights had raw HF offset values (e.g., `pre_fc_norm_embedding.weight` mean = -0.44 instead of +0.56).

## Benchmark (Qwen3.6-27B dense, mixed 6/8-bit)

| Metric | Before Fix | After Fix |
|--------|-----------|-----------|
| MTP acceptance | 0% | 76-88% |
| Throughput (counting) | 9.5 tok/s | 24 tok/s |
| Throughput (creative) | 9.5 tok/s | 22.9 tok/s |
| Baseline (no MTP) | 22 tok/s | 22 tok/s |

### 1. RMSNorm +1.0 offset fixup
Apply +1.0 to all 1D MTP weights (which are all RMSNorm gammas) after loading, matching the convention used by `sanitize()` for the main model.

### 2. scales_key guard
Add `scales_key != key` check to prevent treating non-quantized weight keys (where `key.replace(".weight", ".scales")` equals `key` for keys without `.weight` suffix) as quantized triplets.

### 3. Fused expert format conversion
Qwen3.6 MTP uses fused expert keys (`experts.gate_up_proj` with shape `[n_experts, 2*intermediate, hidden]`). Split these into the `switch_mlp.{gate_proj,up_proj}` format expected by mlx_lm's `DecoderLayer`.

## Test plan

- [x] Qwen3.6-27B (dense): MTP acceptance 76-88%, throughput restored
- [x] Thinking parsing (streaming + non-streaming): reasoning_content correctly separated
- [x] Tool call parsing (single, multiple, complex schemas): all working
- [x] Chunked prefill: compatible with MTP

### DIFF
--- a/tests/test_qwen35_mtp_patch.py
+++ b/tests/test_qwen35_mtp_patch.py
@@ -1,0 +1,47 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Regression tests for Qwen3.5/3.6 MTP weight fixups."""
+
+import mlx.core as mx
+
+from vllm_mlx.patches.qwen3_5_mtp import (
+    _apply_qwen_mtp_rmsnorm_offset_fixups,
+)
+
+
+def test_qwen_mtp_raw_offset_norm_weights_shift_once():
+    weights = {
+        "pre_fc_norm_hidden.weight": mx.array([-0.4, 0.0], dtype=mx.float32),
+        "layers.0.input_layernorm.weight": mx.array([-0.2, 0.1], dtype=mx.float32),
+    }
+
+    shifted = _apply_qwen_mtp_rmsnorm_offset_fixups(weights)
+
+    assert shifted == 2
+    assert mx.allclose(
+        weights["pre_fc_norm_hidden.weight"],
+        mx.array([0.6, 1.0], dtype=mx.float32),
+    )
+    assert mx.allclose(
+        weights["layers.0.input_layernorm.weight"],
+        mx.array([0.8, 1.1], dtype=mx.float32),
+    )
+
+
+def test_qwen_mtp_actual_gamma_norm_weights_are_not_shifted_again():
+    original = mx.array([0.56, 0.82], dtype=mx.float32)
+    weights = {"pre_fc_norm_embedding.weight": original}
+
+    shifted = _apply_qwen_mtp_rmsnorm_offset_fixups(weights)
+
+    assert shifted == 0
+    assert mx.allclose(weights["pre_fc_norm_embedding.weight"], original)
+
+
+def test_qwen_mtp_non_norm_one_dimensional_weights_are_not_shifted():
+    original = mx.array([-0.4, 0.0], dtype=mx.float32)
+    weights = {"layers.0.mlp.shared_expert_gate.weight": original}
+
+    shifted = _apply_qwen_mtp_rmsnorm_offset_fixups(weights)
+
+    assert shifted == 0
+    assert mx.allclose(weights["layers.0.mlp.shared_expert_gate.weight"], original)
--- a/vllm_mlx/patches/qwen3_5_mtp.py
+++ b/vllm_mlx/patches/qwen3_5_mtp.py
@@ -217,7 +217,7 @@ def inject_mtp_support(model: Any, model_path, config: dict) -> bool:
         scales_key = key.replace(".weight", ".scales")
         biases_key = key.replace(".weight", ".biases")
 
-        if scales_key in raw_mtp and biases_key in raw_mtp:
+        if scales_key != key and scales_key in raw_mtp and biases_key in raw_mtp:
             # Quantized triplet → dequantize to BF16
             dq = mx.dequantize(
                 raw_mtp[key],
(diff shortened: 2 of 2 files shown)


waybarrios:
@janhilgard take a look on this: https://github.com/bstnxbt/dflash-mlx if you use it you can ack him :) It could be useful

waybarrios:
Could we review this: https://github.com/ml-explore/mlx-lm/discussions/890 and get some inspiration in order to improve the MTP process?

waybarrios:
@janhilgard could you put here the huggingface link of the model you use?

Thump604 (changes requested):
I think the root cause is real, but I do not think this should merge as written.

Blocking issue: the runtime injector now unconditionally adds `+1.0` to every 1D MTP tensor. That is only correct for raw HF offset weights. It is not correct for already-converted MLX MTP bundles produced by the repo scripts.

The existing extraction paths already apply the RMSNorm shift before writing `model-mtp.safetensors` or `mtp/weights.safetensors`:

- `scripts/add_mtp_weights_qwen35.py` lines 180-210 define the norm suffixes and add `+1.0`
- `scripts/add_mtp_weights.py` lines 139-144 does the same for the older path

I checked the local Qwen3.6-27B MLX MTP bundle that was produced through that pipeline. Its norm tensors are already in actual-gamma space, not raw HF offset space:

```text
mtp.pre_fc_norm_embedding.weight mean=0.5600 min=0.2617 max=0.8281
mtp.pre_fc_norm_hidden.weight    mean=0.8289 min=0.6211 max=1.4141
mtp.norm.weight                  mean=2.2740 min=0.7969 max=3.0000
```

If the injector applies this PR on top of that bundle, those become `1.5600`, `1.8289`, `3.2740`, etc. That looks like a double-shift and would silently corrupt a valid converted artifact.

I think the fix needs one of these contracts before merge:

1. Keep the shift exclusively in the extraction/conversion scripts and do not apply it again at runtime.
2. If runtime must support both raw HF-style MTP files and already-converted MLX MTP files, add explicit convention detection before shifting. It should be key-suffix based, not `ndim == 1`, and it should skip weights that are already in actual-gamma space.
3. Add regression tests for both cases: raw offset norm weights should be shifted once, already-shifted converted weights should remain unchanged.

The `scales_key != key` guard and fused expert remap both look directionally correct. My objection is specifically the unconditional runtime `+1.0`, because it conflicts with the current repo conversion contract and the existing generated artifacts.

Thump604:
I looked at both references Wayner linked.

`bstnxbt/dflash-mlx` looks useful as an architecture reference, but it is solving a different layer than this PR. The speedups come from changing the verifier shape: block-diffusion draft tokens, one target verify pass over a block, recurrent rollback for GatedDeltaNet state, and a verify-specialized M=16 int4 matmul path. That is a speculative decoding architecture, not a reason to shift norm tensors at runtime.

The MLX-LM discussion points in the same direction. The prototype there only showed a small gain until cache layout, `position_ids`, and tree or block verification are addressed. That matches what I see locally: simple MTP toggles are not enough on Apple Silicon when the request contract includes thinking/parser constraints and long outputs.

My recommendation is to keep this PR scoped to the MTP weight convention fix. The safe contract should be one of these:

1. Extraction writes actual-gamma norm weights and runtime does not shift them again.
2. Runtime supports raw HF-style MTP files too, but only with explicit convention detection and tests for raw-offset and already-shifted bundles.

DFlash/EAGLE-style work is worth a separate design PR with acceptance metrics, rollback/cache invariants, and quality gates. I would not mix that with the RMSNorm offset fix.

janhilgard (commented):
@Thump604 You're right — the unconditional runtime `+1.0` would double-shift weights produced by the repo extraction scripts. Our production Qwen3.6-27B MTP weights were extracted manually (direct `mx.load` from the HF BF16 model) rather than through `add_mtp_weights_qwen35.py`, so they were still in raw HF offset space. That's why the runtime fixup appeared to work in our setup.

Updated approach:
- **Removed** the unconditional runtime `+1.0` shift entirely
- **Added** a diagnostic warning that fires when any 1D norm weight has mean < 0.5 (indicating raw HF offset space), pointing the user to re-run the extraction script
- **Contract**: MTP weights MUST be in actual-gamma space (extraction scripts handle the `+1.0` shift at conversion time, runtime does not touch them)
- **Fixed** our production weights separately (one-time `+1.0` applied to the stored safetensors)

The scales_key guard and fused expert conversion are unchanged.

Agree that DFlash/EAGLE-style work belongs in a separate design PR.

janhilgard:
@waybarrios

**Re: dflash-mlx** — Thanks for the reference! Block-diffusion drafting with a verify-specialized matmul path is an interesting architecture, but as Thump604 noted it's solving a different layer than this PR. Worth exploring in a separate design PR with acceptance metrics.

**Re: mlx-lm #890 (EAGLE-3)** — Reviewed it. The prototype shows only ~1.05x speedup on Apple Silicon, which aligns with what we've seen — simple MTP toggles hit diminishing returns when cache layout and tree verification aren't addressed. Our current MTP approach (79-85% acceptance) is more practical for Apple Silicon. Worth revisiting if the EAGLE implementation matures.

**Re: HuggingFace model link** — The BF16 source for MTP weight extraction is [Qwen/Qwen3.5-27B](https://huggingface.co/Qwen/Qwen3.5-27B) (same architecture for Qwen3.6-27B MTP weights). The quantized model is a local mixed 6/8-bit conversion.

janhilgard:
Updated the approach based on testing:

**Problem with warn-only**: Removing the runtime shift entirely broke acceptance for MTP weights extracted directly from HuggingFace BF16 models (not through the repo scripts). These weights are in raw HF offset space and *need* the +1.0 at load time.

**New approach — conditional detection**: Instead of unconditionally shifting ALL 1D weights, the code now checks `mean < 0.5` on each norm weight to distinguish:
- **Raw HF offset** (mean ≈ -0.44 to 0.04): apply +1.0 → correct
- **Already converted** (mean ≈ 0.56 to 2.28): skip → no double-shift

This handles both contracts:
1. Weights from `add_mtp_weights_qwen35.py` (pre-shifted) → no runtime shift
2. Weights extracted directly from HF BF16 (raw offset) → runtime shifts only the ones that need it

Verified: 85-90% acceptance on Qwen3.6-27B with raw HF MTP weights.

Thump604 (approved):
I re-reviewed after the follow-up commit `58f2b2e`.

The previous blocker is addressed:

- runtime no longer shifts every 1D MTP tensor
- fixup is limited to known Qwen MTP RMSNorm keys
- raw HF offset norm weights shift once
- already converted actual-gamma weights are left unchanged
- non-norm 1D weights are protected from false-positive shifts

Local validation:

```text
pytest -q tests/test_qwen35_mtp_patch.py
pytest -q tests/test_qwen35_mtp_patch.py tests/test_qwen35_mllm_patch.py
black --check vllm_mlx/patches/qwen3_5_mtp.py tests/test_qwen35_mtp_patch.py
ruff check vllm_mlx/patches/qwen3_5_mtp.py tests/test_qwen35_mtp_patch.py --select E,F,W --ignore E402,E501,E731,F811,F841
git diff --check
```

CI is green on run `24888725105`.

This keeps the PR scoped to MTP weight convention handling. DFlash, EAGLE, and tree verification should stay separate.

PR #405, opened 2026-04-24
Fix sampling defaults and short prefix-cache reuse

Thump604:
## Summary
- propagate extended sampling defaults (`top_k`, `min_p`, presence penalty, repetition penalty) through server and CLI paths instead of silently falling back to hardcoded values
- preserve explicit request values such as `top_k=0` or `presence_penalty=0.0` as overrides
- reject/store/load prefix-cache entries shorter than 128 tokens to avoid expensive and unsafe short prompt reuse

## Tests
- `python -m pytest tests/test_cli.py tests/test_server.py tests/test_memory_cache.py tests/test_kv_cache_quantization.py -q`

## Local evidence
- Runtime lane launched with `--default-top-k 20` now reaches MLLM as `top_k=20`
- Existing persisted 15-20 token cache entries are skipped on startup and resident cache starts at `entry_count=0`

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -4,6 +4,7 @@ from types import SimpleNamespace
 def _serve_args(**overrides):
     args = {
         "api_key": None,
+        "auto_unload_idle_seconds": 0.0,
         "cache_memory_mb": None,
         "cache_memory_percent": 0.2,
         "chunked_prefill_tokens": 0,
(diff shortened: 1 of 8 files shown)


Thump604:
I could not refresh the fork branch after upstream CI workflow changes because the current OAuth token cannot update workflow files on the fork. I pushed the rebased branch to the upstream repo and opened #418 with the same scope, current main base, and fresh validation. Closing this one to avoid keeping a dirty duplicate in the review queue.

PR #408, opened 2026-04-24
fix: add warm-up to batching performance test

sjswerdloff:
The test_batched_faster_than_sequential test intermittently fails on loaded machines because the first run of each path pays one-time Metal kernel compilation overhead. When the batched path happens to compile during timing, measured throughput drops well below the 0.9x threshold (observed: 0.39x on an M3 Ultra under load).

Add untimed warm-up runs for both sequential and batched paths before the measured runs, ensuring kernel compilation and cache priming are excluded from the throughput comparison.

Also fix pre-existing B007 (unused loop variable `run` → `_run`).

### DIFF
--- a/tests/test_batching_deterministic.py
+++ b/tests/test_batching_deterministic.py
@@ -7,8 +7,9 @@ Run with: pytest tests/test_batching_deterministic.py -v
 """
 
 import asyncio
-import pytest
 import time
+
+import pytest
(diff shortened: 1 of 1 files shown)


perry2of5 (approved):
We need this. Changes make sense. Looks clean. Runs successfully.

PR #410, opened 2026-04-24
Update .pre-commit-config.yaml and pyproject.toml to match ci.yml

### LINKED ISSUE #239: CI lint uses black but pre-commit config uses ruff-format
## Description

The CI workflow (`.github/workflows/*.yml`) runs `black --check` for formatting validation, but `.pre-commit-config.yaml` uses `ruff-format` (via `ruff-pre-commit`).

While ruff-format is *nearly* black-compatible, there are edge cases where they disagree. This means:
- Code can pass pre-commit locally but fail CI
- Contributors need to run both `black` and `ruff format` to be safe

## Options

1. **Switch CI to ruff-format** — align with pre-commit config, drop black dependency
2. **Switch pre-commit to black** — align with CI, add `mirrors-black` hook
3. **Run both in CI** — belt and suspenders (current state, but pre-commit only runs ruff)

No preference on direction — just flagging the mismatch.

Found while fixing lint failures on PR #232.

perry2of5:
Update .pre-commit-config.yaml and pyproject.toml to use ruff and black like ci.yml

Fixes #239

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,6 @@ repos:
     hooks:
       - id: ruff
         args: [--select, E,F,W, --ignore, E402,E501,E731,F811,F841]
-      - id: ruff-format
 
   - repo: https://github.com/psf/black
     rev: 24.1.1
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,8 +13,15 @@ repos:
     rev: v0.1.9
     hooks:
       - id: ruff
-        args: [--fix, --exit-non-zero-on-fix]
+        args: [--select, E,F,W, --ignore, E402,E501,E731,F811,F841]
       - id: ruff-format
+
+  - repo: https://github.com/psf/black
+    rev: 24.1.1
+    hooks:
+      - id: black
+        args: [--check]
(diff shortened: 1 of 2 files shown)


perry2of5:
On my latest commit, I'm getting a ton of issues from mypy. I'm guessing people generally aren't using pre-commit....

sjswerdloff (approved):
This will really help to minimise the black failures in CI.

sjswerdloff:
this would be so nice to have merged.
I'm constantly having to ignore mypy to make a commit (on code I didn't change).
I'll set aside some time to grind through those mypy issues, but that's straight forward and serves a purpose.
The conflicts between pre-commit and CI are just friction without purpose.
Thank you @perry2of5

Thump604 (approved):
This is clean now. I pushed 370d0ea to remove the extra ruff-format hook so the local pre-commit stack matches CI instead of adding a second formatter path.

The refreshed CI run is green: https://github.com/waybarrios/vllm-mlx/actions/runs/24889142476

The resulting contract is straightforward: ruff check handles lint/imports and black remains the formatter gate, matching ci.yml.

perry2of5:
Thanks for that final cleanup, @Thump604

PR #413, opened 2026-04-24
fix: eval all cache types in MLLM chunked prefill to prevent OOM

sjswerdloff:
Hybrid models like Qwen3.5 use both ArraysCache (GatedDeltaNet linear attention with .state) and KVCache (full MLA attention with .keys/.values).

The existing `mx.eval([c.state for c in cache])` only evaluates ArraysCache layers. KVCache layers silently pass through because they have no .state attribute, letting the lazy computation graph accumulate across chunks. On long prompts (96K+ tokens), this causes OOM as the graph grows unbounded.

Fix: check for both cache types between chunks:
- KVCache: eval .keys and .values
- ArraysCache: eval .state entries

Affects both _run_chunked_text_prefill (text-only long prompts) and the prefix-cache partial-hit path (cached prefix + remaining tokens).

### DIFF
--- a/tests/test_mllm_continuous_batching.py
+++ b/tests/test_mllm_continuous_batching.py
@@ -37,6 +37,76 @@ pytestmark = pytest.mark.skipif(not HAS_MLX, reason="MLX not available")
 TEST_IMAGE_B64 = "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNk+M9QDwADhgGAWjR9awAAAABJRU5ErkJggg=="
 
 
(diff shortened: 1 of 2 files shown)


Thump604 (approved):
I pushed 5205267 to keep the hybrid-cache OOM fix but remove the duplicated cache-eval logic and add focused regression coverage for KVCache plus ArraysCache style state.

Validation is green on the refreshed CI run: https://github.com/waybarrios/vllm-mlx/actions/runs/24889491289

This is the right safety fix for Qwen hybrid long-prefill paths: evaluate both key/value attention cache tensors and recurrent state tensors between chunks so the lazy graph does not grow unbounded.

PR #414, opened 2026-04-24
feat: MLLM chunked prefill — non-blocking interleaved prefill/decode

janhilgard:
## Summary

- Add `install_chunked_prefill_mllm()` that splits long text-only prefills into fixed-size chunks (default 1024 tokens), yielding the event loop between chunks so health/status/metrics endpoints remain responsive
- Fix event loop blocking on idle servers: previously `num_active == 0` bypassed interleaved chunked prefill entirely, causing the whole prefill to run synchronously (blocking all HTTP endpoints for minutes on 100K+ token requests)
- Interleave prefill chunks with generation steps when an active batch exists, maintaining 30-50 tok/s for concurrent requests
- Make `_preprocess_request()` idempotent for text-only requests so chunked prefill can safely re-enter without double-processing
- Use `_eval_prompt_cache()` helper for hybrid model (KVCache + ArraysCache) compatibility
- Fix exact prefix-cache hit path: use `_trim_cache_offset(kv, 1)` to correctly reduce cache offset (previously used `_copy_prefix_cache` which didn't trim)

## Test plan

- [x] 50K-token prefill on idle Qwen3.6-27B: 377/377 health probes responded (avg 3ms, max 794ms) vs complete unresponsiveness before the fix
- [x] Short prompt completion works unchanged (falls through to `_orig_next()`)
- [x] Concurrent requests during long prefill maintain generation throughput
- [x] Client disconnect during chunked prefill aborts cleanly
- [x] Prefix cache hit reduces chunk count correctly

### DIFF
--- a/tests/test_mllm_continuous_batching.py
+++ b/tests/test_mllm_continuous_batching.py
@@ -1284,3 +1284,236 @@ class TestPreprocessIdempotent:
         # Should NOT return early — will try to import prepare_inputs
         with pytest.raises(Exception):
             gen._preprocess_request(req)
+
+
+class TestChunkedPrefillCacheHandling:
+    """Tests for chunked prefill prefix cache handling paths."""
+
+    def _make_fake_batch_gen(self):
+        """Build a minimal fake MLLMBatchGenerator for chunked prefill tests."""
+        from vllm_mlx.mllm_batch_generator import MLLMBatchGenerator, MLLMBatchStats
+
+        gen = MLLMBatchGenerator.__new__(MLLMBatchGenerator)
+        gen._stats = MLLMBatchStats()
+        gen._pending_error_responses = []
+        gen._aborted_request_ids = set()
+        gen._prefill_progress = {}
+        gen.active_batch = None
+        gen.stop_tokens = set()
+        gen.unprocessed_requests = []
+        gen._think_suffix_len = 0
+
+        # _has_empty_rotating_cache — always False for test caches
+        gen._has_empty_rotating_cache = lambda cache: False
+
+        return gen
+
+    def _make_fake_kv_cache(self, offset=100):
+        """Create a real KVCache with the given offset."""
+        from mlx_lm.models.cache import KVCache
+
+        cache = KVCache()
+        # Populate cache by updating with dummy data
+        if offset > 0:
+            k = mx.zeros((1, 1, offset, 4))
+            v = mx.zeros((1, 1, offset, 4))
+            cache.update_and_fetch(k, v)
+            mx.eval(cache.keys, cache.values)
+        return cache
+
+    def test_exact_hit_uses_trim_cache_offset(self, monkeypatch):
+        """Exact prefix-cache hit must use _trim_cache_offset(kv, 1),
+        NOT _copy_prefix_cache, to reduce the cache offset by 1."""
+        from vllm_mlx.mllm_batch_generator import (
+            MLLMBatchRequest,
+            install_chunked_prefill_mllm,
+        )
+
+        gen = self._make_fake_batch_gen()
+
+        # Track which functions are called
+        trim_calls = []
+        copy_calls = []
+
+        import vllm_mlx.mllm_batch_generator as bg_mod
+
+        orig_trim = bg_mod._trim_cache_offset
+
+        def tracking_trim(cache, n):
+            trim_calls.append(n)
+            return orig_trim(cache, n)
+
+        monkeypatch.setattr(bg_mod, "_trim_cache_offset", tracking_trim)
+
+        gen._copy_prefix_cache = lambda kv: (copy_calls.append(1), kv)[1]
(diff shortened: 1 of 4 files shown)


Thump604 (changes requested):
I would not merge this yet. The direction is valuable, but I see a correctness issue in the prefix-cache exact-hit path and the manual test plan still has the hard cases unchecked.

For an exact prefix-cache hit, `MemoryAwarePrefixCache.fetch()` returns the cached KV state plus `remaining_ids == []`. This PR handles that by keeping the full cached state and replaying the final prompt token:

```python
request_cache = batch_gen._copy_prefix_cache(cached_kv)
remaining = input_ids[:, -1:]
cached_count = total_tokens - 1
remaining_count = 1
```

Unless the cache is trimmed by one token first, this replays the last prompt token on top of a cache that already contains that token. That shifts positions and makes the logits come from a duplicated suffix rather than the original prompt. The existing prefix path only feeds uncached remaining tokens. If exact-hit needs to synthesize last logits by replaying one token, the cache should be trimmed back to `total_tokens - 1` first, and there should be a regression test for exact-hit generation equivalence.

I am also not comfortable with the checked state of the test plan yet. The PR body still has the important safety cases unchecked: short prompt unchanged, concurrent generation throughput, client disconnect during chunked prefill, and prefix cache hit chunk-count behavior. Those are exactly the paths this wrapper changes.

I would split or tighten this before merge:

1. add unit coverage for exact prefix hit, prefix partial hit, and abort cleanup,
2. prove first-token logits match the original `_process_prompts()` path for short and exact-cache-hit prompts,
3. keep the live health-probe evidence, but do not use that as a substitute for cache-position correctness.

janhilgard:
## Reviewer feedback addressed

### 1. Exact prefix-cache hit bug — fixed

The exact-hit path in `install_chunked_prefill_mllm` now uses `_trim_cache_offset(cached_kv, 1)` instead of `_copy_prefix_cache() + _trim_rotating_caches()`. This matches the canonical `_process_prompts` path:

```python
# Before (broken — no offset reduction):
request_cache = batch_gen._copy_prefix_cache(cached_kv)
batch_gen._trim_rotating_caches(request_cache)

# After (correct — offset reduced by 1):
request_cache = _trim_cache_offset(cached_kv, 1)
```

**Note on severity:** The bug didn't trigger in practice because `remaining_count = 1` is always `<= budget (1024)`, so the code fell through to `_orig_next()` → `_process_prompts()` which correctly calls `_trim_cache_offset(cached_kv, 1)`. Fixed for code correctness and to prevent future regressions.

### 2. Unit tests added

Added `TestChunkedPrefillCacheHandling` with 4 tests:
- **`test_exact_hit_uses_trim_cache_offset`** — verifies exact cache hit uses `_trim_cache_offset(kv, 1)` and does NOT call `_copy_prefix_cache`
- **`test_partial_hit_uses_copy_prefix_cache`** — verifies partial hit uses `_copy_prefix_cache` and does NOT call `_trim_cache_offset`
- **`test_abort_cleans_up_partial_prefill`** — verifies `_partial` is cleared, `_prefill_progress` is cleaned up, and abort response is returned
- **`test_short_prompt_falls_through_to_orig_next`** — verifies short prompts (< budget) fall through to `_orig_next()`

janhilgard:
@Thump604 Friendly ping — I have a fix ready for the exact-hit path (`_trim_cache_offset` instead of `_copy_prefix_cache`) and unit tests covering exact hit, partial hit, and abort cleanup. Would you prefer I push the update to this PR, or would it be easier to review a fresh rebase on current main?

Thump604:
@janhilgard fresh rebase on current main would be easiest to review. A lot has changed since this was opened and a clean diff will be faster to evaluate.

janhilgard:
@Thump604 Done — fresh rebase on current main (single squashed commit).

Changes from previous version:
1. **Exact-hit path fixed**: uses `_trim_cache_offset(cached_kv, 1)` instead of `_copy_prefix_cache()` — matches the canonical `_process_prompts` path
2. **Unit tests added**: `TestChunkedPrefillCacheHandling` with 4 tests covering exact hit, partial hit, abort cleanup, and short prompt fallthrough
3. **Clean diff**: single commit on top of current main including #450 (streaming timeout)

CI running now.

Thump604 (approved):
Clean rebase, clean diff. The exact-hit path fix using _trim_cache_offset(kv, 1) instead of _copy_prefix_cache is correct. The interleaved prefill/decode design keeps health/status endpoints responsive during long prefills without blocking the event loop.

Test coverage is solid: exact hit, partial hit, abort cleanup, and short prompt fallthrough all verified. CI all green.

One thing to watch in production: the budget=1024 default means a 64K prompt takes ~63 steps to fully prefill. With the scheduler yielding between steps, that is 63 context switches. If there is measurable overhead from the yield, it would show up as increased TTFT on long prompts under load. Probably fine at this budget size but worth a bench-serve run after merge.

PR #415, opened 2026-04-24
Fix O(n^2) performance in JSONSchemaLogitsProcessor

janhilgard:
## Summary

- **Fix O(n^2) performance** in `JSONSchemaLogitsProcessor` that caused server hangs on constrained decoding requests
- The processor re-decoded the entire generated suffix and re-scanned the full decoded text on **every generation step**, making overall complexity O(n^2) in output length
- With `max_tokens=262144` and 4 concurrent requests, this caused the server to hang for 280+ seconds with throughput dropping to ~21 tok/s

### Changes

1. **Incremental suffix decoding**: only decode newly appended tokens and concatenate to cached text (O(1) amortised per step instead of O(n))
2. **Incremental JSON context tracking**: scan only new characters for bracket/brace depth and string state (O(1) per step instead of O(n))
3. **Bracket depth pre-check** in `_suffix_is_complete_json`: skip the expensive `json.loads` call when brackets are unbalanced (~99% of steps)
4. **numpy-based `_build_allow_mask`**: use numpy for C-level mask construction instead of Python loop over `vocab_size` elements
5. **Fix O(n) list comparison** in `__call__`: use `prompt_len` directly instead of comparing `suffix == tokens_list`

Overall complexity reduced from **O(n^2) to O(n)** in output length.

## Test plan

- [x] All 24 constrained decoding tests pass (including 5 new tests for incremental caching)
- [x] Full test suite: 1772 passed, 11 skipped
- [ ] Manual test with constrained decoding + `max_tokens=262144` to verify no more hangs

### DIFF
--- a/tests/test_constrained_decoding.py
+++ b/tests/test_constrained_decoding.py
@@ -429,5 +429,191 @@ class TestSimplifySchema:
                 pytest.fail(f"Nested anyOf still present in branch: {branch!r}")
 
 
+# ---------------------------------------------------------------------------
+# Incremental caching — O(n) suffix decode + context tracking.
+# ---------------------------------------------------------------------------
+
+
+class _BPETokenizer(_FakeTokenizer):
+    """Fake tokenizer that simulates BPE whitespace-prefix behaviour.
+
+    In real BPE tokenizers (Mistral, Llama, Qwen), a token like ``world``
+    decodes to ``" world"`` (with leading space) when preceded by another
+    token, but ``"world"`` when decoded alone.  This means::
+
+        decode([tok_hello]) + decode([tok_world]) != decode([tok_hello, tok_world])
+
+    This tokenizer adds multi-character tokens with context-dependent
+    whitespace to exercise prefix-stability in ``_decode_suffix``.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+        # Add multi-char tokens that simulate BPE merges.
+        # Token ids continue from the parent's vocab.
+        self._multi = {
+            self.vocab_size: "Hello",
+            self.vocab_size + 1: " world",  # note: leading space in context
+            self.vocab_size + 2: " value",
+        }
+        self._multi_alone = {
+            self.vocab_size: "Hello",
+            self.vocab_size + 1: "world",  # NO leading space when decoded alone
+            self.vocab_size + 2: "value",
+        }
+        self.vocab_size += len(self._multi)
+
+    def encode(self, text: str) -> list[int]:
+        # Simple: just use parent for JSON chars, multi tokens for words.
+        return super().encode(text)
+
+    def decode(self, ids: list[int]) -> str:
+        parts: list[str] = []
+        for idx, tok_id in enumerate(ids):
+            if tok_id in self._multi:
+                # Simulate BPE: leading space only when preceded by another token.
+                if idx > 0:
+                    parts.append(self._multi[tok_id])
+                else:
+                    parts.append(self._multi_alone[tok_id])
+            elif 0 <= tok_id < len(self._id_to_tok):
+                tok = self._id_to_tok[tok_id]
+                if not tok.startswith("<"):
+                    parts.append(tok)
+        return "".join(parts)
+
+
+@pytestmark_lmfe
+class TestIncrementalCaching:
+    """Verify that the incremental decode / context-tracking optimisations
+    produce correct results identical to a full re-scan on every step."""
+
+    def _make_processor(self, schema: dict | None = None, tokenizer=None):
+        from vllm_mlx.constrained import JSONSchemaLogitsProcessor
+
+        tok = tokenizer or _FakeTokenizer()
+        return JSONSchemaLogitsProcessor(schema, tok), tok
+
+    def test_incremental_decode_matches_full_decode(self):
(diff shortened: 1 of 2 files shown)


Thump604 (changes requested):
I do not think this is safe to merge yet.

The performance direction is right, but `_decode_suffix()` now assumes that decoded text is appendable across token boundaries:

```python
self._cached_suffix_text += self._tokenizer.decode(new_tokens)
```

That assumption is false for supported tokenizer families. I verified it locally with a cached Mistral tokenizer:

```text
path: /Users/David/ai-models/huggingface/hub/models--mlx-community--Mistral-7B-v0.2-4bit/snapshots/175637a2b3e5751f0378132143e95eba50b86734
text: 'Hello world'
ids: [22557, 1526]
full decode:   'Hello world'
per-token join: 'Helloworld'
```

If the JSON context scanner sees `Helloworld` instead of `Hello world`, the brace/string/key tracking can diverge from what the model actually generated. That can change allowed-token filtering and EOS gating, which makes this a correctness bug, not just a tokenizer edge case.

I would keep the bracket-depth and mask-building optimizations, but the suffix text cache needs a tokenizer-safe strategy. Options I would trust:

1. keep full `tokenizer.decode(suffix)` for tokenizer families without a proven incremental decoder,
2. gate append-only decoding behind an explicit tokenizer capability check with tests using a non-concatenative tokenizer,
3. use a tokenizer-provided incremental/streaming decoder if one exists for the active backend.

The new fake-tokenizer tests do not catch this because the fake tokenizer is concatenative by construction. Please add a regression test that proves `_decode_suffix()` matches full decode for a non-concatenative tokenizer boundary before this merges.

janhilgard:
@Thump604 Good catch — you're absolutely right. The per-token `decode(new_tokens)` concatenation is incorrect for BPE/SentencePiece tokenizers where whitespace is encoded as a token prefix (e.g. `▁world` → `" world"` in context, but `"world"` when decoded alone).

**Fix plan:**

Revert `_decode_suffix()` to always use full `tokenizer.decode(suffix)` — correct for all tokenizer families. Keep a length-based cache to avoid redundant decodes within the same step (where `_decode_suffix` is called from both `_get_json_context` and `_suffix_is_complete_json`).

The **incremental JSON context scanning** in `_get_json_context` remains safe because:
- `decode(tokens[:n])` is a stable prefix of `decode(tokens[:n+1])` for all supported tokenizers (deterministic decode, new characters only appended at end)
- The scanner resumes from `_json_ctx_scanned_len` and only processes new characters
- Will add a prefix-stability assertion as a safety net — if it ever fails, reset and re-scan from zero

The main performance wins are unaffected by this change:
1. **Incremental JSON scanning** — O(1) per step instead of O(n) re-scan ← this was the biggest contributor to the 280s hang
2. **Bracket depth pre-check** — skips `json.loads` on ~99% of steps
3. **numpy mask construction** — C-level instead of Python loop
4. **Reordered key filter** — context state updated before EOS guard

Will push the fix with a regression test using a non-concatenative tokenizer boundary.

Thump604:
The fix plan sounds right. The incremental JSON scanning is the real performance win and that stays safe since decode(tokens[:n]) is a stable prefix. Reverting _decode_suffix to full tokenizer.decode(suffix) with a length-based cache is the correct conservative fix for BPE/SentencePiece correctness.

The prefix-stability assertion as a safety net is a good idea. Looking forward to the regression test with a non-concatenative tokenizer boundary.

janhilgard:
@Thump604 Both items from your review are implemented in the rebased branch:

1. **`_decode_suffix` reverted to full `tokenizer.decode(suffix)`** — correct for all tokenizer families (BPE, SentencePiece, etc.). Length-based cache avoids redundant decodes within the same step.

2. **Prefix-stability assertion** — if `decode(tokens[:n+1])` doesn't start with `decode(tokens[:n])`, the incremental JSON context scanner resets and re-scans from zero (safety net).

3. **Regression test `test_non_concatenative_tokenizer_decode`** — uses a `_BPETokenizer` fake that simulates whitespace-prefix behavior (`decode([hello]) + decode([world]) = "Helloworld"` vs `decode([hello, world]) = "Hello world"`). Verifies `_decode_suffix` always matches the full decode.

Branch is cleanly rebased on current `main` (including #375), all 9 CI checks pass. Could you re-review?

Thump604 (approved):
Reviewed the rebased implementation. The key changes look correct:

1. `_decode_suffix` uses full `tokenizer.decode()` with length-keyed caching and prefix stability validation -- if `decode(tokens[:n])` is not a prefix of `decode(tokens[:n+1])`, the incremental context state resets. This handles BPE whitespace-prefix behavior correctly.

2. `_get_json_context` now scans incrementally from `_json_ctx_scanned_len` instead of from 0 on every call. Amortized O(1) per step.

3. `_suffix_is_complete_json` uses cached bracket/brace depth as a fast pre-check before the expensive `json.loads` call.

4. `_build_allow_mask` moves from Python loop to numpy vectorized assignment -- good for large vocabularies.

5. `_get_json_context` call moved before the EOS guard so depth counters are current for the pre-check.

Test coverage with `_BPETokenizer` is a nice addition -- validates the prefix stability assumption that underpins the incremental approach.

All 9 CI checks green.

PR #416, opened 2026-04-24
Prevent MLLM preprocessing from blocking event loop

janhilgard:
## Summary

- Run text-only request preprocessing (Jinja2 template rendering + tokenization) in a thread-pool executor via `run_in_executor()` so the event loop stays responsive during long prompts (40K+ tokens)
- Add chunked prefill progress logging (start, every 10 chunks, completion) for operator visibility
- Add slow-step warnings (>2s) for easier debugging of event loop blocking

## Problem

The MLLM scheduler's `_process_loop()` runs directly on the asyncio event loop. When a long text-only request (40K+ tokens) arrives, `step()` calls `_chunked_next()` which calls `_preprocess_request()` synchronously. For long conversations, Jinja2 template rendering + tokenization can take 10-30+ seconds, completely blocking the event loop during that time. Health checks, new connections, and streaming responses for existing requests all fail.

## Solution

Run `_preprocess_request()` via `await loop.run_in_executor(None, ...)` in a thread-pool worker **before** `step()`. Since preprocessing is CPU-bound (no MLX GPU work) and HuggingFace tokenizers are thread-safe (backed by Rust `tokenizers`), this is safe to offload. The call is idempotent — the subsequent call from `_chunked_next()` or `_process_prompts()` returns immediately when `input_ids` is already set.

## Test plan

- [x] All 1768 existing tests pass (4 pre-existing failures unrelated to this change)
- [x] Verify health checks respond during long prompt preprocessing on a live server (20/20 OK, max 323ms, avg 20ms on Qwen3.6-27B with 202-message conversation)
- [x] Check log output shows chunked prefill progress (start, periodic, completion messages)

### DIFF
--- a/tests/test_mllm_continuous_batching.py
+++ b/tests/test_mllm_continuous_batching.py
@@ -1228,3 +1228,59 @@ class TestBatchedMLLMConfigWiring:
         asyncio.run(engine._start_mllm())
 
         assert captured["config_kwargs"]["prefill_step_size"] == 256
+
+
+class TestPreprocessIdempotent:
+    """_preprocess_request must be idempotent for text-only requests.
+
+    The scheduler offloads preprocessing to a thread-pool executor so
+    the event loop stays responsive.  _process_prompts then calls
+    _preprocess_request again — the second call must be a no-op.
+    """
+
+    def test_text_only_not_preprocessed_twice(self):
+        """When input_ids is already set (executor did it), skip."""
+        from vllm_mlx.mllm_batch_generator import MLLMBatchRequest
+
+        req = MLLMBatchRequest(
+            uid=0,
+            prompt="Hello",
+            request_id="test-idem",
+        )
+        # Simulate executor having already set input_ids
+        req.input_ids = mx.array([[1, 2, 3]])
+
+        # Build a minimal batch generator with the method
+        from vllm_mlx.mllm_batch_generator import MLLMBatchGenerator
+
+        gen = MLLMBatchGenerator.__new__(MLLMBatchGenerator)
+        gen._preprocess_request = MLLMBatchGenerator._preprocess_request.__get__(
+            gen, MLLMBatchGenerator
+        )
+
+        # Must return immediately without touching prepare_inputs
+        gen._preprocess_request(req)
+        assert req.input_ids.shape == (1, 3)
+
+    def test_vision_request_not_skipped(self):
+        """Vision requests should NOT be skipped even with input_ids set."""
+        from vllm_mlx.mllm_batch_generator import MLLMBatchRequest
+
+        req = MLLMBatchRequest(
+            uid=0,
+            prompt="Describe",
+            request_id="test-vis",
+            images=["fake.png"],
+        )
+        req.input_ids = mx.array([[1, 2, 3]])
+
+        from vllm_mlx.mllm_batch_generator import MLLMBatchGenerator
+
+        gen = MLLMBatchGenerator.__new__(MLLMBatchGenerator)
+        gen._preprocess_request = MLLMBatchGenerator._preprocess_request.__get__(
+            gen, MLLMBatchGenerator
+        )
+
+        # Should NOT return early — will try to import prepare_inputs
+        with pytest.raises(Exception):
+            gen._preprocess_request(req)
--- a/vllm_mlx/mllm_batch_generator.py
+++ b/vllm_mlx/mllm_batch_generator.py
@@ -772,10 +772,17 @@ class MLLMBatchGenerator:
         3. Running vision encoder to get features
 
         Uses vision cache to skip processing for repeated images.
+        Idempotent: if input_ids is already set, returns immediately.
 
         Args:
             request: Request to preprocess
         """
+        # Already preprocessed (e.g. by early executor offloading in
+        # _process_loop).  Only skip for text-only requests; vision
+        # requests need pixel cache lookup even if input_ids was set.
+        if request.input_ids is not None and not request.images and not request.videos:
+            return
+
         from mlx_vlm.utils import prepare_inputs
 
         tic = time.perf_counter()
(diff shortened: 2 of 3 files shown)


Thump604 (changes requested):
I do not think this fixes the stated single-request event-loop blocker yet.

The PR moves `_preprocess_request(req)` earlier in `_process_loop()`, but it still calls it synchronously on the event loop:

```python
bg._preprocess_request(req)
await asyncio.sleep(0)
```

That yields only after the full Jinja render plus tokenization has already completed for that request. For the case described in the PR body, one 40K+ token text-only request can still block the event loop for the full 10-30 seconds before health checks, new connections, and active streams get a chance to run. The slow-step logging helps detect the stall after the fact, but it does not prevent it.

I would trust one of these shapes instead:

1. split preprocessing so CPU-only template render/tokenization runs off the event loop, then create MLX arrays back on the model thread,
2. add a real async/chunked preprocessing path that yields during the long operation, not only between requests,
3. narrow the PR to progress logging only and avoid claiming event-loop responsiveness until the live health-probe test is green.

The test plan still has live health-check verification unchecked. I would not merge this as a responsiveness fix until a single long prompt is proven not to starve `/health` during preprocessing.

janhilgard:
Thanks for the review — you're right, the previous version only yielded *between* requests.

Updated to use `loop.run_in_executor(None, bg._preprocess_request, req)` so the entire Jinja2 template rendering + tokenization runs in a thread-pool worker. Since preprocessing is CPU-bound (no MLX GPU operations) and HuggingFace tokenizers are thread-safe (backed by Rust `tokenizers`), this is safe to offload.

**Live test results** (Qwen3.6-27B, 202-message / ~20K-token conversation, port 1236):

```
Health check #0:  323ms — status 200   ← during preprocessing
Health check #1:    4ms — status 200
Health check #2:    3ms — status 200
...
Health check #19:   4ms — status 200

Results: 20/20 OK, 0 timeouts
Max latency: 323ms
Avg latency: 20ms
PASS: Health endpoint stayed responsive
```

Health checks respond in <400 ms even during active preprocessing.

> Note: `step()` itself still blocks for ~3.5 s per chunk on dense 27B models (synchronous `mx.eval`). This is inherent to MLX's thread-local stream model — the MLLM forward pass must run on the event-loop thread. Health checks survive this because `await asyncio.sleep(0)` between steps gives the event loop a window. A future PR could explore running `step()` in a dedicated executor with `bind_generation_streams()`, similar to the text-only `EngineCore` path.

Thump604 (changes requested):
Thanks for the revision. I still cannot approve this version because the early preprocessing result is not consumed as the exclusive preprocessing path.

The scheduler now offloads `bg._preprocess_request(req)` for text-only requests, but `_process_prompts()` still calls `self._preprocess_request(req)` unconditionally for every request before prefill. That means the expensive Jinja/tokenization path still runs inside `step()`, on the event-loop scheduler path, after the executor copy has already completed. In other words, the health responsiveness improvement is not guaranteed by this patch because the blocking work is duplicated at the original call site.

I verified this from the current diff: `mllm_scheduler.py` offloads at lines 823-825, then `mllm_batch_generator.py` calls `_preprocess_request(req)` again at line 1128 without checking `req.input_ids` or another preprocessed marker. The existing `tests/test_mllm_continuous_batching.py` suite still passes, but it does not cover this invariant.

The fix I would expect is one of these shapes:

1. Make `_process_prompts()` skip `_preprocess_request()` when the request was already preprocessed, with a regression test that proves `_preprocess_request` is not called twice for a text-only request.
2. Better, split text-only CPU template/tokenization from MLX array creation so only CPU work runs in the executor and MLX arrays are finalized on the scheduler/model thread.

I am fine with the overall direction, but this version can still block exactly where the PR is trying to avoid blocking.

janhilgard:
Thanks for the careful review — you caught a real gap.

**What changed in the latest push:**

1. **`_preprocess_request` is now idempotent** (the key fix you asked for):
   Added an early-return guard at the top of `_preprocess_request`:
   ```python
   if request.input_ids is not None and not request.images and not request.videos:
       return
   ```
   When the executor sets `input_ids` in `_process_loop`, the subsequent call in `_process_prompts` inside `step()` hits this guard and returns immediately — no duplicate work. Vision requests are NOT skipped (they need pixel cache lookup even with `input_ids` set).

2. **Adaptive yield after slow steps**:
   When `step()` takes >1s (dense models with chunked prefill, `mx.eval` blocking), yield for 50ms instead of `asyncio.sleep(0)`. This ensures `asyncio.wait()` timeouts, heartbeat timers, and disconnect polls actually fire between heavy GPU steps. For fast steps (<1s, typical MoE models), still uses `asyncio.sleep(0)` for zero overhead.

3. **Regression tests** for both invariants:
   - `test_text_only_not_preprocessed_twice`: verifies idempotency
   - `test_vision_request_not_skipped`: verifies vision requests still go through full preprocessing

All existing tests pass (96 passed, 10 deselected).

janhilgard:
### Updated: Multi-yield event loop fix

Replaced the single `asyncio.sleep()` yield with multiple `asyncio.sleep(0)` cycles between steps.

**Root cause**: A single `asyncio.sleep(0.05)` gives only ONE `_run_once()` cycle. But an HTTP health check needs ~3 event loop cycles minimum (accept TCP → read/parse HTTP → run handler/write response). With heavy `mx.eval()` steps blocking ~1s each, health checks could take 3-5 yield windows to complete, occasionally exceeding the 5s timeout.

**Fix**: Yield 5-10 `asyncio.sleep(0)` cycles after each step. Each cycle triggers one `_run_once()` iteration with zero-timeout I/O poll, giving the event loop enough cycles to complete pending HTTP requests. Total overhead is negligible (<1ms per yield batch).

**Test results** (50K-token cold prefill on dense 27B model):

| Metric | Before | After |
|--------|--------|-------|
| Health checks OK | 77/120 | **120/120** |
| Timeouts | 43 | **0** |
| Max latency | 5656ms | **3023ms** |
| Avg latency | 1502ms | **848ms** |

Thump604 (approved):
The idempotent guard in `_preprocess_request` cleanly resolves the duplication concern. Text-only requests preprocessed by the executor skip the second call; vision requests are explicitly not skipped. The chunked prefill progress logging and slow-step warnings are good operator visibility.

Tested the logic path: executor offloads CPU-bound Jinja2 + tokenization, `step()` stays on the event loop thread for MLX GPU work, and the idempotent guard prevents double-processing. Tests cover both cases.

PR #423, opened 2026-04-24
Add model artifact workflow CLI

Thump604:
## Summary

This adds a first-class `vllm-mlx model` workflow for model setup before serving:

- `vllm-mlx model inspect <path-or-hf-id>` reads config, HF file inventory, source revision, size estimates, MLX/conversion signals, and large-context warnings without loading weights.
- `vllm-mlx model acquire <hf-id>` downloads through Hugging Face with resumable `snapshot_download`, uses `HF_HUB_ENABLE_HF_TRANSFER=1` when `hf_transfer` is installed, supports staged finalization into `--target-dir`, and writes `vllm_mlx_model_manifest.json`.
- `vllm-mlx model convert <source> --output <path>` wraps `mlx-lm convert`, records the exact quantization recipe and command, supports dry-run, and writes `vllm_mlx_conversion_manifest.json` after successful conversion.

## Why

Model setup is currently an operator workflow: manually inspect a repo, download files, choose conversion flags, remember the source revision, edit local config, then hope the artifact is complete. This PR makes the pre-serve part auditable and scriptable without trying to replace serving policy or qualification.

This intentionally stops before registry/preset mutation. Acquisition and conversion produce manifests. A separate follow-up can wire those manifests into registry insertion and `bench-serve --workload` qualification once that workload contract PR lands.

## Validation

```bash
uv run --extra dev pytest -q tests/test_model_workflow.py tests/test_download.py
uv run --extra dev black --check vllm_mlx/model_workflow.py vllm_mlx/cli.py tests/test_model_workflow.py
uvx ruff check vllm_mlx/model_workflow.py vllm_mlx/cli.py tests/test_model_workflow.py --select E,F,W --ignore E402,E501,E731,F811,F841
git diff --check
uv run --extra dev python -m vllm_mlx.cli model inspect mlx-community/Llama-3.2-1B-Instruct-4bit
uv run --extra dev python -m vllm_mlx.cli model convert /Users/David/code/vllm-mlx-model-workflow --output /tmp/vllm-mlx-dry-run-out --quantize --q-bits 4 --q-group-size 64 --q-mode affine --dry-run
```

Result: `19 passed`; formatting clean; ruff clean; live HF metadata inspect returned revision `08231374eeacb049a0eade7922910865b8fce912` without loading weights.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,19 @@ python examples/tts_multilingual.py "Hola mundo" --lang es --play
 vllm-mlx bench-serve --url http://localhost:8000 --concurrency 5 --prompts prompts.txt --output results.csv
 ```
 
+### Model acquisition and conversion
+
+```bash
+# Inspect repo metadata, file sizes, config, and rough fit before downloading weights
+vllm-mlx model inspect mlx-community/Llama-3.2-3B-Instruct-4bit
+
+# Acquire with resumable Hugging Face transfer and write a local artifact manifest
+vllm-mlx model acquire mlx-community/Llama-3.2-3B-Instruct-4bit --target-dir ./models/llama-3b-4bit
+
+# Wrap mlx-lm conversion and record the exact recipe in the converted artifact
+vllm-mlx model convert meta-llama/Llama-3.2-3B-Instruct --output ./models/llama-3b-mlx-q4 --quantize --q-bits 4 --q-group-size 64 --q-mode affine
+```
+
 ### Prometheus metrics
 
 ```bash
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -81,6 +81,13 @@ Check your internet connection and HuggingFace access. Some models require authe
 huggingface-cli login
 ```
 
+You can inspect and stage models before serving:
+```bash
+vllm-mlx model inspect mlx-community/Llama-3.2-3B-Instruct-4bit
+vllm-mlx model acquire mlx-community/Llama-3.2-3B-Instruct-4bit \
+  --target-dir ./models/llama-3b-4bit
+```
+
 ### Out of memory
 
 Use a smaller quantized model:
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -5,6 +5,7 @@
 | Command | Description |
 |---------|-------------|
 | `vllm-mlx serve` | Start OpenAI-compatible server |
+| `vllm-mlx model` | Inspect, acquire, or convert model artifacts |
 | `vllm-mlx-bench` | Run performance benchmarks |
 | `vllm-mlx-chat` | Start Gradio chat interface |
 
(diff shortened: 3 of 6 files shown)


Thump604:

janhilgard:
@Thump604 Thanks for putting this together — the manifest-driven workflow is a clean idea and the frozen dataclass pattern is solid. A few things I'd flag:

### Must fix

**1. Double print on conversion failure**

In `cli.py` `model_command`, the failed-conversion path prints the payload *and* calls `sys.exit`, but the success path also prints unconditionally at the end of the function. If `sys.exit` is caught (test harness, wrapper), the payload prints twice. Simpler to always fall through to one print:

```python
print(json.dumps(payload, indent=2))
if payload.get("status") == "failed":
    sys.exit(payload.get("returncode") or 1)
```

**2. GPTQ/AWQ false positive in `has_mlx_signals`**

```python
has_mlx_signals = bool(family.get("quantization")) or has_name_signal
```

A PyTorch GPTQ model will have `quantization_config` in its config, making `has_mlx_signals = True` and `needs_conversion = False` — even though it absolutely needs conversion to MLX format. This is a meaningful false negative that will silently tell users a non-MLX model is ready to serve.

Consider checking for MLX-specific quantization keys (e.g. `"bits"` + `"group_size"` without `"quant_method": "gptq"`) or looking for `.safetensors` files with MLX weight naming patterns.

**3. Missing test: conversion failure path**

`model_command` has special handling for `status == "failed"` (`sys.exit`), but no test covers it. The success and dry-run paths are tested, but the failure path is not.

### Should fix

**4. `--local-files-only` on `acquire` is contradictory**

`acquire` is explicitly about downloading. Passing `--local-files-only` to `snapshot_download` means "don't download" — this is confusing. Remove it from `acquire`, or at minimum document it as "use cached version without network access."

**5. No retry logic in `acquire_model`**

The existing `ensure_model_downloaded` in `utils/download.py` has exponential backoff and configurable retries. The new `acquire_model` calls `snapshot_download` with no retries — for multi-GB downloads that take hours, a single network hiccup fails the entire operation.

**6. `_read_json` only catches `FileNotFoundError`**

A corrupted `config.json` will raise `json.JSONDecodeError` and propagate as an unhandled exception. Either catch it too and surface a warning, or document this is intentional.

**7. Relationship with existing `download` command**

Both `vllm-mlx download` and `vllm-mlx model acquire` call `snapshot_download` with the same patterns. The overlap isn't documented — should users use one or the other? Is `download` being deprecated?

**8. `convert` positional arg named `source` vs `model` elsewhere**

`inspect` and `acquire` use `model` as the positional arg name. `convert` uses `source`. Minor inconsistency but users will notice.

### Minor / nice to have

- No validation on `--q-bits` range — `--q-bits 0` will pass through to `mlx-lm` and fail with an unhelpful error
- `--trust-remote-code` help text understates the risk — consider mentioning "SECURITY RISK" since it enables arbitrary code execution from the repo
- The module uses `print` via the CLI layer but no `logging` — importing `model_workflow` as a library gives no progress feedback, unlike the existing `download.py`
- Staging cleanup on `shutil.move` failure is untested

janhilgard (changes requested):
### Must fix

1. **Double print on conversion failure** — failed-conversion path prints payload *and* calls `sys.exit`, but success path also prints unconditionally. If `sys.exit` is caught, payload prints twice.

2. **GPTQ/AWQ false positive in `has_mlx_signals`** — PyTorch GPTQ model with `quantization_config` triggers `has_mlx_signals = True` → `needs_conversion = False`, silently telling users a non-MLX model is ready to serve.

3. **Missing test: conversion failure path** — `status == "failed"` / `sys.exit` path is untested.

### Should fix

4. `--local-files-only` on `acquire` contradicts the download purpose
5. No retry logic in `acquire_model` (existing `ensure_model_downloaded` has exponential backoff)
6. `_read_json` only catches `FileNotFoundError`, not `json.JSONDecodeError`
7. Overlap with existing `download` command undocumented
8. `convert` uses `source` as positional arg, `inspect`/`acquire` use `model`

### Minor

- No validation on `--q-bits` range
- `--trust-remote-code` help text understates security risk
- No `logging` — importing as library gives no progress feedback

See [full review comment](https://github.com/waybarrios/vllm-mlx/pull/423#issuecomment-4320003879) for details and code examples.

Thump604:
@janhilgard Verified all items against the code. Here is where I landed:

**Must-fix (1):** Correct, the double-print on conversion failure is a real bug. Will restructure so there's a single `print(json.dumps(payload))` at the end, with `sys.exit` after it.

**Must-fix (2):** Good catch. PyTorch GPTQ/AWQ models have `quantization_config` which triggers a false positive. I'll tighten the check to look for MLX-specific signals (presence of `bits` + `group_size` without `quant_method` being `gptq`/`awq`, or the presence of MLX-format weight shards).

**Must-fix (3):** Will add a test for the conversion failure path.

**Should-fix (4):** Fair point. `--local-files-only` on `acquire` is confusing. I'll remove it from that subcommand and document that `inspect` supports it for offline cache queries.

**Should-fix (5):** I intentionally kept `acquire` simple since it's a thin wrapper, but you're right that multi-GB downloads without retry is asking for trouble. I'll add basic retry with backoff.

**Should-fix (6):** Will catch `json.JSONDecodeError` in `_read_json` and surface it as a warning.

**Should-fix (7-8):** Will document the overlap with `download` (acquire is the manifest-aware replacement, download stays for quick grabs) and unify the positional arg name to `model`.

janhilgard (approved):
All three must-fix items from my previous review are addressed:

1. **Double print on conversion failure** — `return` after `sys.exit` prevents the fallthrough to the second `print(json.dumps(payload))`. Verified in `cli.py` model_command.

2. **GPTQ/AWQ false positive** — `_is_mlx_quantization()` now filters `_NON_MLX_QUANT_METHODS` (gptq, awq, squeezellm, marlin, fp8) before checking for `"bits"` key. Covered by `test_inspect_gptq_model_is_not_detected_as_mlx`.

3. **Conversion failure test** — `test_convert_model_failure_reports_status_and_stderr` verifies `status == "failed"`, `returncode`, `stderr`, and absence of `output_inspection`/`manifest_path`.

Additionally, `_read_json` now catches `json.JSONDecodeError` alongside `FileNotFoundError` (should-fix #6).

Remaining should-fix items (positional arg naming, `--local-files-only` on acquire, retry logic, download overlap docs) are minor and don't block merge.

LGTM — clean, well-tested, ready to merge.

PR #424, opened 2026-04-24
Fix sampling defaults and short prefix-cache reuse

Thump604:
Supersedes #405 because the fork branch cannot be refreshed through the current OAuth token after upstream CI workflow changes landed. This branch is rebased on current `main` and carries the same scope:

- propagate CLI/server defaults for `top_k`, `min_p`, `presence_penalty`, and `repetition_penalty`
- preserve server default chat-template kwargs from current `main`
- avoid reusing short prefix-cache entries that are too small to be useful and can contaminate benchmark runs
- update focused tests for the current upstream helper/refactor shape

Local validation:

```bash
uv run --extra dev pytest -q tests/test_cli.py tests/test_server.py tests/test_server_cache_controls.py tests/test_memory_cache.py tests/test_memory_cache_mlx.py tests/test_prefix_cache.py
# 186 passed, 3 deselected

uv run --extra dev black --check vllm_mlx/cli.py vllm_mlx/server.py tests/test_cli.py tests/test_server.py
uvx ruff check vllm_mlx/cli.py vllm_mlx/server.py tests/test_cli.py tests/test_server.py --select E,F,W --ignore E402,E501,E731,F811,F841
git diff --check
# clean
```

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -4,6 +4,7 @@ from types import SimpleNamespace
 def _serve_args(**overrides):
     args = {
         "api_key": None,
+        "auto_unload_idle_seconds": 0.0,
         "cache_memory_mb": None,
         "cache_memory_percent": 0.2,
         "chunked_prefill_tokens": 0,
@@ -19,6 +20,7 @@ def _serve_args(**overrides):
         "download_timeout": 1,
         "embedding_model": None,
         "enable_auto_tool_choice": False,
+        "enable_metrics": False,
         "enable_mtp": False,
         "enable_prefix_cache": True,
         "gpu_memory_utilization": 0.9,
@@ -31,6 +33,7 @@ def _serve_args(**overrides):
         "max_num_seqs": 32,
         "max_tokens": 16,
         "mcp_config": None,
+        "mllm_prefill_step_size": None,
         "mllm": False,
         "model": "local-test-model",
         "mtp_num_draft_tokens": 1,
@@ -53,6 +56,7 @@ def _serve_args(**overrides):
         "stream_interval": 1,
         "tool_call_parser": None,
         "timeout": 300,
+        "lazy_load_model": False,
         "use_paged_cache": False,
     }
     args.update(overrides)
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,0 +1,98 @@
+from types import SimpleNamespace
+
+
+def _serve_args(**overrides):
+    args = {
+        "api_key": None,
+        "cache_memory_mb": None,
+        "cache_memory_percent": 0.2,
+        "chunked_prefill_tokens": 0,
+        "continuous_batching": False,
+        "default_min_p": None,
+        "default_presence_penalty": None,
+        "default_repetition_penalty": None,
+        "default_temperature": None,
+        "default_top_k": None,
+        "default_top_p": None,
+        "disable_prefix_cache": False,
+        "download_retries": 0,
+        "download_timeout": 1,
+        "embedding_model": None,
+        "enable_auto_tool_choice": False,
+        "enable_mtp": False,
+        "enable_prefix_cache": True,
+        "gpu_memory_utilization": 0.9,
+        "host": "127.0.0.1",
+        "kv_cache_min_quantize_tokens": 256,
+        "kv_cache_quantization": False,
+        "kv_cache_quantization_bits": 8,
+        "kv_cache_quantization_group_size": 64,
+        "max_cache_blocks": 1000,
+        "max_num_seqs": 32,
+        "max_tokens": 16,
+        "mcp_config": None,
+        "mllm": False,
+        "model": "local-test-model",
+        "mtp_num_draft_tokens": 1,
+        "mtp_optimistic": False,
+        "no_memory_aware_cache": False,
+        "offline": True,
+        "paged_cache_block_size": 64,
+        "port": 8000,
+        "prefill_batch_size": 8,
+        "prefill_step_size": 512,
+        "prefix_cache_size": 100,
+        "rate_limit": 0,
+        "reasoning_parser": None,
+        "served_model_name": None,
+        "specprefill": False,
+        "specprefill_backbone_pct": 0.0,
+        "specprefill_draft_model": None,
+        "specprefill_keep_pct": 0.3,
+        "specprefill_threshold": 8192,
+        "stream_interval": 1,
+        "tool_call_parser": None,
+        "timeout": 300,
+        "use_paged_cache": False,
+    }
+    args.update(overrides)
+    return SimpleNamespace(**args)
+
+
+def test_serve_command_propagates_all_sampling_defaults(monkeypatch):
+    from vllm_mlx import cli, server
+    from vllm_mlx.utils import download
+
+    monkeypatch.setattr(
+        download, "ensure_model_downloaded", lambda *args, **kwargs: "local-test-model"
+    )
+    monkeypatch.setattr(server, "load_model", lambda *args, **kwargs: None)
+    monkeypatch.setattr("uvicorn.run", lambda *args, **kwargs: None)
+
+    for attr in (
+        "_default_temperature",
+        "_default_top_p",
+        "_default_top_k",
+        "_default_min_p",
+        "_default_presence_penalty",
+        "_default_repetition_penalty",
+    ):
+        monkeypatch.setattr(server, attr, None)
+
+    cli.serve_command(
+        _serve_args(
+            default_temperature=0.6,
+            default_top_p=0.95,
+            default_top_k=20,
+            default_min_p=0.0,
+            default_presence_penalty=0.0,
+            default_repetition_penalty=1.0,
+        )
+    )
+
+    assert server._default_temperature == 0.6
+    assert server._default_top_p == 0.95
+    assert server._default_top_k == 20
+    assert server._default_min_p == 0.0
+    assert server._default_presence_penalty == 0.0
+    assert server._default_repetition_penalty == 1.0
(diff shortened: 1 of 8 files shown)


Thump604:

janhilgard:
@Thump604 Good scoping — the `_resolve_*` pattern is consistent with the existing temperature/top_p resolvers, and the prefix cache floor is a clean addition. Some notes:

### Should fix

**1. Responses API misses the new sampling defaults**

`_prepare_responses_request` (~line 1413 in `server.py`) only resolves `temperature` and `top_p`. It does not call `_resolve_top_k`, `_resolve_min_p`, `_resolve_presence_penalty`, or `_resolve_repetition_penalty`. If a user sets CLI defaults for these, they will have no effect on `/v1/responses` requests. Either wire them in or document the omission as intentional.

**2. Verify all Anthropic code paths go through `_prepare_anthropic_invocation`**

The diff updates `_prepare_anthropic_invocation`, but in earlier versions of the codebase there were separate inline `or 0` / `or 1.0` patterns in `create_anthropic_message` and `_stream_anthropic_messages`. If the base branch hasn't consolidated these, the new resolvers won't cover all Anthropic paths.

### Prefix cache floor

**3. No way to tune `min_prefix_tokens` from CLI**

The 128-token default is reasonable for chat workloads (system prompts are typically 500-2000 tokens), but for classification or short-prompt use cases it could be too aggressive. A `--min-prefix-tokens` flag would let operators tune this without code changes — or at minimum log the value at startup so operators know what they're getting.

**4. Missing test: LCP guard below floor**

`test_short_prefix_reuse_is_rejected` tests the store rejection and early fetch bail, but not the case where a long token sequence has an LCP match that falls below the 128-token floor (diff line ~428). That's a distinct code path worth covering.

**5. Missing test: `load_from_disk` short-entry skipping**

The guard that skips short entries during cache restoration (diff line ~458) has no test coverage. A test that writes a short entry to disk, then loads with a higher floor and verifies it's skipped would close this gap.

### Minor

- `test_server.py` uses manual `try/finally` to restore globals — `monkeypatch.setattr` would be more pytest-idiomatic and auto-cleanup
- The `"miss_short_prefix"` and `"miss_short_lcp"` match types are nice for diagnostics but aren't surfaced in `get_stats()` as separate counters — they just increment `misses`. Consider exposing them if you want observability into how often the floor triggers
- Worth documenting the rationale for 128 as the default (e.g. "below ~128 tokens the KV restore overhead exceeds the prefill savings")

Overall this is clean and the semantics around `None` vs. falsy (`top_k=0` means "user said 0", `top_k=None` means "use server default") are a real improvement over the old `or 0` pattern. The prefix cache floor will help avoid wasting cycles on tiny entries.

janhilgard (commented):
### Should fix

1. **Responses API misses new sampling defaults** — `_prepare_responses_request` only resolves `temperature` and `top_p`, not `top_k`, `min_p`, `presence_penalty`, or `repetition_penalty`.
2. **Verify all Anthropic code paths** go through `_prepare_anthropic_invocation` — earlier versions had inline `or 0` patterns elsewhere.

### Prefix cache floor

3. No CLI flag for `min_prefix_tokens` — operators can't tune the 128-token default without code changes.
4. Missing test: LCP match below floor (distinct from early fetch bail).
5. Missing test: `load_from_disk` short-entry skipping.

### Minor

- `test_server.py` uses manual try/finally — `monkeypatch.setattr` would be more robust
- `"miss_short_prefix"` / `"miss_short_lcp"` match types not exposed in `get_stats()` counters

Overall clean — the `None` vs falsy semantics (`top_k=0` means "user said 0") are a real improvement.

See [full review comment](https://github.com/waybarrios/vllm-mlx/pull/424#issuecomment-4320019902) for details.

Thump604:
Jan, assigning this to you for review. It propagates extended sampling defaults (top_k, min_p, presence/repetition penalty) end to end and adds a short-prefix-cache reuse guard. Mergeable and CI green.

janhilgard (commented):
Thanks for the ping. The sampling-default propagation and prefix-cache floor are both clean additions. A few items to address before merge:

---

### Must fix

**1. Responses API missing new sampling defaults**

`_prepare_responses_request` (line 1723-1727) builds `chat_kwargs` with only `temperature` and `top_p` resolved. It's missing `top_k`, `min_p`, `presence_penalty`, and `repetition_penalty`:

```python
# Current (incomplete):
chat_kwargs = {
    "max_tokens": chat_request.max_tokens or _default_max_tokens,
    "temperature": _resolve_temperature(chat_request.temperature),
    "top_p": _resolve_top_p(chat_request.top_p),
}

# Should be:
chat_kwargs = {
    "max_tokens": chat_request.max_tokens or _default_max_tokens,
    "temperature": _resolve_temperature(chat_request.temperature),
    "top_p": _resolve_top_p(chat_request.top_p),
    "top_k": _resolve_top_k(chat_request.top_k),
    "min_p": _resolve_min_p(chat_request.min_p),
    "presence_penalty": _resolve_presence_penalty(chat_request.presence_penalty),
    "repetition_penalty": _resolve_repetition_penalty(chat_request.repetition_penalty),
}
```

Without this, `--default-top-k 20` etc. won't take effect on the `/v1/responses` endpoint.

---

### Should fix

**2. `CacheStats` doesn't surface short-prefix rejections**

`miss_short_prefix` and `miss_short_lcp` are tracked as `_last_match_type` strings but they're counted as generic `misses` in `CacheStats`. Adding dedicated counters (or at least including `_last_match_type` distribution in `to_dict()`) would help operators understand *why* cache hit rate dropped after upgrading — they'd see "128 misses, of which 90 were short-prefix rejections" vs. just "128 misses".

**3. No CLI flag for `min_prefix_tokens`**

The 128-token default is reasonable, but operators can't tune it without code changes. Adding `--prefix-cache-min-tokens` (defaulting to 128) would follow the pattern of the other cache knobs like `--kv-cache-min-quantize-tokens`.

---

### Nits

**4. `test_server.py` try/finally vs monkeypatch**

The `TestSamplingDefaults` tests manually save/restore globals with try/finally. Since the test file already uses pytest, `monkeypatch.setattr` would be cleaner and safer if a test fails mid-execution:

```python
def test_extended_sampling_defaults(self, monkeypatch):
    monkeypatch.setattr(server, "_default_top_k", 20)
    monkeypatch.setattr(server, "_default_min_p", 0.05)
    ...
```

**5. Squash suggestion**

4 commits could be squashed into 1-2 for a cleaner history (one for sampling defaults, one for prefix cache floor), but this is maintainer preference.

---

### What's good

- The `None` vs falsy semantics fix is a real improvement — `top_k=0` now means "user explicitly requested 0" rather than being silently replaced.
- Anthropic endpoint now correctly uses `_resolve_temperature`/`_resolve_top_p` — this was a bug before.
- `min_prefix_tokens` enforcement is thorough: store, fetch (early bail + LCP check), and load_from_disk all covered.

CI is green 9/9. No merge conflicts.

Thump604:
Pushed 832735d addressing the review feedback.

**Must fix:**
- Responses API now resolves all four extended sampling defaults (top_k, min_p, presence_penalty, repetition_penalty) in `_prepare_responses_request`. Both streaming and non-streaming paths use the same `chat_kwargs` dict so both are covered.

**Should fix:**
- Added `--prefix-cache-min-tokens` CLI flag (serve + bench commands, default 128). Threaded through `SchedulerConfig`, `MLLMSchedulerConfig`, and into both `MemoryCacheConfig` instantiation sites (scheduler.py and mllm_scheduler.py). Logged at startup so operators know the active value.
- `CacheStats` now exposes `misses_short_prefix` and `misses_short_lcp` as dedicated counters in `to_dict()`, incremented alongside the existing `_last_match_type` assignments.

**Tests:**
- Added `test_lcp_match_below_floor_is_rejected`: stores a long entry, queries with a short LCP overlap, verifies miss with `misses_short_lcp` counter.
- Added `test_load_from_disk_skips_short_entries`: creates on-disk cache layout with a 10-token entry, loads with `min_prefix_tokens=64`, verifies the entry is skipped before `load_prompt_cache` is called.
- Converted `TestSamplingDefaults` from manual try/finally to `monkeypatch.setattr`.

Validation: 46/46 `test_memory_cache.py`, 2/2 `TestSamplingDefaults`, all files pass `black --check` and `py_compile`.

Thump604:
Rechecked the current head after the review-addressing push. The requested Responses API propagation, prefix-cache-min-tokens CLI surface, short-prefix counters, LCP-below-floor test, load-from-disk short-entry test, and sampling-default test cleanup are present.\n\nFocused verification:\n- `uv run --extra dev pytest tests/test_memory_cache.py tests/test_memory_cache_mlx.py tests/test_kv_cache_quantization.py tests/test_server.py::TestSamplingDefaults tests/test_cli.py` -> 85 passed\n- `uv run --extra dev black --check vllm_mlx/cli.py vllm_mlx/memory_cache.py vllm_mlx/server.py tests/test_cli.py tests/test_kv_cache_quantization.py tests/test_memory_cache.py tests/test_memory_cache_mlx.py tests/test_server.py` -> passed\n\nRequested review from @waybarrios and @janhilgard.

janhilgard (approved):
Both features are clean and well-tested.

### Sampling defaults propagation

The `_resolve_*` pattern (request > CLI default > fallback) is applied uniformly across all four API paths: Chat Completions, Completions streaming, Anthropic adapter, and Responses. The four new `--default-*` CLI flags wire through `serve_command` correctly. Test coverage confirms both propagation and request-override semantics.

### Short prefix-cache reuse gating

`min_prefix_tokens` (default 128) is checked in all relevant paths:
- `store()` — rejects caching short sequences
- `fetch()` exact match — rejects fetching short token lists
- `fetch()` LCP — rejects returning LCP matches below the floor
- `load_from_disk()` — skips short entries on cache recovery

The default of 128 is sensible — avoids polluting cache with trivially short prefixes (system prompts, greetings) that provide minimal prefill savings.

CI 9/9. LGTM.

waybarrios:
I was reviewing this and it seems look! Nice work @Thump604 It is all set to merge!

waybarrios (approved):

PR #426, opened 2026-04-24
Add request cancellation and streaming timeout enforcement

Thump604:
## Summary

I added an explicit request cancellation surface so an operator can abort an active or queued request without restarting the server.

The patch adds:
- `BaseEngine.abort_request(...)` as the default async engine contract
- async-aware `BatchedEngine.abort_request(...)` routing for both MLLM scheduler cancellation and text `AsyncEngineCore.abort_request(...)`
- `POST /v1/requests/{request_id}/cancel`
- `DELETE /v1/requests/{request_id}` as an alias

## Why

The existing streaming disconnect path is useful but not sufficient as the only safety surface. A client can disappear or be killed while the server still has a long generation active. In that state, the safe operator action should be cancelling the request id exposed by `/v1/status`, not restarting the resident process.

## Validation

```bash
/opt/ai-runtime/venv-live/bin/python -m pytest -q tests/test_batched_engine.py tests/test_server.py
# 82 passed, 3 deselected

/opt/ai-runtime/venv-live/bin/black --fast --check \
  vllm_mlx/engine/base.py \
  vllm_mlx/engine/batched.py \
  vllm_mlx/server.py \
  tests/test_batched_engine.py \
  tests/test_server.py

git diff --check
```

I also promoted the same change locally and validated it against a live resident MLLM lane by cancelling an in-flight streaming request from `/v1/status`; the server returned to `num_running=0`, `num_waiting=0` without a restart.

### DIFF
--- a/tests/test_batched_engine.py
+++ b/tests/test_batched_engine.py
@@ -124,3 +124,61 @@ class TestBatchedEngineCacheRestore:
         assert loaded == 2
         scheduler._ensure_batch_generator.assert_called_once_with()
         prefix_cache.load_from_disk.assert_called_once_with("/tmp/cache")
+
+
+class TestBatchedEngineAbortRequest:
+    @pytest.mark.anyio
+    async def test_abort_request_routes_to_mllm_scheduler(self):
+        from vllm_mlx.engine.batched import BatchedEngine
+
+        with patch("vllm_mlx.engine.batched.is_mllm_model", return_value=True):
+            engine = BatchedEngine("test-mllm")
+
+        engine._mllm_scheduler = MagicMock()
+        engine._mllm_scheduler.abort_request.return_value = True
+
+        assert await engine.abort_request("req-1") is True
+        engine._mllm_scheduler.abort_request.assert_called_once_with("req-1")
+
+    @pytest.mark.anyio
+    async def test_abort_request_routes_to_text_engine(self):
+        from vllm_mlx.engine.batched import BatchedEngine
+
+        with patch("vllm_mlx.engine.batched.is_mllm_model", return_value=False):
+            engine = BatchedEngine("test-model")
+
+        engine._loaded = True
+        engine._is_mllm = False
+        engine._engine = MagicMock()
+        engine._engine.abort_request.return_value = True
+
+        assert await engine.abort_request("req-1") is True
+        engine._engine.abort_request.assert_called_once_with("req-1")
+
+    @pytest.mark.anyio
+    async def test_abort_request_routes_to_async_text_engine(self):
+        from vllm_mlx.engine.batched import BatchedEngine
+
+        with patch("vllm_mlx.engine.batched.is_mllm_model", return_value=False):
+            engine = BatchedEngine("test-model")
+
+        engine._loaded = True
+        engine._is_mllm = False
+        engine._engine = MagicMock()
+        engine._engine.abort_request = AsyncMock(return_value=True)
+
+        assert await engine.abort_request("req-1") is True
+        engine._engine.abort_request.assert_awaited_once_with("req-1")
+
+    @pytest.mark.anyio
+    async def test_abort_request_returns_false_without_supported_engine(self):
+        from vllm_mlx.engine.batched import BatchedEngine
+
+        with patch("vllm_mlx.engine.batched.is_mllm_model", return_value=False):
+            engine = BatchedEngine("test-model")
+
+        engine._loaded = True
+        engine._is_mllm = False
+        engine._engine = None
+
+        assert await engine.abort_request("req-1") is False
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -2412,6 +2412,62 @@ class TestReasoningAndToolCallsNonStreaming:
 
 
 # =============================================================================
+# Request Cancellation Tests
+# =============================================================================
+
+
+class TestRequestCancellationEndpoint:
+    @pytest.mark.anyio
+    async def test_cancel_request_routes_to_loaded_engine(self):
+        import vllm_mlx.server as server
+
+        class DummyEngine:
+            is_mllm = False
+
+            async def abort_request(self, request_id):
+                self.request_id = request_id
+                return True
+
+        engine = DummyEngine()
+        old_engine = server._engine
+        old_model_name = server._model_name
+        try:
+            server._engine = engine
+            server._model_name = "test-model"
+
+            result = await server.cancel_request("req-123")
+
+            assert result["cancelled"] is True
+            assert result["id"] == "req-123"
+            assert result["model"] == "test-model"
+            assert engine.request_id == "req-123"
+        finally:
+            server._engine = old_engine
+            server._model_name = old_model_name
+
+    @pytest.mark.anyio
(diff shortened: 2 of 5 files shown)


Thump604:

janhilgard:
@Thump604 The abort routing via `inspect.isawaitable` is clean and the auth gating is correct. Some findings:

### Must fix

**1. Request IDs are not discoverable by API callers**

The cancel endpoint accepts `request_id`, but internal request UUIDs (generated in `add_request()`) are never exposed in API responses. The IDs the client sees (`chatcmpl-xxx`, `cmpl-xxx`) are response-level IDs generated in the formatting layer, not engine-level request IDs.

This means the endpoint is effectively unusable from the client side — it only works if you know the internal UUID (e.g., from server logs or `/v1/status`). Either:
- Expose the internal `request_id` in streaming chunks
- Accept the response-level ID and map it back internally
- Document this explicitly as an operational/admin endpoint

### Should fix

**2. Text scheduler always returns `True` on abort — even for non-existent requests**

`Scheduler.abort_request()` enqueues the ID into `_pending_abort_ids` and always returns `True`, regardless of whether the request exists. This means `POST /v1/requests/nonexistent/cancel` returns 200 with `cancelled: true` when using the text engine. The MLLM scheduler correctly returns `False` for unknown IDs.

The text scheduler should check whether the request is known before returning `True`, or at minimum the return value should reflect "enqueued" vs "confirmed."

**3. Duplicate timeout check logic**

The timeout is enforced at two nearly identical points — before `asyncio.wait` (diff ~lines 285-299) and after (diff ~lines 309-321). The second check is largely dead code because `wait_timeout` is already capped at `remaining_s`, so the post-wait check only fires in a negligible race window. Consolidating to one check at the top of the loop would reduce duplication.

**4. Missing test: `DELETE /v1/requests/{request_id}`**

The DELETE alias endpoint is untested. Low risk since it delegates to `cancel_request`, but a basic route test would confirm the alias is wired correctly.

### Consider

**5. Response code 202 vs 200**

For the text scheduler, abort is deferred (enqueued) — the request may still generate for a few more steps. 200 implies the cancellation is complete, while 202 (Accepted) would better communicate the deferred nature. For the MLLM scheduler, 200 is correct since abort is synchronous.

**6. Additional test coverage**

Missing tests for: 500 error path (abort_request raises), 503 engine-not-loaded, and partial streaming before timeout (current test has 0 chunks before timeout fires).

**7. Test cleanup pattern**

`test_cancel_request_routes_to_loaded_engine` uses manual try/finally to save/restore `server._engine` and `server._model_name`. `monkeypatch.setattr` would be more robust and auto-cleanup.

### What's good

- `inspect.isawaitable` for sync/async polymorphism — correctly handles both MLLM (sync) and text (async) abort
- Auth gating with `verify_api_key` + `check_rate_limit` on both endpoints
- `schedule_removal()` instead of eager `remove()` in MLLM abort — avoids Metal thread-safety issues
- Streaming timeout cancellation properly awaits the cancelled task with `CancelledError`/`StopAsyncIteration` handling
- No input validation concerns — `request_id` is only used as a dict key lookup

janhilgard (changes requested):
### Must fix

1. **Request IDs are not discoverable by API callers** — cancel endpoint accepts internal UUIDs (`request_id`) that are never exposed in API responses. Client-visible IDs (`chatcmpl-xxx`) are response-level, not engine-level. The endpoint is effectively unusable from the client side.

### Should fix

2. **Text scheduler always returns `True` on abort** — even for non-existent requests. MLLM scheduler correctly returns `False` for unknown IDs.
3. **Duplicate timeout check logic** — timeout enforced at two nearly identical points; post-wait check is largely dead code.
4. **Missing test: `DELETE /v1/requests/{request_id}`** — alias endpoint untested.

### Consider

5. Response code 202 vs 200 for deferred (text scheduler) abort
6. Missing tests: 500 error path, 503 engine-not-loaded, partial streaming before timeout
7. Test cleanup: `monkeypatch.setattr` instead of manual try/finally

See [full review comment](https://github.com/waybarrios/vllm-mlx/pull/426#issuecomment-4320036742) for details.

Thump604:
@janhilgard Verified all items against the code. Findings:

**Must fix (1):** Correct. Internal request UUIDs are never exposed in streaming chunks or response objects. The cancel endpoint is currently admin-only (discoverable via /v1/status). Fix: expose request_id in the streaming response metadata so clients can cancel their own requests. Will also document the /v1/status discovery path for operator use.

**Should fix (2):** Correct. Text Scheduler.abort_request() unconditionally returns True without checking if the request exists, while MLLM scheduler correctly returns False for unknown IDs. Fix: check _active_requests / _pending_requests before enqueuing, return False if unknown.

**Should fix (3):** Correct. The post-wait timeout check is effectively dead code since wait_timeout is already capped at remaining_s. Will consolidate to one check at loop top.

**Should fix (4):** Correct. Will add a basic route test for the DELETE alias.

**Consider (5):** Good call on 202 for the text scheduler's deferred abort. Will use 202 for text (enqueued) and 200 for MLLM (synchronous).

**Consider (6-7):** Will add the missing test paths and switch to monkeypatch.setattr.

Working on fixes now.

janhilgard:
@Thump604 All sounds good. The 202/200 split for text vs MLLM scheduler is a nice touch — makes the async vs sync semantics explicit in the response.

janhilgard:
@Thump604 Heads up — PR #450 (streaming timeout in `_disconnect_guard`) was merged into `main` today. Your branch now has merge conflicts because the commit "Enforce streaming request timeouts" overlaps with #450's implementation.

Specifically, #450 already adds:
- `timeout` parameter to `_disconnect_guard()` with absolute wall-clock check
- Deferred `generator.aclose()` to ensure `stream_outputs()` finally-block fires `abort_request()` (prevents scheduler deadlock after timeout)
- All streaming endpoints pass `total_timeout`

Your cancel endpoint + request ID exposure (the unique parts of this PR) are still valuable. To resolve:

1. Rebase on `main`
2. Drop the streaming timeout commit (it's already in main)
3. Keep the cancel endpoint, request ID in SSE chunks, and abort routing

The cancel endpoint complements the timeout nicely — timeout is the automatic safety net, cancel is the explicit operator action.

Thump604:
@janhilgard rebased onto current main with #450 included. Dropped the streaming timeout commit entirely since your implementation covers that cleanly. What remains is just the cancel endpoint, abort routing, and tests.

janhilgard (approved):
Clean rebase — streaming timeout correctly dropped (now in main via #450). The cancel endpoint implementation looks good:

- `BaseEngine.abort_request()` with default False + BatchedEngine routing to MLLM scheduler or text engine
- `POST /v1/requests/{request_id}/cancel` + `DELETE` alias
- Proper 404 for unknown IDs, 500 for engine failures
- Test coverage for routing, 404, and server integration

One note: the request ID discoverability works because the `chatcmpl-xxx` ID from SSE chunks is the same ID passed to the scheduler — client gets it from the first streaming chunk. Worth a brief mention in the endpoint docstring.

LGTM.

Thump604:
Added the docstring note Jan suggested. Approved and CI green. Wayner, review and merge when ready?

PR #427, opened 2026-04-24
Add bench-serve workload contracts

Thump604:
## What changed

- Added `vllm-mlx bench-serve --workload` for declarative product-style serving contracts.
- Workload cases can define request settings, quality checks, tags, and comparison-only `policy_timeout_ms`.
- The runner records runtime provenance, hardware fingerprint, `/metrics` deltas, `/v1/status` Metal/cache data, quality failures, and policy-timeout pass/fail separately.
- Added workload output formats using the existing `--format` surface: JSON, CSV, SQL, and table.
- Added an example workload and docs for using contract workloads as model and feature-stack qualification inputs.

## Why

Prompt sweeps are useful for raw serving performance, but they are not enough to qualify real application contracts. This adds a durable upstream path for repeatable workload qualification where quality gates, cache metrics, runtime metadata, and product policy comparisons are captured in one artifact.

`policy_timeout_ms` is deliberately reported as comparison evidence, not as an implicit hardware SLA. That avoids turning arbitrary app guardrails into benchmark truth.

## Validation

- `uvx black --fast --check vllm_mlx/bench_serve.py vllm_mlx/cli.py tests/test_bench_serve.py`
- `python -m pytest -q tests/test_bench_serve.py`
- `python -m py_compile vllm_mlx/bench_serve.py vllm_mlx/cli.py`
- `uvx ruff check --select I vllm_mlx/bench_serve.py vllm_mlx/cli.py tests/test_bench_serve.py`
- CLI smoke for workload JSON, CSV, and SQL output against an unavailable local URL to verify error and policy-timeout recording

## Note

I also tried `python -m pytest -q tests/test_metrics.py tests/test_bench_serve.py`; `tests/test_metrics.py` is blocked in this local checkout by missing `fastapi`. CI installs `fastapi`, so I am leaving that to CI rather than changing dependency setup in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,9 @@ python examples/tts_multilingual.py "Hola mundo" --lang es --play
 
 ```bash
 vllm-mlx bench-serve --url http://localhost:8000 --concurrency 5 --prompts prompts.txt --output results.csv
+
+# Product-style workload with quality checks and metrics deltas
+vllm-mlx bench-serve --url http://localhost:8000 --workload workload.json --output results.json
 ```
 
 ### Model acquisition and conversion
--- a/README.md
+++ b/README.md
@@ -190,7 +190,7 @@ python examples/tts_multilingual.py "Hola mundo" --lang es --play
 vllm-mlx bench-serve --url http://localhost:8000 --concurrency 5 --prompts prompts.txt --output results.csv
 
 # Product-style workload with quality checks and metrics deltas
-vllm-mlx bench-serve --url http://localhost:8000 --workload workload.json --output results.json
+vllm-mlx bench-serve --url http://localhost:8000 --workload workload.json --repetitions 5 --output results.json
 ```
 
 ### Model acquisition and conversion
--- a/README.md
+++ b/README.md
@@ -191,6 +191,9 @@ vllm-mlx bench-serve --url http://localhost:8000 --concurrency 5 --prompts promp
 
 # Product-style workload with quality checks and metrics deltas
 vllm-mlx bench-serve --url http://localhost:8000 --workload workload.json --repetitions 5 --output results.json
+
+# Append workload rows into SQLite for longitudinal comparisons
+vllm-mlx bench-serve --url http://localhost:8000 --workload workload.json --repetitions 5 --format sqlite --output bench.db
 ```
 
 ### Model acquisition and conversion
--- a/docs/benchmarks/README.md
+++ b/docs/benchmarks/README.md
@@ -35,6 +35,8 @@ vllm-mlx bench-serve --url http://localhost:8000 \
 running OpenAI-compatible server. This is intended for model and feature-stack
 qualification, where raw speed is not enough and every run needs provenance,
 quality checks, Prometheus metric deltas, and policy-timeout evidence.
+Use `--repetitions` to measure variance; workload summaries report per-case
+sample counts, failure rates, and min/median/max latency and throughput.
 
 Example workload:
 
@@ -94,6 +96,11 @@ policy?" after first measuring what the model and serving stack can actually do.
 Workload output defaults to JSON for full provenance. Use `--format csv` for
 flat per-case rows or `--format sql` to emit importable SQL for a local
 benchmark database.
+
+```bash
+vllm-mlx bench-serve --url http://localhost:8000 \
+  --workload ./workload.json --repetitions 5 --output workload-results.json
+```
 
 ## Standalone Test Defaults
 
--- a/docs/benchmarks/README.md
+++ b/docs/benchmarks/README.md
@@ -64,6 +64,29 @@ Example workload:
 }
 ```
 
+Cases can also reference an existing OpenAI-compatible request JSON instead of
+duplicating a large prompt body:
+
+```json
+{
+  "name": "writing-contract",
+  "cases": [
+    {
+      "id": "resume-golden-1",
+      "request_path": "./fixtures/job543_resume_precise_request.json",
+      "checks": {
+        "finish_reason": "stop",
+        "forbidden_regex": ["<think>"]
+      }
+    }
+  ]
+}
+```
+
+When `request_path` is used, `messages`, `max_tokens`, `enable_thinking`, and
+extra request-body fields such as `thinking_token_budget` are read from that
+file. Case-level `extra_body` values override request-file values.
+
 `policy_timeout_ms` is recorded as comparison evidence. It is not treated as a
 hardware capability claim. Use it to answer "would this run fit my product
 policy?" after first measuring what the model and serving stack can actually do.
--- a/docs/benchmarks/README.md
+++ b/docs/benchmarks/README.md
@@ -19,7 +19,58 @@ vllm-mlx-bench --model mlx-community/Qwen3-VL-8B-Instruct-4bit
 
 # Video benchmark
 vllm-mlx-bench --model mlx-community/Qwen3-VL-8B-Instruct-4bit --video
+
+# Running-server prompt sweep with Prometheus metric deltas
+vllm-mlx bench-serve --url http://localhost:8000 --prompts short,long \
+  --concurrency 1,4 --output bench.json --format json
+
+# Running-server product-style workload with quality checks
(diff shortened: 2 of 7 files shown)


Thump604:

janhilgard:
## Review: PR #427 — bench-serve workload contracts

Solid feature scope — declarative workload contracts with quality checks, policy-timeout evidence, and SQLite longitudinal output is a real upgrade over raw prompt sweeps. The record schema (provenance, hardware fingerprint, cache deltas, quality issues) is well thought out. A few things to address before merge:

---

### Must fix

#### 1. `checks` override instead of merge — inconsistent with `extra_body`

```python
# bench_serve.py, load_workload()
checks = item.get("checks", defaults.get("checks"))
```

Case-level `checks` completely **replaces** default checks — no merge. But `extra_body` IS merged (`request_extra.update(extra_body)`). This inconsistency will bite users:

```json
{
  "defaults": {
    "checks": { "finish_reason": "stop", "forbidden_regex": ["<think>"], "min_chars": 100 }
  },
  "cases": [{
    "id": "code-case",
    "checks": { "required_regex": ["def "] }
  }]
}
```

Result: case loses `finish_reason`, `forbidden_regex`, and `min_chars` from defaults. User expects merge (like `extra_body`), gets full replacement.

Fix: merge checks like extra_body:
```python
merged_checks = dict(defaults.get("checks") or {})
if item.get("checks"):
    merged_checks.update(item["checks"])
checks = merged_checks or None
```

For list-valued keys (`required_regex`, `forbidden_regex`) you may want to concatenate rather than replace, but at minimum the top-level dict keys should merge.

#### 2. `--workload` silently ignores sweep-only args

```bash
vllm-mlx bench-serve --workload w.json --prompts long --concurrency 8 --warmup 5
# --prompts, --concurrency, --warmup all silently ignored
```

Users who mix modes by mistake will get confusing results. Either:
- Emit a warning when sweep-only args are non-default with `--workload`
- Or use `argparse` mutually exclusive groups

---

### Should fix

#### 3. `note` string duplicated in every record

```python
"note": "comparison-only unless your product contract explicitly requires it",
```

This string appears in **every** result record. For a 50-case × 5-rep workload, that is 250 copies of the same sentence in the JSON output. Move it to `payload["policy"]["note"]` at the top level.

#### 4. `DELETE /v1/cache` endpoint dependency undocumented

`clear_runtime_cache()` calls `DELETE /v1/cache`, but this endpoint is not mentioned in docs or the PR body. Users who set `cache_policy: "before-case"` against a server without this endpoint will get silent `ok: false` events. Add a note in the docs about the endpoint requirement.

#### 5. No progress indication for long workloads

For a workload with 10 cases × 5 repetitions, the runner is completely silent until all 50 requests complete. Add a simple progress line:
```python
print(f"  [{repetition+1}/{repetitions}] {case.case_id} ...", end="", flush=True)
# after completion:
print(f" {result[metrics][e2e_latency_ms]:.0f}ms {✓ if result[ok] else ✗}")
```

---

### Test coverage gaps

#### 6. Invalid regex in checks

`validate_quality_checks` catches `re.error` and appends to issues (good!), but no test covers this path:

```python
def test_invalid_regex_in_checks(self):
    ok, issues = validate_quality_checks(
        "stop", "content", {"required_regex": ["[unclosed"]}
    )
    assert ok is False
    assert any("invalid required_regex" in i for i in issues)
```

#### 7. Checks override behavior untested

No test verifies the current override semantics. Add a test that demonstrates a case with `checks` loses default checks, so the behavior is at least documented by tests (and will catch future changes).

#### 8. Schema / column count assertion

`WORKLOAD_RESULT_COLUMNS` (39 entries) and `_WORKLOAD_SQL_SCHEMA` are maintained separately. If one is updated without the other, SQLite writes silently break. Add:

```python
def test_workload_columns_match_sql_schema():
    schema_cols = [col.split()[0] for col in _WORKLOAD_SQL_SCHEMA.split(", ")]
    assert schema_cols == WORKLOAD_RESULT_COLUMNS
```

---

### Minor / consider

- **`request_path` + `messages` precedence**: if both are provided, case `messages` wins. This is reasonable but the docs should mention it explicitly — currently only says "Cases can also reference an existing request JSON **instead of** duplicating a large prompt body" (emphasis on "instead of").

- **Workload name in SQL**: workload name goes into SQL output (`format_workload_sql`) but is not validated during `load_workload`. A name containing a single quote will produce invalid SQL. Consider sanitizing or validating early.

- **`_message_list` validation strictness**: `_require_message_list` requires both `role` and `content` in every message. Tool-call messages (`role: "assistant"` with `tool_calls` instead of `content`) would be rejected. For benchmarking fresh conversations this is fine, but `request_path` might reference multi-turn conversations with tool calls.

---

### What is good

- Policy timeout as comparison-only evidence (not a hard gate) — clean separation of measurement vs. policy
- `_validate_sql_identifier` — solid defense against SQL injection
- Interleaved repetitions (case-a rep0, case-b rep0, case-a rep1, ...) — reduces cache contamination
- Legacy metal key support (`active_gb` → `active_memory_gb` fallback)
- `_normalize_cache_policy` with underscore→hyphen normalization
- Test coverage is thorough — 637 new test lines cover loading, quality checks, runner, formatters, and SQLite

janhilgard (changes requested):
Two items need fixing before merge:

1. **`checks` override instead of merge** — case-level `checks` completely replaces defaults (inconsistent with `extra_body` which merges). Users will silently lose default quality gates when adding case-specific checks.

2. **`--workload` silently ignores sweep-only args** — `--prompts`, `--concurrency`, `--warmup` are discarded without warning when `--workload` is set.

See full review details in the [comment above](https://github.com/waybarrios/vllm-mlx/pull/427#issuecomment-4320050202) (additional should-fix items, test coverage gaps, and minor notes).

Thump604:
@janhilgard Agree with the two blockers and the should-fix items.

**Must-fix (1):** `checks` should merge with defaults the same way `extra_body` does. I'll do a shallow dict merge, and for list-valued keys (`required_regex`, `forbidden_regex`) I'll concatenate rather than replace. That matches the mental model of "case adds constraints on top of defaults."

**Must-fix (2):** Will emit a warning when sweep-only args (`--prompts`, `--concurrency`, `--warmup`) are non-default alongside `--workload`. Mutually exclusive groups felt too rigid since someone might want `--warmup` to apply globally in a future iteration, but the warning makes the discard visible.

**Should-fix (3):** Agreed, the per-record `note` string is wasteful at scale. Moving it to `payload["policy"]["note"]`.

**Should-fix (4):** Will document the `DELETE /v1/cache` dependency in the workload contract docs and have the runner surface a clear error when the endpoint returns non-200.

**Should-fix (5):** Will add a progress line per case/repetition.

Will address the test coverage gaps in the same push.

janhilgard (approved):
Both must-fix items from my previous review are addressed:

1. **Checks merge** — Case-level checks now properly merge with defaults: list-valued keys (`required_regex`, `forbidden_regex`) are concatenated, scalars are overridden. Covered by `TestWorkloadChecksMerge` (4 tests).

2. **Sweep-only args warning** — `cli.py` lines 2080-2095 detect non-default `--prompts`/`--concurrency`/`--warmup` when `--workload` is set and print explicit stderr warning listing the ignored args.

The rest looks solid:
- SQL injection prevention via whitelist regex in `_validate_sql_identifier()`
- Legacy metal metric key backward compat (`active_memory_gb` / `active_gb`)
- Comprehensive quality check validation (catches invalid regex patterns gracefully)
- Request path resolution relative to workload file directory
- Clear policy_timeout_ms semantics (comparison evidence, not hardware SLA)
- Excellent test coverage (90%+ of new code paths)

LGTM.

janhilgard:
Merged all the other approved PRs. This one has merge conflicts now — could you rebase onto main? Thanks.

Thump604:
Follow-up from the two-week merge-readiness audit: I opened #482 to track the process failure here. This PR was self-merged before the post-rebase CI completed, and that CI then failed because main already contained the shared raw_request issue from #438. The concrete code failure was later fixed by 0c0dd3d, but the self-merge / green-CI gate needs to be corrected as policy.

PR #428, opened 2026-04-24
feat(engine): add raw_output to preserve unstripped text for parsers

Thump604:
## Summary

- Adds `raw_output` parameter to `generate()`, `chat()`, `stream_generate()`, and `stream_chat()` on both `SimpleEngine` and `BatchedEngine`
- When `raw_output=True`, `clean_output_text()` is skipped so reasoning/tool parsers receive unstripped model output
- Server non-streaming chat paths (OpenAI `/v1/chat/completions` and Anthropic `/v1/messages`) now pass `raw_output=True` to `engine.chat()`, fixing a double-clean bug where special tokens were stripped before the reasoning parser could see them

## Motivation

The reasoning parser (`extract_reasoning()`) and tool call parser need to see raw model output including special tokens like `<think>`, `</think>`, `<|channel|>`, etc. Previously, `engine.chat()` applied `clean_output_text()` which stripped these tokens, and then the server applied it again on the result. This caused:
1. Reasoning tokens stripped before the parser could extract them (non-streaming path)
2. Redundant double-cleaning

The streaming path (`stream_chat`) was already effectively raw — this change adds the parameter for API consistency and ensures it doesn't leak into sub-calls.

Supersedes #255 with a narrower, single-commit change rebased on current main.

## Test plan

- [x] `tests/test_simple_engine.py` — 8/8 pass
- [ ] Manual: non-streaming chat with reasoning model returns correct `reasoning_content`
- [ ] Manual: non-streaming chat with tool calls returns correct `tool_calls`
- [ ] Verify streaming path behavior unchanged

### DIFF
--- a/vllm_mlx/engine/batched.py
+++ b/vllm_mlx/engine/batched.py
@@ -576,6 +576,8 @@ class BatchedEngine(BaseEngine):
         if not self._loaded:
             await self.start()
 
+        raw_output = bool(kwargs.pop("raw_output", False))
+
         if self._is_mllm and self._mllm_scheduler:
             # Use MLLM scheduler for all requests when model is multimodal.
             # MLLM models only initialise the _mllm_scheduler (not _engine),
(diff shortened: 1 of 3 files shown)


Thump604:
Superseded by #435, which is a clean re-implementation against current main. This PR remains open for reference.

Thump604:
Closing -- superseded by #435 which has the updated implementation.

PR #430, opened 2026-04-24
Add Gemma 4 audio support to chat completions

Thump604:
## Summary
- add audio extraction to chat-completions preprocessing and batched multimodal scheduling
- teach the MLLM chat and streaming paths to accept OpenAI-style `audio_url` / base64 audio content parts
- plumb `audio` through BatchedEngine, the MLLM scheduler, and batch request preprocessing so Gemma 4 processors can emit `input_features` and `input_features_mask`
- add regression coverage for audio content extraction, audio input processing, and batched message preparation

## Validation
- `PYTHONPATH=/Users/David/code/vllm-mlx /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_api_utils.py tests/test_server.py tests/test_native_tool_format.py tests/test_mllm.py tests/test_mllm_mtp_routing.py -q`
- real `/v1/chat/completions` smoke against local `gemma-4-E2B-it-5bit` on both engine paths with a short WAV request
  - BatchedEngine response: `Hello world from Gemma audio test`
  - SimpleEngine response: `Hello world from Gemma audio test.`

## Note on local validation target
The local `gemma-4-E2B-it-5bit` MLX export has the audio tower weights, but its exported `processor_config.json` was missing the official `feature_extractor` block. For the endpoint smoke only, I used a temporary validation copy of the model directory with the official Gemma 4 E2B audio feature extractor config restored. The code in this PR does not depend on that workaround; it just made the local smoke exercise the real audio path end-to-end.

### DIFF
--- a/tests/test_api_utils.py
+++ b/tests/test_api_utils.py
@@ -201,17 +201,18 @@ class TestExtractMultimodalContent:
             Message(role="system", content="You are helpful."),
             Message(role="user", content="Hello"),
         ]
-        processed, images, videos = extract_multimodal_content(messages)
+        processed, images, videos, audios = extract_multimodal_content(messages)
 
         assert len(processed) == 2
         assert processed[0] == {"role": "system", "content": "You are helpful."}
         assert processed[1] == {"role": "user", "content": "Hello"}
         assert images == []
         assert videos == []
+        assert audios == []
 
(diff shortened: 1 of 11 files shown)


Thump604:
Closing -- superseded by #439 which has the updated implementation.

PR #432, opened 2026-04-25
fix: don't suppress whitespace-only deltas after partial-markup check

### LINKED ISSUE #431: streaming: whitespace-only deltas dropped when tool parser is active, breaking markdown layout
## Bug Description

In streaming mode, whitespace-only content deltas are silently dropped under some `--tool-call-parser` configurations. Multi-paragraph prose, code blocks, and bullet lists collapse into one unbroken blob client-side. Non-streaming responses are unaffected.

## Environment

- macOS, Apple Silicon (M3 Ultra)
- vllm-mlx built from `waybarrios/vllm-mlx@93f9fb8` (current `main`)
- Reproduced on three model / parser combinations:

| model | `--tool-call-parser` | affected? |
| --- | --- | --- |
| `mlx-community/Qwen3.6-35B-A3B-8bit` | `qwen` | yes |
| `mlx-community/MiniMax-M2.7-8bit` | `minimax` | yes |
| `mlx-community/Qwen3-Coder-Next-8bit` | `qwen3_coder` | no |

## Reproduction

```bash
curl -sN -X POST http://127.0.0.1:8002/v1/chat/completions \
  -d '{
    "model": "mlx-community/Qwen3.6-35B-A3B-8bit",
    "messages": [{"role": "user", "content": "write three short sentences on three lines"}],
    "max_tokens": 500,
    "stream": true
  }' \
  | sed -n 's/^data: //p' \
  | jq -r '.choices[0].delta.content // empty'
```

Expected: three sentences separated by blank lines.

Actual on the `qwen` and `minimax` parsers: every line printed in the output is non-whitespace content (`The`, ` sun`, `.`, `Birds`, ...). Whitespace-only tokens such as `"\n\n"` never arrive.

Sanity check that the model does emit whitespace tokens: the same prompt non-streaming on minimax returns `"Hello.  \nNice to meet you.  \nGoodbye."`.

## Diagnosis

`vllm_mlx/server.py:~4884`:

```python
if (
    not tool_markup_possible
    and not _streaming_tool_markup_possible(tool_accumulated_text + content)
):
    tool_accumulated_text += content
    if not content.strip():
        continue   # drops legitimate prose whitespace
```

This branch runs only when `_streaming_tool_markup_possible()` has already determined the text cannot be the start (or partial prefix) of tool-call XML. At that point the suppression isn't protecting any tool-call use case; it's only stripping legitimate whitespace.

The "stray newlines before tool call XML" rationale in the comment appears to predate `_streaming_tool_markup_possible()` (which now does smart partial-markup detection upstream). Possibly a leftover from when the simpler `"<" not in content` check was replaced.

The `qwen3_coder` parser appears to bypass the bug, likely because something in its parser path causes `_streaming_tool_markup_possible()` to return `True` earlier and push down the `else` branch. Didn't dig deeper into why.

I'll put up a PR with a proposed fix shortly.

luizribeiro:
Fixes #431.

The whitespace suppression block in `chat_completions` streaming runs after `_streaming_tool_markup_possible()` has already determined the buffered text cannot be the start (or partial prefix) of tool-call XML. At that point it isn't guarding against any tool-call use case; it only strips legitimate prose whitespace, collapsing multi-paragraph output into one unbroken blob client-side.

Removing the four-line block. Buffered text is still accumulated for downstream tool-call detection.

## Verification

Reproducer from the linked issue, against `mlx-community/Qwen3.6-35B-A3B-8bit` with `--tool-call-parser qwen`:

```bash
curl -sN -X POST http://127.0.0.1:8002/v1/chat/completions \
  -d '{
    "model": "mlx-community/Qwen3.6-35B-A3B-8bit",
    "messages": [{"role": "user", "content": "write three short sentences on three lines"}],
    "max_tokens": 500,
    "stream": true
  }' \
  | sed -n 's/^data: //p' \
  | jq -r '.choices[0].delta.content // empty'
```

Before: every emitted line is non-whitespace content; `"\n\n"` deltas missing.

After: whitespace-only deltas (`"\n\n"`, `"  \n"`, etc.) appear in the stream as expected.

Tool-call streaming verified intact on `--tool-call-parser qwen` and `--tool-call-parser minimax` in production.

### DIFF
--- a/vllm_mlx/server.py
+++ b/vllm_mlx/server.py
@@ -5015,10 +5015,7 @@ async def stream_chat_completion(
                         )
                     ):
                         tool_accumulated_text += content
-                        # Suppress whitespace-only content when tools are active;
-                        # avoids emitting stray newlines before tool call XML.
-                        if not content.strip():
-                            continue
+                        # Emit as-is; no tool markup is even possible yet.
                     else:
                         if not tool_markup_possible:
                             tool_markup_possible = True


Thump604 (approved):
Cleaner than our #464. Correct reasoning — suppression is unnecessary after the markup-possible check already passed. Should merge.

PR #434, opened 2026-04-25
Fix qwen3_coder parser: wire Qwen3XMLToolParser

Thump604:
## Summary

- `qwen3_coder` was incorrectly aliased to `HermesToolParser` (Hermes JSON format)
- Qwen 3.5/3.6 uses XML parameter format: `<function=name><parameter=key>value</parameter></function>`
- Add `Qwen3XMLToolParser` (ported from vLLM PR #25028, Qwen API team)
- Register as `qwen3_xml`, `qwen3.5`, `qwen3_coder`
- Remove `qwen3_coder` from `HermesToolParser` registration
- Import in `__init__.py`

## Test plan

- [x] 4 registration tests verifying all aliases resolve correctly
- [x] `hermes`/`nous` still resolve to `HermesToolParser`

### DIFF
--- a/tests/test_qwen3_xml_parser.py
+++ b/tests/test_qwen3_xml_parser.py
@@ -1,0 +1,213 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Functional tests for Qwen3XMLToolParser parsing logic."""
+
+import json
+
+import pytest
+
+pytest.importorskip("transformers")
+
+from vllm_mlx.tool_parsers.qwen3_xml_tool_parser import (
+    Qwen3XMLToolParser,
+    StreamingXMLToolCallParser,
+)
+
+
+@pytest.fixture
+def parser():
+    return Qwen3XMLToolParser(None)
+
+
+@pytest.fixture
+def streaming_parser():
+    return StreamingXMLToolCallParser()
+
+
+class TestNonStreamingExtraction:
+    """Non-streaming extract_tool_calls on complete text."""
+
+    def test_single_tool_call(self, parser):
+        text = (
+            "<tool_call>\n"
+            "<function=get_weather>\n"
+            "<parameter=city>Tokyo</parameter>\n"
+            "</function>\n"
+            "</tool_call>"
+        )
+        result = parser.extract_tool_calls(text)
+        assert result.tools_called
+        assert len(result.tool_calls) == 1
+        assert result.tool_calls[0]["name"] == "get_weather"
+        args = json.loads(result.tool_calls[0]["arguments"])
+        assert args["city"] == "Tokyo"
+
+    def test_multiple_tool_calls(self, parser):
+        text = (
+            "<tool_call>\n"
+            "<function=read_file>\n"
+            "<parameter=path>/a.py</parameter>\n"
+            "</function>\n"
+            "</tool_call>\n"
+            "<tool_call>\n"
+            "<function=write_file>\n"
+            "<parameter=path>/b.py</parameter>\n"
+            "<parameter=content>hello</parameter>\n"
+            "</function>\n"
+            "</tool_call>"
+        )
+        result = parser.extract_tool_calls(text)
+        assert result.tools_called
+        assert len(result.tool_calls) == 2
+        assert result.tool_calls[0]["name"] == "read_file"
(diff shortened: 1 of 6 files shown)


janhilgard (changes requested):
@Thump604 Thanks for porting the Qwen3 XML parser — the expat-based streaming approach with type coercion and auto-closing is solid, and the adapter pattern for vllm-mlx is clean. A few things:

### Must fix

**1. No functional tests for 1389-line parser**

`test_qwen3_xml_registration.py` only covers parser registration (4 tests). The actual parsing logic — `StreamingXMLToolCallParser` with its preprocessing, deferred parsing, type coercion, auto-closing of malformed XML, multi-tool-call handling, and streaming state machine — has zero functional test coverage.

For a parser this complex with this many code paths (deferred vs. immediate parsing, string/int/float/bool/object type handling, XML error recovery, inter-tool-call text buffering), some key tests are needed:

- Single tool call extraction (non-streaming)
- Multi-tool-call in one response
- Streaming: incremental delta output across chunk boundaries
- Type coercion: int, float, bool, object/array (deferred path), string with special characters
- Malformed XML: unclosed tags, missing function name, empty parameters
- `strip_think_tags` interaction with tool call content

**2. `qwen3_coder` alias change is undocumented**

Moving `qwen3_coder` from `HermesToolParser` to `Qwen3XMLToolParser` is a breaking behavioral change. Users with `--tool-call-parser qwen3_coder` in their configs will silently switch from Hermes-style JSON tool calls to XML-style tool calls. This should be called out in the PR description and ideally in a CHANGELOG or migration note.

### Should fix

**3. No docstring on `StreamingXMLToolCallParser`**

The class is 600+ lines with complex state management (preprocessing pass, main XML parsing, deferred parameter evaluation, auto-closing). A class-level docstring explaining the overall architecture would help future maintainers.

**4. Preprocessing regex is fragile for edge cases**

The `_preprocess_before_xml_parse` method uses regex to extract parameter types from raw XML before feeding to expat. While expat handles the actual parsing, the regex could fail on edge cases like attribute values containing quotes or angle brackets. This works for well-formed model output but is a concern for malformed XML — which the parser explicitly tries to handle.

### Minor

- `_ToolDef`/`_FunctionDef`/`_ParamsDef` wrapper classes could be replaced with `types.SimpleNamespace` for less boilerplate
- Some comments read like machine translations (e.g., "If about to start new tool_call, and there are unclosed tool_calls, close tool_call first") — fine functionally but could be polished
- `expat.ParserCreate()` without explicit encoding — consider `ParserCreate('utf-8')` for robustness
- `repair_param_type` has broad type matching (`uint`, `long`, `short`, `unsigned`, etc.) that seems overengineered for JSON Schema types — only `string`, `integer`, `number`, `boolean`, `object`, `array` are standard

Thump604:
Addressed in 9aae4f6:

**Must fix 1 (functional tests):** Added 11 tests covering: single/multi tool call extraction, type coercion (int, bool with schema), think tag stripping, multiline parameters, special characters (>, &), streaming delta production, malformed XML (unclosed tool_call, missing function close), and no-tool-calls case. 15 total tests now (4 registration + 11 functional).

**Must fix 2 (breaking change):** Documented in commit message. Users with `--tool-call-parser qwen3_coder` will silently switch to the correct XML parser for Qwen 3.5/3.6. The previous Hermes JSON parser was wrong for these models.

**Should fix 3 (docstring):** Added class-level docstring to `StreamingXMLToolCallParser` covering the four-stage architecture (preprocessing, expat parsing, type coercion, auto-closing) and streaming protocol.

**Should fix 4 (regex fragility):** Acknowledged -- the preprocessing regex works on well-formed model output. The auto-closing logic handles most malformed cases. The expat parser is the actual parsing authority. Added two malformed XML tests to demonstrate the recovery works.

janhilgard (approved):
All items addressed in 9aae4f6. 11 functional tests covering the key paths (single/multi tool call, type coercion, streaming, malformed XML, think tag stripping), class docstring added, breaking change documented. LGTM.

Thump604:
@waybarrios This has been reviewed and approved by @janhilgard. When you have a chance, could you take a look and merge if it looks good?

Thump604:
This is approved and CI green. Wayner, could you take a look and merge when ready?

PR #435, opened 2026-04-25
Add raw_output to preserve unstripped text for parsers

Thump604:
## Summary

- Add `raw_output: bool = False` to `generate()`, `chat()`, `stream_generate()`, `stream_chat()` on both `SimpleEngine` and `BatchedEngine`
- When `raw_output=True`, `clean_output_text()` is skipped so reasoning/tool parsers receive the model's original special tokens
- Server non-streaming chat paths (OpenAI `/v1/chat/completions`, Anthropic `/v1/messages`, Responses API) now pass `raw_output=True`

Fixes a double-clean bug where special tokens were stripped before the reasoning parser could see them in non-streaming responses.

Re-implementation of #428 against current main.

## Test plan

- [x] 146 existing tests pass (test_simple_engine + test_reasoning_parser)
- [ ] Manual: non-streaming chat with reasoning model returns correct `reasoning_content`
- [ ] Manual: streaming path behavior unchanged

### DIFF
--- a/vllm_mlx/engine/base.py
+++ b/vllm_mlx/engine/base.py
@@ -154,6 +154,7 @@ class BaseEngine(ABC):
         temperature: float = 0.7,
         top_p: float = 0.9,
         stop: list[str] | None = None,
+        raw_output: bool = False,
         **kwargs,
     ) -> GenerationOutput:
         """
@@ -180,6 +181,7 @@ class BaseEngine(ABC):
         temperature: float = 0.7,
         top_p: float = 0.9,
         stop: list[str] | None = None,
+        raw_output: bool = False,
         **kwargs,
     ) -> AsyncIterator[GenerationOutput]:
         """
@@ -208,6 +210,7 @@ class BaseEngine(ABC):
         tools: list[dict] | None = None,
         images: list[str] | None = None,
         videos: list[str] | None = None,
+        raw_output: bool = False,
         **kwargs,
     ) -> GenerationOutput:
         """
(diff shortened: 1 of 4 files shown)


janhilgard (commented):
@Thump604 Clean and focused change — the `raw_output` parameter is a good solution for the double-clean problem where `clean_output_text()` was stripping tokens that downstream parsers (reasoning, tool call) need to see. A couple of notes:

### Should fix

**1. Streaming server paths still clean output**

The diff adds `raw_output=True` to non-streaming server paths (chat completions, Anthropic messages, Responses API) but not to streaming paths. If `clean_output_text()` strips content that streaming parsers need (e.g., think tags in accumulated `output_text`), streaming and non-streaming will produce different output for the same input.

If streaming doesn't need `raw_output=True` because it operates on raw `new_text` deltas before any cleaning, it's worth a brief comment in the code explaining why.

**2. No test coverage for the new parameter**

A simple test that verifies `clean_output_text()` is skipped when `raw_output=True` (and called when `False`) would close this gap. Even a unit test on the engine methods would suffice — doesn't need a full server integration test.

### Minor

- `SimpleEngine.stream_generate` accepts `raw_output` in the signature but the diff doesn't show it being used inside the streaming loop — verify it's not silently ignored
- The change is one-directional: server always passes `raw_output=True`. Consider whether this should be the default for server paths (i.e., flipping the default to `True` in server context and `False` for direct engine callers) — would be cleaner than passing the flag everywhere

Overall this is a solid fix for a real problem. The pattern of letting the caller decide whether to clean output is the right abstraction.

Thump604:
Closing this PR. The raw_output feature is needed but this is the 4th attempt (#255, #358, #428, #435) and each time the broad 4-file scope conflicts with other merges before review lands.

Will refile as a narrower PR against current main after #466, #469, #470 merge. The engine-side change and server-side change should be separate PRs to reduce conflict surface.

Tracking in the runtime STATUS.md open tasks table.

PR #379, opened 2026-04-20
feat: thinking-aware logits processor with token budget

Thump604:
## Summary

- Add `ThinkingAwareLogitsProcessor` with a four-phase state machine
  (IDLE/THINKING/TRANSITIONING/CONTENT) that manages the thinking lifecycle
  for reasoning models (Qwen 3.x, DeepSeek-R1, etc.)
- `thinking_token_budget` request parameter caps reasoning tokens as a
  sub-limit of `max_tokens` via logits-level enforcement
- Thinking-aware constrained decoding: JSON/schema enforcer only activates
  during CONTENT phase, replacing the force-thinking-off workaround
- `usage.reasoning_tokens` in responses for token accounting
- `--default-thinking-token-budget` server CLI argument
- `/v1/messages` (Anthropic) parity via adapter forwarding

## Motivation

Thinking models can spend their entire `max_tokens` budget on reasoning
without ever producing content (vllm-project/vllm#38894). The current
workaround forces thinking off when constrained decoding is active, which
loses reasoning quality. This feature makes thinking and constrained
decoding coexist correctly.

## Design

When `thinking_token_budget` is set, a logits processor tracks tokens
inside `<think>...</think>`. When the budget is exhausted, it forces the
reasoning end sequence by masking logits, then transitions to content
phase. During content phase, it delegates to an inner JSON/schema
enforcer if present.

Budget within `max_tokens`: if `max_tokens=32768` and
`thinking_token_budget=8192`, the model gets up to 8K for thinking and the
remaining 24K+ for content.

## Phase 2 roadmap

Budget-only requests (no inner JSON constraint) can re-enable MTP after
the reasoning phase ends, recovering speculative decoding speedup for
content generation.

## Test plan

- [ ] 40 offline unit tests covering state machine transitions, budget
  enforcement, inner processor delegation, bounded suffix matcher, API
  models, Anthropic adapter forwarding, and decision tree coverage
- [ ] 2 env-gated integration tests (thinking budget cap, thinking+JSON)
- [ ] Verify no regressions in existing test suites

PR #436, opened 2026-04-25
Add thinking-aware logits processor with token budget

Thump604:
## Summary

- `ThinkingAwareLogitsProcessor` with a four-phase state machine (IDLE/THINKING/TRANSITIONING/CONTENT) that manages the thinking lifecycle for reasoning models
- `thinking_token_budget` request parameter caps reasoning tokens as a sub-limit of `max_tokens` via logits-level enforcement
- Thinking-aware constrained decoding: JSON/schema enforcer only activates during CONTENT phase
- `usage.reasoning_tokens` in responses for token accounting
- `--default-thinking-token-budget` server CLI argument

## Motivation

Thinking models can spend their entire `max_tokens` budget on reasoning without ever producing content (vllm-project/vllm#38894). The current workaround forces thinking off when constrained decoding is active, which loses reasoning quality. This feature makes thinking and constrained decoding coexist correctly.

## Design

When `thinking_token_budget` is set, the processor tracks tokens inside `<think>...</think>`. When the budget is exhausted, it forces the reasoning end sequence by masking logits, then transitions to content phase. During content phase, it delegates to an inner JSON/schema enforcer if present.

Re-implementation of #379 against current main. Narrowed to core processor, OpenAI API, and server wiring. Anthropic adapter can follow separately.

## Test plan

- [x] 21 unit tests covering: suffix matcher, lifecycle transitions, budget enforcement, budget-zero edge case, multi-token end sequences, inner delegation, content-phase control token masking, speculative rollback, 2-D logits shapes
- [x] Existing tests pass (test_simple_engine + test_reasoning_parser: 146 passed)

### DIFF
--- a/tests/test_thinking_processor.py
+++ b/tests/test_thinking_processor.py
@@ -1,0 +1,419 @@
+# SPDX-License-Identifier: Apache-2.0
+"""
+Unit tests for ThinkingAwareLogitsProcessor.
+
+Covers:
+* Phase state machine transitions (IDLE -> THINKING -> CONTENT)
+* Budget enforcement (THINKING -> TRANSITIONING -> CONTENT)
+* Budget-zero edge case (immediate TRANSITIONING)
+* Inner processor delegation in CONTENT phase
+* Content-phase control token masking
+* BoundedSuffixMatcher correctness
+* Snapshot/restore for speculative decoding rollback
+* prompt_has_think_tag fast-start paths
+
+These tests are pure-logic: they use synthetic token IDs and small
+vocab sizes. No model loading required.
+"""
+
+from __future__ import annotations
+
+import mlx.core as mx
+import pytest
+
+from vllm_mlx.constrained.thinking_processor import (
+    BoundedSuffixMatcher,
+    Phase,
+    ThinkingAwareLogitsProcessor,
+)
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+VOCAB_SIZE = 32
+# Synthetic token IDs for <think> and </think>
+THINK_START = [10]
+THINK_END = [11]
+# Multi-token end sequence for testing
+THINK_END_MULTI = [11, 12]
+
+
+def _make_logits(vocab_size: int = VOCAB_SIZE) -> mx.array:
+    """Return uniform logits (all zeros)."""
+    return mx.zeros((vocab_size,))
+
+
+def _tokens(*ids: int) -> mx.array:
+    """Build a 1-D token array from integer IDs."""
+    return mx.array(list(ids), dtype=mx.int32)
+
+
+def _argmax(logits: mx.array) -> int:
+    """Return the argmax of a logits vector."""
+    return int(logits.argmax().item())
+
+
+# ---------------------------------------------------------------------------
+# BoundedSuffixMatcher
+# ---------------------------------------------------------------------------
+
+
+class TestBoundedSuffixMatcher:
+    def test_single_token_match(self):
+        m = BoundedSuffixMatcher([5])
+        assert not m.feed(4)
+        assert m.feed(5)
+
+    def test_multi_token_match(self):
+        m = BoundedSuffixMatcher([1, 2, 3])
+        assert not m.feed(1)
+        assert not m.feed(2)
+        assert m.feed(3)
+
+    def test_no_false_positive(self):
+        m = BoundedSuffixMatcher([1, 2])
+        m.feed(1)
+        m.feed(9)  # breaks the sequence
+        assert not m.feed(2)  # 9,2 != 1,2
+
+    def test_overlapping_prefix(self):
+        """The rolling buffer catches overlapping prefixes."""
+        m = BoundedSuffixMatcher([1, 1])
+        m.feed(1)
+        # Buffer is [1]; next 1 gives [1,1]
+        assert m.feed(1)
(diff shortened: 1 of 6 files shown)


janhilgard (changes requested):
@Thump604 The state machine design is clean — the 4-phase model with snapshot/restore for speculative rollback is well thought out, and the test coverage for the processor itself is excellent. A few issues:

### Must fix

**1. `prompt_has_think_tag=True` is hardcoded — breaks non-thinking requests**

In `_build_thinking_processor()` (server.py ~line 821):

```python
proc = ThinkingAwareLogitsProcessor(
    ...
    prompt_has_think_tag=True,  # always True
)
```

This means the processor always starts in `THINKING` (or `TRANSITIONING` if budget=0), assuming `<think>` is already in the prompt. But this depends on `enable_thinking` being `True` for the request AND the chat template actually injecting the tag.

If a CLI default `--default-thinking-token-budget` is set, ALL requests get the processor — including those with `enable_thinking=False`. The processor will start in THINKING phase, count content tokens as thinking tokens, and eventually force `</think>` into the middle of regular content.

This should be conditioned on the actual `enable_thinking` state:

```python
prompt_has_think_tag=bool(chat_kwargs.get("enable_thinking", True)),
```

**2. `reasoning_tokens` field is added to Usage but never populated**

`api/models.py` adds `reasoning_tokens: int | None = None` to `Usage`, but no code in server.py extracts `proc.thinking_tokens` and sets this field in the response. The field will always be `None`. Either wire it up or remove it from this PR.

### Should fix

**3. Snapshot memory grows unbounded through CONTENT phase**

`_sync_to_tokens` continues appending to `_snapshots` and `_processed_token_ids` even in CONTENT phase (where `_advance_with_token` is a no-op). For long content generations (10K+ tokens), this is wasted memory.

Consider either:
- Stopping snapshots after entering CONTENT (check `_state == Phase.CONTENT` before appending)
- Or documenting that the engine should drop the processor when `is_retired` is True

**4. `_content_phase_mask_ids` only masks first token of each sequence**

```python
self._content_phase_mask_ids = tuple(
    dict.fromkeys([start_token_ids[0], end_token_ids[0]])
)
```

For multi-token sequences, only the first token is masked. This is sufficient in practice (most tokenizers encode `<think>` as a single special token), but inconsistent with the multi-token support in `BoundedSuffixMatcher`. Either mask all tokens in both sequences, or add a comment explaining why masking only the first token is sufficient.

**5. `_sync_to_tokens` converts full array to list every call**

```python
token_ids = [int(tokens[idx].item()) for idx in range(target_len)]
```

This is O(n) per step over the entire sequence. Use `tokens.tolist()` which is significantly faster for MLX arrays (single C++ call vs. n Python-level `.item()` calls). The common-prefix comparison afterwards will still find the divergence point quickly.

### Consider

**6. No integration with streaming paths**

The `is_retired` signal isn't consumed anywhere in the server — the engine doesn't know it can drop the processor and re-enable MTP for the remaining content generation. This is mentioned as "Phase 2 optimization" in the docstring, which is fine for now, but worth noting as a follow-up.

**7. Budget interaction with `enable_thinking=False`**

If a user sets `enable_thinking=False` AND a thinking budget is set (via CLI default), the budget should arguably be ignored since thinking is disabled. The current code would still build the processor. Consider adding a guard:

```python
if thinking_budget is not None and chat_kwargs.get("enable_thinking") is not False:
```

### What's good

- BoundedSuffixMatcher with rolling deque — handles overlapping prefixes correctly
- Snapshot/restore for speculative decoding rollback — proven pattern from MTP work
- Inner processor delegation — composable with JSON schema processor
- Test coverage is thorough: lifecycle, budget, multi-token, masking, rollback, 2D logits

Thump604:
@janhilgard Good catches. Going through each:

**Must-fix (1):** Correct. Hardcoding `prompt_has_think_tag=True` breaks requests where `enable_thinking=False` but a CLI default budget is set. Will condition on the actual `enable_thinking` state from `chat_kwargs`.

**Must-fix (2):** `reasoning_tokens` is dead weight right now. I'll wire it up to `proc.thinking_tokens` in the response builder rather than removing it, since the data is already tracked.

**Should-fix (3):** Good point. Snapshots are useless after entering CONTENT phase. I'll gate on `_state != Phase.CONTENT` before appending.

**Should-fix (4):** You're right this is inconsistent with `BoundedSuffixMatcher`'s multi-token support. In practice Qwen3 tokenizes `<think>` as a single special token, but I'll add an explicit comment explaining why first-token masking is sufficient and note where this assumption would break.

**Should-fix (5):** `tokens.tolist()` is the right call. The per-element `.item()` loop was an early debugging leftover. Will fix.

**Consider (6-7):** Agree on both. The `is_retired` signal consumption is tracked as Phase 2. For (7), I'll add a guard so `enable_thinking=False` skips processor construction entirely regardless of CLI budget defaults.

janhilgard (approved):
All review items addressed:

1. `prompt_has_think_tag` now correctly wired from `enable_thinking` — non-thinking requests won't enter THINKING phase spuriously.
2. `reasoning_tokens` removed from Usage (no dead field).
3. Snapshot growth bounded in CONTENT phase with safe `_restore_snapshot` clamping.

Test coverage is thorough (419 lines). LGTM.

Thump604:
Wayner, review feedback from Jan is addressed in the latest commits. When you get a chance, review and merge if it looks good?

PR #425, opened 2026-04-24
Add model registration and qualification handoff

Thump604:
Stacked on #417. This keeps the first model artifact workflow PR reviewable, then adds the next handoff layer without mutating any local Ops registry.

Scope:

- add `vllm-mlx model register <artifact>` to write a portable `vllm_mlx_registration_manifest.json`
- include artifact inspection, acquisition/conversion source manifests, parser policy, serving defaults, feature flags, preset alias, and explicit `production_ready=false`
- add `vllm-mlx model qualify <model-id>` to create or run a `bench-serve` qualification handoff command and optional request manifest
- keep qualification as a handoff artifact, not a green-light. Production readiness still requires the standing workload contract and review of the resulting evidence

Local validation:

```bash
uv run --extra dev pytest -q tests/test_model_workflow.py tests/test_download.py
# 23 passed

uv run --extra dev black --check vllm_mlx/model_workflow.py vllm_mlx/cli.py tests/test_model_workflow.py
uvx ruff check vllm_mlx/model_workflow.py vllm_mlx/cli.py tests/test_model_workflow.py --select E,F,W --ignore E402,E501,E731,F811,F841
git diff --check
# clean
```

CLI smokes:

```bash
vllm-mlx model qualify qwen-test --workload /tmp/workload.json --dry-run --extra-arg=--tag --extra-arg smoke
vllm-mlx model register <tmp-artifact> --model-id qwen-test --served-model-name qwen-test --preset-alias fast-qwen --mllm --default-temperature 0.6 --default-top-p 0.95 --default-top-k 20 --default-min-p 0.0 --default-presence-penalty 0.0 --default-repetition-penalty 1.0 --default-chat-template-kwargs '{"enable_thinking":true}' --feature-flag prefix_cache
```

Dependency note: `model qualify --workload` is intended to compose with #406's workload contract support. Without that bench-serve support, the dry-run manifest is still useful, but running the generated command with `--workload` depends on #406 or equivalent landing.

PR #437, opened 2026-04-25
Add model artifact workflow with registration and qualification

Thump604:
## Summary

Fresh re-implementation combining the content from #423 (model artifact workflow CLI) and #425 (registration + qualification handoff), with all review feedback from both PRs addressed.

- `vllm-mlx model inspect` reads config, HF metadata, size estimates, and MLX/conversion signals without loading weights
- `vllm-mlx model acquire` downloads via snapshot_download with staged finalization, hf_transfer, and artifact manifest
- `vllm-mlx model convert` wraps mlx-lm with recorded recipe, dry-run support, and conversion manifest
- `vllm-mlx model register` writes a portable registration manifest with serving defaults, parser policy, feature flags, and source manifest chain
- `vllm-mlx model qualify` creates or runs a bench-serve qualification handoff with environment metadata and timeout support

## Review fixes applied

From #423:
- Fix double print on conversion failure (single stderr print, then sys.exit)
- Fix GPTQ/AWQ false positive in has_mlx_signals (check for MLX-specific quant keys, reject quant_method: gptq/awq)
- Add test for conversion failure path
- Remove --local-files-only from acquire (contradicts download purpose)
- Catch json.JSONDecodeError in _read_json
- Document relationship with existing download command (retry gap noted)
- Rename convert positional arg from source to model for consistency
- Add help= text to bare --q-bits and --q-group-size arguments

From #425:
- Base branch fixed (fresh branch from upstream/main)
- Add --no-mllm flag (store_false counterpart via mutually exclusive group)
- Test: register_model with minimal/default options
- Test: qualification success path (returncode=0)
- Test: NotADirectoryError path
- Add help= text to all bare arguments
- Assert zero-valued defaults survive _drop_none
- Document artifact_path portability limitation in docstring
- Add --timeout to qualify (subprocess timeout)
- Document extra_args override risk in --extra-arg help text
- Add environment metadata to qualification manifest (Python/vllm-mlx/MLX version)

## Validation

```
pytest -q tests/test_model_workflow.py   # 19 passed
black --check                            # clean
ruff check                               # clean
```

## Test plan

- [ ] Verify `model inspect` against a local MLX artifact and a HF model id
- [ ] Verify `model acquire` with --target-dir produces a staged download with manifest
- [ ] Verify `model convert --dry-run` records the command without executing
- [ ] Verify `model register` with minimal options writes a valid manifest
- [ ] Verify `model qualify --dry-run` records the bench-serve command
- [ ] Verify GPTQ model is correctly detected as needing conversion (not an MLX artifact)

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -156,6 +156,8 @@ manifest recording, and `hf_transfer` support but does not include retry logic.
 vllm-mlx model inspect <path-or-hf-model-id>
 vllm-mlx model acquire <hf-model-id> [--target-dir <path>]
 vllm-mlx model convert <path-or-hf-model-id> --output <path> [--quantize]
+vllm-mlx model register <artifact-dir> [--model-id <id>] [--mllm]
+vllm-mlx model qualify <model-id> [--workload <path>] [--dry-run]
 ```
 
 ### Options
@@ -173,6 +175,18 @@ vllm-mlx model convert <path-or-hf-model-id> --output <path> [--quantize]
 | `convert` | `--quant-predicate` | `mlx-lm` mixed-bit quantization recipe |
 | `convert` | `--dtype` | Dtype for non-quantized parameters |
 | `convert` | `--dry-run` | Print command and manifest without executing conversion |
+| `register` | `--model-id` | Model identifier (defaults to directory name) |
+| `register` | `--served-model-name` | Name for the /v1/models endpoint |
+| `register` | `--mllm` / `--no-mllm` | Mark as multimodal or text-only |
+| `register` | `--default-temperature`, `--default-top-p`, ... | Serving defaults |
+| `register` | `--tool-call-parser`, `--reasoning-parser` | Parser policy |
+| `register` | `--feature-flag` | Feature flag (repeatable) |
+| `qualify` | `--url` | Server URL (default: `http://127.0.0.1:8080`) |
+| `qualify` | `--workload` | bench-serve workload contract path |
+| `qualify` | `--repetitions` | bench-serve repetitions |
+| `qualify` | `--timeout` | Max seconds before aborting |
+| `qualify` | `--dry-run` | Print command without executing |
+| `qualify` | `--extra-arg` | Extra arg passed to bench-serve (repeatable) |
 
 ### Examples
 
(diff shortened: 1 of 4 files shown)


janhilgard (commented):
@Thump604 Clean consolidation of #423 and #425. I verified every review item from both PRs — all addressed. Some notes on the combined result:

### Should fix

**1. `os.environ` mutation in `acquire_model` is not thread-safe**

Lines 1391-1435 of `model_workflow.py` modify `os.environ` globally to set `HF_HUB_ENABLE_HF_TRANSFER`. The `finally` block restores the old value, but between `update` and restore, any concurrent code (threads, async tasks, or library callers importing `acquire_model`) sees the modified env. For a CLI-only tool this is fine, but since this is an importable library function, consider using `subprocess.run(env={**os.environ, ...})` or documenting the thread-safety limitation.

**2. Use `pytest.raises` instead of manual try/except**

`test_acquire_model_refuses_existing_target`, `test_register_model_requires_local_directory`, and `test_register_model_rejects_file_path` all use the pattern:

```python
try:
    func()
except SomeError:
    pass
else:
    raise AssertionError("expected SomeError")
```

The idiomatic pytest pattern is simpler and more expressive:

```python
with pytest.raises(SomeError):
    func()
```

### Minor

- `_environment_metadata()` imports `importlib.metadata` twice (lines 1559 and 1565) — hoist to one import at the top of the function
- `--trust-remote-code` help text on `convert` could mention security implications (arbitrary code execution from the HF repo), matching the recommendation from the #423 review
- `_existing_manifests` still reads JSON without checking `schema_version` — fine for v1 but worth noting for future schema changes
- No test for `TimeoutExpired` path in `qualify_model` — the timeout handling (lines 1676-1690) is untested. A mock that raises `subprocess.TimeoutExpired` would close this gap

### What's good

- All review items from #423 and #425 verified as addressed
- `_is_mlx_quantization()` correctly rejects GPTQ/AWQ by checking for `quant_method` key
- Frozen dataclasses throughout — clean immutable API
- `_drop_none` semantics are correct (preserves 0, 0.0, False, "" — only strips None) with explicit test coverage
- Staged download with cleanup on failure
- Environment metadata in qualification manifests (Python, vllm-mlx, MLX versions)
- `--no-mllm` via mutually exclusive group on register
- Module docstring clearly documents the relationship with `vllm-mlx download`
- 19 tests with good coverage of success, failure, and edge case paths

Thump604:
Closing — #423 (model artifact workflow CLI) and #425 (registration + qualification handoff) are both merged, covering the full scope of this PR.

PR #440, opened 2026-04-25
Fix/glm4 add validation tests from b2ornot2b

perry2of5:
Merge glm4 tests from b2ornot2b from PR #246.


janhilgard (approved):
Clean test additions from b2ornot2b's PR #246. Good coverage for GLM4 reasoning parser registration and GLM47 tool parser — especially the zero-argument edge case which is a common failure mode.

CI 9/9 pass. Minor: 2 commits could be squashed.

PR #441, opened 2026-04-26
Fix RoPE attribute mismatch and extractor dispatch in SpecPrefill

Thump604:
## Summary

- SpecPrefill checked only `attn.rope` for RoPE, missing mlx_vlm models that use `attn.rotary_emb`. When cooperative SpecPrefill ran on a VLM language model, positional encoding was silently wrong. Added `_get_rope`/`_set_rope` helpers that check both attributes.
- Query extractor dispatch used fragile attribute sniffing (`v_norm` -> Gemma4, `q_norm` -> Qwen3.5, no `rope` -> Nemotron-H). Replaced with explicit `model_type` registry with RoPE-based fallback for unknown architectures.

## Test plan

- [ ] Verify `score_tokens` selects correct extractor for each model_type
- [ ] Verify `sparse_prefill` patches/restores RoPE on models using `rotary_emb`
- [ ] Verify `cleanup_rope` restores both `rope` and `rotary_emb` attributes
- [ ] Existing SpecPrefill tests pass unchanged

### DIFF
--- a/vllm_mlx/specprefill.py
+++ b/vllm_mlx/specprefill.py
@@ -330,8 +330,6 @@ def score_tokens(
     if query_extractor is None:
         model_type = getattr(getattr(model, "config", None), "model_type", "")
         _EXTRACTOR_REGISTRY = {
-            "gemma4": _gemma4_extract_queries,
-            "gemma4_text": _gemma4_extract_queries,
             "qwen3_5": _qwen35_extract_queries,
             "qwen3_5_moe": _qwen35_extract_queries,
             "qwen3_vl": _qwen35_extract_queries,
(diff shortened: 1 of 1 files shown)


janhilgard (approved):
LGTM — clean, well-focused fix.

The `_get_rope`/`_set_rope` abstraction correctly handles both `attn.rope` (mlx_lm) and `attn.rotary_emb` (mlx_vlm) with consistent priority order. The model_type registry is a clear improvement over attribute sniffing — easier to extend, no silent misclassification on new architectures.

Verified:
- Qwen3.6 models reuse `model_type: qwen3_5`/`qwen3_5_moe` → correctly mapped to `_qwen35_extract_queries`
- Gemma 4 (standard RoPE, no `q_norm` at attn level) falls through to `_llama_extract_queries` via the RoPE fallback — correct behavior
- `original_ropes` storing `(attn, rope)` tuples guarantees correct restore target
- `cleanup_rope` correctly uses `_get_rope`/`_set_rope` for both attribute variants

Minor: `_EXTRACTOR_REGISTRY` defined inside `score_tokens()` (recreated per call) — moving to module-level would be marginally cleaner, but dict is tiny so no perf impact.

PR #446, opened 2026-04-26
fix: remove duplicate repetition_penalty declarations silently ignored by Pydantic

clickbrain:

## Problem

Both `ChatCompletionRequest` and `CompletionRequest` declare `repetition_penalty` twice. Pydantic V2 uses the **last** declaration and silently discards the first. Any validators, descriptions, or default overrides on the first declaration are silently lost.

## Fix

Remove the first (undocumented) duplicate from both models, keeping only the annotated declaration in the sampling penalties block.

### DIFF
--- a/vllm_mlx/api/models.py
+++ b/vllm_mlx/api/models.py
@@ -162,7 +162,6 @@ class ChatCompletionRequest(BaseModel):
     top_k: int | None = None
     min_p: float | None = None
     presence_penalty: float | None = None
-    repetition_penalty: float | None = None
     max_tokens: int | None = Field(default=None, gt=0)
     stream: bool = False
     stream_options: StreamOptions | None = (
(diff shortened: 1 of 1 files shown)


Thump604 (approved):
Verified the file has duplicate `repetition_penalty` declarations in both `ChatCompletionRequest` (lines 165 and 183) and `CompletionRequest` (lines 266 and 271). The PR removes the first bare declaration and keeps the second annotated one (with the mlx-lm style comment). Pydantic V2 uses last-writer-wins, so the kept declaration is the one that was actually active. Straightforward cleanup. LGTM.

Thump604 (approved):
Reviewed. Clean fix.

perry2of5 (approved):
Looks right to me.

janhilgard (approved):
Clean fix. Pydantic V2 silently shadows the first declaration with the last one, so the duplicate served no purpose. Removing it prevents confusion.

Verified: both `ChatCompletionRequest` and `CompletionRequest` retain exactly one `repetition_penalty` field (inherited from `CommonCompletionFields`). CI 9/9 pass.

PR #447, opened 2026-04-26
fix: use 'is not None' check for execution_time_ms in MCP audit log

clickbrain:

## Problem

```python
if execution_time_ms:  # falsy — 0.0 is falsy in Python
    ...
```

Tool calls completing in under ~0.5ms (rounds to 0.0) are silently logged without timing data, making the audit log incomplete for fast tools.

## Fix

```python
if execution_time_ms is not None:
    ...
```


perry2of5 (approved):
looks reasonable to me.

janhilgard (approved):
Correct fix. `0.0` is falsy in Python, so sub-0.5ms tool calls silently lost their timing data in the audit log. `is not None` is the right guard here.

CI is pending — likely needs a rebase onto current main after the recent batch of merges.

PR #448, opened 2026-04-26
perf: replace deque with OrderedDict for O(1) LRU in PrefixCacheManager

clickbrain:

## Problem

`PrefixCacheManager` uses `collections.deque` for LRU tracking. Updating access order on a cache hit requires `deque.remove()`, which is **O(N)** — it scans the entire deque. Under load with a large prefix cache this becomes the bottleneck.

## Fix

Replace with `collections.OrderedDict`:

| Operation | Before | After |
|-----------|--------|-------|
| Cache hit / touch | `deque.remove()` O(N) | `move_to_end()` O(1) |
| New entry | `deque.append()` O(1) | `dict[key] = None` O(1) |
| Eviction | `deque.popleft()` O(1) | `popitem(last=False)` O(1) |

No change to external API or eviction semantics. `clear()` and `len()` work identically.

### DIFF
--- a/vllm_mlx/prefix_cache.py
+++ b/vllm_mlx/prefix_cache.py
@@ -13,7 +13,7 @@ This module provides two implementations:
 import copy
 import logging
 import time
-from collections import deque
+from collections import OrderedDict
 from dataclasses import dataclass
 from typing import Any, Dict, List, Optional, Tuple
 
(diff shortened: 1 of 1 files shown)


Thump604 (approved):
Good optimization. `deque.remove()` is O(N) and was the LRU bottleneck on cache hits. `OrderedDict.move_to_end()` is O(1) for the same operation. The implementation is correct: `move_to_end(key)` for touches, `dict[key] = None` for inserts, `popitem(last=False)` for eviction. The `_touch_lru` method now also handles the case where the key is missing (adds it), which is more defensive than the old try/except ValueError pattern. Store logic properly separates update vs insert paths. Clean swap with no API change. LGTM.

Thump604 (approved):
Reviewed. Clean fix.

PR #449, opened 2026-04-26
perf: O(1) tool lookup in ToolExecutor via lazily-cached name index

clickbrain:

## Problem

`_get_tool_by_name()`, `_tool_exists()`, and `_get_server_for_tool()` each call `manager.get_all_tools()` and iterate the full list on every invocation. With many MCP tools registered, each tool call pays an O(N) scan — sometimes multiple times per call path.

## Fix

Add `_get_tool_index()` which builds a `dict[str, MCPTool]` indexed by both `full_name` and short `name`. The index is rebuilt only when the tool count changes (e.g. after `refresh_tools()`), so steady-state lookups are O(1).

All three lookup helpers now delegate to `_get_tool_index().get()` or `in` membership — both O(1).

### DIFF
--- a/vllm_mlx/mcp/executor.py
+++ b/vllm_mlx/mcp/executor.py
@@ -95,6 +95,10 @@ class ToolExecutor:
         self.default_timeout = default_timeout or manager.config.default_timeout
         self.validate_arguments = validate_arguments
         self.sandbox = sandbox or get_sandbox()
+        # Lazily-built lookup index: full_name and short name → MCPTool.
+        # Rebuilt whenever the tool count changes (e.g. after refresh_tools).
+        self._tool_index: dict[str, "MCPTool"] = {}
+        self._tool_index_size: int = -1
 
     async def execute_tool_calls(
         self,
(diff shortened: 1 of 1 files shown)


vllm_mlx/mcp/executor.py
```diff
@@ -119,17 +123,21 @@ async def execute_tool_calls(
         else:
             return await self._execute_sequential(tool_calls)
 
+    def _get_tool_index(self) -> dict[str, "MCPTool"]:
+        """Return a name→tool index, rebuilding it if the tool list changed."""
+        tools = self.manager.get_all_tools()
+        if len(tools) != self._tool_index_size:
```
perry2of5, line 129: Is it reasonable to assume that a tool wouldn't be removed simultaneously to a different tool being added?

Thump604 (changes requested):
I do not think this is merge-ready yet.

The O(1) lookup direction is good, but the cache invalidation rule is too weak. `_get_tool_index()` only rebuilds when `len(manager.get_all_tools())` changes. `MCPClient.refresh_tools()` and reconnect can replace the discovered tools with the same count, including different names, schemas, or server ownership. In that case the executor keeps returning the old `MCPTool` objects: `_tool_exists()` can accept removed tools, `_validate_tool_call()` can validate against stale schemas, and `_get_server_for_tool()` can route short-name calls to a stale server.

A safe version needs a stronger invalidation signal, for example a manager/client tool-generation counter bumped after refresh/reconnect/disconnect, or a cheap signature over the current tool identities such as `(server_name, name)` before reusing the index. It also needs a regression test where the manager's tool list changes from one tool to a different one with the same count and the executor stops accepting the stale tool.

vllm_mlx/mcp/executor.py
waybarrios: I dont think we should have that assumption

vllm_mlx/mcp/executor.py
```diff
@@ -119,17 +123,21 @@ async def execute_tool_calls(
...
+            idx: dict[str, "MCPTool"] = {}
+            for t in tools:
```
waybarrios, line 131: I think this for loop adds some latency, we could rewrite it more efficient.

waybarrios:
Looping back on this. Is there any blocker? Happy to help!

Thump604:
Yes, the remaining blocker is the cache invalidation rule from the earlier review.

The O(1) lookup direction is good, but the current index reuse is keyed only on tool count. A refresh/reconnect can replace the tool set with the same count but different names, schemas, or server ownership. In that case the executor can keep stale MCPTool objects and route or validate against stale data.

A merge-ready version needs either:
- a tool-generation/version counter bumped on refresh/reconnect/disconnect, or
- a cheap signature over current tool identities, e.g. `(server_name, name)`, before reusing the index.

It also needs a regression test where the manager’s tool list changes from one tool to a different tool with the same count and the executor stops accepting/routing the stale tool.

Happy to re-review once that is addressed.

waybarrios:
Ook this is pretty old, I decided to close it. Everyone is welcome to update it and re-open it.

PR #450, opened 2026-04-26
fix: add absolute timeout to streaming disconnect guard

janhilgard:
## Summary

- `_disconnect_guard()` previously had no server-side timeout for streaming requests, relying solely on `is_disconnected()` which can fail to detect dropped connections
- This caused zombie streaming requests to run indefinitely (observed 3650+ seconds in production with `disconnected=False` the entire time)
- Add a `timeout` parameter that enforces an absolute maximum wall-clock duration for any streaming request
- Falls back to `_default_timeout` (configurable via `--timeout`) when no explicit timeout is passed
- All streaming endpoints (`/v1/completions`, `/v1/chat/completions`, `/v1/responses`, `/v1/messages`) now pass their request budget as the timeout
- Add deferred `generator.aclose()` in the finally block to ensure `stream_outputs()` cleanup fires `abort_request()` — prevents scheduler deadlock after timeout or disconnect

## Context

In production, we observed 4 concurrent streaming requests running for over 3650 seconds (>1 hour) with `is_disconnected()` consistently returning `False`. The heartbeat mechanism forces ASGI writes every 5s which should trigger broken-pipe detection, but in some cases the TCP connection stays open despite the client being gone (e.g. client behind a proxy that keeps the connection alive).

The non-streaming path (`_wait_with_disconnect()`) already has a timeout parameter — this PR brings parity to the streaming path.

### Scheduler deadlock fix

When `_disconnect_guard()` broke out of the loop (timeout or disconnect), it previously only called `anext_task.cancel()`. This cancels one `__anext__()` awaitable but does NOT close the underlying async generator. As a result, `stream_outputs()` in `mllm_scheduler.py` never reached its `finally` block → `abort_request()` was never called → the scheduler believed the request was still active → all subsequent requests got prefill but 0 decode tokens (deadlock).

The fix adds a deferred `generator.aclose()` (delayed 0.5s to avoid a Metal thread-safety race with `run_in_executor`) which ensures the scheduler properly cleans up.

## Test plan

- [x] Verified syntax and import
- [x] Deployed to production (ports 1236, 1237)
- [x] Confirmed `timeout=900s` appears in disconnect_guard START logs
- [x] Streaming requests complete normally before timeout
- [x] Client disconnect mid-stream → scheduler recovers, next request generates normally
- [x] CI tests pass (9/9)

### DIFF
--- a/vllm_mlx/server.py
+++ b/vllm_mlx/server.py
@@ -3368,6 +3368,7 @@ async def _disconnect_guard(
     poll_interval: float = 0.5,
     heartbeat_interval: float = 5.0,
     cleanup=None,
+    timeout: float | None = None,
 ) -> AsyncIterator[str]:
     """Wrap streaming generator to abort on client disconnect.
 
@@ -3377,6 +3378,11 @@ async def _disconnect_guard(
     heartbeat.  This forces an ASGI write which triggers broken-pipe
     detection — without heartbeats, ``is_disconnected()`` stays False
     during long prefill because no data is written to the socket.
+
+    If *timeout* is set, the stream is forcefully stopped once the
+    elapsed wall-clock time exceeds the given number of seconds.
+    Without this, zombie streaming requests can run indefinitely when
+    ``is_disconnected()`` never fires.
 
     On disconnect, the cancellation propagates to stream_outputs()
     finally-block → abort_request() → abort_prefill().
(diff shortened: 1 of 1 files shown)


Thump604 (approved):
Solid fix for a real production issue. The absolute timeout on streaming requests fills a clear gap: `is_disconnected()` can stay False indefinitely behind proxies that keep TCP alive. The fallback to `_default_timeout` is correct, and all three streaming endpoints now pass `total_timeout`.

The deferred `generator.aclose()` with 0.5s delay is a good solution to the Metal thread-safety race. Without it, `stream_outputs()` never hits its finally block, `abort_request()` never fires, and the scheduler deadlocks. The delay avoids concurrent `mlx::core::eval` on two threads.

One minor note: the 0.5s sleep is a heuristic. If a long eval is still running after 0.5s the race is still possible in theory, but in practice scheduler steps are fast and this is a reasonable tradeoff. LGTM.

PR #451, opened 2026-04-26
fix: dequantize KV cache before disk persistence (#443)

### LINKED ISSUE #443: bug: --kv-cache-quantization breaks prefix-cache persistence — _QuantizedCacheWrapper missing .state for save_prompt_cache
## Summary

When the server is started with `--kv-cache-quantization`, on shutdown the prefix cache cannot be persisted to disk. Every entry beyond entry 0 fails with: `'_QuantizedCacheWrapper' object has no attribute 'state'`. The `--ssd-cache-dir` tier and the in-process prefix cache become effectively non-persistable, defeating the SSD-tier value across restarts.

## Reproduction

```bash
vllm-mlx serve mlx-community/Qwen3-Coder-30B-A3B-Instruct-4bit \
  --continuous-batching \
  --ssd-cache-dir /tmp/ssdcache \
  --kv-cache-quantization \
  --kv-cache-quantization-bits 8

# Drive enough traffic to populate >1 prefix cache entries
# (e.g. via Claude Code with 30K+ token system prompt)

# Stop the server (Ctrl-C or SIGTERM)
```

Expected: All prefix-cache entries serialized to `~/.cache/vllm-mlx/prefix_cache/<model>/entry_*.safetensors`.

Actual (server log on shutdown):
```
INFO:vllm_mlx.memory_cache:[cache_persist] saved entry 0: 16 tokens, 1.5MB KV, file=...
WARNING:vllm_mlx.memory_cache:[cache_persist] failed to save entry 1: '_QuantizedCacheWrapper' object has no attribute 'state'
WARNING:vllm_mlx.memory_cache:[cache_persist] failed to save entry 2: '_QuantizedCacheWrapper' object has no attribute 'state'
... (every other entry)
INFO:vllm_mlx.memory_cache:[cache_persist] SAVED 1/7 entries
```

Captured with vllm-mlx pip build (M4 Max, macOS 14, Qwen3-Coder-30B-A3B-Instruct-4bit) on 2026-04-26.

## What Does NOT Work

- ❌ Disabling `--ssd-cache-dir` does not help — the in-memory prefix cache uses the same `save_prompt_cache` serialization path on shutdown.
- ❌ The existing `try/except Exception` catches the failure but only emits a `WARNING` per entry. Most users will never notice 6/7 entries silently fail; they just see a slow first-prompt on next start.

## Implementation Approach

Two reasonable paths:

1. **Implement `state` on `_QuantizedCacheWrapper`** so it can be passed to `mlx_lm.save_prompt_cache` directly. The wrapper would expose its quantized arrays in the shape `save_prompt_cache` understands.

2. **Dequantize on the persistence path**: inside `_persist_to_disk()`, detect quantized cache layers and dequantize them before calling `save_prompt_cache`. Trades persist-time CPU for compatibility. There's already a dequantization helper at `memory_cache.py:537` (`# Dequantize _QuantizedCacheWrapper layers...`) — the persistence code should reuse the same path.

Code locations:
- `vllm_mlx/memory_cache.py:484` — `class _QuantizedCacheWrapper`
- `vllm_mlx/memory_cache.py:537` — existing dequantization helper
- `vllm_mlx/memory_cache.py:1198-1229` — failing `_persist_to_disk()` loop

## Test Scenarios

- **Fresh start, KV-quant on**: server populates >1 entry; shutdown saves all entries to disk.
- **Restart, KV-quant on**: persisted entries are reloaded; first-call prefill is fast.
- **KV-quant off**: behavior unchanged (regression check).
- **Mid-life flip**: server started without quant, requests run, then a quantized restart loads the prior cache (or clearly errors on incompatibility — does not silently lose data).
- **Test added**: extend `tests/test_memory_cache_mlx.py` (line ~287 already references the combination but does not assert on persistence success) to exercise persist→reload round-trip with quantization on.

## Acceptance Criteria

- [ ] With `--kv-cache-quantization` enabled, shutdown logs `[cache_persist] SAVED N/N entries` (not `1/N`).
- [ ] On restart, persisted quantized entries are reloaded and re-served from cache.
- [ ] Round-trip persist→reload test passes for both quantized and non-quantized configurations.
- [ ] No regression in non-quantized behavior.

## Severity

Medium — silent partial-failure on shutdown. Affects users combining `--kv-cache-quantization` with `--ssd-cache-dir` (or simply enabling the in-memory cache persistence). The two flags otherwise work in-process; the breakage is purely on the disk-persistence path.

## Related

- #388 (closed) — `tests: kv_cache_quantization asserts stale QuantizedKVCache contract on v0.2.8` (different scope: stale test contract vs runtime persistence).
- #60 (closed) — original `--kv-cache-quantization` feature.

Thump604:
## Summary

- When `--kv-cache-quantization` is enabled, `save_to_disk` fails for every entry after entry 0 because `save_prompt_cache` calls `.state` and `.meta_state` on each cache layer, but `_QuantizedCacheWrapper` does not implement these properties.
- Fix: dequantize quantized layers before calling `save_prompt_cache`, reusing the existing `_dequantize_cache` helper. This restores the original KVCache types that implement the required `.state` / `.meta_state` interface.

## Test plan

- [ ] Start server with `--kv-cache-quantization --ssd-cache-dir /tmp/ssdcache`
- [ ] Drive enough traffic to populate >1 prefix cache entry
- [ ] Stop the server (SIGTERM)
- [ ] Verify shutdown log shows `[cache_persist] SAVED N/N entries` (not `1/N`)
- [ ] Restart and confirm persisted entries reload successfully
- [ ] Verify non-quantized mode is unchanged (regression check)

Fixes #443

### DIFF
--- a/vllm_mlx/memory_cache.py
+++ b/vllm_mlx/memory_cache.py
@@ -1198,9 +1198,18 @@ class MemoryAwarePrefixCache:
         for i, (tokens_key, entry) in enumerate(self._entries.items()):
             entry_path = os.path.join(cache_dir, f"entry_{i}.safetensors")
             try:
+                # Dequantize _QuantizedCacheWrapper layers before saving.
+                # save_prompt_cache requires .state and .meta_state which
+                # the wrapper does not provide; dequantizing restores the
+                # original cache types that do.
+                persist_cache = (
+                    _dequantize_cache(entry.cache)
+                    if any(isinstance(c, _QuantizedCacheWrapper) for c in entry.cache)
+                    else entry.cache
+                )
                 save_prompt_cache(
                     entry_path,
-                    entry.cache,
+                    persist_cache,
                     metadata={"num_tokens": str(len(tokens_key))},
                 )
                 # Save tokens separately (can be 100K+ ints → binary is smaller)


janhilgard (approved):
LGTM — minimal and correct fix.

The `any(isinstance(c, _QuantizedCacheWrapper) ...)` guard avoids unnecessary dequantization for non-quantized entries. The temporary `persist_cache` goes out of scope after `save_prompt_cache`, so no memory leak. Memory overhead from dequantization is acceptable since this only runs at shutdown.

Round-trip is clean: `save_to_disk` dequantizes → saves standard KVCache → `load_from_disk` loads standard KVCache → re-quantized on store via normal flow.

Issue #420, opened 2026-04-24
Test failures on main


perry2of5:
I am running tests with 

```
rm -rf .venv uv.lock
uv venv python --python 3.13
uv sync
uv pip install -e ".[dev,vision]"
uv run pytest
```

Failing tests:
FAILED tests/test_continuous_batching.py::TestContinuousBatchingIntegration::test_single_request - AssertionError: assert 0 > 0
FAILED tests/test_continuous_batching.py::TestContinuousBatchingIntegration::test_concurrent_requests - assert False
FAILED tests/test_continuous_batching.py::TestContinuousBatchingIntegration::test_batching_improves_throughput - assert 0.0 > 100
FAILED tests/test_engine_core_stream_safety.py::test_engine_core_no_cross_thread_stream_error - AssertionError: scheduler logged cross-thread stream errors: ['Error in batch generation step: There is no Stream(gpu, 5) in current thread.\nTraceback (most recent call last):\n  File "/Users/tperry/code/llm/vllm-mlx/vllm_mlx/scheduler.py", ...
FAILED tests/test_engine_core_thread_streams.py::test_engine_core_runs_all_scheduler_steps_on_one_worker_thread - AttributeError: 'module' object at vllm_mlx.engine_core has no attribute 'bind_generation_streams'
FAILED tests/test_engine_core_thread_streams.py::test_mllm_scheduler_runs_steps_on_model_load_thread - TimeoutError
FAILED tests/test_model_registry.py::TestMultiEngine::test_sequential_engines_with_close - AssertionError: assert 0 > 0
FAILED tests/test_model_registry.py::TestMultiEngine::test_sequential_engines_without_close - AssertionError: assert 0 > 0
FAILED tests/test_model_registry.py::TestCacheRecovery::test_recovery_from_simulated_cache_corruption - AssertionError: assert 0 > 0
FAILED tests/test_model_registry.py::TestBenchmarkScenario::test_benchmark_like_usage - AssertionError: assert 0 > 0
FAILED tests/test_model_registry.py::TestBenchmarkScenario::test_multiple_models_sequentially - AssertionError: assert 0 > 0
FAILED tests/test_rerank.py::TestClassifierForward::test_classifier_forward_returns_logits_shape - RuntimeError: There is no Stream(gpu, 5) in current thread.
FAILED tests/test_rerank.py::TestClassifierForward::test_classifier_forward_different_num_labels - RuntimeError: There is no Stream(gpu, 5) in current thread.

[...]

Issue #407, opened 2026-04-24
`RuntimeError: There is no Stream(gpu, N) in current thread` when calling `/v1/messages` or `/v1/chat/completions` on Qwen3.5 MLX models (MLLM path runs in worker thread)


luopeixiang:
`RuntimeError: There is no Stream(gpu, N) in current thread` when calling `/v1/messages` or `/v1/chat/completions` on Qwen3.5 MLX models (MLLM path runs in worker thread)

## Summary

On Apple Silicon, any model whose name matches the MLLM pattern `Qwen3.5-` (e.g. `mlx-community/Qwen3.5-4B-4bit`, `mlx-community/Qwen3.5-9B-4bit`) fails at inference time for chat-style endpoints with:

```
RuntimeError: There is no Stream(gpu, 0) in current thread.
```

(In `--continuous-batching` mode the error becomes `Stream(gpu, 2)` at a different site, same root cause.)

`/v1/completions` (raw prompt) works fine. `/v1/chat/completions` and `/v1/messages` both fail. In non-streaming mode `/v1/messages` currently masks the failure and returns `HTTP 200` with empty `content` and `input_tokens/output_tokens = 0`, which is a secondary bug.

## Environment

- OS: macOS, Apple Silicon (unified memory ~38.7 GB)
- Python: 3.13.12 (cpython-3.13.12-macos-aarch64-none, from uv)
- Installed via: `uv tool install git+https://github.com/waybarrios/vllm-mlx.git` (latest)
- Models reproduced on:
  - `mlx-community/Qwen3.5-4B-4bit`
  - `mlx-community/Qwen3.5-9B-4bit`
- Both are cached locally; running with `--offline`.

### A. Continuous-batching mode → `Stream(gpu, 2)`

Start server:

```bash
vllm-mlx serve mlx-community/Qwen3.5-9B-4bit \
  --port 8000 \
  --continuous-batching \
  --reasoning-parser qwen3 \
  --enable-auto-tool-choice \
  --tool-call-parser qwen3_coder \
  --mllm-prefill-step-size 8192 \
  --offline
```

Call:

```bash
curl -s http://localhost:8000/v1/messages \
  -H "Content-Type: application/json" \
  -d '{
    "model": "mlx-community/Qwen3.5-9B-4bit",
    "max_tokens": 256,
    "messages": [{"role": "user", "content": "Hello, please introduce yourself."}]
  }'
```

Client sees `HTTP 200` with empty content:

```json
[...]

PR #452, opened 2026-04-26
fix: bind MLX streams in generate_batch_sync and test conftest (#420)

Thump604:
## Summary

- `generate_batch_sync` calls `scheduler.step()` directly without binding MLX generation streams to the calling thread. When the model was loaded on a different thread, this causes `RuntimeError: There is no Stream(gpu, N) in current thread`, which surfaces as 0 completion tokens in tests.
- Tests that create worker threads with `bind_generation_streams` (e.g. `test_engine_core_thread_streams`) can leave MLX stream state polluted for later tests running on the main thread.

Two changes:

1. Call `bind_generation_streams()` at the top of `generate_batch_sync` so the calling thread owns the required MLX streams before stepping.
2. Add a session-scoped autouse conftest fixture that creates a fresh default MLX stream on the main thread at test session start, preventing cross-test stream pollution.

Together these address the majority of test failures reported in #420:
- `test_model_registry` (`assert 0 > 0` in sequential engine tests, cache recovery, benchmark scenarios)
- `test_continuous_batching` integration tests (`assert 0 > 0` in single/concurrent request tests)
- `test_rerank` classifier_forward tests (`Stream(gpu, 5)` error)
- `test_engine_core_stream_safety` (cross-thread stream errors with Llama model)

## Remaining test failures from #420 that need separate investigation

- `test_engine_core_thread_streams::test_engine_core_runs_all_scheduler_steps_on_one_worker_thread`: monkeypatch `AttributeError` for `bind_generation_streams`. This may be Python 3.13-specific or related to import ordering. The attribute is set at module import time via `from .mlx_streams import bind_generation_streams`.
- `test_engine_core_thread_streams::test_mllm_scheduler_runs_steps_on_model_load_thread`: TimeoutError. Likely needs the `MLLMScheduler._process_loop` to be stubbed differently.

## Test plan

- [ ] Run `pytest tests/test_model_registry.py tests/test_continuous_batching.py tests/test_rerank.py tests/test_engine_core_stream_safety.py` on Apple Silicon with a small model available
- [ ] Verify `test_sequential_engines_with_close`, `test_sequential_engines_without_close`, `test_benchmark_like_usage` pass
- [ ] Verify `test_classifier_forward_returns_logits_shape` and `test_classifier_forward_different_num_labels` pass
- [ ] Verify no regressions in existing passing tests

Addresses #420

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -56,3 +56,26 @@ def server_url(request):
 def anyio_backend():
     """Run anyio-marked tests on asyncio only."""
     return "asyncio"
+
+
+@pytest.fixture(autouse=True, scope="session")
+def _bind_mlx_default_stream():
+    """Ensure the main-thread default MLX stream is set once at session start.
+
+    Tests that create worker threads with ``bind_generation_streams`` allocate
+    new MLX streams and call ``mx.set_default_stream``.  Because MLX streams
+    are thread-local, this is harmless to the main thread.  However, if a test
+    *on the main thread* calls ``bind_generation_streams`` (e.g. via
+    ``generate_batch_sync`` or a monkeypatch), the default stream changes and
+    later tests that assume the original default stream will fail with
+    ``RuntimeError: There is no Stream(gpu, N) in current thread``.
+
+    Binding once at session start gives all main-thread tests a known-good
+    default stream.
+    """
+    try:
+        import mlx.core as mx
+
+        mx.set_default_stream(mx.new_stream(mx.default_device()))
+    except ImportError:
+        pass
--- a/vllm_mlx/engine_core.py
+++ b/vllm_mlx/engine_core.py
@@ -588,6 +588,11 @@ class EngineCore:
             self.scheduler.add_request(request)
             request_ids.append(request_id)
 
+        # Bind MLX generation streams to the calling thread so that
+        # scheduler.step() can evaluate KV cache state without hitting
+        # "There is no Stream(gpu, N) in current thread" errors.
+        bind_generation_streams()
+
         # Process until all done - direct scheduler access, no async overhead
         results: Dict[str, RequestOutput] = {}
         while self.scheduler.has_requests():


Thump604:
Wayner, this has a real user reproducer now: #407 (comment from ChristopherA). The exact sequence stream/non-stream/stream on a text-only LLM hits "There is no Stream(gpu, N) in current thread" every time. This fix binds MLX streams in generate_batch_sync() which closes it. Would appreciate a quick review so we can unblock that report.

perry2of5:
I can still reproduce the issue as reported using this code with python 3.13. Working on a fix.

perry2of5:
I fixed both the issue I found and another issue where switching from streaming false to streaming true would break. Please see my PR against your branch.

Thump604:
Rebased onto current main (includes all the recent merges). Perry's additions from PR #2 are included.

This addresses the Stream(gpu, N) thread safety issue by binding MLX streams explicitly in `generate_batch_sync` and test conftest. Ready for review.

perry2of5:
I don't see my changes in this PR. Should they be???

I've been running my changes stacked on top of these changes and I haven't had issues.

janhilgard (approved):
Minimal and well-targeted fix. `bind_generation_streams()` in `generate_batch_sync` is the right pattern — same approach used in other engine entry points. The session-scoped conftest fixture cleanly prevents cross-test MLX stream pollution with proper ImportError handling for non-Apple CI. Honest disclosure of remaining #420 failures is appreciated. LGTM.

Thump604:
Perry, you are right. I checked the current diff and my earlier comment was wrong: this PR currently contains only the narrow `generate_batch_sync()` stream binding fix plus the `tests/conftest.py` session stream setup. Your broader stacked changes are not in this branch after the rebase.

Given Jan approved the narrow patch, I see two clean options:

1. Keep #452 narrow and merge it as the minimal `generate_batch_sync()` fix.
2. Open or restore a separate follow-up PR for your broader stream ownership fixes around SimpleEngine startup, residency preload, and stream/non-stream switching.

I do not want to silently widen this PR after review. My preference is option 1 for #452, then option 2 as a separate reviewable follow-up if those broader fixes are still needed.

perry2of5:
okay, I opened a PR against main for them. I can't use opencode without them. Not sure what else breaks without them.

arindamroy-eng:
Is this released in vlm-mlx 0.4.0?

Thump604:
Yes. This fix is included in vllm-mlx 0.4.0. I verified the v0.4.0 tag contains both parts of #452: `bind_generation_streams()` in `generate_batch_sync()` and the session-level default MLX stream fixture.

If you can still reproduce the stream/non-stream switching failure on 0.4.0, please share the Python version, macOS/MLX versions, model, and minimal request sequence so we can distinguish a remaining path from the issue fixed here.

Issue #442, opened 2026-04-26
vllm-mlx / MLX backend grows wired/Metal memory over time and eventually aborts with `kIOGPUCommandBufferCallbackErrorOutOfMemory`


seanpianka:
## Summary

Running `vllm-mlx` with `Qwen3.5-35B-A3B-4bit` on Apple Silicon appears to grow native wired/Metal memory over sustained OpenAI-compatible `/v1/chat/completions` traffic. Memory pressure rises from under ~100 GB used to ~220 GB+ over a few hours. The Python process RSS significantly underreports the memory pressure, while macOS `wired` memory grows very large. Eventually the backend aborts with a Metal OOM exception, leaving a stale pid file and causing the router to return connection failures.

This looks related to, but not identical to, existing MLX-LM reports:
- `mlx_lm.server` unbounded memory growth / IOGPU crash: https://github.com/ml-explore/mlx-lm/issues/883
- `mlx_lm.server` crashes on Metal OOM instead of returning HTTP error: https://github.com/ml-explore/mlx-lm/issues/854

## Environment

- Hardware: Apple Silicon Mac Studio, 256 GB unified memory
- macOS: 26.2, build `25C56`
- Kernel: Darwin `25.2.0`
- Python: 3.12
- `vllm-mlx`: 0.2.9
- `mlx-vlm`: 0.4.4
- `mlx-lm`: 0.31.1
- `mlx`: 0.31.1
- Model: `mlx-community/Qwen3.5-35B-A3B-4bit`
- Served model name: `./models/Qwen3.5-35B-A3B-4bit`

## Start command

```bash
uv run python -u -m vllm_mlx.cli serve \
  /Volumes/Data/llms/models/Qwen3.5-35B-A3B-4bit \
  --host 0.0.0.0 \
  --port 8002 \
  --served-model-name ./models/Qwen3.5-35B-A3B-4bit \
  --max-tokens 262144 \
  --max-request-tokens 262144 \
  --max-num-seqs 6 \
  --cache-memory-percent 0.25 \
  --timeout 300 \
  --continuous-batching \
  --reasoning-parser qwen3
```

## Traffic pattern

The server receives sustained OpenAI-compatible chat traffic through a LiteLLM/nginx router.

Typical requests include:

1. Structured extraction requests:
   - `response_format.type=json_schema`
   - `temperature=0.1`
   - prompt size roughly 2.8k-3.5k chars
   - constrained decoding enabled

2. Summary requests:
[...]

PR #453, opened 2026-04-26
fix: release stale cache references to prevent Metal memory growth (#442)

Thump604:
## Summary

Under sustained traffic, Metal wired memory grows unbounded because request objects retain references to MLX arrays that have already been copied into the BatchGenerator. The Metal buffer cache cannot free these buffers while Python references exist, even after `mx.clear_cache()`.

Three sources of leaked references, all fixed in this PR:

- **`request.prompt_cache`** (LLM scheduler path): Set during prefix cache fetch, never cleared after the cache is passed to `BatchGenerator.insert()`. Every request with a cache hit retains a full KV cache reference for the duration of generation plus until the request object is GC'd. For a 35B model this can be 100MB+ per request.

- **`request._extracted_cache`** (LLM scheduler path): Cleared for memory_aware_cache storage but not for the legacy prefix_cache path, not on storage failure, and not in the abort path.

- **`MLLMBatchRequest.pixel_values` / `attention_mask` / `image_grid_thw` / `extra_kwargs`** (MLLM batch generator): Set during preprocessing, never cleared after vision encoding. For multi-image VLM requests, `pixel_values` can be hundreds of MB and stays pinned in Metal memory for the entire generation duration.

All references are now released at the earliest safe point:
- `prompt_cache` cleared immediately after successful `BatchGenerator.insert()`
- Both `_extracted_cache` and `prompt_cache` cleared in `_cleanup_finished` and `_do_abort_request`
- Vision inputs cleared after each vision encoding / prefill path and at the end of `_process_prompts`

## Test plan

- [ ] Run the server with sustained traffic (6+ concurrent requests) for 2+ hours
- [ ] Monitor `mx.get_active_memory()`, `mx.get_cache_memory()`, and system wired memory via `vm_stat`
- [ ] Verify Metal memory stabilizes and does not grow monotonically
- [ ] Verify prefix cache hits still work correctly (cache is used before release)
- [ ] Verify MLLM (VLM) requests still generate correct output after vision input cleanup
- [ ] Verify abort/timeout paths properly release Metal buffers

Addresses #442

### DIFF
--- a/vllm_mlx/mllm_batch_generator.py
+++ b/vllm_mlx/mllm_batch_generator.py
@@ -1000,6 +1000,11 @@ class MLLMBatchGenerator:
             self._prefill_progress[request.request_id] = (total, total)
             output = self.language_model(input_ids, cache=cache)
             request.vision_encoded = True
+            # Release preprocessed inputs after encoding (issue #442)
+            request.pixel_values = None
+            request.attention_mask = None
+            request.image_grid_thw = None
+            request.extra_kwargs.clear()
             if hasattr(output, "logits"):
                 return output.logits
             return output
@@ -1053,6 +1058,11 @@ class MLLMBatchGenerator:
         last_chunk = input_ids[:, processed:]
         output = self.language_model(last_chunk, cache=cache)
         request.vision_encoded = True
+        # Release preprocessed inputs after encoding (issue #442)
+        request.pixel_values = None
+        request.attention_mask = None
+        request.image_grid_thw = None
+        request.extra_kwargs.clear()
         self._prefill_progress[request.request_id] = (total, total)
 
         if chunk_count > 0:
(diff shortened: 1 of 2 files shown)


janhilgard (approved):
LGTM — well-analyzed memory leak fix, purely additive (46+/0-).

All three leak sources are addressed at the earliest safe point:

1. **`prompt_cache`** cleared in `_schedule_waiting` immediately after `BatchGenerator.insert()` — the generator has its own copy, request's reference is stale.
2. **`_extracted_cache`** correctly deferred to `_cleanup_finished` / `_do_abort_request` (still needed for memory_aware_cache store path during generation).
3. **Vision inputs** (`pixel_values`, `attention_mask`, `image_grid_thw`, `extra_kwargs`) cleared in all encoding paths + catch-all in `_process_prompts`.

The redundant cleanup across 4 MLLM paths is harmless (None-on-None is a no-op, `.clear()` on empty dict is a no-op) and provides defense-in-depth against any future code path changes.

Minor style note: the 4 repeated `request.pixel_values = None; ...` blocks could be a helper, but not worth a revision for 4 lines.

PR #439, opened 2026-04-25
Add Gemma 4 audio support to chat completions

Thump604:
## Summary

Re-implementation of #430 against current main. Threads audio inputs through the full chat completion pipeline so Gemma 4 (and future audio-capable models) can process audio content alongside text and images.

- **api/utils.py**: `extract_multimodal_content` now returns a 4-tuple `(messages, images, videos, audios)`, handling both `audio` and `audio_url` content types
- **engine/batched.py**: Audio parameter plumbed through `generate`, `stream_generate`, `chat`, `stream_chat`; `_prepare_mllm_messages` converts `audio_url` to HuggingFace `audio` format; `_extract_media_from_messages` extracts audio alongside images/videos
- **mllm_scheduler.py / mllm_batch_generator.py**: `audio` field on `MLLMRequest` and `MLLMBatchRequest` dataclasses; audio passed through scheduling and preprocessing
- **models/mllm.py**: New helpers `is_base64_audio`, `decode_base64_audio`, `download_audio`, `process_audio_input`, `_prepare_audio`, `_collect_audio_inputs`; refactored `download_video` into shared `_download_media` with extension mapping; audio support in `generate`, `stream_generate`, `chat`, `stream_chat`; cache disabled for audio inputs
- **server.py**: All `extract_multimodal_content` call sites updated for 4-tuple; `audio`/`audio_url` added to MLLM media type detection

## Test plan

- [x] `test_api_utils.py`: 22 tests pass -- all `extract_multimodal_content` callers updated to 4-tuple, new `test_multimodal_with_audio_url` and `test_multimodal_with_audio_type`
- [x] `test_mllm.py`: New `TestAudioProcessing` class with `test_process_audio_input_local_file`, `test_process_audio_input_dict_format`, `test_collect_audio_inputs`; new `test_is_base64_audio`
- [x] `test_mllm_mtp_routing.py`: New `test_batched_prepare_mllm_messages_converts_audio_url`
- [x] `test_native_tool_format.py`: All 3-tuple unpacking updated to 4-tuple with `audios` assertions
- [x] `test_server.py`: New `test_extract_multimodal_content_with_audio_url`; existing tests updated for 4-tuple; `fake_extract` stubs return 4-tuple
- [x] `test_lifecycle_server.py`: `fake_extract` stubs return 4-tuple
- [x] `black --check` passes on all changed files
- [ ] Integration test with Gemma 4 audio model (requires model weights)

PR #433, opened 2026-04-25
Promote tool_call blocks from reasoning to content

Thump604:
## Summary

- Teach `BaseThinkingReasoningParser` to detect `<tool_call>` blocks inside reasoning and reclassify them as content so the downstream tool parser can find them
- Fixes Failure Mode B (tool calls inside `<think>` blocks lost by parser pipeline), confirmed on Qwen 3.5 27B/35B/122B/397B and Qwen 3.6 30B/35B
- Both non-streaming and streaming paths
- Related: vllm-project/vllm#39055 (non-streaming only, known bugs)

## Non-streaming

`_promote_tool_calls(reasoning, content)` scans reasoning after `_extract_complete_reasoning` splits it from content:

- **Closed blocks** (`<tool_call>...</tool_call>`): appended to content
- **Unclosed blocks** (`<tool_call>\s*[\{<]...` spanning `</think>` boundary): prepended to content to reassemble with continuation
- Structural `\s*[\{<]` guard on unclosed regex prevents false positives on prose that mentions `<tool_call>`
- `content = (content or "")` handles the `content=None` case
- Logs warning when promotion fires (model misbehavior recovery path)

## Streaming

Tool-call buffering state machine in the `thinking` phase:

- `<tool_call>` detected: start buffering, stop emitting as reasoning
- `</tool_call>` detected: emit buffer as content, resume reasoning
- `</think>` while buffering: flush buffer as content, transition to content phase
- Stream ends while buffering: `finalize_stream()` flushes as content
- `_transition_to_content` applies non-streaming promotion as catch-all for single-delta cases

## Test plan

- [x] 11 non-streaming tests (closed/unclosed/boundary/prose/JSON/multiple/warning)
- [x] 7 streaming tests (full-text, large-chunk, think-ends-while-buffering, finalize, multiple)
- [x] 2 composition tests (preserves trailing content, tool_choice=required)
- [x] 1 composition test (reasoning parser + tool parser end-to-end, skipped without transformers)
- [x] 121 existing parser tests: 0 regressions

PR #455, opened 2026-04-26
Add --max-kv-size CLI flag for per-sequence KV cache cap

Thump604:
## Summary

- Adds `--max-kv-size <int>` flag to the `serve` subcommand that caps per-sequence KV cache size in tokens
- When set, the scheduler creates `RotatingKVCache` instances (from mlx-lm) that evict older entries beyond the limit, bounding memory usage
- Connects the CLI flag through `SchedulerConfig` to the scheduler's request insertion path, where caches are pre-created with the size limit before being passed to `BatchGenerator`

Addresses the scenario in #442 where multiple concurrent long-context sequences cause Metal OOM. The internal `max_kv_size` plumbing in mlx-lm already exists but was not exposed through the CLI.

**Usage:**
```bash
vllm-mlx serve my-model --continuous-batching --max-kv-size 131072
```

## Changes

- `vllm_mlx/scheduler.py`: Add `max_kv_size` field to `SchedulerConfig`; pre-create bounded `RotatingKVCache` when inserting sequences without an existing cache
- `vllm_mlx/cli.py`: Add `--max-kv-size` argument to `serve` subcommand; propagate to `SchedulerConfig`; print at startup when set
- `tests/test_max_kv_size.py`: CLI argument parsing tests and SchedulerConfig propagation tests

## Test plan

- [x] `black --check` passes on all changed files
- [x] `pytest tests/test_max_kv_size.py` passes (4 CLI tests pass, 3 SchedulerConfig tests skip gracefully without mlx)
- [ ] Manual: `vllm-mlx serve <model> --continuous-batching --max-kv-size 8192` starts and shows "Max KV cache size per sequence: 8192 tokens"
- [ ] Manual: verify cache objects are `RotatingKVCache` when flag is set

### DIFF
--- a/tests/test_max_kv_size.py
+++ b/tests/test_max_kv_size.py
@@ -1,0 +1,64 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for --max-kv-size CLI flag and SchedulerConfig plumbing."""
+
+import pytest
+
+
+class TestMaxKvSizeCLI:
+    """Verify --max-kv-size argparse integration."""
+
+    def _parse(self, args_list):
+        """Build the serve parser and parse args."""
+        from vllm_mlx.cli import create_parser
+
+        parser = create_parser()
+        return parser.parse_args(["serve", "test-model"] + args_list)
+
+    def test_default_is_none(self):
+        args = self._parse([])
+        assert args.max_kv_size is None
+
+    def test_explicit_value(self):
+        args = self._parse(["--max-kv-size", "65536"])
+        assert args.max_kv_size == 65536
+
+    def test_with_continuous_batching(self):
+        args = self._parse(["--continuous-batching", "--max-kv-size", "32768"])
+        assert args.max_kv_size == 32768
+        assert args.continuous_batching is True
+
+
+class TestSchedulerConfigMaxKvSize:
+    """Verify SchedulerConfig accepts and stores max_kv_size."""
+
+    def test_default_zero(self):
+        pytest.importorskip("mlx.core")
+        from vllm_mlx.scheduler import SchedulerConfig
+
+        config = SchedulerConfig()
+        assert config.max_kv_size == 0
+
+    def test_set_value(self):
+        pytest.importorskip("mlx.core")
+        from vllm_mlx.scheduler import SchedulerConfig
+
+        config = SchedulerConfig(max_kv_size=65536)
+        assert config.max_kv_size == 65536
+
+
+class TestMLLMSchedulerConfigMaxKvSize:
+    """Verify MLLMSchedulerConfig accepts and stores max_kv_size."""
+
+    def test_default_zero(self):
+        pytest.importorskip("mlx.core")
+        from vllm_mlx.mllm_scheduler import MLLMSchedulerConfig
+
+        config = MLLMSchedulerConfig()
+        assert config.max_kv_size == 0
+
+    def test_set_value(self):
+        pytest.importorskip("mlx.core")
+        from vllm_mlx.mllm_scheduler import MLLMSchedulerConfig
+
+        config = MLLMSchedulerConfig(max_kv_size=32768)
+        assert config.max_kv_size == 32768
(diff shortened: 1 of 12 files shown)


waybarrios:
Appreciate the quick turnaround @Thump604. The CLI direction itself is right, this is purely about how the flag interacts with the rest of the engine. Mind taking another look at what I wrote on #442 (issuecomment-4322954244) before we iterate further? A handful of those points aren't addressed in the current diff yet, and I'd rather we work through them in one go than do it piecemeal.

Thump604:
All five items addressed in 98ce04b:

1. **MLLM path plumbing** -- `max_kv_size` now flows from `SchedulerConfig` through `engine/batched.py` into `MLLMSchedulerConfig`, then through `MLLMScheduler` into `MLLMBatchGenerator.__init__`, where it is passed to `make_prompt_cache(self.language_model, max_kv_size=...)`. The dead `create_mllm_prompt_cache` helper in `multimodal_processor.py` (zero callers) is deleted.

2. **SpecPrefill draft model** -- Verified: `specprefill.py` does not exist in this repo. That call site is in our downstream overlay, not upstream. Nothing to change here.

3. **SchedulerConfig field** -- Changed from `Optional[int] = None` to `int = 0`. All call sites convert `0 -> None` via `or None` before passing to mlx-lm, so we never accidentally send 0 as a cache size. Tests updated accordingly.

4. **Prefix cache log line** -- Added `logger.debug("Prefix cache: skipping reuse of %d-token entry (RotatingKVCache is not trimmable)", ...)` in `PrefixCacheManager.lookup_cache` when `_can_trim_cache` returns false.

5. **Reasoning model floor doc** -- CLI help for `--max-kv-size` now includes: "For reasoning models (Qwen3, DeepSeek-R1) set this high enough to contain the full think block (recommend >= 32768)."

`black --check` passes on all 8 changed files.

Thump604:
@waybarrios addressed all your points in two commits:

1. SchedulerConfig uses int=0 (0 means unbounded, converted to None at call sites)
2. MLLM batch generator plumbed via MLLMSchedulerConfig
3. Dead create_mllm_prompt_cache helper deleted
4. Prefix cache log line added when RotatingKVCache skips reuse
5. CLI help text notes >= 32768 floor for reasoning models

Second commit closes the remaining gaps:
6. mllm.py (SimpleEngine MLLM path) now accepts and passes max_kv_size through all 3 make_prompt_cache call sites
7. SimpleEngine constructor accepts max_kv_size, server passes it from scheduler_config
8. BatchedEngine passes max_kv_size from scheduler_config to MLXMultimodalLM constructor

All 5 make_prompt_cache call sites in the tree now pass max_kv_size. Verified with grep, no unplumbed paths remain.

Re: per-request override and single-process gate, happy to do those as follow-up PRs to keep this one focused on the server-level flag.

janhilgard (approved):
LGTM — thorough plumbing of existing mlx-lm `max_kv_size` through the full stack.

Verified all `make_prompt_cache` call sites are updated (5 in simple.py, 2 in mllm_batch_generator.py, 3 in models/mllm.py, 1 new in scheduler.py). The `max_kv_size or None` pattern correctly maps 0 → None (unbounded) for the mlx-lm API.

Prefix cache interaction is sound: `_can_trim_cache()` already returns False for RotatingKVCache, and the new debug log makes the skip visible. No type mismatch risk — fetched entries are standard KVCache, new misses create RotatingKVCache, both share the generation interface.

Clean removal of unused `create_mllm_prompt_cache`. Good help text with the reasoning model caveat (>= 32768).

Test coverage is adequate: CLI parsing, config propagation, and existing test fixtures properly updated with `generator.max_kv_size = 0`.

Thump604:
Thanks for the review @janhilgard. This one plus #453, #451, #441, #439, #436, #434, #433, #427, #426, and #423 are all approved. Could you merge the ones you have access to? Some were assigned to Wayner but if you can proceed that would unblock a lot of downstream work. Appreciate it.

janhilgard:
Done — merged #423, #426, #433, #434, #436, #441, #451, #453, and #455 (9 out of 11). #427 and #439 have merge conflicts after the batch merge; left a note on both asking for a rebase.

PR #457, opened 2026-04-27
Fix client disconnect detection for non-streaming requests

janhilgard:
## Summary

- Starlette's `is_disconnected()` always returns `False` for non-streaming requests due to an immediately-cancelled `anyio.CancelScope` that prevents the ASGI `receive()` from executing
- Replace with `_is_client_disconnected()` that reads uvicorn's internal `disconnected` flag directly from the `RequestResponseCycle`, walking through middleware wrapper closures
- Cache the cycle reference on the request object for O(1) subsequent polls

## Root cause

```python
# Starlette's is_disconnected() — broken:
with anyio.CancelScope() as cs:
    cs.cancel()                        # immediately cancel
    message = await self._receive()    # never executes
# message stays {}, disconnect never detected
```

Meanwhile uvicorn correctly sets `cycle.disconnected = True` in `connection_lost()` — the flag is there, Starlette just can't read it.

## Before/After

| Metric | Before | After |
|--------|--------|-------|
| Non-streaming disconnect detection | Never (relies on timeout) | ~1s after TCP close |
| Zombie request lifetime | Until timeout (120s+) | Aborted within 1-2s |
| GPU waste on disconnected clients | Full generation | Minimal |

## Test plan

- [x] Non-streaming request, kill client after 3s → scheduler aborts within 2s
- [x] Streaming requests unaffected (heartbeat detection still works as before)
- [x] Normal request completion unaffected
- [x] Production verified on Apple M3 Ultra with Qwen3.6-35B-A3B

### DIFF
--- a/vllm_mlx/server.py
+++ b/vllm_mlx/server.py
@@ -3728,6 +3728,84 @@ async def _ensure_sse_terminal(
             yield terminal_frame
 
 
+def _find_uvicorn_cycle(obj, depth=0, visited=None):
+    """Walk through middleware wrappers to find uvicorn's RequestResponseCycle.
+
+    This relies on uvicorn's internal ``RequestResponseCycle.disconnected``
+    attribute and Starlette's middleware closure layout.  Tested against
+    uvicorn 0.34-0.40 and starlette 0.44-0.46.  If either changes the
+    internal layout, this function returns None and disconnect detection
+    silently falls back to timeout-only behaviour.
+    """
+    if depth > 8:
+        return None
+    if visited is None:
+        visited = set()
+    obj_id = id(obj)
+    if obj_id in visited:
+        return None
+    visited.add(obj_id)
+
+    # Direct hit: object has 'disconnected' bool attr (RequestResponseCycle)
+    if hasattr(obj, "disconnected") and isinstance(getattr(obj, "disconnected"), bool):
+        return obj
+
+    # Check __self__ of bound methods
+    self_obj = getattr(obj, "__self__", None)
+    if self_obj is not None:
+        result = _find_uvicorn_cycle(self_obj, depth + 1, visited)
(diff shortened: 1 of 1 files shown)


janhilgard:
@Thump604 Could you review this one when you get a chance? Thanks!

Thump604 (approved):
Clean fix for a real problem. Starlette's CancelScope is broken for non-streaming disconnect detection, and reading uvicorn's RequestResponseCycle.disconnected directly is the right workaround.

Implementation is solid:
- Depth-limited closure walk (8 levels) prevents runaway recursion
- Caches the cycle reference on the request object for O(1) subsequent polls
- Falls back safely to False if cycle isn't found (no regression for non-uvicorn servers)
- Sync function replacing the broken async `is_disconnected()` is correct since we're just reading a bool flag

One fragility note for future maintainers: `_find_uvicorn_cycle` depends on uvicorn's internal `RequestResponseCycle.disconnected` attribute and Starlette's middleware closure layout. If either changes, the function returns None and disconnect detection silently stops working (reverts to timeout-only behavior). Worth a comment at the top of the function noting the uvicorn version this was tested against.

Minor: the PR description says "Generated with Claude Code" -- per our conventions, external-facing text should be first person with no agent attribution.

janhilgard:
Addressed both review notes:
1. Added version comment to `_find_uvicorn_cycle` documenting tested uvicorn (0.34-0.40) and starlette (0.44-0.46) versions, plus the silent fallback behavior.
2. Removed the Claude Code attribution from the PR description.

Thump604:
Approved. Main had a lint regression from a recent merge (F821 + black formatting) that should now be fixed. Once CI re-runs clean, this should be mergeable. You may need to rebase onto current main to trigger a fresh CI run.

PR #458, opened 2026-04-27
fix: token duplication in MLLM logits processor context

janhilgard:
## Summary

Fixes degenerated JSON output when `response_format` with `json_schema` is used with the MLLM batched engine.

### Root cause: token duplication

The last generated token was **duplicated** when passed to logits processors in `_step()`:

1. The caller builds `output_tokens` as `list(req.output_tokens) + [token]` — already includes the current token.
2. `_step()` then appends the same token again: `full_context = output_tokens[e] + [cur_tok]`.

This means the lm-format-enforcer sees `[{, {]` instead of `[{]`, which corrupts its internal parser state. It crashes, disables itself, and the model generates freely — producing degenerated whitespace/brace loops instead of valid JSON.

**Fix:** Remove the redundant `+ [cur_tok]` in `_step()`, since `output_tokens[e]` already contains the current token.

### Secondary: thinking-aware JSON constraining

For thinking models (Qwen 3.5/3.6, DeepSeek-R1, etc.), two additional improvements:

- **`_ThinkingAwareLogitsProcessor`**: Wraps the JSON schema processor so it only activates *after* the model emits `</think>`, letting it reason freely first. Without this, `enable_thinking` is forced to `False` when constrained decoding is active, which degrades output quality.
- **Softer system prompt rules**: When `--reasoning-parser` is configured, the system prompt avoids demanding `{` as the very first character (which conflicts with `<think>` blocks) and instead allows the model to reason before outputting JSON.

## Changes

- **`vllm_mlx/mllm_batch_generator.py`**: Remove duplicate token append in `_step()` logits processor context.
- **`vllm_mlx/server.py`**: Add `_ThinkingAwareLogitsProcessor` wrapper; use it when a reasoning parser is configured; thread `thinking_model` flag to system prompt builder.
- **`vllm_mlx/api/tool_calling.py`**: Add `thinking_model` parameter to `build_json_system_prompt()` for softer output rules.

## Test plan

- [ ] Verify JSON schema constrained decoding produces valid JSON (no degenerated output)
- [ ] Verify thinking models can reason in `<think>` blocks before JSON output
- [ ] Verify non-thinking models still get strict JSON rules and `enable_thinking=False`
- [ ] Verify the lm-format-enforcer does not crash/disable itself during generation

### DIFF
--- a/vllm_mlx/api/tool_calling.py
+++ b/vllm_mlx/api/tool_calling.py
@@ -849,6 +849,8 @@ def parse_json_output(
 
 def build_json_system_prompt(
     response_format: Optional[Union[ResponseFormat, Dict[str, Any]]] = None,
+    *,
+    thinking_model: bool = False,
 ) -> Optional[str]:
     """
     Build a system prompt instruction for JSON output.
@@ -858,6 +860,10 @@ def build_json_system_prompt(
 
     Args:
         response_format: ResponseFormat specification
+        thinking_model: When ``True`` use softer output rules that allow
+            the model to reason (think) before emitting JSON.  Strict rules
+            that demand ``{`` as the very first character conflict with
+            ``<think>`` blocks and cause degenerated output.
 
     Returns:
         System prompt instruction string, or None if not needed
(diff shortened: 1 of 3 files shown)


Thump604 (approved):

PR #2, opened 2026-01-08
Fix asyncio.get_event_loop() deprecation for Python 3.10+

waybarrios:
## Summary

Fixes #1

- Replace deprecated `asyncio.get_event_loop()` with `asyncio.new_event_loop()` + `asyncio.set_event_loop()` for Python 3.10+ compatibility
- This resolves the issue when running `vllm-mlx serve` in simple mode

## Changes

- `vllm_mlx/server.py`: Use `new_event_loop()` pattern instead of deprecated `get_event_loop()`

## Test

```bash
vllm-mlx serve mlx-community/Llama-3.2-3B-Instruct-4bit --port 8000
```

Thanks @rhychung for reporting and suggesting the fix!

PR #459, opened 2026-04-27
fix: JSON schema enforcer stuck on arrays of objects

### LINKED ISSUE #1: # Simple mode (single user, max throughput) failed with % vllm-mlx serve mlx-community/Llama-3.2-3B-Instruct-4bit --port 8000
suggest to modify line 176 in server.py from _**asyncio.get_event_loop().run_until_complete(_engine.start())**_ to **asyncio.new_event_loop().run_until_complete(_engine.start())**

janhilgard:
## Summary

Two fixes for constrained JSON decoding (`response_format: json_schema`):

1. **Array context bug**: `_get_json_context()` returned `"key_start"` after commas inside JSON arrays (e.g. `[{...},`), not just inside objects. The schema-aware key filter then removed `{` from allowed tokens (since it doesn't match any property name), causing the enforcer to get stuck with an empty allowed set. The processor disabled itself and the model generated freely, producing degenerate output.

2. **Force EOS on disable**: When the enforcer gets stuck and disables itself, the processor now forces EOS immediately instead of passing logits through unchanged. Previously, with `max_tokens=262144`, the model would continue generating useless output for minutes/hours, blocking the inference slot and never returning a response to the client.

**Fix #1**: Added `_container_stack` to track whether the current nesting context is an object (`"o"`) or array (`"a"`). `_get_json_context` now returns `"key_start"` / `"in_key"` only inside objects.

**Fix #2**: When `_disabled=True`, return an EOS-only mask instead of unmodified logits. Generation stops immediately and the server returns whatever was generated (the JSON extraction layer can often recover valid JSON from partial output).

## Test plan

- [x] 3 new tests in `TestJsonContextArrayVsObject`
- [x] All 28 constrained decoding tests pass
- [x] Live tested on 2 production servers (Qwen3.6-35B-A3B, Qwen3.6-27B) — all schemas produce valid JSON
- [x] Verified stuck requests now return in seconds instead of blocking for hours
- [x] 4 concurrent requests with `max_tokens=262144` all return quickly with valid JSON

### DIFF
--- a/tests/test_constrained_decoding.py
+++ b/tests/test_constrained_decoding.py
@@ -615,5 +615,93 @@ class TestIncrementalCaching:
                 assert arr[i] == -np.inf, f"Position {i} should be -inf"
 
 
+# ---------------------------------------------------------------------------
+# JSON context — array vs object discrimination.
+# ---------------------------------------------------------------------------
+
+
+@pytestmark_lmfe
+class TestJsonContextArrayVsObject:
+    """Verify ``_get_json_context`` distinguishes array from object context.
+
+    Regression test: after ``},`` inside a JSON array, the processor returned
+    ``"key_start"`` and ``_filter_key_start_tokens`` removed ``{``, preventing
+    the next array element from being generated.  The fix uses a container
+    stack to track nesting so ``"key_start"`` is only returned inside objects.
+    """
+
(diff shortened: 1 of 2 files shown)


Thump604 (approved):

PR #460, opened 2026-04-27
Fix generation_tps always 0 in batch stats, expose in /v1/status

janhilgard:
## Summary

- **Fix `batch_generator` stats not propagated**: `BatchedEngine.get_stats()` promotes selected keys from MLLM scheduler stats to top-level, but `batch_generator` was missing from the promoted keys list. This caused `generation_tps` and `prompt_tps` to always be 0 in `/v1/status`.
- **Fix batch extension time attribution**: Mid-generation batch extension (text-only requests joining an active batch) set `prompt_processing=True`, causing the entire step time to be attributed to `prompt_time` even though the dominant cost is generating tokens for existing requests. This made `generation_tps` permanently 0 while `prompt_tps` was artificially inflated.
- **Expose `generation_tps` and `prompt_tps`** from `MLLMBatchStats` in the `/v1/status` endpoint so monitoring dashboards can consume server-side throughput metrics.

## Test plan

- [x] Verify `generation_tps > 0` in `/v1/status` response during active generation (23.3 tok/s on 27B, 38.0 tok/s on 35B-A3B)
- [x] Verify `prompt_tps` reflects actual prompt processing time
- [ ] Run existing test suite

### DIFF
--- a/vllm_mlx/engine/batched.py
+++ b/vllm_mlx/engine/batched.py
@@ -1046,6 +1046,7 @@ class BatchedEngine(BaseEngine):
                 "memory_aware_cache",
                 "paged_cache",
                 "prefix_cache",
+                "batch_generator",
                 "requests",
             ):
                 if key in mllm_stats:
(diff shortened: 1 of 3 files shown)


Thump604 (approved):

PR #464, opened 2026-04-28
fix: streaming parser drops content newlines when tool parser active

### LINKED ISSUE #462: Streaming parser suppresses content newlines when tool parser is configured
## Bug

In `server.py`, the streaming chat completion path suppresses ALL whitespace-only content chunks when a tool call parser is configured:

```python
if not content.strip():
    continue
```

This is intended to skip leading whitespace between `</think>` and the first real content token. But it runs unconditionally on every chunk, so newlines between list items, paragraphs, and other content are silently dropped.

## Impact

Streaming responses with `enable_thinking=true` and a tool parser produce concatenated content:

```
1. London2. Paris3. Berlin4. Rome5. Madrid
```

Instead of:

```
1. London
2. Paris
3. Berlin
4. Rome
5. Madrid
```

Non-streaming responses are unaffected (separate path). Responses without a tool parser are unaffected.

## Fix

Gate the whitespace suppression on a `_first_content_emitted` flag. Only suppress whitespace-only chunks before real content starts. After the first non-whitespace content chunk, pass all chunks through including newlines.

Thump604:
## Summary

- Streaming whitespace suppression ate ALL whitespace-only chunks when a tool parser was configured, not just the leading whitespace between `</think>` and first content
- Newlines between list items, paragraphs, and other structured content were silently dropped
- Gate on `_first_content_emitted` flag so only pre-content whitespace is suppressed

## Before/After

**Before:** `1. London2. Paris3. Berlin4. Rome5. Madrid`
**After:** `1. London\n2. Paris\n3. Berlin\n4. Rome\n5. Madrid`

Non-streaming and responses without tool parser were unaffected.

## Test plan

- [ ] Stream a numbered list request with `enable_thinking=true` and a tool parser configured
- [ ] Verify newlines appear between list items in the streamed content
- [ ] Verify leading whitespace between `</think>` and first content is still suppressed
- [ ] Verify tool call XML parsing still works (not broken by passing through whitespace)

Closes #462

### DIFF
--- a/vllm_mlx/server.py
+++ b/vllm_mlx/server.py
@@ -4994,6 +4994,7 @@ async def stream_chat_completion(
     tool_accumulated_text = ""
     tool_calls_detected = False
     tool_markup_possible = False  # Fast path: skip parsing until markers appear
+    _first_content_emitted = False
     tool_parser = _get_streaming_tool_parser(request, engine)
 
     try:
@@ -5052,9 +5053,7 @@ async def stream_chat_completion(
                         )
                     ):
                         tool_accumulated_text += content
-                        # Suppress whitespace-only content when tools are active;
-                        # avoids emitting stray newlines before tool call XML.
-                        if not content.strip():
+                        if not content.strip() and not _first_content_emitted:
                             continue
                     else:
                         if not tool_markup_possible:
@@ -5134,6 +5133,8 @@ async def stream_chat_completion(
                         if flush:
                             content = content + flush
 
+                if content and not _first_content_emitted:
+                    _first_content_emitted = True
                 chunk = ChatCompletionChunk(
                     id=response_id,
                     model=_model_name,


janhilgard (approved):
LGTM — clean and minimal fix.

The `_first_content_emitted` flag is correctly placed after fence_stripper processing (line 5133), so empty-after-stripping content won't prematurely set it. Reasoning-only chunks (thinking models) also don't trigger it since `content` stays falsy.

No conflicts with any open PRs (#458, #459, #460) — none of them touch the streaming path in `stream_chat_completion`.

Thump604:
Duplicate of #432 which has a cleaner approach — removes the suppression block entirely rather than adding a gate. @waybarrios feel free to close this in favor of #432.

Thump604:
Closing — #432 (luizribeiro) merged with a cleaner fix for the same streaming whitespace issue.

PR #466, opened 2026-04-28
fix: enforce workload timeouts in bench serve

Thump604:
## Summary

I changed bench-serve workload execution so a case-level policy_timeout_ms now acts as an execution boundary for that case, not just post-hoc metadata. When a streaming request exceeds the timeout, bench-serve closes the stream, attempts best-effort server-side cancellation using the OpenAI stream id, and records the case as a request error instead of allowing the qualification run to hang indefinitely.

This also preserves the stream id from SSE chunks so the cancellation endpoint can be addressed when the server supports /v1/requests/{id}/cancel. Servers without that endpoint continue to work because cancellation is best-effort.

## Why

Long-thinking qualification cases can otherwise pin a workload run well past the intended case budget. That makes feature-matrix results hard to trust because one pathological case can block stage progress and leave operators guessing whether the harness or the runtime is stuck.

## Tests

- python -m pytest -q tests/test_bench_serve.py
- uvx ruff check vllm_mlx/bench_serve.py tests/test_bench_serve.py --select E,F,W --ignore E402,E501,E731,F811,F841
- black --check --target-version py312 vllm_mlx/bench_serve.py tests/test_bench_serve.py

### DIFF
--- a/tests/test_bench_serve.py
+++ b/tests/test_bench_serve.py
@@ -38,6 +38,7 @@ from vllm_mlx.bench_serve import (
     parse_status_response,
     run_bench_serve_workload,
     run_workload_case,
+    stream_chat_completion,
     summarize_workload_results,
     validate_quality_checks,
     validate_response,
@@ -622,6 +623,7 @@ class TestSSEParsing:
     def _make_line(self, delta_content=None, finish_reason=None, usage=None):
         """Build a synthetic SSE data line."""
         chunk: dict = {
+            "id": "chatcmpl-test",
             "choices": [
                 {
                     "delta": (
@@ -629,7 +631,7 @@ class TestSSEParsing:
                     ),
                     "finish_reason": finish_reason,
                 }
-            ]
+            ],
         }
         if usage is not None:
             chunk["usage"] = usage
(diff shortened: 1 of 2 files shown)


Thump604:
Jan, could you review this one when you have a chance?

This is a narrow follow-up to the workload-contract work: `policy_timeout_ms` now bounds the actual case execution instead of only being recorded as metadata, and streaming timeout paths best-effort cancel the server request when an SSE id is available.

I kept it scoped to `bench_serve.py` plus tests. CI is green, including Apple Silicon. The main behavior to scrutinize is the timeout/cancel path and whether the SSE id capture is the right level of best-effort cancellation for bench workloads.

janhilgard (approved):
Clean implementation. The `_consume_stream()` extraction + `asyncio.timeout()` wrapper is the right approach. Best-effort cancellation via `/v1/requests/{id}/cancel` degrades gracefully for servers without the endpoint. Test coverage for the timeout + cancel URL flow is solid. LGTM.

Thump604:
Hey Jan -- this one is approved and mergeable. Can you merge it when you get a chance? Also, #470 and #469 are in the same state (approved, mergeable, clean CI) if you can take those too. Thanks.

PR #467, opened 2026-04-28
Fix/test stream thread safety additions

perry2of5:
These got dropped from @Thump604's branch fix/test-stream-thread-safety, but are needed on main.  They were merged into Thump604's branch, but then somehow dropped.

### DIFF
--- a/tests/test_chat_template_kwargs.py
+++ b/tests/test_chat_template_kwargs.py
@@ -245,24 +245,31 @@ async def test_simple_engine_mllm_chat_forwards_chat_template_kwargs():
         engine = SimpleEngine("test-model")
         engine._loaded = True
         engine._is_mllm = True
+        engine._text_model = None
         engine._model = MagicMock()
-        engine._model.chat = MagicMock(
-            return_value=SimpleNamespace(
-                text="OK",
-                prompt_tokens=5,
-                completion_tokens=1,
-                finish_reason="stop",
+        engine._model.stream_chat = MagicMock(
+            return_value=iter(
+                [
+                    SimpleNamespace(
+                        text="OK",
+                        prompt_tokens=5,
+                        finish_reason="stop",
+                    )
+                ]
             )
         )
 
-        await engine.chat(
+        output = await engine.chat(
             [{"role": "user", "content": "Hello"}],
             chat_template_kwargs={"enable_thinking": False},
         )
 
-        assert engine._model.chat.call_args.kwargs["chat_template_kwargs"] == {
+        assert output.text == "OK"
+        assert engine._model.stream_chat.call_args.kwargs["chat_template_kwargs"] == {
             "enable_thinking": False
         }
+        # Text-only MLLM non-stream chat now aggregates the streaming path.
+        assert engine._model.chat.call_count == 0
 
 
 @pytest.mark.anyio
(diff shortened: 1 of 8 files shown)


Thump604 (approved):
Reviewed the full diff and the fork PR context. This appears to restore the broader stream-thread-safety changes that were missing from the narrower #452 path.

I think this is mergeable as a correctness fix. The important parts are covered: default load/preload stay on the same thread as later stream execution, stream/non-stream mode switching gets explicit regression coverage, text-only MLLM non-stream avoids the broken mlx_vlm non-stream path, and pending removals no longer have the snapshot/clear lost-enqueue window. CI is green and the tests target the right failure mode.

One follow-up I would keep separate: the `mlx_streams` lock serializes rebinding of module-level `generation_stream` references, but it does not make module-level generation streams a complete multi-engine concurrency boundary while generation is in progress. That is not a reason to block this PR, but I would avoid treating that lock as proof that concurrent engines cannot clobber each other. The stronger long-term answer is either a per-engine generation owner thread or avoiding mutable module-level stream state in the generation path.

perry2of5:
I keep thinking we should move things into a central worker that processes a queue to keep concurrent engines from clobbering each other.

perry2of5:
It seems like there are two obvious solutions if we don't move to a centralized worker:

Use threading.local() inside bind_generation_streams to store a per-thread reference, and have each generation call retrieve it from the TLS slot, bypassing the module global entirely.

OR 

document that _generation_lock must be an application-wide singleton lock (not one per engine) so that at most one engine enters generation at a time, making the module global safe again.

Thump604:
I think TLS is the better of those two short-term directions.

A process-wide singleton generation lock is the safest emergency guardrail, but it turns module-global stream state into a scalability bottleneck and makes multi-engine serving effectively serialized. That may be acceptable as a temporary defensive fallback, but I would not want it to become the architecture.

The stronger invariant I would aim for is:

one engine -> one MLX generation owner -> one stream-binding domain

That could be implemented as a per-engine generation owner thread/executor, with load/preload/prefill/decode/stream and non-stream generation all entering MLX through that owner. TLS or engine-owned stream context can then remove the mutable module-global clobbering problem without forcing unrelated engines through one global lock.

For this PR, I would keep the scope tactical and merge the concrete regression fixes. I’d put the owner-thread/TLS design in #468 so this PR does not grow into an architecture rewrite.

perry2of5:
That sounds like a good approach to me. Merging this PR soon has concrete real-world benefits.

janhilgard (approved):
Thorough fix for MLX thread-local stream ownership drift across four entry points (startup, nonstream→stream transition, MLLM text-only chat, residency preload). The approach is sound — keeping model load and stream generation on the same thread preserves MLX stream-local ownership.

Key observations:

1. **Inline `prepare_for_start`** blocks the event loop during model load, but this is acceptable — it happens once at startup before any requests are served, and the alternative (threaded) path caused the ownership mismatch.

2. **Atomic swap with lock** for `_pending_removal_uids` correctly closes the race window between snapshot and clear. The `_InjectOnExhaustionSet` test deterministically exercises the exact lost-enqueue window — nice approach.

3. **`_STREAM_REBIND_LOCK`** in `mlx_streams.py` prevents concurrent engines from clobbering module-level `generation_stream` references during rebinding.

4. **MLLM text-only broadening** (removing `self._text_model is not None` guard) is the right trade-off — safety over potential non-stream `chat()` optimization.

Minor: 4 commits could be squashed for cleaner history, but not a blocker.

CI 9/9 pass.

PR #469, opened 2026-04-29
Fix MLLM prefix cache disable wiring

Thump604:
## Summary

Fixes the MLLM continuous-batching path so cache disable flags actually reach `MLLMSchedulerConfig`.

Before this change, the serve CLI computed `enable_prefix_cache = args.enable_prefix_cache and not args.disable_prefix_cache`, but `BatchedEngine._start_mllm()` did not forward that value into the MLLM scheduler. As a result, `--disable-prefix-cache` could still create `MemoryAwarePrefixCache` on MLLM continuous batching.

This also forwards `use_memory_aware_cache`. MLLM currently only has a memory-aware prefix cache implementation, so `--no-memory-aware-cache` disables MLLM prefix caching instead of silently creating the memory-aware cache anyway.

## Why

This addresses the latest #442 telemetry where current main plus `--max-kv-size 65536` still showed `memory_aware_cache` entries growing even when the server was started with `--disable-prefix-cache`.

## Tests

- `python -m pytest tests/test_batched_engine_mllm_config.py -q`
- `python -m pytest tests/test_mllm_continuous_batching.py::TestBatchedMLLMConfigWiring::test_batched_engine_forwards_prefill_step_size_to_mllm_scheduler -q -rs` (skips locally without MLX, should run in CI)

## Notes

This does not claim to close all Metal memory growth in #442. It fixes the first concrete control-plane bug exposed by the new data: cache-off flags were not actually honored by the MLLM continuous-batching path. After this lands, the clean next reproduction is current main with `--max-kv-size 65536 --disable-prefix-cache` and sustained traffic to see whether `memory_aware_cache.entry_count` remains zero and whether Metal active memory plateaus.

### DIFF
--- a/tests/test_batched_engine_mllm_config.py
+++ b/tests/test_batched_engine_mllm_config.py
@@ -1,0 +1,79 @@
+"""BatchedEngine MLLM scheduler configuration wiring tests.
+
+These tests avoid model loading and MLX imports. They validate that CLI-level
+SchedulerConfig fields survive the BatchedEngine -> MLLMSchedulerConfig bridge.
+"""
+
+import asyncio
+import sys
+import types
+from types import SimpleNamespace
+
+
+def test_start_mllm_forwards_prefix_cache_disable_to_mllm_scheduler(monkeypatch):
+    from vllm_mlx.engine.batched import BatchedEngine
+
+    captured = {}
+
+    class FakeMLXMultimodalLM:
+        def __init__(self, model_name, trust_remote_code=True, **kwargs):
+            self.model_name = model_name
+            self.model = object()
+            self.processor = object()
+
+        def load(self):
+            return None
+
+    class FakeMLLMSchedulerConfig:
+        def __init__(self, **kwargs):
+            captured["config_kwargs"] = kwargs
+            self.__dict__.update(kwargs)
+
+    class FakeMLLMScheduler:
+        def __init__(self, model, processor, config):
+            captured["scheduler_config"] = config
+
+        async def start(self):
+            return None
+
+    import vllm_mlx.engine.batched as batched_mod
+
+    fake_mllm_scheduler = types.ModuleType("vllm_mlx.mllm_scheduler")
+    fake_mllm_scheduler.MLLMScheduler = FakeMLLMScheduler
+    fake_mllm_scheduler.MLLMSchedulerConfig = FakeMLLMSchedulerConfig
+    fake_mllm_model = types.ModuleType("vllm_mlx.models.mllm")
+    fake_mllm_model.MLXMultimodalLM = FakeMLXMultimodalLM
+    monkeypatch.setitem(sys.modules, "vllm_mlx.mllm_scheduler", fake_mllm_scheduler)
+    monkeypatch.setitem(sys.modules, "vllm_mlx.models.mllm", fake_mllm_model)
+    monkeypatch.setattr(
+        batched_mod.BatchedEngine, "_inject_mtp_mllm", lambda self: None
+    )
+
+    engine = BatchedEngine(
+        model_name="fake-qwen",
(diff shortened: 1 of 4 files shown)


Thump604:
Local validation on M2 Ultra 128GB using Qwen3.6-35B-A3B-VLM-MTP-8bit:

1. Rebuilt the local runtime overlay with this patch.
2. Started a temporary single-server process with:

```text
--mllm
--continuous-batching
--disable-prefix-cache
--no-memory-aware-cache
--prefill-step-size 256
```

3. Checked `/v1/status` before and after a smoke request. Cache stats stayed disabled/empty:

```json
{
  "hits": 0,
  "misses": 0,
  "current_memory_mb": 0.0,
  "max_memory_mb": 0.0,
  "entry_count": 0
}
```

The pre-fix behavior from #442 was that the same cache-off profile still reported `memory_aware_cache` entries growing. With this patch, the MLLM scheduler no longer constructs that cache when the cache-off flags are set.

This is not a long-run plateau claim. It only verifies the control-plane bug fixed here: the cache-off flags now reach the MLLM continuous-batching path.

Thump604:
Follow-up validation after the cache-off control fix:

I found one important correction and one dependency issue while validating this PR locally.

Correction: text-only traffic on my Qwen 3.6 35B VLM lane routes through the direct TextModel path, so text-only sustained traffic does not prove sustained MLLM media-path behavior. I reran with an image payload and verified response metadata `routing=mllm_batch` before treating it as MLLM evidence.

The media-path smoke then exposed a separate owned-dependency bug in `mlx-qsdpa` when `--max-kv-size 65536` and 8-bit KVQ are active together:

- `QuantizedRotatingSDPACache` was missing `merge()` for MLLM per-request cache merging.
- `BatchQuantizedRotatingSDPACache.make_mask()` used logical multimodal offset for mask width instead of visible K/V buffer length.
- Compact merged rotating buffers did not grow before decode writes.

I fixed that in `mlx-qsdpa` and opened https://github.com/Thump604/mlx-qsdpa/pull/1. Local package validation: `python -m pytest tests -q` -> `95 passed`.

After vendoring that fix locally, the PR #469 cache-off envelope produced this evidence:

```text
model: qwen3.6-35b-a3b VLM
isolated port: :8094
flags: --mllm --continuous-batching --max-num-seqs 3 --disable-prefix-cache --no-memory-aware-cache --kv-cache-quantization --kv-cache-quantization-bits 8 --kv-cache-quantization-group-size 64
bounded KV: max_kv_size=65536 via runtime mode
live envelope: PASS
media smoke: PASS, routing=mllm_batch, kv_quantize=true
sustained media run: 600 MLLM requests at concurrency 3
result: 200/200 bench result rows validated, representing 600 streamed MLLM requests total, no server errors
/v1/status cache: entry_count=0, current_memory_mb=0.0, max_memory_mb=0.0, hits=0, misses=0, evictions=0
Metal: active returned to 41.0GB, peak 43.59GB
```

I still do not want to overstate this as closing #442 globally. It does show that this PR makes the cache-off controls real on the MLLM scheduler, and with the `mlx-qsdpa` rotating-cache fix in place, the MLLM media path no longer shows the cache-retention pattern on this local sustained run.

janhilgard (approved):
Clean fix for a real control-plane bug — `--disable-prefix-cache` was silently ignored on the MLLM continuous-batching path because `_start_mllm()` never forwarded the cache flags to `MLLMSchedulerConfig`.

The fix is well-scoped: forward the three missing fields, gate `MemoryCacheConfig` creation on both `enable_prefix_cache` and `use_memory_aware_cache`, and preserve existing defaults (`True`) for backward compatibility.

Good test coverage with both a dedicated wiring test and extension of the existing `TestBatchedMLLMConfigWiring`.

CI 9/9 pass.

PR #472, opened 2026-04-29
fix: skip preamble text before activating JSON schema enforcer

janhilgard:
## Summary

- Add a "waiting for JSON start" phase to `_ThinkingAwareLogitsProcessor` that scans for the first `{` or `[` token before activating the JSON schema enforcer
- Prevents enforcer crashes (`KeyError: ''`) when models emit preamble text (e.g. `"Here's a "`, `"Thinking Process:\n"`) before the actual JSON object
- Includes a 50-token safety limit that force-activates the enforcer if no JSON start character is found

### Problem

After `</think>` or the 3-token detection window, the enforcer activated immediately on the next token. But thinking models sometimes generate preamble text before JSON — the enforcer saw this as invalid JSON, crashed with `KeyError`, got disabled, and the model generated freely up to `max_tokens` (262144). This caused 30-60 minute blocked requests and 504 timeouts.

### Fix

Extend the state machine from 3 to 4 phases:

```
[Detection] → [Thinking] → [Waiting for JSON] → [Active]
                   ↓
             [Waiting for JSON] → [Active]
```

Key invariant: `self._inner.__call__` is never invoked until a `{` or `[` token is found, so the enforcer stays in its bootstrap state and cannot crash on preamble text.

## Test plan

- [x] 8 unit tests covering all state transitions (preamble skip, direct JSON, think+preamble, safety limit, array start, BPE tokens)
- [x] 10/10 integration tests on production server with `response_format: json_schema` — all valid JSON, zero enforcer crashes
- [x] Existing test suite passes (no regressions)

### DIFF
--- a/tests/test_thinking_aware_processor.py
+++ b/tests/test_thinking_aware_processor.py
@@ -1,0 +1,359 @@
+"""Tests for _ThinkingAwareLogitsProcessor preamble bypass fix.
+
+The processor must wait for the first JSON-start character ({ or [) before
+activating the inner JSONSchemaLogitsProcessor.  Without this, preamble text
+like "Here's a " crashes the enforcer with KeyError: ''.
+"""
+
+# ---------------------------------------------------------------------------
+# Minimal fakes
+# ---------------------------------------------------------------------------
+
+
+class _FakeTokenizer:
+    """Maps token IDs → strings via a simple lookup table."""
+
+    def __init__(self, vocab: dict[int, str]):
+        self._vocab = vocab
+
+    def decode(self, ids: list[int]) -> str:
+        return "".join(self._vocab.get(i, "") for i in ids)
+
+
+class _FakeInnerProcessor:
+    """Stands in for JSONSchemaLogitsProcessor.
+
+    Records every call so tests can assert when/how activation happened.
+    """
+
+    def __init__(self, tokenizer):
+        self._tokenizer = tokenizer
+        self._prompt_len: int | None = None
+        self._disabled = False
+        self.calls: list[tuple[list[int], str]] = []  # (tokens, logits)
+
+    def __call__(self, tokens, logits):
+        self.calls.append((list(tokens), logits))
+        return "CONSTRAINED"
+
+    @property
+    def schema(self):
+        return {}
+
+
+# ---------------------------------------------------------------------------
+# Import the class under test
+# ---------------------------------------------------------------------------
+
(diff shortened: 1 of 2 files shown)


Thump604 (changes requested):
I think this needs one more state boundary before merge.

The new waiting phase currently scans from `_base_prompt_len` every time. That is fine for the no-thinking path, but after `</think>` it means `_scan_for_json_start()` also scans all hidden reasoning tokens. If the reasoning contains `{` or `[` before the final answer, the processor can activate at a token inside the reasoning span and set `_inner._prompt_len` to the wrong offset. The same issue applies to the 50-token safety limit: it counts thinking tokens, so a normal long reasoning block can force-activate immediately after `</think>` instead of allowing the post-thinking preamble window.

I would track a separate JSON search start offset. For the no-thinking path, initialize it to `_base_prompt_len`. For the thinking path, initialize it only when `</think>` is observed, at the token position after the closing thinking marker. Then scan and count safety tokens from that offset, not from the original generation start.

A regression test should include hidden reasoning with JSON-like text before `</think>`, for example `<think>consider {"bad": true}</think>Here is {"ok": true}`. The enforcer should activate at the final answer `{`, not at the brace inside the thinking span.

janhilgard:
Good catch. Added `_json_scan_offset` that tracks where the JSON search begins:

- **No-thinking path**: set to `_base_prompt_len` (same as before)
- **Thinking path**: set to `n` (current token count) when `</think>` is detected — so `{`/`[` inside the reasoning span are invisible to the scanner
- **Safety limit**: counts tokens from `_json_scan_offset`, not from `_base_prompt_len` — long reasoning blocks don't eat into the 50-token budget

Added 3 regression tests:
- `test_json_inside_thinking_ignored`: `<think>consider {"bad": true}</think>Here is {"ok": true}` — enforcer activates at answer `{`, not reasoning `{`
- `test_json_array_inside_thinking_ignored`: `[` inside thinking ignored, answer `[` triggers enforcer
- `test_safety_limit_counts_from_scan_offset`: 10 thinking tokens (some with `{`) + 5 post-think tokens → still waiting (only 5 tokens past scan offset, not 50)

Thump604 (approved):
The _json_scan_offset fix is exactly what I asked for. Reasoning-phase braces are invisible to the scanner, safety limit counts from the right offset, and the three regression tests cover the edge cases. Clean to merge.

PR #477, opened 2026-04-30
fix: suppress thinking for json_schema constrained decoding

janhilgard:
## Summary

- Fix `TokenEnforcerTokenizerData` constructor to pass the required `use_bitmask` and `vocab_size` parameters (added in lm-format-enforcer v0.11.3). Without these the enforcer silently fails to build and json_schema constraints are ignored.
- Only activate `_ThinkingAwareLogitsProcessor` when the user explicitly sets `enable_thinking=True`. For all other json_schema requests, suppress thinking (`enable_thinking=False`) so the template emits an empty `<think></think>` block and the enforcer constrains output from the first token.

### Problem

On thinking models (Qwen 3.5/3.6) with `response_format: json_schema`:

1. The enforcer failed to build due to missing constructor args — model generated unconstrained text.
2. When the enforcer did build, `_ThinkingAwareLogitsProcessor` searched for `<think>` in generated output, but Qwen 3.5/3.6 templates inject `<think>\n` directly into the prompt. The tag never appeared in model output, causing the processor to either crash the enforcer on non-JSON preamble or let the model exhaust `max_tokens` in unconstrained thinking.

### Changes

- `constrained/cache.py`: Add `use_bitmask=False, vocab_size=vocab_size` to `TokenEnforcerTokenizerData()`
- `server.py`: Default to `enable_thinking=False` for json_schema requests (OpenAI + Anthropic paths); only use `ThinkingAwareLogitsProcessor` when `enable_thinking` is explicitly `True`; add `prompt_has_think_tag` parameter to handle template-injected thinking tags

## Test plan

- [x] Tested on Qwen3.6-27B (dense) and Qwen3.6-35B-A3B (MoE) — 20/20 JSON schema tests passing
- [x] Tested schemas: person extraction, sentiment analysis, list extraction, system prompt, preamble bait
- [x] Both OpenAI and Anthropic endpoints produce valid constrained JSON

### DIFF
--- a/vllm_mlx/constrained/cache.py
+++ b/vllm_mlx/constrained/cache.py
@@ -168,6 +168,8 @@ def get_tokenizer_data(tokenizer: Any) -> Any | None:
             regular_tokens,
             decode_fn,
             eos_token_id,
+            use_bitmask=False,
+            vocab_size=vocab_size,
         )
     except Exception as exc:
         logger.warning("Failed to build TokenEnforcerTokenizerData: %s", exc)
--- a/vllm_mlx/server.py
+++ b/vllm_mlx/server.py
@@ -440,10 +440,14 @@ class _ThinkingAwareLogitsProcessor:
     think before answering.
     """
 
-    def __init__(self, inner):
+    def __init__(self, inner, prompt_has_think_tag: bool = False):
         self._inner = inner
         self._active = False
-        self._in_thinking: bool | None = None  # None = not yet determined
+        # When the chat template already injects ``<think>\n`` into the prompt,
+        # the model's output starts INSIDE the thinking block — so ``<think>``
+        # will never appear in the generated tokens.  Pre-set ``_in_thinking``
+        # to skip the 3-token detection and go straight to ``</think>`` scan.
+        self._in_thinking: bool | None = True if prompt_has_think_tag else None
         self._waiting_for_json = False  # Waiting for { or [ before activating
         self._base_prompt_len: int | None = None
         self._json_scan_offset: int | None = None  # Absolute offset to start scanning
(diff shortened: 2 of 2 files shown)


Thump604 (changes requested):
I think this is the right direction, but I would not merge it as-is yet.

The `TokenEnforcerTokenizerData` constructor fix looks necessary. The default behavior for `json_schema` also makes sense operationally: if the user did not explicitly request thinking, constrained decoding should suppress thinking and constrain JSON from token one.

The blocker is that the implementation currently reasons about `request.enable_thinking`, but the rendered Qwen prompt can still be controlled through `chat_template_kwargs` or server defaults. The engines apply `chat_template_kwargs` after the top-level `enable_thinking` value, so a request or server default with `chat_template_kwargs={"enable_thinking": true}` can still render `<think>` while the enforcer is active from token one. That recreates the class of failure this PR is fixing.

I would make the effective thinking decision explicit:

- If JSON schema suppression is selected, force both `chat_kwargs["enable_thinking"] = False` and `chat_kwargs["chat_template_kwargs"]["enable_thinking"] = False`.
- If explicit thinking is requested through either top-level `enable_thinking=True` or `chat_template_kwargs={"enable_thinking": true}`, use the thinking-aware wrapper with `prompt_has_think_tag=True`.
- Apply the same rule in the Anthropic path. Right now OpenAI has an explicit-thinking escape hatch and Anthropic does not.
- Add regressions that assert the rendered prompt state, not just the request fields, for OpenAI and Anthropic with server defaults and request `chat_template_kwargs`.

This also lines up with the Qwen 3.5/3.6 template issue being discussed here: https://huggingface.co/froggeric/Qwen-Fixed-Chat-Templates. The important lesson for this PR is not necessarily to adopt that template wholesale; it is that the rendered Jinja prompt is the source of truth. JSON schema enforcement should not trust the request object alone.

janhilgard:
Thanks for the thorough review. Updated the PR to address the `chat_template_kwargs` gap:

**Changes:**

1. **Effective thinking detection** — both OpenAI and Anthropic paths now check `request.enable_thinking` AND `chat_template_kwargs["enable_thinking"]` to determine the rendered prompt state. A server default like `--default-chat-template-kwargs '{"enable_thinking": true}'` is correctly detected.

2. **Suppression covers both layers** — when suppressing thinking for JSON schema, `enable_thinking=False` is forced in both `chat_kwargs["enable_thinking"]` and `chat_kwargs["chat_template_kwargs"]["enable_thinking"]`. This prevents the Jinja template from rendering an open `<think>` block regardless of where the setting originated.

3. **Anthropic path parity** — the Anthropic endpoint now has the same explicit-thinking escape hatch as OpenAI. If `chat_template_kwargs={"enable_thinking": true}` is set, the ThinkingAware wrapper is used; otherwise thinking is suppressed.

Tested 20/20 on Qwen3.6-27B (dense) and Qwen3.6-35B-A3B (MoE) after the update.

Thump604 (commented):
The update fixes the main gap on the OpenAI path. Checking effective thinking from both top-level `enable_thinking` and `chat_template_kwargs["enable_thinking"]`, and forcing both layers off during suppression, is the right change.

I still see one remaining blocker in the Anthropic path. It currently derives `effective_thinking` only from `chat_template_kwargs["enable_thinking"]`:

```python
ctk = chat_kwargs.get("chat_template_kwargs") or {}
effective_thinking = ctk.get("enable_thinking") is True
```

The OpenAI path correctly treats either top-level `request.enable_thinking is True` or `chat_template_kwargs["enable_thinking"] is True` as an explicit-thinking request. Anthropic should do the same. Otherwise a request that sets top-level `enable_thinking=True` can still diverge from the constrained-decoding path on Anthropic, which leaves the parity issue only partially fixed.

I’d make Anthropic use the same rule as OpenAI:

```python
effective_thinking = (
    openai_request.enable_thinking is True
    or ctk.get("enable_thinking") is True
)
```

Once that is aligned, this looks close.

janhilgard:
Good catch — fixed in the latest push. The Anthropic path now uses the same two-source detection as OpenAI:

```python
effective_thinking = (
    openai_request.enable_thinking is True
    or ctk.get("enable_thinking") is True
)
```

Both paths are now symmetric: top-level `enable_thinking` and `chat_template_kwargs["enable_thinking"]` are treated identically, with two-layer suppression when JSON schema is active without a reasoning parser.

All 20/20 JSON schema tests pass on both servers.

Thump604 (approved):
I re-reviewed the latest commit and I’m good with this change.

The remaining Anthropic parity issue I called out is addressed: effective thinking is now derived from both top-level enable_thinking and chat_template_kwargs["enable_thinking"], matching the OpenAI path. The suppression path also forces both layers off before attaching the JSON logits processor.

I also spot-checked the prepared invocation paths locally with mocked constrained decoding, including:
- OpenAI path with server-default chat_template_kwargs enable_thinking=true wraps with prompt_has_think_tag.
- OpenAI path with request chat_template_kwargs enable_thinking=false suppresses both layers.
- Anthropic path with top-level enable_thinking=true wraps with prompt_has_think_tag.

Focused local tests passed where available:
- tests/test_chat_template_kwargs.py
- tests/test_constrained_decoding.py, with lm-format-enforcer-dependent cases skipped in this local venv
- the two Anthropic chat_template_kwargs server tests in tests/test_server.py

CI is green on the PR. I’m not merging this myself.

PR #478, opened 2026-04-30
fix(simple): use persistent MLX worker thread to fix thread-local stream crash

### LINKED ISSUE #407: `RuntimeError: There is no Stream(gpu, N) in current thread` when calling `/v1/messages` or `/v1/chat/completions` on Qwen3.5 MLX models (MLLM path runs in worker thread)

xykong:
## Summary

Replaces `asyncio.to_thread()` + `bind_generation_streams()` in `SimpleEngine` with a persistent `MLXWorkerThread` that loads the model and runs all inference on the **same** OS thread, fixing `RuntimeError: There is no Stream(gpu, N) in current thread` on MLX >= 0.31.2.

## Problem

MLX 0.31.2 ([ml-explore/mlx#3348](https://github.com/ml-explore/mlx/pull/3348)) made Metal `CommandEncoder` storage thread-local. The current `SimpleEngine._run_blocking_serialized()` spawns transient threads via `asyncio.to_thread()` and calls `bind_generation_streams()` which creates a *new* stream index via `mx.new_stream()`. However, the model's compiled Metal kernels reference the original stream index from the loading thread — that index does not exist in the transient worker thread's TLS.

**Reproduction**: Run `vllm-mlx` with `SimpleEngine` on Apple Silicon with MLX >= 0.31.2 and any model that triggers Metal kernel compilation during load (most models). The server crashes on first inference request.

## Root Cause

1. Model loading creates Metal compiled kernels associated with stream index N (created by C++ layer during compilation, not by Python `mx.new_stream()`)
2. `asyncio.to_thread()` spawns a new thread each time
3. `bind_generation_streams()` calls `mx.new_stream()` → creates stream M (M ≠ N)
4. Forward pass references stream N → **crash**: `There is no Stream(gpu, N) in current thread`

## Fix

Introduce `MLXWorkerThread` — a persistent single-threaded executor:
- Model loads on this thread (stream N created here)
- All inference runs on this same thread (stream N always available)
- No stream rebinding needed

This matches the proven approach in:
- [jundot/omlx PR #891](https://github.com/jundot/omlx/pull/891) — `ThreadPoolExecutor(max_workers=1)` with initializer
- [MLX maintainer recommendation](https://github.com/ml-explore/mlx/issues/3216) — persistent thread for Metal ops

## Changes

- **`vllm_mlx/mlx_streams.py`**: Added `MLXWorkerThread` class (persistent daemon thread with `queue.SimpleQueue` task dispatch). `bind_generation_streams()` retained for backward compatibility but deprecated.
- **`vllm_mlx/engine/simple.py`**: 
  - `start()` loads model via `_mlx_worker.submit()` instead of event-loop-thread load
  - `_run_blocking_serialized()` submits to `_mlx_worker` instead of `asyncio.to_thread()`
  - `stream_generate()` and `stream_chat()` MLLM text-only fallback route through worker
  - Removed `_bind_worker_generation_streams()` and `run_blocking_startup_work` usage
- **`tests/test_mlx_worker_thread.py`**: New test file verifying thread persistence, FIFO ordering, exception propagation, and MLX Metal ops.

## Testing

Tested on Apple M4 Max 128GB, macOS Tahoe 26.3.1, MLX 0.31.2, with `mlx-community/gemma-4-26b-a4b-it-4bit`:

| Metric | Before (crash) | After |
|--------|----------------|-------|
| Server start | ✅ | ✅ |
| Health check | ❌ RuntimeError | ✅ 200 OK |
| Short prompt (tok/s) | N/A | 116.0 |
| Medium prompt (tok/s) | N/A | 102.1 |
| TTFT (ms) | N/A | 8.5–17.6 |

## Related Issues

- Closes #407
- Related: [ml-explore/mlx#3348](https://github.com/ml-explore/mlx/pull/3348) (root cause)
- Related: [ml-explore/mlx#3216](https://github.com/ml-explore/mlx/issues/3216) (discussion)
- Related: [jundot/omlx#891](https://github.com/jundot/omlx/pull/891) (same fix in oMLX)

### DIFF
--- a/tests/test_mlx_worker_thread.py
+++ b/tests/test_mlx_worker_thread.py
@@ -1,0 +1,106 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for MLXWorkerThread persistent worker thread.
+
+Verifies that model loading and inference on the same persistent thread avoids
+the ``RuntimeError: There is no Stream(gpu, N) in current thread`` crash that
+occurs when MLX >= 0.31.2 thread-local CommandEncoders are accessed from
+transient worker threads (e.g. asyncio.to_thread).
+"""
+
+import asyncio
+import threading
+
+import pytest
+
+
+def _mlx_available() -> bool:
+    try:
+        import mlx.core as mx
+
+        return mx.metal.is_available()
+    except (ImportError, AttributeError):
+        return False
+
+
+@pytest.mark.anyio
+async def test_mlx_worker_thread_runs_on_persistent_thread():
+    """All submissions execute on the same OS thread."""
+    from vllm_mlx.mlx_streams import MLXWorkerThread
+
+    worker = MLXWorkerThread(name="test-worker")
+    loop = asyncio.get_event_loop()
+
+    thread_ids = []
+    for _ in range(5):
+        tid = await worker.submit(loop, threading.get_ident)
+        thread_ids.append(tid)
+
+    assert len(set(thread_ids)) == 1, "Worker must use a single persistent thread"
+    assert thread_ids[0] != threading.get_ident(), (
+        "Worker thread must differ from event loop thread"
+    )
+    worker.shutdown()
+
+
+@pytest.mark.anyio
+async def test_mlx_worker_thread_preserves_exception():
+    """Exceptions from submitted callables propagate correctly."""
+    from vllm_mlx.mlx_streams import MLXWorkerThread
+
+    worker = MLXWorkerThread(name="test-exc")
+    loop = asyncio.get_event_loop()
+
+    def raise_value_error():
+        raise ValueError("test error")
+
+    with pytest.raises(ValueError, match="test error"):
+        await worker.submit(loop, raise_value_error)
+
+    worker.shutdown()
+
+
+@pytest.mark.anyio
+async def test_mlx_worker_thread_sequential_execution():
+    """Tasks execute sequentially (FIFO) on the single worker thread."""
+    from vllm_mlx.mlx_streams import MLXWorkerThread
+
+    worker = MLXWorkerThread(name="test-seq")
+    loop = asyncio.get_event_loop()
+
+    results = []
+
+    def append_value(val):
+        results.append(val)
+        return val
+
+    futs = [worker.submit(loop, append_value, i) for i in range(10)]
+    await asyncio.gather(*futs)
+
+    assert results == list(range(10)), "Tasks must execute in submission order"
+    worker.shutdown()
+
+
+@pytest.mark.anyio
+@pytest.mark.skipif(
+    not _mlx_available(),
+    reason="MLX not available",
+)
+async def test_mlx_ops_on_worker_thread_no_stream_error():
+    """MLX array operations on worker thread do not raise stream errors."""
+    from vllm_mlx.mlx_streams import MLXWorkerThread
+
+    worker = MLXWorkerThread(name="test-mlx")
+    loop = asyncio.get_event_loop()
+
+    def mlx_matmul():
+        import mlx.core as mx
+
+        a = mx.ones((32, 32))
+        b = mx.ones((32, 32))
+        c = a @ b
+        mx.eval(c)
+        return c[0, 0].item()
+
+    result = await worker.submit(loop, mlx_matmul)
+    assert result == 32.0
+    worker.shutdown()
--- a/vllm_mlx/engine/simple.py
+++ b/vllm_mlx/engine/simple.py
@@ -22,16 +22,12 @@ from .base import (
     BaseEngine,
     GenerationOutput,
     cleanup_startup_cancellation,
-    run_blocking_startup_work,
 )
-from ..mlx_streams import bind_generation_streams
+from ..mlx_streams import MLXWorkerThread
 
 logger = logging.getLogger(__name__)
 
-
-def _bind_worker_generation_streams() -> None:
-    """Rebind mlx generation streams inside the current worker thread."""
-    bind_generation_streams()
+_mlx_worker = MLXWorkerThread()
 
 
 def _seed_logits_processors(
@@ -224,26 +220,14 @@ class SimpleEngine(BaseEngine):
 
         self._model.load()
 
-    def _uses_default_prepare_for_start(self) -> bool:
-        """Return True when prepare_for_start is the class implementation."""
-        method = getattr(self.prepare_for_start, "__func__", None)
-        return method is SimpleEngine.prepare_for_start
-
     async def start(self) -> None:
         """Start the engine (load model if not loaded)."""
         if self._loaded:
             return
         try:
             if self._model is None:
-                if self._uses_default_prepare_for_start():
-                    # MLX generation streams are thread-local. Keep model load on
-                    # the event-loop thread so default LLM stream_generate() runs
-                    # on the same thread that owns model-associated streams.
-                    self.prepare_for_start()
-                else:
-                    # Test doubles and custom overrides may block; preserve the
-                    # cancellation-safe threaded startup helper for those cases.
-                    await run_blocking_startup_work(self.prepare_for_start)
+                loop = asyncio.get_event_loop()
+                await _mlx_worker.submit(loop, self.prepare_for_start)
             self._loaded = True
 
             if self._mtp and self._mtp_num_draft_tokens != 1:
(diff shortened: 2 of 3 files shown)


janhilgard (changes requested):
Good problem analysis and clean `MLXWorkerThread` implementation. The root cause (MLX 0.31.2 thread-local CommandEncoders) is real, and a persistent worker thread is the right approach.

However, this PR introduces a **critical streaming regression** and has a few additional concerns:

---

### 1. Critical: `stream_generate` and `stream_chat` text-only fallback lost true streaming

The old code yielded chunks **as they were generated** (true token-by-token streaming):

```python
    for chunk in self._model.stream_generate(...):
        yield GenerationOutput(...)  # client sees each token immediately
```

The new code **collects ALL chunks into a list first**, then yields them after generation is complete:

```python
def _run_stream_generate():
    results = []
    for chunk in self._model.stream_generate(...):
        results.append(chunk)
    return results

chunks = await self._run_blocking_serialized(_run_stream_generate)
for chunk in chunks:
    yield GenerationOutput(...)  # client sees everything only after full generation
```

This means **TTFT = total generation time** — for a 500-token response at 100 tok/s, the client waits 5 seconds before seeing any output instead of seeing the first token in ~50ms. This is a major UX regression for any SSE/streaming client.

The fix would be to use an `asyncio.Queue` or similar mechanism to bridge the synchronous generator on the worker thread with the async generator on the event loop, yielding chunks as they're produced.

### 2. Module-level singleton starts thread at import time

```python
_mlx_worker = MLXWorkerThread()
```

This starts a daemon thread the moment `simple.py` is imported, even if `SimpleEngine` is never instantiated (e.g., when using `BatchedEngine` or `MllmBatchGenerator`). Consider lazy initialization or moving the worker into the `SimpleEngine` instance.

### 3. Missing `generation_stream` rebinding on worker thread

The old code called `bind_generation_streams()` before each inference to rebind `mlx_lm.generate.generation_stream` and `mlx_vlm.generate.generation_stream` to the current thread. The new code skips this entirely.

These module-level `generation_stream` variables are created at import time (main thread). If `mlx_lm` internally uses `mx.stream(generation_stream)` to schedule Metal operations, those operations reference a stream whose CommandEncoder lives in the main thread's TLS — which would crash on the worker thread.

This may work in your test because either (a) the default stream is special, or (b) `mlx_lm` doesn't reference `generation_stream` in the code path you tested. But it should be verified more broadly, or a single `bind_generation_streams()` call should be added to the worker thread's init.

### 4. Minor: `asyncio.ensure_future` on a Future is a no-op

```python
task = asyncio.ensure_future(
    _mlx_worker.submit(loop, func, *args, **kwargs)
)
```

`worker.submit()` already returns an `asyncio.Future`. `ensure_future` on a Future just returns it. The variable name `task` is misleading since it's a Future, not a Task. `asyncio.shield()` works on both, so it's functionally correct but confusing.

---

### Tests

The new tests are well-structured and cover the right cases (thread persistence, FIFO ordering, exception propagation, MLX ops). No issues there.

### Summary

The `MLXWorkerThread` itself is solid. The main blocker is the streaming regression — `stream_generate` and `stream_chat` must preserve true token-by-token streaming, not batch-then-yield.

waybarrios:
Closing as superseded. The stream/thread ownership bug was fixed on main by 3396981, which keeps load and generation on the event-loop thread for the default path, and the remaining VLM TextModel gap is covered by #543 with a more targeted approach. Thanks for the root-cause analysis, it helped map the problem.

PR #479, opened 2026-05-01
fix: run BatchedEngine MLLM on dedicated MLXWorkerThread to prevent cross-thread stream errors

xykong:
## Summary

Fixes `RuntimeError: There is no Stream(gpu, N) in current thread` when running BatchedEngine (`--continuous-batching`) with MLX >= 0.31.2 on Apple Silicon.

## Problem

MLX 0.31.2 made Metal CommandEncoders thread-local. When `BatchedEngine._prepare_mllm_model()` loads the model on the event-loop thread but `MLLMScheduler._process_loop()` runs `step()` on a different thread, the runtime raises:

```
RuntimeError: There is no Stream(gpu, N) in current thread
```

This happens because MLX arrays created during model loading are tagged with the loading thread's stream, which doesn't exist in the inference thread's TLS.

## Fix

1. **`mlx_streams.py`**: Add `MLXWorkerThread` — a persistent single-thread executor that guarantees model load + inference share the same thread-local stream context.

2. **`engine/batched.py`**: Make `_prepare_mllm_model()` async and run model loading on `MLXWorkerThread`. Pass the worker to `MLLMScheduler`.

3. **`mllm_scheduler.py`**: Accept `mlx_worker` parameter. When provided, submit `step()` and preprocessing to the worker thread. Falls back to the legacy event-loop path (with `bind_generation_streams()`) when no worker is provided.

4. **`mllm_batch_generator.py`**: Remove `MLLMBatchGenerator._stream` (class-level `mx.new_stream()`) in favor of the worker thread's default stream, eliminating the last source of cross-thread stream references.

## Testing

Tested on M4 Max 128GB with `mlx-community/gemma-4-26b-a4b-it-4bit`:

| Metric | Result |
|--------|--------|
| Single request (short) | 93 tok/s |
| Single request (long) | 82 tok/s |
| 4x concurrent aggregate | **179 tok/s** |
| TTFT (short) | 35ms |

Previously this configuration crashed immediately on the first request.

## Backward Compatibility

- The legacy event-loop path (no `MLXWorkerThread`) is preserved as fallback when `mlx_worker=None`
- SimpleEngine path is unaffected
- LLM-only BatchedEngine path is unaffected

### DIFF
--- a/vllm_mlx/engine/batched.py
+++ b/vllm_mlx/engine/batched.py
@@ -27,6 +27,7 @@ from .base import (
     cleanup_startup_cancellation,
     run_blocking_startup_work,
 )
+from ..mlx_streams import MLXWorkerThread
 
 logger = logging.getLogger(__name__)
 
@@ -216,6 +217,7 @@ class BatchedEngine(BaseEngine):
         self._tokenizer = None  # For LLM
         self._engine = None  # AsyncEngineCore for LLM
         self._mllm_scheduler = None  # MLLMScheduler for MLLM
+        self._mlx_worker = None  # Shared MLXWorkerThread for MLLM load + inference
         self._mllm_instance = None  # MLXMultimodalLM instance
         self._loaded = False
 
@@ -282,17 +284,43 @@ class BatchedEngine(BaseEngine):
         method = getattr(self.prepare_for_start, "__func__", None)
         return method is BatchedEngine.prepare_for_start
 
-    def _prepare_mllm_model(self) -> None:
-        """Load the MLLM model before scheduler startup."""
+    async def _prepare_mllm_model(self) -> None:
+        """Load the MLLM model on a dedicated MLXWorkerThread.
+
+        MLX >= 0.31.2 makes Metal command encoders thread-local.  Model
+        loading creates arrays and streams that are bound to the thread
+        that executes them.  If the model is loaded on the event-loop
+        thread but inference runs on a different thread, ``mx.eval()``
+        raises ``RuntimeError: There is no Stream(gpu, N) in current
+        thread``.
+
+        By loading the model on the same ``MLXWorkerThread`` that will
+        later run ``step()`` calls, we guarantee all MLX state lives on
+        a single thread.
+        """
+        import mlx.core as mx
+
         from ..models.mllm import MLXMultimodalLM
 
+        self._mlx_worker = MLXWorkerThread(name="mllm-worker")
+        loop = asyncio.get_event_loop()
+
         max_kv_size = getattr(self._scheduler_config, "max_kv_size", 0)
-        self._mllm_instance = MLXMultimodalLM(
-            self._model_name,
-            trust_remote_code=self._trust_remote_code,
-            max_kv_size=max_kv_size,
-        )
-        self._mllm_instance.load()
+        model_name = self._model_name
+        trust_remote_code = self._trust_remote_code
+
+        def _load_on_worker():
+            mx.new_thread_local_stream(mx.gpu)
+            inst = MLXMultimodalLM(
+                model_name,
+                trust_remote_code=trust_remote_code,
+                max_kv_size=max_kv_size,
+            )
+            inst.load()
+            logger.info("MLLM model loaded on MLXWorkerThread")
+            return inst
+
+        self._mllm_instance = await self._mlx_worker.submit(loop, _load_on_worker)
         self._model = self._mllm_instance.model
         self._processor = self._mllm_instance.processor
 
@@ -329,7 +357,7 @@ class BatchedEngine(BaseEngine):
         from ..mllm_scheduler import MLLMScheduler, MLLMSchedulerConfig
 
         if self._model is None or self._processor is None:
-            self._prepare_mllm_model()
+            await self._prepare_mllm_model()
 
         # Create MLLM scheduler config with batch generator support
         if self._scheduler_config and hasattr(self._scheduler_config, "max_num_seqs"):
@@ -403,6 +431,7 @@ class BatchedEngine(BaseEngine):
             model=self._model,
             processor=self._processor,
             config=mllm_config,
+            mlx_worker=self._mlx_worker,
         )
         await self._mllm_scheduler.start()
 
--- a/vllm_mlx/engine/batched.py
+++ b/vllm_mlx/engine/batched.py
@@ -239,12 +239,18 @@ class BatchedEngine(BaseEngine):
         return self._tokenizer
 
     def prepare_for_start(self) -> None:
-        """Load heavyweight model state off the serving event loop."""
+        """Load heavyweight model state off the serving event loop.
+
+        For MLLM models this is a no-op: the async _prepare_mllm_model()
+        requires the MLXWorkerThread and event loop, so model loading is
+        deferred to _start_mllm() which is properly awaited.
+        """
         if self._model is not None:
             return
 
         if self._is_mllm:
-            self._prepare_mllm_model()
+            # MLLM loading is async and handled by _start_mllm().
+            return
         else:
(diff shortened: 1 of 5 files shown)


janhilgard (changes requested):
The overall approach is solid — model load + inference on the same persistent `MLXWorkerThread` is the right fix for MLX 0.31.2 thread-local CommandEncoders. The legacy fallback path is a nice touch for backward compatibility. No streaming regression here (unlike #478 for SimpleEngine).

A few issues to address:

---

### 1. `if True:` replaces `with mx.stream(...)` — should dedent instead

```python
if True:  # use thread-default stream
    logits = self.language_model(last_token, cache=request_cache)
```

This appears 4 times. The `if True:` block is dead code that exists only to preserve indentation. Just remove the `with mx.stream(...)` wrapper and dedent the body. A `# noqa` or a comment is fine, but a no-op `if True:` is a code smell that will confuse future readers.

### 2. Preprocessing unnecessarily serialized on the MLX worker thread

The old code ran text-only preprocessing in the default thread-pool executor (`loop.run_in_executor(None, ...)`):

```python
await loop.run_in_executor(None, bg._preprocess_request, req)
```

The new worker-thread path submits preprocessing to the *same* single MLX worker:

```python
await self._mlx_worker.submit(loop, bg._preprocess_request, req)
```

The comment says "Preprocessing creates MLX arrays (input_ids) so it must run on the same worker thread" — but `_preprocess_request` does Jinja2 template rendering + `tokenizer.encode()`, which produces Python lists, not `mx.array`. The `mx.array()` conversion happens inside `_process_prompts()` called from `step()`. 

Running preprocessing on the single MLX worker thread serializes it with `step()`, which means a 30-second preprocessing of a 40K+ token conversation blocks ALL inference for 30 seconds. The old approach of offloading to the thread pool kept inference unblocked during preprocessing. Consider keeping `run_in_executor` for preprocessing.

### 3. Lost slow-preprocessing logging and dynamic `n_yields`

The old code logged slow preprocessing:
```python
elapsed = time.perf_counter() - tic
if elapsed > 1.0:
    logger.info(f"Preprocessing {req.request_id[:12]}: {n_tok} tokens in {elapsed:.2f}s")
```

And had dynamic yield counts:
```python
n_yields = 10 if elapsed > 1.0 else 5
```

The worker-thread path drops both — no timing, always 5 yields. Please preserve the slow-preprocessing logging (it's very useful for debugging production issues) and the dynamic yield count.

### 4. `mx.new_thread_local_stream(mx.gpu)` should live in `MLXWorkerThread._run()`, not in `_load_on_worker`

Currently the thread-local stream is initialized inside `_load_on_worker()`:
```python
def _load_on_worker():
    mx.new_thread_local_stream(mx.gpu)
    inst = MLXMultimodalLM(...)
```

But the `MLXWorkerThread` docstring promises "mx.new_stream() is called exactly once (during thread init)." If someone submits a task before model loading, there's no stream. Move `mx.new_thread_local_stream(mx.gpu)` into `MLXWorkerThread._run()` at the top, before the task loop.

### 5. Overlap with PR #478 on `mlx_streams.py`

Both #478 and #479 add `MLXWorkerThread` to `mlx_streams.py`. If both land, there will be a merge conflict. Consider coordinating — either land #478 first and rebase #479 on top, or extract `MLXWorkerThread` into a shared PR that both depend on.

### 6. Comments stripped from legacy path

Several explanatory comments were removed from the legacy path (early preprocessing phase rationale, health-check yield explanation). These comments document *why* the code works the way it does and are valuable for maintainability. Please keep them.

---

### Summary

The `MLXWorkerThread` + BatchedEngine integration is architecturally sound. Main blockers: (1) preprocessing should stay on the thread pool to avoid blocking inference, (2) `if True:` should be proper dedent, (3) preserve the slow-step diagnostics logging. The rest are minor cleanups.

CarloArpini:
Thanks for the great work on this PR which builds on top of #478; the MLLM path fix is solid and the two-phase preprocessing split is a nice touch.

I wanted to flag that the **LLM-only** `BatchedEngine` path (`--continuous-batching` with a text-only model, i.e. `mllm=False`) is still crashing with the same `RuntimeError: There is no Stream(gpu, N) in current thread` on MLX 0.31.2. The crash occurs in `scheduler.step()` when it calls `self.batch_generator.next()` on a thread that doesn't own the GPU stream created during `BatchGenerator` initialization.

Full traceback:
```
File "vllm_mlx/scheduler.py", line 2455, in step
  File "mlx_lm/generate.py", line 1855, in next
    return self._next()
  File "mlx_lm/generate.py", line 1841, in _next
    self._prompt_batch.prompt(prompts)
  File "mlx_lm/generate.py", line 1161, in prompt
    mx.eval([c.state for c in self.prompt_cache])
```

Server startup confirms LLM-only path: `BatchedEngine loaded: <model> (mllm=False)`.

TBH I don't have enough expertise to confirm this, but maybe the fix can follow the same pattern as this PR: the `SchedulerCore` (or wherever `step()` is dispatched) needs to submit `batch_generator.next()` to the same persistent `MLXWorkerThread` on which the `BatchGenerator` was created, mirroring what you've done here for `MLLMScheduler._process_loop()`.

Environment: Apple M3 Ultra 512GB, macOS, MLX 0.31.2, mlx-lm 0.31.3, DeepSeek-R1-Distill-Llama-8B (text-only). Reproduced consistently across the original `waybarrios/main`, `adriangalilea` fork (PR #478), and this branch.

waybarrios:
Hi @xykong, thanks for chasing the cross-thread stream bug. The `MLXWorkerThread` approach looks solid and the test results on Gemma look great. CI is red on three small issues; happy to push a follow-up commit to your branch if you'd prefer, otherwise here's what to fix.

The merge conflict is in `vllm_mlx/mlx_streams.py` and it's documentation only. `main` extended the `bind_generation_streams` docstring with a paragraph about admission/ownership semantics. Your PR added a `.. deprecated::` block. Both texts are complementary, so the cleanest resolution after `git fetch origin main && git merge origin/main` is to keep both:

```python
    MLX streams are thread-local. If a model is loaded on one thread and
    generation runs on another, module-level generation streams created during
    import can point at a stream that does not exist in the worker thread.

    This intentionally creates a fresh stream for the current worker call and
    replaces module-level generation_stream handles under a process-local lock.
    It is an admission/ownership fix, not a batching optimization; callers
    should invoke it at worker-entry boundaries rather than inside token loops.

    .. deprecated::
        Prefer ``MLXWorkerThread`` which keeps model load and inference on the
        same persistent thread, avoiding the need to rebind streams entirely.
    """
```

Then the lint failure on `vllm_mlx/mllm_batch_generator.py:787`. The `from mlx_vlm.utils import prepare_inputs` is unused after your preprocessing refactor in `a157942` (the function now calls `tokenizer(...)` directly). Just delete that import; `ruff check --fix` will do it for you.

Finally, the failing test in `tests/test_mllm_continuous_batching.py:1259`. The mock `FakeMLLMScheduler.__init__` doesn't accept the new `mlx_worker` kwarg that `BatchedEngine._start_mllm` now passes. One line:

```python
        class FakeMLLMScheduler:
            def __init__(self, model, processor, config, mlx_worker=None):
                captured["scheduler_config"] = config
                captured["mlx_worker"] = mlx_worker
```

Capturing `mlx_worker` matches the existing capture style for `config_kwargs` and leaves the door open for a future assertion that the worker is wired correctly. This same change fixes the 3.11, 3.13, and `tests` aggregator jobs.

Let me know which path you want and I'll re-run CI when it lands.

waybarrios:
Closing as superseded, same story as #478. Main's design keeps MLLM step() on the event-loop thread (see mllm_scheduler) and prepare_for_start runs inline since 3396981, so the dedicated worker thread approach moves in the opposite direction from what landed.

PR #481, opened 2026-05-02
fix: harden SSD cache review follow-ups

### LINKED ISSUE #480: Track unresolved SSD cache tiering review concerns from PR #309
PR #309 merged with several concrete review concerns from @janhilgard that were not converted into tracking issues. That was a process miss on my side; this issue is to make the remaining work explicit and auditable rather than leaving it buried in a merged PR thread.

Source comment/review context:

- PR: https://github.com/waybarrios/vllm-mlx/pull/309
- Review comment thread follow-up: https://github.com/waybarrios/vllm-mlx/pull/309#issuecomment-4362805030
- Merge commit: `50e1dedb9a1a27042206237e5d8d1d854063d7fe`

Current read against `main`: several concerns still appear present. This issue should track audit/fix disposition for each one:

1. `SSDIndex.lookup_prefix()` appears to scan all entries with `num_tokens <= ?` and compare token blobs in Python. This can become O(N) per prefix lookup as the SSD tier grows. We should either add an indexed prefilter/signature or document/limit the intended scale.

2. Scheduler SSD promotion code reaches into `MemoryAwarePrefixCache` private fields such as `_current_memory` and `_max_memory`. This should move behind a cache-owned reserve/release API.

3. `_reconstruct_ssd_layers()` creates an ad-hoc dynamic `ArraysCacheLayer` object for `ArraysCache` state. We need to verify this is compatible with downstream cache users or reconstruct the real expected type.

4. RAM budget accounting in SSD promotion reads/writes cache memory fields without going through a cache-owned lock/API. This needs a race/accounting review.

5. `_ssd_tier` initialization was partially improved, but there are still defensive `hasattr(self, "_ssd_tier")` checks. We should make the scheduler invariant explicit and use direct `None` checks.

6. `promote_from_ssd()` reserves budget via `async_promote()`, then releases before `store()` re-accounts. The temporary accounting window and failure paths around reconstruction need review.

7. `SSDCacheTier.__init__()` opens `SSDIndex`; if later init work fails, the index should be closed. Add a close-on-partial-init-failure path if needed.

Proposed disposition:

- Keep SSD tiering treated as opt-in/experimental until these are resolved or explicitly accepted with documented rationale.
- Prefer focused fix PRs if the feature remains viable.
- If the audit shows unsafe exposure that cannot be fixed cleanly, consider reverting PR #309 with a normal revert PR rather than leaving hidden debt.

This is not meant as a criticism of the feature direction. The architecture is useful, but the review debt should have been tracked when the PR merged.

Thump604:
## Summary

This follows up issue #480 / PR #309 review debt with a narrow source patch. It keeps the SSD tier opt-in as before, but hardens the implementation details called out in review.

Changes:
- add a bounded prefix-hash filter to SSD prefix lookup, including legacy index backfill
- initialize the scheduler SSD tier invariant directly instead of relying on hasattr checks
- move SSD promotion RAM reservation behind MemoryAwarePrefixCache APIs
- reconstruct real ArraysCache instances instead of ad-hoc dynamic cache objects
- close the SQLite index if SSDCacheTier initialization fails after opening it
- add regression coverage for the above paths

## Why

PR #309 merged before the review concerns were converted into explicit follow-up work. This PR addresses the source-level concerns without changing the public opt-in surface for SSD tiering.

## Validation

- `PYTHONPATH=/opt/ai-runtime/worktrees/vllm-mlx/issue-480-ssd-cache-review-debt AI_RUNTIME_BYPASS_SAFETY_GATE=1 /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_ssd_cache.py tests/test_memory_cache.py`
- `python -m black --check --target-version py312 vllm_mlx/ssd_cache.py vllm_mlx/memory_cache.py vllm_mlx/scheduler.py tests/test_ssd_cache.py tests/test_memory_cache.py`
- `git diff --check`

Closes #480

### DIFF
--- a/tests/test_memory_cache.py
+++ b/tests/test_memory_cache.py
@@ -1,6 +1,8 @@
 # SPDX-License-Identifier: Apache-2.0
 """Tests for memory-aware prefix cache."""
 
+import threading
+import time
 from unittest.mock import MagicMock, patch
 
 import pytest
@@ -395,6 +397,43 @@ class TestMemoryAwarePrefixCache:
     def test_try_reserve_memory_denies_over_limit(self, small_cache):
         assert small_cache.try_reserve_memory(2 * 1024 * 1024) is False
         assert small_cache.get_stats()["current_memory_mb"] == 0
+
+    def test_memory_mutations_wait_for_memory_lock(self, small_cache, mock_kv_cache):
+        def assert_waits_for_lock(operation):
+            result = {}
+            errors = []
+            small_cache._memory_lock.acquire()
+
+            def run_operation():
+                try:
+                    result["value"] = operation()
+                except Exception as exc:  # pragma: no cover - surfaced below
+                    errors.append(exc)
+
+            thread = threading.Thread(target=run_operation)
+            thread.start()
+            try:
+                time.sleep(0.05)
+                assert thread.is_alive()
+            finally:
+                small_cache._memory_lock.release()
+            thread.join(timeout=1)
+
+            assert not thread.is_alive()
+            assert errors == []
+            return result.get("value")
+
+        assert assert_waits_for_lock(
+            lambda: small_cache.store([10], mock_kv_cache(1000))
+        )
+
+        small_cache.store([20], mock_kv_cache(1000))
+        assert assert_waits_for_lock(lambda: small_cache.remove([20]))
+
+        small_cache.store([30], mock_kv_cache(1000))
+        assert_waits_for_lock(small_cache.clear)
+        assert len(small_cache) == 0
+        assert small_cache.memory_usage_mb == 0
 
     def test_reset_stats(self, small_cache, mock_kv_cache):
         small_cache.store([1, 2, 3], mock_kv_cache(1000))
--- a/tests/test_memory_cache.py
+++ b/tests/test_memory_cache.py
@@ -385,6 +385,17 @@ class TestMemoryAwarePrefixCache:
         assert stats["misses"] == 1
         assert stats["entry_count"] == 1
 
+    def test_try_reserve_and_release_memory(self, small_cache):
+        assert small_cache.try_reserve_memory(512 * 1024) is True
+        assert small_cache.get_stats()["current_memory_mb"] == 0.5
+
+        small_cache.release_reserved_memory(128 * 1024)
+        assert small_cache.get_stats()["current_memory_mb"] == 0.38
+
+    def test_try_reserve_memory_denies_over_limit(self, small_cache):
+        assert small_cache.try_reserve_memory(2 * 1024 * 1024) is False
+        assert small_cache.get_stats()["current_memory_mb"] == 0
+
     def test_reset_stats(self, small_cache, mock_kv_cache):
         small_cache.store([1, 2, 3], mock_kv_cache(1000))
         small_cache.fetch([1, 2, 3])
--- a/tests/test_ssd_cache.py
+++ b/tests/test_ssd_cache.py
@@ -88,6 +88,7 @@ class TestSSDCacheStats:
 
 import os
 
+from vllm_mlx import ssd_cache as ssd_cache_module
 from vllm_mlx.ssd_cache import SSDIndex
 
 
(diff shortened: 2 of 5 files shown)


janhilgard (approved):
Solid review-debt cleanup. Each change is a clear improvement over the original PR #309 code.

**Prefix-hash filter** — the most substantive change. Correctness analysis:

- Stored entry of length L: `prefix_hash = hash(stored[:min(L, 16)])`
- Query generates up to 16 candidate hashes: `{hash(query[:1]), hash(query[:2]), ..., hash(query[:16])}`
- For a stored entry that IS a prefix of the query: `stored[:min(L,16)] == query[:min(L,16)]`, so its hash is in the candidate set — no false negatives
- False positives (same first 16 tokens, different suffix) are caught by the full blob comparison that follows
- The compound index `(prefix_hash, num_tokens)` makes the SQL query efficient
- `_PREFIX_FILTER_TOKENS = 16` is a good balance — small enough to keep the IN clause manageable, large enough to prune most non-matches

**Legacy backfill** — handles schema migration cleanly. `_ensure_column` + `_backfill_prefix_hashes` run once at startup, backfilling NULL prefix_hash values from the stored token blob. One-time cost, no ongoing overhead.

**`try_reserve_memory` / `release_reserved_memory`** — proper encapsulation with `_memory_lock`. The old code directly accessed `_current_memory` / `_max_memory` from the scheduler, which broke encapsulation and wasn't thread-safe. One observation (non-blocking): other `MemoryAwarePrefixCache` methods (`store`, `evict`) still modify `_current_memory` without `_memory_lock` — the lock only covers the new reservation API. Not introduced by this PR, but worth noting for a future pass if the async `promote_from_ssd` path becomes more concurrent.

**Real `ArraysCache` instances** — the old `type("ArraysCacheLayer", (), {"state": ...})()` hack would fail any `isinstance` check. Using the real `ArraysCache` class is correct.

**Init failure cleanup** — `_closed = True` initially, `False` only after full success, with `except` block closing the index. The `_writer_thread = None` initialization ensures `close()` is safe even on partial init.

**`hasattr` → direct `None` check** — `self._ssd_tier = None` in `__init__` makes all the `hasattr(self, "_ssd_tier")` guards redundant. Clean.

Tests cover all the new paths well — prefix hash candidate filtering, legacy backfill, init failure cleanup, reservation API, and real ArraysCache reconstruction.

waybarrios:
A couple of follow-ups on top of what @janhilgard  already covered in the approval.

The new try/except in `SSDCacheTier.__init__` will swallow the original exception if `index.close()` itself raises. Since this cleanup path runs precisely when initialization is failing (DB corrupted, lock state bad, file system flaky), `close()` is the call most likely to throw — and when it does, the bare `raise` at the end is never reached because the active exception has been replaced.

```python
# vllm_mlx/ssd_cache.py:606-610
except Exception:
    index = getattr(self, "_index", None)
    if index is not None:
        index.close()   # if this raises, we re-raise the close error and lose the root cause
    raise
```

https://github.com/waybarrios/vllm-mlx/blob/9c8026f5f0488f0731167f5e2a2b87c80866a375/vllm_mlx/ssd_cache.py#L591-L611

A small wrap fixes it without changing semantics:

```python
except Exception:
    index = getattr(self, "_index", None)
    if index is not None:
        try:
            index.close()
        except Exception:
            logger.exception("ssd_cache: failed to close index during init cleanup")
    raise
```

The other thing worth flagging is concrete file:line cites for the partial-locking gap janhilgard mentioned. The new `_memory_lock` only guards the reservation API:

```python
# vllm_mlx/memory_cache.py:1102-1115
def try_reserve_memory(self, nbytes: int) -> bool:
    with self._memory_lock:
        if self._current_memory + nbytes > self._max_memory:
            return False
        self._current_memory += nbytes
        self._stats.current_memory_bytes = self._current_memory
        return True

def release_reserved_memory(self, nbytes: int) -> None:
    with self._memory_lock:
        self._current_memory = max(0, self._current_memory - nbytes)
        self._stats.current_memory_bytes = self._current_memory
```

https://github.com/waybarrios/vllm-mlx/blob/9c8026f5f0488f0731167f5e2a2b87c80866a375/vllm_mlx/memory_cache.py#L1100-L1117

Every existing mutation site bypasses it: `store` (985, 1006), `_evict_lru` (1036), `remove` (1065), `clear` (1076), `load_from_disk` (1352). Concrete failure modes:

- `try_reserve_memory` checks `_current_memory + nbytes <= _max_memory`, decides there's room, but a concurrent `store()` (unlocked) bumps `_current_memory` before the increment lands. Promotion proceeds and total memory exceeds `_max_memory`.
- `release_reserved_memory` clamps with `max(0, _current_memory - nbytes)`. If `_evict_lru` runs unlocked and decrements first, the release silently saturates at zero. The counter drifts below true usage and triggers false memory pressure on the next reservation.
- The class docstring still reads "NOT thread-safe; use external locking if needed". A reader seeing `_memory_lock` on two methods will assume the rest of the class is safe.

https://github.com/waybarrios/vllm-mlx/blob/9c8026f5f0488f0731167f5e2a2b87c80866a375/vllm_mlx/memory_cache.py#L984-L1010

waybarrios:
@Thump604 could you take a look on this?

Thump604:
@waybarrios I checked both follow-ups you flagged and pushed a narrow patch.

Changes:
- `SSDCacheTier.__init__` now preserves the original setup exception if cleanup `index.close()` also fails. The cleanup failure is logged, but the root setup failure is re-raised.
- `MemoryAwarePrefixCache` now uses the same memory lock for the reservation API and the memory-accounting mutation paths: `store`, `_evict_lru`, `remove`, `clear`, `reset_stats`, and `load_from_disk`.
- I kept this scoped to memory accounting consistency. This does not claim the entire cache class is globally thread-safe beyond the documented cache behavior.

Validation:
- Added regression coverage for cleanup failure preservation.
- Added regression coverage proving memory mutation paths wait on `_memory_lock`.
- `AI_RUNTIME_BYPASS_SAFETY_GATE=1 /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_ssd_cache.py tests/test_memory_cache.py -q`
- Black/Ruff checks on touched files
- `git diff --check`

waybarrios:
just reviewed it. all set!

PR #485, opened 2026-05-02
fix: bound streaming tool marker precheck

### LINKED ISSUE #484: Audit streaming tool parser marker detection from PRs #304/#305
## Context

PRs #304 and #305 changed streaming tool-call detection and auto-parser behavior. Both were merged in the 2026-04-18 batch after comment-only reviews that called out concrete complexity/performance concerns.

## Review concerns to resolve

- #304: streaming marker detection scans accumulated text (`tool_accumulated_text + delta`) and may have performance cost on long streams with tools.
- #304/#305: overlapping marker-detection changes were merged close together and should be audited as one combined parser path.
- #305: bare-bracket streaming detection uses compound marker / partial-pattern logic that the review called fragile and hard to reason about.

## Current spot check

Current `upstream/main` still has accumulated-text marker scanning via `_streaming_tool_markup_possible()` in several streaming paths, and `auto_tool_parser.py` still has compound bare-bracket marker logic.

## Work to do

- Add focused streaming parser tests for ordinary text containing brackets, markdown links, JSON arrays, partial bracket text, and actual bare-bracket tool calls.
- Add a long-stream/perf regression test or bounded-complexity assertion for marker scanning.
- Simplify marker detection if tests show false positives or unbounded accumulated scanning cost.
- Document the intended streaming parser contract for fallback parser activation.

## Scope note

This is parser/performance review debt from #304/#305. It is not a broad tool-parser failure claim.

Thump604:
## Summary

- Bound pre-marker streaming tool markup checks to a small boundary window instead of rescanning the full accumulated stream on ordinary text chunks.
- Keep full parser behavior once tool markup is possible.
- Add false-positive coverage for ordinary bracket text, markdown links, JSON arrays, and actual bare-bracket tool-call streaming.

Closes #484.

## Validation

- `python -m black --check vllm_mlx/server.py tests/test_tool_parsers.py tests/test_server.py`
- `python -m compileall vllm_mlx/server.py vllm_mlx/tool_parsers/auto_tool_parser.py tests/test_tool_parsers.py tests/test_server.py`
- `AI_RUNTIME_BYPASS_SAFETY_GATE=1 /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_tool_parsers.py tests/test_server.py::TestStreamChatCompletion::test_streaming_tool_markup_precheck_uses_bounded_boundary tests/test_server.py::TestStreamChatCompletion::test_auto_parser_streams_bare_bracket_tool_calls -q` — 104 passed

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -1906,6 +1906,22 @@ class TestStreamChatCompletion:
         assert delta.get("content") is None
         assert tool_payloads[0]["choices"][0]["finish_reason"] == "tool_calls"
 
+    def test_streaming_tool_markup_precheck_uses_bounded_boundary(self):
+        """Pre-marker checks should not need the full accumulated stream."""
+        import vllm_mlx.server as server
+
+        long_prefix = "ordinary text " * 1000
+
+        assert not server._streaming_tool_markup_possible_after_delta(
+            long_prefix, "with [ordinary brackets] and no tool call"
+        )
+        assert server._streaming_tool_markup_possible_after_delta(
+            long_prefix, '[read({"file_path": "/tmp/test.py"})]'
+        )
+        assert server._streaming_tool_markup_possible_after_delta(
+            long_prefix + "[read(", '{"file_path": "/tmp/test.py"}'
+        )
+
     @pytest.mark.anyio
     async def test_reasoning_stream_emits_structured_tool_calls(self, monkeypatch):
         """Tool markup after </think> should emit tool_calls chunks."""
--- a/tests/test_tool_parsers.py
+++ b/tests/test_tool_parsers.py
@@ -597,6 +597,23 @@ class TestAutoToolParser:
 
         assert not result.tools_called
 
+    @pytest.mark.parametrize(
+        "text",
+        [
+            "Read [the docs](https://example.test) before changing this.",
+            "The output array was [1, 2, 3], not a tool call.",
+            'Here is JSON: [{"label": "alpha", "score": 0.7}]',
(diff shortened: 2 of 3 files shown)


janhilgard (approved):
Good performance fix. The old code was O(n^2) — `_streaming_tool_markup_possible(tool_accumulated_text + delta_text)` rescanned the entire accumulated response on every chunk. For a 10K-token response, each successive chunk scans more text than the last.

The fix bounds the scan to a 512-char trailing window + delta, making it O(1) per chunk. 512 chars is conservative — all tool markup patterns (`<tool_call>`, `<function=`, `[TOOL_CALLS]`, `\[\w+\(\{`) are < 50 chars, so even a 64-char window would suffice for correctness. The generous buffer handles any realistic chunk boundary split.

**Correctness preserved**: the bounded function only gates the *pre-marker* check. Once markup is detected, the full accumulated text still goes to the parser. The early `if not delta_text: return False` correctly handles empty deltas.

**All 5 callsites updated consistently** — OpenAI streaming, Anthropic streaming (x2), and stream_chat_completion (x2).

**Tests are thorough**:
- `test_streaming_tool_markup_precheck_uses_bounded_boundary` covers: long prefix + ordinary brackets (False), long prefix + real tool call (True), and partial marker split across boundary (True)
- Parametrized false-positive tests for both non-streaming and streaming parsers: markdown links, JSON arrays, square brackets in prose, tool-like text without JSON args — all correctly pass through as content

waybarrios:
I reviewed it without any issues, and everything looks good to me. You should be all set to move forward with the merge.

PR #486, opened 2026-05-02
fix: honor reranker hidden_act config

### LINKED ISSUE #483: Follow up PR #308 reranker forward/config support contract
## Context

PR #308 added `/v1/rerank` support and was merged in the 2026-04-18 batch. The review was comment-only and called out a concrete maintenance/config concern before merge: the custom BERT forward pass hard-codes GELU and ignores `hidden_act` from config.

## Current spot check

Current `upstream/main` still appears to use `_gelu()` directly in `vllm_mlx/rerank_forward.py` and does not appear to honor `hidden_act` or explicitly reject unsupported activations.

## Work to do

- Define the supported reranker architecture/config contract.
- Either implement supported `hidden_act` handling or fail explicitly for unsupported activations.
- Add tests for supported and unsupported activation configs.
- Document supported reranker families and the custom forward-pass limitation.

## Scope note

This is review debt from #308. It is not a runtime qualification claim and not a revert request by itself.

Thump604:
## Summary

- Apply the reranker encoder `hidden_act` from config instead of silently hard-coding GELU.
- Support common BERT-family activations: `gelu`, `gelu_new`/`gelu_fast`, `relu`, and `silu`/`swish`.
- Fail explicitly for unsupported activations so custom reranker architectures add a dedicated adapter.
- Document the supported reranker forward-path activation contract.

Closes #483.

## Validation

- `python -m black --check vllm_mlx/rerank_forward.py tests/test_rerank.py`
- `python -m compileall vllm_mlx/rerank_forward.py tests/test_rerank.py`
- `AI_RUNTIME_BYPASS_SAFETY_GATE=1 /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_rerank.py -q` — 39 passed

### DIFF
--- a/README.es.md
+++ b/README.es.md
@@ -164,6 +164,12 @@ curl http://localhost:8000/v1/rerank -H 'Content-Type: application/json' -d '{
 }'
 ```
 
+La ruta integrada de reranking MLX admite pesos de clasificación de secuencia
+BERT/XLM-RoBERTa estándar con valores `hidden_act` `gelu`,
+`gelu_new`/`gelu_fast`, `relu` o `silu`/`swish`. Otras activaciones fallan de
+forma explícita para que las arquitecturas personalizadas agreguen un adaptador
+dedicado en lugar de usar GELU silenciosamente.
+
 ### Embeddings
 
 ```bash
--- a/README.fr.md
+++ b/README.fr.md
@@ -164,6 +164,12 @@ curl http://localhost:8000/v1/rerank -H 'Content-Type: application/json' -d '{
 }'
 ```
 
+Le chemin de reranking MLX intégré prend en charge les poids de classification
+de séquence BERT/XLM-RoBERTa standard avec les valeurs `hidden_act` `gelu`,
+`gelu_new`/`gelu_fast`, `relu` ou `silu`/`swish`. Les autres activations échouent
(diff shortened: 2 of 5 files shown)


janhilgard (approved):
Clean, well-scoped fix. The old code silently hard-coded GELU regardless of what the model config specified — this reads `hidden_act` from config and fails explicitly for unsupported activations.

**Activation dispatch** — covers the standard BERT-family set: `gelu` (exact), `gelu_new`/`gelu_fast` (tanh approximation), `relu`, `silu`/`swish`. The `isinstance(hidden_act, dict)` guard handles configs that nest activation in `{"type": "gelu"}` format. Good.

**`_gelu_new` implementation** — the tanh-approximation constants (0.7978845608028654, 0.044715) match the HuggingFace `gelu_new` exactly. Correct.

**Fail-explicit** — `ValueError` with a clear message for unsupported activations is the right choice. Silent GELU fallback would produce wrong scores with no indication of the problem.

**Tests** — parametrized test covers all 6 supported activation names, plus a dedicated test for unsupported (`mish`) that verifies the ValueError. Good coverage.

One minor observation (non-blocking): `_relu` could use `nn.relu(x)` for consistency with `_gelu` using `nn.gelu(x)`, but `mx.maximum(x, 0)` is equivalent and fine.

waybarrios:
No issues!

PR #490, opened 2026-05-02
fix: sanitize remote media safety errors

### LINKED ISSUE #488: Follow up PR #325 SSRF review notes
## Context

PR #325 added remote media SSRF protection and was self-merged after comment-only review. The review was positive on the core mitigation but called out follow-up concerns that should be tracked explicitly.

## Review notes to resolve

- DNS TOCTOU/rebinding limitation: `_validate_url_safety()` resolves the host, but `requests` performs its own connection-time resolution. Full mitigation may require a pinned-IP/custom adapter or documented limitation.
- Error detail leakage: `UnsafeRemoteURLError` can expose blocked resolved IPs in exception text. Public HTTP responses should avoid leaking internal topology while logs may retain sanitized diagnostic detail.

## Work to do

- Decide whether to implement pinned-IP fetching or document DNS-rebinding as a known limitation.
- Ensure client-visible errors do not expose resolved internal/private IP details.
- Add tests for sanitized public error behavior.

## Scope note

This is follow-up review debt, not a claim that PR #325 failed to improve SSRF safety.

Thump604:
Closes #488.

This is a narrow follow-up for the SSRF review notes on remote media handling. It does not attempt to implement pinned-IP transport; instead it makes the current behavior explicit and removes the client-visible leakage from URL-safety failures.

Changes:
- add a public-safe `UnsafeRemoteURLError.public_message`
- validate remote media URLs during chat/Responses request preparation so unsafe streaming requests fail before emitting SSE data
- return a generic 400 detail for blocked media URLs while logging the internal diagnostic server-side
- document the remaining DNS rebinding / split-horizon DNS limitation and recommend egress controls or a trusted proxy for environments where that is in scope
- add regression coverage for safe public errors and endpoint-level sanitization

Validation:
- `AI_RUNTIME_BYPASS_SAFETY_GATE=1 /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_mllm.py tests/test_server.py -q -k 'unsafe_remote_url_error_has_safe_public_message or validate_url_safety or request_with_safe_redirects or blocks_unsafe_url_before_request or chat_completion_sanitizes_remote_media_safety_errors'`
- `AI_RUNTIME_BYPASS_SAFETY_GATE=1 /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_mllm.py -q`
- `AI_RUNTIME_BYPASS_SAFETY_GATE=1 /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_server.py -q`
- `AI_RUNTIME_BYPASS_SAFETY_GATE=1 /opt/ai-runtime/venv-live/bin/python -m compileall -q vllm_mlx/models/mllm.py vllm_mlx/server.py`
- `black --target-version py312 --check vllm_mlx/models/mllm.py vllm_mlx/server.py tests/test_mllm.py tests/test_server.py`
- `uvx ruff check vllm_mlx/models/mllm.py vllm_mlx/server.py tests/test_mllm.py tests/test_server.py --select E,F,W --ignore E402,E501,E731,F811,F841`
- `git diff --check`

### DIFF
--- a/docs/guides/multimodal.md
+++ b/docs/guides/multimodal.md
@@ -162,6 +162,18 @@ curl http://localhost:8000/v1/chat/completions \
 | Local file | `{"type": "video", "video": "/path/to/video.mp4"}` |
 | Base64 | `{"type": "video_url", "video_url": {"url": "data:video/mp4;base64,..."}}` |
 
+### Remote URL safety
+
+Remote image, video, and audio URLs are checked before each fetch and redirect
+hop. URLs that resolve to localhost, link-local, private, or otherwise
+non-global addresses are rejected with a generic client error while detailed
+diagnostics stay in server logs.
+
+This validation does not pin the IP address used by the later HTTP transport
+connection. In environments where DNS rebinding or split-horizon DNS is in
+scope, run vllm-mlx behind network egress controls or fetch media through a
+trusted proxy that enforces the destination policy at connect time.
+
 ## Python API
 
 ```python
--- a/tests/test_mllm.py
+++ b/tests/test_mllm.py
@@ -170,6 +170,17 @@ class TestMLLMHelperFunctions:
 
         _validate_url_safety("https://8.8.8.8/image.jpg")
 
+    def test_unsafe_remote_url_error_has_safe_public_message(self):
+        """Public safety errors should not disclose resolved hosts or IPs."""
+        from vllm_mlx.models.mllm import UnsafeRemoteURLError, _validate_url_safety
+
+        with pytest.raises(UnsafeRemoteURLError) as exc_info:
+            _validate_url_safety("http://169.254.169.254/latest/meta-data/")
+
+        assert "169.254.169.254" in str(exc_info.value)
+        assert exc_info.value.public_message == "Remote media URL is not allowed"
+        assert "169.254.169.254" not in exc_info.value.public_message
+
     def test_request_with_safe_redirects_blocks_unsafe_redirect(self, monkeypatch):
         """Test that redirect hops are validated before a second request."""
         from vllm_mlx.models import mllm
--- a/tests/test_responses_api.py
+++ b/tests/test_responses_api.py
@@ -152,6 +152,50 @@ class TestResponsesEndpoint:
             "enable_thinking": False
         }
 
+    def test_streaming_responses_validates_remote_media_once(self, client, monkeypatch):
+        import vllm_mlx.server as srv
+
+        engine = _mock_engine()
+        engine._stream_outputs = [_stream_output("Hello there", finish_reason="stop")]
+        srv._engine = engine
+
+        validate_calls = []
+
+        def fake_validate(messages):
+            validate_calls.append(messages)
+
+        def fake_responses_to_chat(_request):
+            return srv.ChatCompletionRequest(
+                model="test-model",
+                messages=[
+                    srv.Message(
+                        role="user",
+                        content=[
+                            {"type": "text", "text": "describe this"},
+                            {
+                                "type": "image_url",
+                                "image_url": {"url": "https://example.com/image.png"},
+                            },
+                        ],
+                    )
+                ],
+                max_tokens=8,
+                stream=True,
+            )
+
+        monkeypatch.setattr(srv, "_validate_remote_media_urls", fake_validate)
+        monkeypatch.setattr(
+            srv, "_responses_request_to_chat_request", fake_responses_to_chat
+        )
+
+        resp = client.post(
+            "/v1/responses",
+            json={"model": "test-model", "input": "describe this", "stream": True},
+        )
+
+        assert resp.status_code == 200
+        assert len(validate_calls) == 1
+
     def test_responses_request_kwargs_override_server_defaults(self, client):
         import vllm_mlx.server as srv
 
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -2619,6 +2619,72 @@ class TestReasoningAndToolCallsNonStreaming:
         assert response.json()["detail"] == "Remote media URL is not allowed"
         assert "169.254.169.254" not in response.text
 
+    def test_anthropic_message_sanitizes_remote_media_safety_errors(
+        self, client, monkeypatch
+    ):
+        """Anthropic preparation should sanitize URL-safety failures too."""
+        import vllm_mlx.server as server
+
+        class FakeEngine:
+            model_name = "fake-mllm"
+            is_mllm = True
+            preserve_native_tool_format = False
+
+            async def chat(self, messages, **kwargs):  # pragma: no cover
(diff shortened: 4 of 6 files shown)


janhilgard (approved):
Good SSRF hardening. The key insight — validation must happen before any SSE data is emitted — is correctly implemented.

**`UnsafeRemoteURLError.public_message`** — clean separation between internal diagnostics (IP, hostname, reason) and client-facing message. The test explicitly verifies `169.254.169.254` appears in the internal error but NOT in `public_message` — exactly the right security boundary assertion.

**Early validation placement:**
- `_prepare_chat_messages()` → catches both streaming and non-streaming chat completions before any response bytes are sent
- `create_response()` streaming path → `_validate_remote_media_urls(chat_request.messages)` runs before `StreamingResponse()` constructor
- `_prepare_responses_request()` → catches non-streaming Responses path

All three entry points validate before the point of no return (SSE start or response body).

**`_raise_remote_media_http_error`** — logs full details server-side via `logger.warning`, returns only the generic `public_message` via `HTTPException(status_code=400)`. No information leakage to the client.

**`_iter_remote_media_urls`** — handles all multimodal content types (`image_url`, `video_url`, `audio_url`, `image`, `video`, `audio`) plus Pydantic model objects. The `is_url()` filter correctly skips base64 data URIs and local paths that don't need SSRF validation.

**Documentation** — honest about the DNS rebinding / split-horizon limitation. Recommending egress controls or a trusted proxy is the right advice rather than claiming complete protection.

**Endpoint test** — the AWS metadata endpoint (`169.254.169.254`) is the classic SSRF target, good choice. The assertion `"169.254.169.254" not in response.text` verifies no internal detail leaks in the full response body.

waybarrios:
The Anthropic `/v1/messages` endpoint is unprotected. The patch wraps `_prepare_chat_completion_invocation` and `_run_responses_request` with `except UnsafeRemoteURLError`, but `create_anthropic_message` calls `_prepare_anthropic_invocation` outside any try block, and that path goes through `_prepare_chat_messages` → `_validate_remote_media_urls`. `UnsafeRemoteURLError` is a `ValueError` subclass with no app-level handler, so a failure surfaces as a 500 with a stack trace that contains the resolved IP.

```python
# vllm_mlx/server.py:4900-4906 (current PR head)
release_on_exit = True
prepared = _prepare_anthropic_invocation(    # raises UnsafeRemoteURLError
    engine,
    openai_request,
    effective_max_tokens,
)
```

https://github.com/waybarrios/vllm-mlx/blob/d486a8d80691815f49e2d86c5437ca67d97e00eb/vllm_mlx/server.py#L4895-L4912

The fix is the same pattern already used twice in this PR:

```python
try:
    prepared = _prepare_anthropic_invocation(
        engine,
        openai_request,
        effective_max_tokens,
    )
except UnsafeRemoteURLError as exc:
    _raise_remote_media_http_error(exc)
```

Today this is partly masked because `anthropic_to_openai._convert_message` silently drops `image` blocks (only `text` / `tool_use` / `tool_result` are converted), so multimodal Anthropic requests never reach the validator in the first place. As soon as that converter is completed for image support, the gap becomes a live IP-leak path. Worth closing now while the patch is in flight.

The second thing worth flagging: the `/v1/responses` streaming path runs the validator twice. Once eagerly here:

```python
# vllm_mlx/server.py:4692-4699
try:
    if request.stream:
        chat_request = _responses_request_to_chat_request(request)
        _validate_remote_media_urls(chat_request.messages)
        return StreamingResponse(
            _disconnect_guard(_stream_responses_request(request), raw_request),
            media_type="text/event-stream",
        )
```

https://github.com/waybarrios/vllm-mlx/blob/d486a8d80691815f49e2d86c5437ca67d97e00eb/vllm_mlx/server.py#L4690-L4708

And once again, lazily, inside the generator via `_prepare_responses_request`:

```python
# vllm_mlx/server.py:2034-2046
_validate_model_name(request.model)
engine = get_engine()
chat_request = _responses_request_to_chat_request(request)
...
_validate_remote_media_urls(chat_request.messages)
```

https://github.com/waybarrios/vllm-mlx/blob/d486a8d80691815f49e2d86c5437ca67d97e00eb/vllm_mlx/server.py#L2030-L2048

The second call runs after `StreamingResponse(...)` has already returned, so any `UnsafeRemoteURLError` raised there is outside the outer `except` and either corrupts the SSE stream or surfaces as a generic 500. Concrete consequences:

- Latency: every streaming `/v1/responses` request with media pays a duplicate DNS-resolving safety check.
- Failure mode is mode-dependent: if the eager call ever gets removed or moved, the late call silently no-ops the security guarantee instead of failing closed.
- The eager call already does the right thing in isolation, so the inner one is dead-as-a-guard.

Pick one site. Either drop the call inside `_prepare_responses_request` for the streaming path, or move all validation inside `_prepare_responses_request` and remove the eager copy — but document why if both stay.

There's also a third item worth noting but not blocking: the Responses API path runs `_validate_remote_media_urls` against messages that already passed through `_response_content_to_text` (`server.py:1649-1666`), which strips everything except `text`/`input_text`/`output_text` into plain text. Any `image_url` carried in `ResponsesRequest.input` is dropped before the validator sees it, so the validator on this path never has anything to validate. Either Responses is text-only by design (in which case the `_validate_remote_media_urls` call there is dead code) or it should preserve media blocks before validation.

https://github.com/waybarrios/vllm-mlx/blob/d486a8d80691815f49e2d86c5437ca67d97e00eb/vllm_mlx/server.py#L1647-L1670

Thump604:
@waybarrios I addressed the follow-ups you flagged and pushed `dda431be993758b234a0ebf4dd37c6f3a4b94e71`.

Changes:
- Anthropic `/v1/messages` now converts `UnsafeRemoteURLError` from request preparation into the same sanitized 400 response as the other endpoints.
- Streaming `/v1/responses` now validates remote media before constructing `StreamingResponse` and skips the duplicate late validation inside the SSE generator path.
- Non-streaming `/v1/responses` still validates through `_prepare_responses_request`.
- I left the current Responses API media-shape note as-is rather than broadening this patch, since the current Responses model is text/tool focused and media preservation there is a separate design surface.

Validation:
- `AI_RUNTIME_BYPASS_SAFETY_GATE=1 PYTHONPATH=/opt/ai-runtime/worktrees/vllm-mlx/issue-488-ssrf-review-notes /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_responses_api.py tests/test_server.py::TestReasoningAndToolCallsNonStreaming tests/test_mllm.py -q` → `59 passed, 6 deselected`
- `ruff check vllm_mlx/server.py tests/test_server.py tests/test_responses_api.py`
- `black --check --target-version py312 vllm_mlx/server.py tests/test_server.py tests/test_responses_api.py`
- `git diff --check`

CI is green on the pushed commit as well.

waybarrios:
Ok this looks better now. You got my approval!

Issue #493, opened 2026-05-02
Qwen 3.6: prevent dangling thinking from trapping tool calls in chat-template history


Thump604:
## Summary

Qwen 3.6 can produce assistant history where a thinking block is opened and raw tool XML appears before the closing `</think>` marker:

```text
<think>
reasoning...
<tool_call>...</tool_call>
```

If that history is passed back through a Jinja chat template unchanged, the next prompt can condition the model as though the tool call is still inside reasoning. In practice this shows up as tool calls being ignored, reasoning leaking across tool turns, or dirty history being preserved when `preserve_thinking=true`.

## Credit / source

This issue is based on Cheuk-Yiu Chan's public write-up and template fix:

- https://allanchan339.github.io/bug-fixes/2026/05/02/Qwen36-27B-updated-jinja.html
- https://github.com/allanchan339/vLLM-Qwen3-3.5-3.6-chat-template-fix

Chan's template-side repair closes a dangling thinking span before `<tool_call>` so the tool payload is rendered outside reasoning.

## Proposed fix

Before calling `apply_chat_template`, normalize assistant message content so that when both `<think>` and `<tool_call>` are present and the latest `<think>` has no matching `</think>`, insert `</think>` immediately before the first `<tool_call>` that follows the dangling opener.

This keeps the guard independent of any one bundled template and protects user-supplied Qwen 3.6 templates as well.

## Acceptance criteria

- Assistant history with dangling `<think>` before raw `<tool_call>` renders with tool XML outside the thinking span.
- The normalization does not mutate the caller's original message list.
- Existing OpenAI `tool_calls[].function.arguments` normalization remains intact.
- Covered by a focused unit test; no model load required.

PR #494, opened 2026-05-02
Fix dangling think before tool calls in templates

Thump604:
## Summary

Refs #493.

This adds a small pre-template message normalization guard for assistant history where a thinking block is left open before raw tool XML:

```text
<think>
reasoning...
<tool_call>...</tool_call>
```

Before `apply_chat_template`, the normalizer now inserts `</think>` before the first `<tool_call>` that follows a dangling `<think>`. That keeps tool XML outside the reasoning span for both bundled and user-supplied templates.

Credit to Cheuk-Yiu Chan for identifying and documenting the Qwen 3.6 template failure mode and template-side repair:

- https://allanchan339.github.io/bug-fixes/2026/05/02/Qwen36-27B-updated-jinja.html
- https://github.com/allanchan339/vLLM-Qwen3-3.5-3.6-chat-template-fix

## Why

Qwen 3.6 can sustain interleaved thinking across turns. If malformed assistant history leaves `<think>` open before `<tool_call>`, the next rendered prompt can condition the model as though the tool call is still reasoning. That can lead to ignored tool calls, reasoning leakage into tool turns, or polluted preserved history.

This patch handles the structural transcript hygiene issue before template rendering, rather than relying on a specific template fork.

## Changes

- Add `vllm_mlx.engine.chat_template_safety.normalize_messages_for_chat_template`.
- Preserve the existing OpenAI `tool_calls[].function.arguments` mapping normalization.
- Apply the normalization in BatchedEngine and SimpleEngine before `apply_chat_template`.
- Add a unit test covering dangling `<think>` before raw `<tool_call>` without mutating caller messages.

## Validation

- `python -m pytest tests/test_batched_engine.py -q` -> 11 passed
- `AI_RUNTIME_BYPASS_SAFETY_GATE=1 /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_batched_engine.py tests/test_chat_template_kwargs.py -q` -> 20 passed

### DIFF
--- a/tests/test_batched_engine.py
+++ b/tests/test_batched_engine.py
@@ -235,3 +235,39 @@ class TestToolCallReplayNormalization:
         assert normalized[0]["tool_calls"][0]["function"]["arguments"] == {
             "value": ["not", "object"]
         }
+
+    def test_closes_dangling_think_before_raw_tool_call(self):
+        messages = [
+            {
+                "role": "assistant",
+                "content": (
+                    "<think>Need the weather tool.\n"
+                    "<tool_call>\n"
+                    "<function=get_weather>\n"
+                    "<parameter=city>\nParis\n</parameter>\n"
+                    "</function>\n"
+                    "</tool_call>"
+                ),
+            }
+        ]
+
+        normalized = _normalize_tool_call_arguments_for_template(messages)
+        content = normalized[0]["content"]
+
+        assert "Need the weather tool.\n</think><tool_call>" in content
+        assert "<tool_call>" not in content.split("</think>", 1)[0]
+        assert messages[0]["content"].startswith("<think>Need")
+
+    def test_normalizes_pydantic_style_messages_without_stringifying(self):
+        class MessageLike:
+            def model_dump(self, exclude_none=False):
+                assert exclude_none is True
+                return {
+                    "role": "assistant",
+                    "content": "plain response",
+                    "unused": None,
+                }
+
+        normalized = _normalize_tool_call_arguments_for_template([MessageLike()])
+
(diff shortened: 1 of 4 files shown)


janhilgard (approved):
Good fix for a real Qwen 3.6 failure mode. Pre-template normalization is the right approach — it handles the structural hygiene issue without forking the template.

**`_close_dangling_think_before_tool_call`** — logic is correct:
- Fast-path early return when no `<think>` or `<tool_call>` present (majority of messages)
- `rfind("<think>")` vs `rfind("</think>")` correctly detects the *last* unclosed think block
- Handles multiple think sections (e.g., `<think>a</think> text <think>b<tool_call>...`) correctly — the second unclosed think gets closed
- Only operates on `role: "assistant"` messages, so user content with these tags is untouched

**`_message_to_dict`** — nice improvement over the pure `json.dumps(default=str)` path. `model_dump(exclude_none=True)` preserves Pydantic objects properly. The `v is not None` comprehension is redundant with `exclude_none=True` but harmless as a defensive guard.

**SimpleEngine integration** — this PR also fixes a pre-existing gap: SimpleEngine's `chat()` and `_stream_generate_text()` previously didn't normalize tool-call arguments at all. Now both engines share the same normalization path.

**Module structure** — extracting to `chat_template_safety.py` is clean. The old `_normalize_tool_call_arguments_for_template` in `batched.py` is now a trivial passthrough — could be inlined later, but that's a minor cleanup.

Tests cover the key scenarios: dangling think insertion, original message immutability, and Pydantic-style message handling.

waybarrios (approved):
I carefully checked. It was good refactor!

PR #504, opened 2026-05-06
Expose MLLM MTP draft counters

Thump604:
## Summary
- track whether MLLM MTP actually attempted drafts on each primary response
- carry attempted and accepted draft counters through MLLMScheduler outputs
- expose those counters on BatchedEngine GenerationOutput for MLLM routes

## Why
The MLLM MTP wrapper could previously make it hard to distinguish an installed MTP path from a step that actually drafted. That can make skipped speculative steps look like zero-acceptance MTP attempts. This PR records attempts only after the draft path actually runs, and counts accepted draft responses separately.

## Scope note
The counters are currently internal scheduler/engine metadata. Client-facing HTTP exposure is intentionally left to a follow-up surface decision, rather than adding a response-shape change in this PR.

## Tests
- `AI_RUNTIME_BYPASS_SAFETY_GATE=1 PYTHONPATH=/opt/ai-runtime/worktrees/vllm-mlx/mllm-mtp-metadata-accounting /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_mllm_mtp_routing.py -q`
- `/opt/ai-runtime/venv-live/bin/python -m py_compile vllm_mlx/mllm_batch_generator.py vllm_mlx/engine/batched.py tests/test_mllm_mtp_routing.py`
- `/opt/ai-runtime/venv/bin/ruff check vllm_mlx/mllm_batch_generator.py vllm_mlx/mllm_scheduler.py vllm_mlx/engine/batched.py vllm_mlx/engine/base.py vllm_mlx/request.py tests/test_mllm_mtp_routing.py`
- `/opt/ai-runtime/venv/bin/black --check --target-version py312 vllm_mlx/mllm_batch_generator.py vllm_mlx/mllm_scheduler.py vllm_mlx/engine/batched.py vllm_mlx/engine/base.py vllm_mlx/request.py tests/test_mllm_mtp_routing.py`
- `git diff --check`

### DIFF
--- a/tests/test_mllm_mtp_routing.py
+++ b/tests/test_mllm_mtp_routing.py
@@ -195,7 +195,7 @@ def test_mllm_mtp_attempt_metadata_only_marks_real_attempts():
         _mark_mtp_attempts_on_primary_responses,
     )
 
-    attempted = {7: 1}
+    attempted = {7: 1, 99: 1}
     attempted_response = MLLMBatchResponse(
         uid=7,
         request_id="req-7",
@@ -218,6 +218,7 @@ def test_mllm_mtp_attempt_metadata_only_marks_real_attempts():
     assert attempted_response.mtp_attempted_count == 1
     assert skipped_response.mtp_attempted is False
     assert skipped_response.mtp_attempted_count == 0
+    assert attempted == {}
 
 
 def test_mllm_scheduler_exposes_mtp_attempts_and_accepts_on_outputs():
--- a/tests/test_mllm_mtp_routing.py
+++ b/tests/test_mllm_mtp_routing.py
@@ -189,6 +189,81 @@ def test_has_media_content_none_content():
 from unittest.mock import MagicMock
 
 
+def test_mllm_mtp_attempt_metadata_only_marks_real_attempts():
+    from vllm_mlx.mllm_batch_generator import (
+        MLLMBatchResponse,
+        _mark_mtp_attempts_on_primary_responses,
+    )
+
+    attempted = {7: 1}
+    attempted_response = MLLMBatchResponse(
+        uid=7,
+        request_id="req-7",
+        token=11,
+        logprobs=None,
+    )
+    skipped_response = MLLMBatchResponse(
+        uid=9,
+        request_id="req-9",
+        token=21,
+        logprobs=None,
+    )
+
+    _mark_mtp_attempts_on_primary_responses(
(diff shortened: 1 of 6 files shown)


janhilgard (commented):
## Review: Expose MLLM MTP draft counters

Clean, well-scoped observability addition. The data flow is straightforward: `MLLMBatchResponse` → `MLLMRequest` → `RequestOutput` → `GenerationOutput`.

### Positive

- Good separation of concerns: attempt tracking at generation level (`_mark_mtp_attempts_on_primary_responses`), acceptance counting at scheduler level (`_process_batch_responses`)
- `from_draft` on responses is a clean way to count accepted tokens without coupling to the MTP internals
- Tests cover both the marking function and the full scheduler flow
- CI 9/9 green, no merge conflicts

### Nits

1. **Internal-only counters**: `mtp_drafts`/`mtp_accepted` propagate to `GenerationOutput` but aren't surfaced to HTTP responses (not in `RequestOutput.usage`, not in `server.py` response construction). This is fine as a building block, but worth noting in the PR description that a follow-up is needed to expose them to clients (custom response header, `/v1/status`, or `usage` extension).

2. **Intermediate variable before return/yield** in `batched.py`:
   ```python
   generation_output = GenerationOutput(...)
   return generation_output
   ```
   vs the simpler original `return GenerationOutput(...)`. No functional impact, just noise.

3. **`_attempted_drafts_by_uid` stale entries**: If a UID is aborted between `_mtp_step()` recording the attempt and `_mark_mtp_attempts_on_primary_responses()` consuming it, the entry persists until `active_batch is None`. Not a real issue since UIDs are unique per request lifetime, but `_mark_mtp_attempts_on_primary_responses` could defensively clear the dict after processing instead of relying on per-entry `pop()`.

4. **`mtp_attempted_count` is always 1**: Currently hardcoded as `_attempted_drafts_by_uid[uid] = 1`. If MTP only ever drafts 1 token per step, a `bool` would be more honest. The current design is forward-compatible for multi-draft, which is fine, but worth a comment noting the intent.

Overall: looks good to merge. Minor polish only.

janhilgard (approved):
Follow-up commit addresses the stale-entry concern (defensive `.clear()` after processing). Comment added for the single-draft-count intent.

CI 9/9. LGTM.

waybarrios (approved):
All concerns were approved!

Issue #495, opened 2026-05-02
Track admission-control invariant for serialized TextModel-direct routes


Thump604:
## Summary

I hit a local Runtime P0 in a TextModel-direct route that is not currently present on `upstream/main`, but it is important to capture before any similar route is proposed or revived upstream.

The unsafe shape was: text-only MLLM requests bypassed the MLLM scheduler and entered a serialized TextModel-direct generation path guarded by one async lock. Concurrent non-streaming requests were allowed to wait behind that lock for up to 120s. Under agent traffic this created request pileup and memory pressure instead of returning a fast retryable admission result.

This is not a model-behavior issue. It is a serving admission/scheduler boundary issue.

## Why this matters upstream

Current `upstream/main` appears to route MLLM text through `MLLMScheduler`, so I do not believe current main has this exact bug. The risk is future upstreaming of a TextModel-direct optimization, or revival of earlier TextModel-direct work, without carrying the admission-control guard.

I authored/ran this route locally, so I want the invariant tracked publicly before it can surprise downstream users.

## Observed local failure mode

- route: text-only MLLM / TextModel-direct
- request style: concurrent non-streaming JSON/structured traffic
- bad behavior: second request waited behind serialized generation lock instead of failing fast
- prior wait bound: 120s
- observed impact: request pileup and abnormal memory pressure
- attribution: runtime route/admission failure, not model failure

## Local containment

The local Runtime overlay now defaults this serialized route to fail-fast admission:

- second concurrent request receives retryable `503 text_generation_busy`
- observed contention canary result: 503 in ~0.002s instead of waiting
- `/v1/status` exposes TextModel-direct lock/admission telemetry
- regression coverage verifies fail-fast behavior

## Proposed upstream invariant

Any serialized TextModel-direct route should be accepted upstream only with:

1. Fail-fast admission by default, returning a retryable 503 when the route is already busy.
2. An explicit opt-in wait mode, if wait mode is needed at all.
3. Status telemetry showing lock state, admission mode, lock timeout, active request count, waiting count, and busy rejection count.
4. A two-request contention regression test.
5. Clear reporting that this route is serialized and must not be used as continuous-batching/concurrency evidence.

## Acceptance criteria for a future fix PR

- A concurrent request behind an occupied serialized TextModel-direct route fails fast rather than waiting for minutes.
- The error is machine-readable, e.g. `text_generation_busy`, with HTTP 503.
- Status exposes enough data to diagnose route contention.
[...]

PR #507, opened 2026-05-06
Wire MLLM assistant drafters for Gemma 4 MTP

Thump604:
## Summary
- wire direct SimpleEngine/MLXMultimodalLM support for activation-sharing MLLM assistant drafters
- add CLI/server flags for `--mllm-draft-model`, `--mllm-draft-kind mtp`, and `--mllm-draft-block-size`
- keep text-only MLLM drafter requests on the mlx-vlm MLLM path by disabling TextModel routing when an MLLM drafter is configured
- expose request-local MLLM drafter counters through `MLLMOutput` / `GenerationOutput`
- reject currently unsupported combinations with continuous batching and lifecycle residency

## Runtime motivation
This PR is driven by Runtime Gemma 4 MTP bring-up. The first Runtime target is E2B: `google/gemma-4-E2B-it` plus the `gemma4_assistant` drafter checkpoint. The assistant loader requires an mlx-vlm build that provides `mlx_vlm.speculative.drafters.gemma4_assistant`.

## Runtime qualification evidence
Runtime E2B non-live server proof now passes for the narrow path this PR enables: SimpleEngine, forced MLLM, no-thinking chat-template request, explicit MLLM assistant drafter, `draft_kind=mtp`, `draft_block_size=3`.

Matched non-streaming probe:
- target-only output and target+assistant output matched exactly
- output SHA256: `1e2b24b02560f124666235e1503e1eeffa14d9849891979edc83db9689398bdb`
- target-only metadata: `routing=textmodel`, `mtp_drafts=0`, `mtp_accepted=0`, latency `0.447822334s`, peak Metal `10.31GB`
- target+assistant metadata: `routing=textmodel_mtp`, `mtp_drafts=28`, `mtp_accepted=6`, `mtp_acceptance_rate=0.2143`, latency `0.698765959s`, peak Metal `10.47GB`

Streaming `bench-serve` probe:
- repetitions: `3`
- quality: `passed=true`, `failure_count=0`, `quality_passed=true`
- latency mean: `545.1278886757791ms`
- TTFT mean: `542.30075001639ms`
- all repetitions reported `routing=textmodel_mtp` and `mtp_acceptance_rate=0.2143`

This is narrow non-live Runtime evidence, not a production/live Gemma lane qualification.

## Scope boundary
This is intentionally narrow upstream wiring. It does not add registry/residency support, live routing, continuous batching support, Jobs routing, Open WebUI routing, or broad Runtime model qualification.

## Validation
- `.venv-test/bin/python -m pytest tests/test_mllm_assistant_drafter.py tests/test_lifecycle_cli.py tests/test_mllm_mtp_routing.py tests/test_simple_engine.py tests/test_mllm.py -q` (`81 passed, 6 deselected`)
- `.venv-test/bin/python -m py_compile vllm_mlx/models/mllm.py vllm_mlx/engine/base.py vllm_mlx/engine/simple.py tests/test_mllm_assistant_drafter.py`
- `.venv-test/bin/python -m black --check vllm_mlx/models/mllm.py vllm_mlx/engine/base.py vllm_mlx/engine/simple.py tests/test_mllm_assistant_drafter.py`
- `git diff --check`

### DIFF
--- a/tests/test_lifecycle_cli.py
+++ b/tests/test_lifecycle_cli.py
@@ -185,6 +185,88 @@ class TestLifecycleCli:
             "mllm_draft_block_size": 4,
         }
 
+    def test_main_rejects_nonpositive_mllm_draft_block_size(self, monkeypatch, capsys):
+        """Drafter block size must be positive at parse time."""
+        import vllm_mlx.cli as cli
+
+        monkeypatch.setattr(
+            cli.sys,
+            "argv",
+            [
+                "vllm-mlx",
+                "serve",
+                "google/gemma-4-E2B-it",
+                "--mllm",
+                "--mllm-draft-model",
+                "assistant",
+                "--mllm-draft-kind",
+                "mtp",
+                "--mllm-draft-block-size",
+                "0",
+            ],
+        )
+
+        with pytest.raises(SystemExit):
+            cli.main()
+
+        assert "--mllm-draft-block-size" in capsys.readouterr().err
+
+    def test_serve_command_rejects_mllm_draft_without_mllm(self, capsys):
+        """Drafter flags should not be silently ignored on text-only models."""
+        import vllm_mlx.cli as cli
+
+        args = SimpleNamespace(
+            model="mlx-community/Qwen3-0.6B-8bit",
+            models_config=None,
+            served_model_name=None,
+            enable_auto_tool_choice=False,
+            tool_call_parser=None,
+            gpu_memory_utilization=0.90,
+            max_tokens=32768,
+            max_request_tokens=32768,
+            mllm_draft_model="assistant",
+            mllm_draft_kind="mtp",
+            mllm_draft_block_size=4,
+            continuous_batching=False,
+            auto_unload_idle_seconds=0.0,
+            lazy_load_model=False,
+            mllm=False,
+        )
+
+        with pytest.raises(SystemExit):
+            cli.serve_command(args)
+
+        assert "--mllm-draft-model requires --mllm" in capsys.readouterr().out
+
+    def test_serve_command_rejects_mllm_draft_with_models_config(self, capsys):
+        """Registry mode must not accept drafter flags that it does not load."""
+        import vllm_mlx.cli as cli
+
+        args = SimpleNamespace(
+            model=None,
+            models_config="/tmp/models.yaml",
+            served_model_name=None,
+            enable_auto_tool_choice=False,
+            tool_call_parser=None,
+            gpu_memory_utilization=0.90,
+            max_tokens=32768,
+            max_request_tokens=32768,
+            mllm_draft_model="assistant",
+            mllm_draft_kind="mtp",
+            mllm_draft_block_size=4,
+            continuous_batching=False,
+            auto_unload_idle_seconds=0.0,
+            lazy_load_model=False,
+            mllm=True,
+        )
+
+        with pytest.raises(SystemExit):
+            cli.serve_command(args)
+
+        assert "--mllm-draft-model cannot be used with --models-config" in (
+            capsys.readouterr().out
+        )
+
     def test_serve_command_wires_auto_unload_idle_seconds_into_load_model(
         self, monkeypatch
     ):
--- a/tests/test_lifecycle_cli.py
+++ b/tests/test_lifecycle_cli.py
@@ -148,6 +148,43 @@ class TestLifecycleCli:
 
         assert captured["lazy_load_model"] is False
 
+    def test_main_parses_mllm_draft_flags(self, monkeypatch):
+        """The top-level CLI should accept MLLM assistant drafter knobs."""
+        import vllm_mlx.cli as cli
+
+        captured = {}
+
+        def fake_serve_command(args):
+            captured["mllm_draft_model"] = args.mllm_draft_model
+            captured["mllm_draft_kind"] = args.mllm_draft_kind
+            captured["mllm_draft_block_size"] = args.mllm_draft_block_size
+
+        monkeypatch.setattr(cli, "serve_command", fake_serve_command)
+        monkeypatch.setattr(
+            cli.sys,
+            "argv",
+            [
+                "vllm-mlx",
+                "serve",
+                "google/gemma-4-E2B-it",
+                "--mllm",
+                "--mllm-draft-model",
+                "google/gemma-4-E2B-it-assistant",
+                "--mllm-draft-kind",
+                "mtp",
+                "--mllm-draft-block-size",
+                "4",
+            ],
+        )
+
+        cli.main()
+
+        assert captured == {
+            "mllm_draft_model": "google/gemma-4-E2B-it-assistant",
+            "mllm_draft_kind": "mtp",
+            "mllm_draft_block_size": 4,
+        }
+
     def test_serve_command_wires_auto_unload_idle_seconds_into_load_model(
         self, monkeypatch
     ):
@@ -156,6 +193,7 @@ class TestLifecycleCli:
 
         import vllm_mlx.cli as cli
         import vllm_mlx.server as srv
+        import vllm_mlx.utils.download as download_mod
 
         captured = {}
 
@@ -165,6 +203,11 @@ class TestLifecycleCli:
 
         monkeypatch.setattr(srv, "load_model", fake_load_model)
         monkeypatch.setattr(uvicorn, "run", lambda *args, **kwargs: None)
+        monkeypatch.setattr(
+            download_mod,
+            "ensure_model_downloaded",
+            lambda *args, **kwargs: "/tmp/model",
+        )
 
         args = SimpleNamespace(
             model="mlx-community/Qwen3-0.6B-8bit",
--- a/tests/test_lifecycle_server.py
+++ b/tests/test_lifecycle_server.py
@@ -320,7 +320,12 @@ class TestCompletionStreamingRelease:
                 raise RuntimeError("generation failed")
 
         async def fake_acquire(
-            raw_request, *, total_timeout=None, deadline=None, count_activity=True
+            raw_request,
+            *,
+            total_timeout=None,
+            deadline=None,
+            count_activity=True,
+            model=None,
         ):
             acquires["count"] += 1
             return FakeEngine()
@@ -393,7 +398,12 @@ class TestCompletionStreamingRelease:
                 )
 
         async def fake_acquire(
-            raw_request, *, total_timeout=None, deadline=None, count_activity=True
+            raw_request,
+            *,
+            total_timeout=None,
+            deadline=None,
+            count_activity=True,
+            model=None,
         ):
             return FakeEngine()
 
@@ -546,7 +556,12 @@ class TestToolParserUsesLocalEngine:
         local_engine = FakeEngine("tok-local")
 
         async def fake_acquire(
-            raw_request, *, total_timeout=None, deadline=None, count_activity=True
+            raw_request,
+            *,
+            total_timeout=None,
+            deadline=None,
+            count_activity=True,
+            model=None,
         ):
             return local_engine
 
@@ -619,7 +634,12 @@ class TestLifecycleFailureHandling:
             preserve_native_tool_format = False
 
         async def fake_acquire(
-            raw_request, *, total_timeout=None, deadline=None, count_activity=True
+            raw_request,
+            *,
+            total_timeout=None,
+            deadline=None,
+            count_activity=True,
+            model=None,
         ):
             calls["acquires"] += 1
             return FakeEngine()
@@ -648,7 +668,12 @@ class TestLifecycleFailureHandling:
             preserve_native_tool_format = False
 
         async def fake_acquire(
-            raw_request, *, total_timeout=None, deadline=None, count_activity=True
+            raw_request,
+            *,
+            total_timeout=None,
+            deadline=None,
+            count_activity=True,
+            model=None,
         ):
             calls["acquires"] += 1
             return FakeEngine()
@@ -3497,7 +3522,12 @@ class TestResponseModelFieldUsesServedName:
         served_name = "my-custom-served-name"
 
         async def fake_acquire(
-            raw_request, *, total_timeout=None, deadline=None, count_activity=True
+            raw_request,
+            *,
+            total_timeout=None,
+            deadline=None,
+            count_activity=True,
+            model=None,
         ):
             return FakeEngine()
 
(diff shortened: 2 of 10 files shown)


waybarrios (changes requested):
In `vllm_mlx/engine/simple.py` the new routing gate sits right under a comment that says the opposite of what the gate now does:

```python
# Build parallel mlx_lm TextModel for text-only routing.
# Even when MTP is disabled, text-only requests should not be trapped
# on the slower mlx_vlm multimodal path.
if self._is_mllm and self._should_route_text_through_text_model():
    ...

def _should_route_text_through_text_model(self) -> bool:
    """Return whether text-only MLLM requests may use mlx_lm TextModel."""
    return self._mllm_draft_model_path is None
```

When `--mllm-draft-model` is set, `_text_model` is never built and every text-only request falls back to `async with self._generation_lock: ... mlx_vlm.stream_chat(...)`. That is the exact pile-up shape issue #495 documented and that PR #514 is pinning an invariant against — text-only requests queueing on the generation lock with no fail-fast 503. Please either narrow this gate to per-request opt-in (a request-side flag indicating the caller actually wants drafter assistance) or add a `text_generation_busy`-style admission gate on the drafter lane so concurrent text traffic doesn't pile up silently on the lock.

In `vllm_mlx/models/mllm.py` `_draft_metrics_since` computes `mtp_drafts` as a static multiplier rather than a real counter:

```python
drafted_per_round = max(block_size - 1, 0)
return {
    "mtp_drafts": drafted_per_round * len(new_accept_lens),
    "mtp_accepted": sum(int(value) for value in new_accept_lens),
}
```

This assumes the drafter always emits exactly `block_size - 1` tokens per round. It doesn't — EOS, sampler trims, and budget caps end rounds early — so `mtp_drafts` over-counts on those rounds. Same shape as the `bypass_reasons` static dict we discussed on #473. Please track the actual drafted count from the drafter rather than back-computing it from `block_size`.

In `vllm_mlx/cli.py` the registry path silently drops the three new flags. `serve_command` forwards them through `load_model(...)` for the single-model branch but `RegistryServeDefaults(...)` doesn't include them:

```python
if models_config:
    defaults = RegistryServeDefaults(
        continuous_batching=args.continuous_batching,
        force_mllm=getattr(args, "mllm", False),
        enable_mtp=args.enable_mtp,
        ...
        specprefill_draft_model=args.specprefill_draft_model,
        stream_interval=args.stream_interval if args.continuous_batching else 1,
        ...
    )
    load_model_registry(models_config, defaults=defaults)
else:
    load_model(
        ...
        mllm_draft_model=mllm_draft_model,
        mllm_draft_kind=mllm_draft_kind,
        mllm_draft_block_size=mllm_draft_block_size,
        ...
    )
```

A user combining `--models-config foo.yaml --mllm-draft-model assistant` accepts both flags and silently runs without the drafter. Either thread `mllm_draft_*` through `RegistryServeDefaults` (and `load_model_registry`), or reject the combination at parse time alongside the existing rejections:

```python
if mllm_draft_model and args.continuous_batching:
    print("Error: --mllm-draft-model is supported only without --continuous-batching")
    sys.exit(1)
if mllm_draft_model and (args.auto_unload_idle_seconds > 0 or args.lazy_load_model):
    print("Error: --mllm-draft-model is not supported with lifecycle residency yet")
    sys.exit(1)
```

While you're there, also reject `--mllm-draft-model` without `--mllm`. `SimpleEngine.prepare_for_start` only forwards the drafter args inside the `_is_mllm` branch, so pointing the flag at a text-only model silently ignores it.

The `--mllm-draft-block-size` parser:

```python
serve_parser.add_argument(
    "--mllm-draft-block-size",
    type=int,
    default=None,
    help="Draft block size passed to mlx-vlm for --mllm-draft-model.",
)
```

accepts `0` and negative ints. `_draft_metrics_since` then computes `max(block_size - 1, 0) = 0`, reporting `mtp_drafts=0` forever, which looks like the drafter is broken when the user has just misconfigured. Please reject `<= 0` at parse time.

`MLXMultimodalLM.generate`, `chat`, `stream_chat`, and `stream_generate` all do:

```python
**self._draft_generation_kwargs(),
**kwargs,
```

If a caller's `kwargs` ever contains `draft_model` / `draft_kind` / `draft_block_size` (test helpers, future HTTP-passthrough payloads), Python raises `TypeError: got multiple values for keyword argument` and the call fails before reaching the drafter. Either pop those keys from `kwargs` first or define a deterministic precedence.

A couple of non-blocking notes:

- `mtp_drafts` and `mtp_accepted` are added to `GenerationOutput` / `MLLMOutput` but the diff doesn't surface them anywhere consumers can see — not in `/v1/status.mtp`, not in OpenAI/Anthropic usage. Same dead-Usage-field shape as #436. Either wire them through or hold off on the fields until they're consumed.
- `load_gemma4_assistant_drafter(model_path)` does `Path(model_path) / "config.json"` and `path.glob("*.safetensors")` directly on a user-supplied string with no `ensure_model_downloaded`-style retry and no constrained root. A multi-GB drafter download with one network hiccup fails serve startup. Worth routing through `ensure_model_downloaded` like the main model.

Thump604:
Pushed 47f575a to address the review feedback.

Concrete changes:
- Text-only MLLM requests keep the TextModel route by default; the MLLM assistant-drafter path is request opt-in via `mllm_draft` when a drafter is configured.
- `--mllm-draft-model` now rejects the unsupported server shapes early: no `--mllm`, `--models-config`, continuous batching, lifecycle residency, and non-positive block size.
- Drafter metrics now count actual `draft_block` output lengths when the drafter exposes that path; accepted tokens still come from `accept_lens`.
- Configured drafter kwargs take precedence by stripping caller-provided `draft_model`, `draft_kind`, and `draft_block_size` before calling mlx-vlm.
- Non-stream text-only aggregation preserves MTP counters from the final streamed output.

Validation run locally:
- `AI_RUNTIME_BYPASS_SAFETY_GATE=1 PYTHONPATH=/opt/ai-runtime/worktrees/vllm-mlx/gemma4-assistant-drafter /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_server.py::TestChatCompletionRequest tests/test_mllm_assistant_drafter.py tests/test_lifecycle_cli.py tests/test_simple_engine.py tests/test_mllm.py -q` -> 76 passed, 6 deselected.
- `AI_RUNTIME_BYPASS_SAFETY_GATE=1 PYTHONPATH=/opt/ai-runtime/worktrees/vllm-mlx/gemma4-assistant-drafter /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_mllm_mtp_routing.py -q` -> 15 passed.
- `/opt/ai-runtime/venv-live/bin/python -m py_compile ...` on touched source/tests -> passed.
- `/opt/ai-runtime/venv/bin/ruff check ...` on touched source/tests -> passed.
- `/opt/ai-runtime/venv/bin/black --check --target-version py312 ...` on touched source/tests -> passed.
- `git diff --check` -> passed.

janhilgard (commented):
Good scoping — SimpleEngine-only, no batching, clear unsupported-combo rejections. A few observations:

---

### Semantics: `mllm_draft` request field is misleading

The `mllm_draft: bool` request field doesn't actually control whether the drafter is used. It only controls **text-only routing** — when `True`, text requests go through the MLLM path (where the drafter happens to be). But:

- If the request has images/video, it goes through MLLM regardless → drafter is always used.
- If `mllm_draft` is `False`/`None` and the request is text-only, it goes through the fast text-model path → drafter is bypassed.
- `_draft_generation_kwargs()` always injects the draft model kwargs when `self._draft_model` is set — there's no check on `mllm_draft`.

This means `mllm_draft=True` really means "route text through MLLM so the drafter can participate", not "enable the drafter". The name and API doc should clarify this, or the behavior should match the name (i.e., `_draft_generation_kwargs` should return `{}` when `mllm_draft` is `False`).

### `_draft_generation_kwargs` mutates input

```python
def _draft_generation_kwargs(self, call_kwargs: dict | None = None) -> dict:
    if call_kwargs is not None:
        for key in _DRAFT_KWARG_NAMES:
            call_kwargs.pop(key, None)  # mutates caller's dict
```

This silently strips keys from the caller's kwargs dict as a side effect. The test covers it (`test_mllm_chat_uses_configured_drafter_over_call_kwargs`), but a comment or docstring on this intentional mutation would help.

### Dependency on unreleased mlx-vlm module

`load_gemma4_assistant_drafter` imports `mlx_vlm.speculative.drafters.gemma4_assistant`. The PR body acknowledges this requires a specific mlx-vlm build. Worth adding a version check or at minimum a log line indicating which mlx-vlm version/build is needed, so users don't hit a cryptic ImportError.

### `_install_draft_metrics_hooks` called twice

`_install_draft_metrics_hooks(self._draft_model)` is called both in `load()` (line ~990) and again in `_draft_generation_kwargs()` (line ~889). The guard (`_vllm_mlx_metrics_hooked`) prevents double-patching, but the call in `_draft_generation_kwargs` runs on every generation. This seems like defensive redundancy — if intentional, a comment explaining why would help.

### Minor

- `_count_draft_tokens` handles shape/len/fallback robustly — good defensive coding.
- The validation in both `cli.py` and `server.py load_model()` is good defense-in-depth.
- Acceptance rate 21.4% in the qualification run is notably low vs. Qwen3.5/3.6 MTP (78-85%). Expected for Gemma 4 assistant or room for improvement?

CI 9/9 green, no merge conflicts.

Thump604:
Pushed cb7b8aa to make `mllm_draft` a real request opt-in instead of only a text-routing hint.

Changes:
- Configured MLLM drafters are no longer injected unless the request passes `mllm_draft=True`.
- Text-only requests still stay on the TextModel route by default; `mllm_draft=True` routes them through MLLM and forwards the opt-in to the MLLM wrapper.
- Media requests still route through MLLM, but drafter kwargs are only passed when `mllm_draft=True`.
- The outbound mlx-vlm kwargs scrub now removes `mllm_draft` plus any caller-provided draft kwargs before forwarding.
- Added a log line that records the mlx-vlm version used when loading the Gemma 4 assistant drafter.

Validation:
- `AI_RUNTIME_BYPASS_SAFETY_GATE=1 PYTHONPATH=/opt/ai-runtime/worktrees/vllm-mlx/gemma4-assistant-drafter /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_mllm_assistant_drafter.py tests/test_mllm_mtp_routing.py tests/test_lifecycle_cli.py -q` -> 33 passed.
- `AI_RUNTIME_BYPASS_SAFETY_GATE=1 PYTHONPATH=/opt/ai-runtime/worktrees/vllm-mlx/gemma4-assistant-drafter /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_server.py::TestChatCompletionRequest tests/test_mllm_assistant_drafter.py tests/test_lifecycle_cli.py tests/test_simple_engine.py tests/test_mllm.py tests/test_mllm_mtp_routing.py -q` -> 93 passed, 6 deselected.
- `/opt/ai-runtime/venv-live/bin/python -m py_compile vllm_mlx/api/models.py vllm_mlx/engine/simple.py vllm_mlx/models/mllm.py tests/test_mllm_assistant_drafter.py` -> passed.
- `/opt/ai-runtime/venv/bin/ruff check ...` on touched source/tests -> passed.
- `/opt/ai-runtime/venv/bin/black --check --target-version py312 ...` on touched source/tests -> passed.
- `git diff --check` -> passed.

Thump604:
Pushed cd208c4 to replace the remaining `pytest.mark.asyncio` marker with the repo-standard `pytest.mark.anyio`.\n\nFocused verification on the PR branch:\n- `uv run --extra dev pytest tests/test_mllm_assistant_drafter.py tests/test_async_markers.py tests/test_mllm_mtp_routing.py tests/test_lifecycle_cli.py` -> 34 passed\n- `uv run --extra dev black --check tests/test_mllm_assistant_drafter.py tests/test_async_markers.py` -> passed\n- `git diff --check` -> passed\n\nRequested re-review from @waybarrios and @janhilgard.

janhilgard (changes requested):
CI is red (1 failed, 528 passed):

```
tests/test_lifecycle_server.py::TestLifecycleFailureHandling::test_chat_completion_prep_error_releases_resident

AttributeError: 'types.SimpleNamespace' object has no attribute 'mllm_draft'
```

The test creates a `SimpleNamespace` mock request that doesn't include the new `mllm_draft` field. `_prepare_chat_completion_invocation` at line 650 now accesses `request.mllm_draft` unconditionally — mock requests in existing tests need the field added, or the access should use `getattr(request, "mllm_draft", None)`.

Fix should be straightforward — either add `mllm_draft=None` to the mock request in the failing test, or make the server access defensive.

Thump604:
Pushed 95d3dbe after inspecting the failed Apple Silicon CI log. The failure was a stale lifecycle test double around the existing `model=` acquisition keyword plus the remaining async marker mismatch.\n\nCurrent focused verification on the PR branch:\n- `uv run --extra dev pytest tests/test_lifecycle_server.py::TestCompletionStreamingRelease::test_completion_nonstreaming_error_releases_active_request tests/test_lifecycle_server.py::TestCompletionStreamingRelease::test_completion_streaming_release_matches_chat_pattern tests/test_lifecycle_server.py::TestToolParserUsesLocalEngine::test_chat_completion_initializes_parser_from_acquired_engine tests/test_lifecycle_server.py::TestLifecycleFailureHandling::test_chat_completion_prep_error_releases_resident tests/test_lifecycle_server.py::TestResponseModelFieldUsesServedName::test_completion_response_uses_served_model_name tests/test_lifecycle_server.py::TestResponseModelFieldUsesServedName::test_chat_completion_response_uses_served_model_name tests/test_lifecycle_server.py::TestResponseModelFieldUsesServedName::test_anthropic_response_uses_served_model_name tests/test_mllm_assistant_drafter.py tests/test_async_markers.py` -> 15 passed\n- `uv run --extra dev pytest tests/test_mllm_assistant_drafter.py tests/test_async_markers.py tests/test_mllm_mtp_routing.py tests/test_lifecycle_cli.py` -> 34 passed\n- `uv run --extra dev black --check vllm_mlx/server.py tests/test_lifecycle_server.py tests/test_mllm_assistant_drafter.py tests/test_async_markers.py` -> passed\n- `uv run --extra dev ruff check vllm_mlx/server.py tests/test_lifecycle_server.py tests/test_mllm_assistant_drafter.py tests/test_async_markers.py --select E,F,W --ignore E402,E501,E731,F811,F841` -> passed\n- `git diff --check` -> passed\n\nCI has been restarted on the new head.

janhilgard (approved):
CI green (9/9) after the compatibility fix in 95d3dbed. Previous review concerns (routing gate, metrics, kwargs precedence) were addressed in earlier commits. LGTM.

waybarrios (approved):
All concerns were addressed

Issue #499, opened 2026-05-05
bench-serve: split workload validation/execution/reporting for release-risk reduction


Thump604:
## Context

While doing an advisory lax complexity scan for unreleased `main` versus `v0.2.9`, I found that `vllm_mlx/bench_serve.py` picked up concentrated workload-path complexity in the unreleased PR set.

This is not a claim that #427, #466, or #470 are wrong. The concern is that bench-serve is now both a workload runner and an evidence artifact producer, so complexity in load/validate/run paths makes release behavior and future review harder to reason about.

## Findings

The advisory scan shows the main complexity concentrated around:

- `load_workload`
- `validate_quality_checks`
- `run_workload_case`
- `run_bench_serve`

The file accounts for a `+7` lax finding delta versus the `v0.2.9` release baseline.

## Suggested follow-up

- Keep this as cleanup/testability work, not a behavior change.
- Split workload loading/normalization from quality-check validation.
- Split per-case execution from aggregate reporting/output formatting.
- Add or preserve focused tests for:
  - default merging
  - timeout handling
  - quality-check validation
  - streaming tool-call qualification output
  - artifact schema compatibility
- Avoid mixing this cleanup with Postgres/external-driver promotion unless that PR is explicitly scoped as a refactor.

## Scope note

This issue is advisory release-risk cleanup. It should not impose `agent-slop-lint` as an upstream requirement, and it should not block release unless maintainers decide the bench-serve contract itself is release-critical.

waybarrios:
Status update.

The advisory complexity scan asked for splits across loading, validation, and per-case execution / aggregate reporting. PRs in flight, all based on `main` so they can land in any order:

- #515 — splits `load_workload` (workload loading/normalisation) plus 27 focused regression tests pinning the runner contract
- #517 — splits `validate_quality_checks` into per-check helpers
- (next) — split `run_workload_case` separating request assembly, transport, and record building
- (next) — split `summarize_workload_results` and the workload formatters

The last to land will close this issue.

waybarrios:
Status update.

All three refactor splits the issue asks for are now in flight, each on its own branch from `main` so they can land in any order without merge conflicts:

- #515 — splits `load_workload` plus 27 focused regression tests pinning the runner contract
- #517 — splits `validate_quality_checks` into per-check helpers
- #518 — splits `run_workload_case` into request, transport, and record helpers

[...]

PR #515, opened 2026-05-07
Split load_workload and harden bench-serve workload runner with regression tests

waybarrios:
Refs #499.

## Summary

Splits the 90-line `load_workload` into a thinner orchestrator plus
three private helpers and adds 27 focused regression tests for the
bench-serve workload runner. The refactor follows the boundaries
issue #499 calls out (workload loading vs normalization vs case
construction), preserves observable behavior on every valid input,
and folds in a small bug fix as a side benefit.

This PR is the first half of #499. The remaining splits the issue
asks for, separating quality-check validation from per-case
execution and per-case execution from aggregate reporting, live in
follow-ups. Issue stays open until those land.

## Refactor: split `load_workload`

Extracted from `vllm_mlx/bench_serve.py` `load_workload`:

- **`_normalize_tags(tags, *, case_id) -> tuple[str, ...]`**. Coerces a
  case's `tags` field (string or list) to a tuple and rejects other
  types with a case-scoped `ValueError`.
- **`_merge_case_checks(default_checks, case_checks, *, case_id) -> dict | None`**.
  Merges a case's checks over the workload defaults. Most keys are
  overridden by the case-level value; `required_regex` and
  `forbidden_regex` are list-concatenated with default patterns first
  and case-level patterns appended. Validates that `case_checks` is a
  dict.
- **`_build_workload_case(item, idx, *, defaults, workload_path) -> WorkloadCase`**.
  Constructs one `WorkloadCase` from a raw entry, composing the
  helpers above and resolving scalar fields (`max_tokens`,
  `enable_thinking`, `policy_timeout_ms`) via `_first_not_none`
  priority (case-level beats request_path defaults beats workload
  defaults).

`load_workload` itself is now about 25 lines covering file I/O, JSON
parse, root validation, and workload assembly. The 16 existing tests
on `load_workload` (8 in this file's `TestLoadWorkloadValidation`, 2
in `TestLoadWorkloadDefaultMerging`, 6 in `test_bench_serve.py`) all
pass without modification, which is the safety net the refactor was
gated on.

### Bug fix folded in

Before this PR, a non-dict truthy `checks` value (a bare string, list,
or int in the JSON) crashed deep inside the loader with a cryptic
`AttributeError: 'str' object has no attribute 'items'` because the
inline merge called `case_checks.items()` without validating type.
After this PR, `_merge_case_checks` validates the type up front and
raises `ValueError("<case_id>: checks must be an object")` so the
operator gets a clear, case-scoped diagnostic.

The dead post-hoc guard `if checks is not None and not isinstance(checks, dict)`
in the original was preserved in the first refactor pass for safety;
it has been removed now that the validation lives in the helper.

## Tests added

`tests/test_bench_serve_workload_hardening.py` (27 cases):

`TestLoadWorkloadValidation` (9):
- root must be a JSON object
- empty cases list rejected
- missing cases key rejected
- defaults must be an object
- case must be an object
- extra_body invalid type rejected
- tags string is normalised to a single-tuple
- tags invalid type (int) rejected
- checks non-dict-truthy now rejected as `ValueError` (covers the bug fix)

`TestLoadWorkloadDefaultMerging` (2):
- case-level `max_tokens` beats `defaults.max_tokens`
- workload name falls back to filename stem when no `name` is set

`TestStreamingToolCallAccumulation` (4):
- name and arguments concatenate across chunk-boundary deltas
- out-of-order indices (2, 0, 1) across multi-call state land sorted
- id set on the first delta survives later deltas that omit it
- omitted index defaults to 0 (mlx-lm style streams sometimes do this)

`TestQualityCheckDiagnostics` (7):
- `tool_call_count` mismatch surfaces the check name and both counts
- `tool_call_args` invalid JSON reports a parse error
- `tool_call_args` non-object arguments are flagged
- missing required keys are listed (and present keys are not)
- `no_tool_calls` combines with `min_chars` to surface both diagnostics
- `finish_reason` list-form accepts any allowed member
- `finish_reason` string-form rejects others

`TestArtifactSchemaCompatibility` (5):
- `summarize_workload_results` top-level dict has the core key set
- `summarize_workload_results` per-case summary entries have the core key set
- `format_workload_table` renders the core column headers
- `format_workload_csv` rows have the core columns
- `format_workload_json` round-trips top-level keys (workload, summary, results)

The contract in the schema tests is core-keys-must-exist, not exact
match. New fields are fine; deleting or renaming a core field breaks
the suite. Diagnostic-string assertions use short stable anchors
instead of full sentences, so future cleanup of error wording does
not cosmetically break the suite.

## Verification

```text
======================== 27 passed in 0.16s ========================
```

`pytest tests/test_bench_serve.py tests/test_bench_serve_workload_hardening.py`
→ 98 + 27 + 2 skipped (live-run integration) = 125/125 passed.

`black --check vllm_mlx/ tests/` is clean.

## Demo: tests catch a real regression

To prove the streaming-accumulation tests aren't vacuous, I temporarily
swapped the `+=` concatenation in `accumulate_tool_calls` for a `=`
overwrite (the regression #499 worries about as the dense state
machine gets touched):

```python
if function_delta.get("name"):
    entry["function"]["name"] = function_delta["name"]   # was: +=
if function_delta.get("arguments"):
    entry["function"]["arguments"] = function_delta["arguments"]  # was: +=
```

The hardening test fired with a precise diagnostic:

```text
FAILED test_concatenates_name_and_arguments_across_deltas
  AssertionError: assert 'weather' == 'get_weather'

  - get_weather
  ? ----
  + weather
========================= 1 failed, 26 passed in 0.23s =========================
```

After restoring `+=` the full suite goes back to 27 passed.

## Note on scope

This PR delivers the test-pinning half of #499 plus the first split
the issue asks for (workload loading). The remaining splits, separating
quality-check validation from per-case execution and per-case
execution from aggregate reporting, are larger surgeries on
`validate_quality_checks` and `run_workload_case` and live in
follow-ups. The tests landed here are the precondition for doing
those refactors safely.

### DIFF
--- a/tests/test_bench_serve_workload_hardening.py
+++ b/tests/test_bench_serve_workload_hardening.py
@@ -1,0 +1,581 @@
+# SPDX-License-Identifier: Apache-2.0
+"""
+Targeted regression tests for the bench-serve workload runner.
+
+The existing ``test_bench_serve.py`` already covers happy paths for workload
+loading, sweep expansion, formatters, runner end-to-end, and basic quality
+checks. Issue #499 highlighted that the load/validate/run/report code paths
+in ``vllm_mlx/bench_serve.py`` are dense and worth pinning further. This
+module fills the remaining corners around the areas the issue calls out:
+
+- ``load_workload`` validation errors. Every guard clause that rejects a
+  malformed workload JSON, plus tag-string normalisation.
+- ``load_workload`` default merging. ``max_tokens``, ``enable_thinking``,
+  and the workload-name fallback to filename stem.
+- Streaming tool-call accumulation. ``accumulate_tool_calls`` and
+  ``finalize_tool_calls`` chunk-boundary behaviour and ordering.
+- ``validate_quality_checks`` diagnostics. Tool-call argument validation
+  edge cases (invalid JSON, non-object, missing keys, count mismatch) and
+  the combination of ``no_tool_calls`` with content-length checks.
+- Artifact schema compatibility. ``summarize_workload_results``,
+  ``format_workload_json``, ``format_workload_csv``, and
+  ``format_workload_table`` are pinned to a stable core set of keys and
+  columns so an accidental rename or deletion during refactor is caught.
+
+Brittleness note: assertions on ``match=`` strings and diagnostic
+substrings deliberately use short, stable anchors instead of full
+sentences. The contract is the diagnostic *intent* (which subject the
+error names, which check failed), not the exact wording.
+"""
+
+from __future__ import annotations
+
+import csv
+import json
+from pathlib import Path
+
+import pytest
+
+from vllm_mlx.bench_serve import (
+    accumulate_tool_calls,
+    finalize_tool_calls,
+    format_workload_csv,
+    format_workload_json,
+    format_workload_table,
+    load_workload,
+    summarize_workload_results,
+    validate_quality_checks,
+)
+
+# ---------------------------------------------------------------------------
+# load_workload — validation guard clauses
+# ---------------------------------------------------------------------------
+
+
+class TestLoadWorkloadValidation:
+    def _write(self, tmp_path: Path, payload) -> Path:
+        f = tmp_path / "workload.json"
+        f.write_text(json.dumps(payload))
+        return f
+
+    def test_root_must_be_object(self, tmp_path: Path):
+        f = tmp_path / "workload.json"
+        f.write_text(json.dumps(["not", "an", "object"]))
+        with pytest.raises(ValueError, match=r"root.*JSON|JSON.*root"):
+            load_workload(f)
+
+    def test_empty_cases_list_rejected(self, tmp_path: Path):
+        f = self._write(tmp_path, {"cases": []})
+        with pytest.raises(ValueError, match="cases"):
+            load_workload(f)
+
+    def test_missing_cases_key_rejected(self, tmp_path: Path):
+        f = self._write(tmp_path, {"defaults": {}})
+        with pytest.raises(ValueError, match="cases"):
+            load_workload(f)
+
+    def test_defaults_must_be_object(self, tmp_path: Path):
+        f = self._write(
+            tmp_path,
+            {
+                "defaults": "not-an-object",
+                "cases": [{"id": "a", "messages": [{"role": "user", "content": "hi"}]}],
+            },
+        )
+        with pytest.raises(ValueError, match="defaults"):
+            load_workload(f)
+
+    def test_case_must_be_object(self, tmp_path: Path):
+        f = self._write(tmp_path, {"cases": ["not-an-object"]})
+        # ``case`` alone would also match the list-level "non-empty cases"
+        # error; pin to the per-item form.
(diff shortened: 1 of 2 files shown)


waybarrios:
Follow-up: PR #517 splits `validate_quality_checks` along the same axis. Independent of this PR (based on `main`, no merge conflict). Together they deliver 2 of the 3 refactor splits issue #499 asks for; the third (`run_workload_case`) is queued next.

waybarrios:
Follow-up: PR #518 splits `run_workload_case` into request, transport, and record helpers. Independent of this PR (based on `main`, no merge conflict). With #517 and this one, three of the three refactor splits #499 asks for are now in flight.

waybarrios:
Follow-up: PR #519 splits `summarize_workload_results` into a per-case helper. With this and #517 / #518, all four refactor splits #499 asks for are now in flight. Independent of this PR (based on `main`, no merge conflict).

Thump604 (approved):
Reviewed the diff and ran the focused suite locally against this PR ref:

```text
PYTHONPATH=/tmp/vllm-mlx-review-bench AI_RUNTIME_BYPASS_SAFETY_GATE=1 /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_bench_serve.py tests/test_bench_serve_workload_hardening.py -q
125 passed, 2 skipped
```

The extraction keeps `load_workload` as an orchestrator and moves the normalization/merge/case-build details behind small helpers without changing the valid-input contract. The new hardening tests are useful here: they pin workload loading, tool-call streaming accumulation, quality-check diagnostics, and artifact schema shape before the follow-up refactors land.

One non-blocking note: the `extra_body` invalid-type path still accepts the existing `dict.update(...)`/sequence-style error shape rather than producing a new clean diagnostic. Since that is not introduced by this PR and the `checks` type bug is the behavior actually fixed here, I do not see it as a blocker.

LGTM.

janhilgard (approved):
Clean first split of #499. The extraction boundaries are well-chosen: `_normalize_tags` isolates the string/list coercion, `_merge_case_checks` owns the regex list-concatenation merge plus the new type validation, and `_build_workload_case` encapsulates the full per-entry construction including `_first_not_none` priority resolution.

The bug fix is correct: the original `if item.get("checks"): ... item["checks"].items()` would crash with `AttributeError` on a non-dict truthy value (bare string or list). The new `_merge_case_checks` validates the type upfront and raises a case-scoped `ValueError`. The dead post-hoc guard (`if checks is not None and not isinstance(checks, dict)`) that could never trigger (since `merged_checks` was always a `dict`) is correctly removed.

I agree with Thump604's non-blocking note about the `extra_body` type guard being removed — the implicit `dict.update()` error is less clean, but the test pins both error paths (`ValueError | TypeError`), so a regression that suppresses the error would be caught.

The 27 new tests are well-structured:
- Short stable anchors instead of full sentences for diagnostic assertions
- `TestArtifactSchemaCompatibility` pins core keys with a superset contract (new fields fine, deletions break) — this is the right precondition before the follow-up refactors
- The streaming accumulation tests proved non-vacuous per the regression demo in the PR body

CI 9/9 green, single commit, Thump604 approved.

PR #517, opened 2026-05-09
Split validate_quality_checks into per-check helpers

waybarrios:
Refs #499. Follow-up to #515.

## Summary

Splits the 122-line `validate_quality_checks` into a thinner
orchestrator plus six private helpers, one per supported check
family. This is the second of three refactor splits issue #499 asks
for in `vllm_mlx/bench_serve.py`. The first (workload loading) lives
in #515; the third (per-case execution) is a follow-up.

Behavior-preserving on every input. The 9 existing tests in
`TestQualityChecks` (`tests/test_bench_serve.py:917`) pass without
modification, which is the safety net this refactor was gated on.

## Refactor: split validate_quality_checks

Extracted from `vllm_mlx/bench_serve.py` `validate_quality_checks`:

- **`_check_finish_reason(allowed, finish_reason) -> list[str]`**.
  Verifies `finish_reason` is in the allowed set when one is configured.
- **`_check_length_bounds(min_chars, max_chars, content) -> list[str]`**.
  Applies `min_chars` / `max_chars` content-length bounds.
- **`_check_regex_patterns(patterns, content, *, kind, expect_match) -> list[str]`**.
  Shared by `required_regex` and `forbidden_regex` via the
  `expect_match` flag, with `kind` reused in diagnostics so operators
  can grep for the failing check.
- **`_check_json_content(should_be_json, content) -> list[str]`**.
  Verifies `content` parses as JSON when `checks['json']` is truthy.
- **`_check_tool_call_count_and_names(checks, tool_calls) -> list[str]`**.
  Applies `no_tool_calls` / `tool_call_count` / `tool_call_names`.
- **`_check_tool_call_args(required_args, tool_calls) -> list[str]`**.
  The complex one: validates parsed JSON arguments include the
  required keys per function, reporting missing tool calls, invalid
  JSON, non-object arguments, and missing keys.

`validate_quality_checks` itself is now about 25 lines: it runs
`validate_response`, normalises the inputs, then `.extend()`s the
issue list with each helper's output.

## Verification

```text
======================== 98 passed in 0.38s ========================
```

Full bench-serve suite is green:
`pytest tests/test_bench_serve.py` → 98 passed + 2 skipped (live-run
integration).

`black --check vllm_mlx/bench_serve.py` is clean.

## Note on scope

This PR is the validation split only. It does not touch
`load_workload` (see #515), `run_workload_case`, or
`summarize_workload_results`. Each remaining split is a separate
independent PR based on `main` so they can land in any order without
[...]

PR #518, opened 2026-05-09
Split run_workload_case into request, transport, and record helpers

waybarrios:
Refs #499. Follow-up to #515 and #517.

## Summary

Splits the 150-line `run_workload_case` into a thinner async
orchestrator plus seven private helpers along the boundaries issue
#499 calls out for `vllm_mlx/bench_serve.py`. This is the third of
three refactor splits in that issue. The first (workload loading)
lives in #515; the second (quality-check validation) in #517.

Behavior-preserving on every input. The `TestWorkloadRunner` test
(`tests/test_bench_serve.py:1190`) passes without modification, which
is the safety net this refactor was gated on.

## Refactor: split run_workload_case

Extracted from `vllm_mlx/bench_serve.py` `run_workload_case`:

- **`_resolve_max_tokens(case, workload) -> int`**. Collapses the
  duplicated `int(case.max_tokens or workload.defaults.get("max_tokens", 256))`
  used in two places (kwargs assembly and the request block of the
  record).
- **`_assemble_case_request_kwargs(case, workload, model) -> dict`**.
  Builds the kwargs dict passed to `stream_chat_completion` in one
  place, including the policy-timeout-to-seconds conversion.
- **`_empty_completion_result() -> dict`**. Zero-valued shape used
  when the transport raises, so downstream code can read
  `result.get(...)` without branching on success vs failure.
- **`_fetch_post_run_status(client, base_url) -> dict`** (async).
  Wraps the `GET /v1/status` call with the swallow-and-empty-dict
  fallback the original inlined.
- **`_compute_within_policy_timeout(timeout_ms, *, error_present, e2e_latency_ms) -> Optional[bool]`**.
  Pins the three-way None / False / comparison logic in one named
  place: None when no policy timeout was configured, False on error,
  otherwise the latency comparison.
- **`_build_tool_calls_summary(tool_calls) -> Optional[dict]`**.
  Returns the compact summary or None when no tool calls were
  emitted, so consumers can distinguish "feature not exercised" from
  "feature exercised, zero calls".
- **`_build_workload_record(...) -> dict`**. Pure function assembling
  the JSON-serializable record from raw inputs and the completion
  result. No I/O, deterministic given its arguments.

`run_workload_case` itself drops from 150 lines to about 50 and
reads as a straight orchestration: pre-scrape, transport, post-scrape,
validate, build record.

## Verification

```text
======================== 98 passed in 0.38s ========================
```

`pytest tests/test_bench_serve.py` → 98 passed + 2 skipped (live-run
integration).

`black --check vllm_mlx/bench_serve.py` is clean.

## Note on scope

This PR is the per-case execution split only. It does not touch
`load_workload` (see #515), `validate_quality_checks` (see #517), or
`summarize_workload_results` / the workload formatters. Each split is
independent of the others, all based on `main`, so they can land in
any order without merge conflicts.

The remaining `summarize_workload_results` / formatters cleanup is a
follow-up. The final split can mark issue #499 complete after all related PRs land.

No new tests required: `TestWorkloadRunner` already exercises the
full record-building path with monkeypatched transport, and the
`_build_workload_record` extraction is covered by the same test
through `run_workload_case`.

### DIFF
--- a/vllm_mlx/bench_serve.py
+++ b/vllm_mlx/bench_serve.py
@@ -1229,94 +1229,120 @@ def _summary_or_empty(values: list[float]) -> dict:
     return compute_summary_stats(values) if values else {}
 
 
-async def run_workload_case(
-    client: httpx.AsyncClient,
-    base_url: str,
+def _resolve_max_tokens(case: WorkloadCase, workload: Workload) -> int:
+    """Return the effective ``max_tokens`` for a case, falling back to
+    workload defaults and finally to 256."""
+    return int(case.max_tokens or workload.defaults.get("max_tokens", 256))
+
+
+def _assemble_case_request_kwargs(
+    case: WorkloadCase, workload: Workload, model: str
+) -> dict:
+    """Build the keyword-arguments dict passed to ``stream_chat_completion``
+    for one case, applying max_tokens fallback and converting
+    ``policy_timeout_ms`` to seconds."""
+    return {
+        "messages": case.messages,
+        "model": model,
+        "max_tokens": _resolve_max_tokens(case, workload),
+        "enable_thinking": case.enable_thinking,
+        "extra_body": case.extra_body,
+        "timeout_s": (
+            case.policy_timeout_ms / 1000
+            if case.policy_timeout_ms is not None
+            else None
+        ),
+    }
+
+
+def _empty_completion_result() -> dict:
+    """Zero-valued completion result used when ``stream_chat_completion``
+    raises. The structure matches a real successful response so downstream
+    code can read ``result.get(...)`` without branching on the failure."""
+    return {
+        "ttft_ms": 0.0,
+        "tpot_ms": 0.0,
+        "e2e_latency_ms": 0.0,
+        "gen_tps": 0.0,
+        "prompt_tps": 0.0,
+        "prompt_tokens": 0,
+        "completion_tokens": 0,
+        "finish_reason": None,
+        "content": "",
+        "tool_calls": [],
+    }
+
+
+async def _fetch_post_run_status(client: httpx.AsyncClient, base_url: str) -> dict:
+    """GET ``/v1/status`` after a case run, swallowing transport errors so
+    a missing or temporarily-unavailable status endpoint does not fail
+    the case record."""
+    try:
+        resp = await client.get(f"{base_url}/v1/status")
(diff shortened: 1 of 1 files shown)


Thump604:
Reviewed the run-case split and ran the focused suite locally against this PR ref:

```text
PYTHONPATH=/tmp/vllm-mlx-review-bench AI_RUNTIME_BYPASS_SAFETY_GATE=1 /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_bench_serve.py -q
98 passed, 2 skipped
```

The extraction boundaries look right: request kwargs, empty-error result shape, post-run status, policy-timeout calculation, tool-call summary, and record assembly are now separately readable while `run_workload_case(...)` still owns the orchestration. I did not see a behavior change in the record contract.

I am holding formal approval only because the PR body contains the literal close keyword in this sentence: `The last to land can Closes #499.` Even as explanatory text, that is a mechanical GitHub close-keyword hazard on a non-final split. Please reword it to avoid the literal close token, for example “the last split to land can close issue #499 after all four PRs merge.” After that wording fix, I expect this to be approvable.

janhilgard (approved):
Clean behavior-preserving refactor. The seven extracted helpers have well-chosen boundaries — each isolates a single concern (max_tokens fallback, kwargs assembly, error-shape, status fetch, policy-timeout logic, tool-call summary, record building), and `run_workload_case` now reads as a straight orchestration flow.

I verified the `_build_tool_calls_summary` extraction preserves the original semantics: the caller passes `result.get("tool_calls") or []`, so `None` and `[]` both map to `[]` at the helper boundary, which returns `None` — matching the original `if result.get("tool_calls") else None` gate.

`_build_workload_record` has a large parameter surface (20 kwargs), but that's the right trade-off for a pure record-builder — all inputs are explicit and the function is deterministic with no I/O.

CI 9/9 green, single commit.

One item from Thump604's earlier comment is still outstanding: the PR body contains `The last to land can Closes #499` — the literal `Closes #499` is a GitHub auto-close keyword. Please reword to avoid accidental issue closure when this (non-final) split merges.

Thump604 (approved):
The close-keyword hazard in the PR body is resolved. Prior focused bench-serve validation and the green CI matrix still cover this behavior-preserving split. No remaining blockers from my review.

PR #519, opened 2026-05-09
Split summarize_workload_results into per-case helper

waybarrios:
Refs #499. Follow-up to #515, #517, #518.

## Summary

Splits the 80-line `summarize_workload_results` into a thinner
top-level aggregator plus two helpers along the same boundaries the
earlier splits in #515, #517, and #518 followed. This is the last
follow-up issue #499 calls out for `vllm_mlx/bench_serve.py`.

Behavior-preserving on every input. The existing
`test_summarize_workload_results_tracks_quality_and_policy` in
`tests/test_bench_serve.py:1145` passes without modification.

## Refactor: split summarize_workload_results

Extracted from `vllm_mlx/bench_serve.py` `summarize_workload_results`:

- **`_group_results_by_case_id(results) -> dict[str, list[dict]]`**.
  Buckets workload case records by their `case_id` field, defaulting
  a missing `case_id` to the empty string so the grouping is stable.
- **`_summarize_case(case_results) -> dict`**. Builds the per-case
  summary block. Mirrors the shape used at the run level (sample
  counts, pass/fail rates, policy-timeout outcome, latency / ttft /
  gen_tps summaries) and adds the case-only fields `sample_count`,
  `repetitions`, `content_chars`.

`summarize_workload_results` itself becomes a clean top-level
aggregator: compute global failures and policy trials, group by case,
build per-case summaries via the helper, return the assembled dict.

The workload formatters (`format_workload_table`,
`format_workload_csv`, `format_workload_sql`,
`format_workload_json`, `format_workload_payload`,
`write_workload_sqlite`) were considered as part of this split too,
but `_workload_record_to_row` (the only repeated logic) is already
factored out, and each formatter is small and single-purpose. No
extraction needed there.

## Verification

```text
======================== 98 passed in 0.40s ========================
```

`pytest tests/test_bench_serve.py` → 98 passed + 2 skipped (live-run
integration).

`black --check vllm_mlx/bench_serve.py` is clean.

## Note on scope

This is the final refactor split for #499. With #515, #517, #518,
and this PR, the four splits the advisory complexity scan asked for
are all in flight, each independent of the others (all based on
`main`, no merge conflicts). The last of the four to land closes the
issue.

No new tests required: the existing
`test_summarize_workload_results_tracks_quality_and_policy` exercises
the per-case summary path through the public function.

### DIFF
--- a/vllm_mlx/bench_serve.py
+++ b/vllm_mlx/bench_serve.py
@@ -1593,11 +1593,63 @@ async def run_workload_case(
     )
 
 
+def _group_results_by_case_id(results: list[dict]) -> dict[str, list[dict]]:
+    """Bucket workload case records by their ``case_id`` field, defaulting
+    a missing ``case_id`` to the empty string so the grouping is stable."""
+    cases: dict[str, list[dict]] = {}
+    for result in results:
+        cases.setdefault(str(result.get("case_id", "")), []).append(result)
+    return cases
+
+
+def _summarize_case(case_results: list[dict]) -> dict:
+    """Build the per-case summary block.
+
+    Mirrors the shape used at the run-level (sample counts, pass/fail
+    rates, policy-timeout outcome, latency / ttft / gen_tps summaries)
+    and adds two case-only fields: ``sample_count`` and ``repetitions``
+    (the sorted set of repetition indices the case was run under), plus
+    ``content_chars`` since content length is more useful per-case than
+    per-run.
+    """
+    quality_failures = [r for r in case_results if not r["quality"].get("ok")]
+    policy_trials = [
+        r for r in case_results if r["policy"].get("within_timeout") is not None
+    ]
+    policy_failures = [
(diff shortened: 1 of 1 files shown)


Thump604 (approved):
Reviewed the summary split and ran the focused suite locally against this PR ref:

```text
PYTHONPATH=/tmp/vllm-mlx-review-bench AI_RUNTIME_BYPASS_SAFETY_GATE=1 /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_bench_serve.py -q
98 passed, 2 skipped
```

The helper boundary is clean: `_group_results_by_case_id` only owns bucketing, `_summarize_case` owns the per-case aggregate shape, and `summarize_workload_results` remains the run-level aggregator. I do not see behavior drift in the result keys, timeout summary, latency summaries, or case ordering.

Also good that the PR body uses `Refs #499` rather than an automatic close keyword. This one can land independently; #499 should only be closed once the full split set has actually merged.

LGTM.

PR #525, opened 2026-05-10
Correct _suffix off-by-one and strip backslashes before non-ASCII in JSON output

yfaney:
This change is to fix running VLLM-MLX with Gemma 4 with:

  - json_schema_processor: `_suffix()` was initialising `_prompt_len` as `len(tokens_list) - 1`, which prepended the last prompt token to every suffix fed to the enforcer.  For Gemma 4 that token is `<channel|>` (the end-of-thinking marker), causing the JSON grammar to fail immediately and produce empty output. 
    - Fix: use `len(tokens_list)` so the prompt is fully excluded.

  - server: add `_strip_backslash_before_unicode()` and apply it at both JSON post-processing call sites (OpenAI and Anthropic endpoints). `lm-format-enforcer` allows `\\` followed by non-ASCII characters as a valid JSON escape sequence, so the model generates `\\빠\\르\\게` for Korean text.  The helper strips the spurious backslashes after `json.loads` decodes them to literal `\; json.dumps` with `ensure_ascii=False` then emits clean Unicode rather than `\\uXXXX` escapes.

### DIFF
--- a/vllm_mlx/constrained/json_schema_processor.py
+++ b/vllm_mlx/constrained/json_schema_processor.py
@@ -348,7 +348,11 @@ class JSONSchemaLogitsProcessor:
     def _suffix(self, tokens_list: list[int]) -> list[int]:
         """Return the slice of ``tokens`` that corresponds to generated output."""
         if self._prompt_len is None:
-            self._prompt_len = max(0, len(tokens_list) - 1)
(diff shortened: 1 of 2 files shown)


waybarrios (approved):
I was reviewing and it seems good. Thank you for this contribution!

PR #527, opened 2026-05-11
Clarify MLLM MTP bypass counter semantics

Thump604:
Refs #473.

## Summary

Adds an explicit `/v1/status.mtp` schema marker for `bypass_counts`:

```json
"bypass_counts_semantics": "per_condition_overlapping_not_total_steps"
```

This preserves the current counter behavior while making it clear to clients that the `bypass_counts` values are per-condition counters. Conditions can overlap, so summing the values is not a total bypassed-step count.

## Validation

- `AI_RUNTIME_BYPASS_SAFETY_GATE=1 PYTHONPATH=/opt/ai-runtime/worktrees/vllm-mlx/issue-473-bypass-counts-schema-note /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_mllm_continuous_batching.py::TestMLLMBatchGeneratorMTPGuards -q` -> 11 passed
- `AI_RUNTIME_BYPASS_SAFETY_GATE=1 PYTHONPATH=/opt/ai-runtime/worktrees/vllm-mlx/issue-473-bypass-counts-schema-note /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_mllm_continuous_batching.py -q` -> 47 passed, 4 deselected
- `python3 -m py_compile vllm_mlx/mllm_batch_generator.py tests/test_mllm_continuous_batching.py`
- `uvx ruff check vllm_mlx/mllm_batch_generator.py tests/test_mllm_continuous_batching.py --select E,F,W --ignore E402,E501,E731,F811,F841`
- `black --check --target-version py312 vllm_mlx/mllm_batch_generator.py tests/test_mllm_continuous_batching.py`
- `git diff --check`

## Out of scope

- No MTP counter behavior changes.
- No changes to acceptance-rate math.
- No performance or routing changes.

### DIFF
--- a/tests/test_mllm_continuous_batching.py
+++ b/tests/test_mllm_continuous_batching.py
@@ -1010,6 +1010,10 @@ class TestMLLMBatchGeneratorMTPGuards:
         assert stats["requested_draft_tokens"] == 4
         assert stats["effective_draft_tokens"] == 1
         assert stats["attempted"] == 0
+        assert (
+            stats["bypass_counts_semantics"]
+            == "per_condition_overlapping_not_total_steps"
+        )
         assert stats["bypass_counts"] == {
             "prefill": 0,
             "no_active_batch": 0,
--- a/vllm_mlx/mllm_batch_generator.py
+++ b/vllm_mlx/mllm_batch_generator.py
@@ -2005,6 +2005,7 @@ def install_mtp_mllm(
             "errors": errors,
             "acceptance_rate": acceptance_rate,
             "bypass_counts": bypass_counts,
+            "bypass_counts_semantics": "per_condition_overlapping_not_total_steps",
         }
 
     batch_gen.get_mtp_stats = _get_mtp_stats


janhilgard (commented):
The change itself is trivial and correct — 1 line in production code, 4 in tests. CI 9/9.

### Question

Is a static constant string in the runtime JSON response the right place for this documentation? The value `"per_condition_overlapping_not_total_steps"` will never change at runtime — it's effectively API documentation embedded in every response payload.

Alternatives:
1. A comment/docstring in `_get_mtp_stats()` explaining the semantics
2. Documentation in the API docs or README
3. Keep it as-is if clients parse this field programmatically to adapt their aggregation logic

If there's a concrete client that reads this field to decide how to aggregate counters, then having it in the payload makes sense. Otherwise it adds bytes to every `/v1/status.mtp` response for documentation that could live elsewhere.

Not blocking — just raising the question for the maintainer's call.

janhilgard (approved):
Trivial, harmless addition — 1 line in production code, 4 in tests. The static field documents counter semantics for API consumers. CI 9/9. LGTM.

Thump604:
For the maintainer call: I would keep this in the status payload rather than moving it to docs only.

The field is intentionally machine-readable metadata for clients that aggregate `/v1/status.mtp`. The bypass counters are per condition and can overlap, so summing them as total bypassed steps would be wrong. A static field is a small cost on a diagnostic endpoint and lets clients avoid hardcoding that convention. If you prefer API docs only, I can move it, but the runtime field is deliberate rather than just inline documentation.

Issue #524, opened 2026-05-10
Prefix cache: generalize prompt canonicalization beyond Anthropic billing header


waybarrios:
## Summary

Generalise the per-request system-prompt sanitization that commit b9f2a5f introduced for one specific client (Claude Code) into a registry that any client family can extend. The current single `re.sub` call only catches the Anthropic billing header; other clients inject equivalent prefix-cache poison that today nukes hit rate the same way.

## Problem

`vllm_mlx/api/anthropic_adapter.py:67` strips the `x-anthropic-billing-header:` line from system prompts via inline `re.sub`. The `cch=HASH` token in that line changes per request and invalidates every prefix-cache block after position ~40, so consecutive turns from the same conversation share zero blocks unless that line is removed first.

The fix from #277 works for that one client. The same failure mode affects other clients with their own per-request volatile bytes, and each of them currently degrades the cache the same way:

| Client | Volatile bytes | Status |
|---|---|---|
| Claude Code | `x-anthropic-billing-header: ...; cch=<hash>;` | covered (b9f2a5f) |
| OpenAI Responses API / certain copilots | session IDs, per-turn timestamps | not covered |
| MCP servers | tool-call invocation UUIDs in echoed context | not covered |
| Locale-formatted timestamps | `2026-05-10T13:42:18.123Z` strings | not covered |

## Impact

Reported on the original fix (b9f2a5f / #277):

> **50s → 3.65s per request (13.7×)** on Gemma 4 26B-A4B with 60K-token prompts.

That delta only applies to Claude Code today. Each additional client family covered by the registry would unlock the same order of speedup for its users.

## Proposed module

New file `vllm_mlx/api/prompt_canonicalize.py`:

```python
def canonicalize_system_prompt(text: str) -> str:
    """Apply every registered stripper to text, in registration order."""

def register_stripper(
    name: str,
    pattern: re.Pattern[str] | str,
    replacement: str = "",
) -> None:
    """Register an additional stripper at runtime. Names must be unique."""

def unregister_stripper(name: str) -> bool: ...

def list_strippers() -> list[str]: ...
```

Built-in strippers (initial set):

- `anthropic_billing_header` — the existing regex from b9f2a5f, moved into the registry.
- `mcp_tool_call_id` — opt-in, once the regex is verified against real MCP traffic.
- `iso_timestamp_in_system` — opt-in (risky: may strip legitimate "current time" hints the model needs).

## Integration options

[...]

PR #528, opened 2026-05-12
Canonicalize volatile system prompt headers in OpenAI paths

Thump604:
Refs #524.

## Summary

- Add a small static system-prompt canonicalization helper with the currently validated `x-anthropic-billing-header` stripper.
- Apply it to prepared system-role messages in Chat Completions and Responses paths before engine execution.
- Cover the helper, Chat Completions preparation, Responses preparation, and existing Anthropic adapter behavior.

## Local repro / observed behavior

On current `waybarrios/vllm-mlx` main (`f068991` when this branch was cut), the Anthropic Messages adapter removes `x-anthropic-billing-header:` from `request.system` in `vllm_mlx/api/anthropic_adapter.py::anthropic_to_openai`, but the OpenAI server paths did not apply the same canonicalization:

- `vllm_mlx/server.py::_prepare_chat_messages`
- `vllm_mlx/server.py::_prepare_responses_request`

A Chat Completions system message or Responses `instructions` value containing:

```text
x-anthropic-billing-header: account=abc; cch=rotating-hash
```

was still present in the prepared system message sent to the engine.

## Expected behavior

The validated billing-header line is non-semantic request metadata and should be removed from system-role text before engine execution. User-role content with the same text is preserved, and user-visible timestamp text is not stripped.

## Minimal patch shape

- New `vllm_mlx/api/prompt_canonicalize.py` module with a static stripper list and `canonicalize_system_prompt()`.
- New `canonicalize_system_messages()` helper that copies only changed system messages.
- Call the helper after existing message normalization in Chat Completions and after Responses input conversion.
- Do not add runtime registration APIs or speculative strippers.

## Explicitly not claimed

- This does not add timestamp, MCP UUID, or session-ID strippers.
- This does not change SimpleEngine system-prefix KV-cache logic from #523.
- This does not include new TTFT or cache-hit-rate benchmark results.
- This does not change media extraction, tool parsing, sampling, or decode controls.

## Verification

```sh
AI_RUNTIME_BYPASS_SAFETY_GATE=1 PYTHONPATH=/opt/ai-runtime/worktrees/vllm-mlx/issue-524-prompt-canonicalization /opt/ai-runtime/venv-live/bin/python -m pytest tests/test_prompt_canonicalize.py tests/test_responses_api.py tests/test_anthropic_adapter.py tests/test_server.py::TestPromptCanonicalization -q
# 74 passed

uvx ruff check vllm_mlx/api/prompt_canonicalize.py vllm_mlx/server.py tests/test_prompt_canonicalize.py tests/test_server.py tests/test_responses_api.py
# All checks passed

/opt/ai-runtime/venv-live/bin/python -m black --check --target-version py312 vllm_mlx/api/prompt_canonicalize.py vllm_mlx/server.py tests/test_prompt_canonicalize.py tests/test_server.py tests/test_responses_api.py
# 5 files would be left unchanged

git diff --check
# clean
```

### DIFF
--- a/tests/test_prompt_canonicalize.py
+++ b/tests/test_prompt_canonicalize.py
@@ -1,0 +1,34 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for system-prompt canonicalization."""
+
+from vllm_mlx.api.prompt_canonicalize import canonicalize_system_prompt
+
+
+def test_canonicalize_system_prompt_strips_anthropic_billing_header_line():
+    text = (
+        "You are a coding assistant.\n"
+        "x-anthropic-billing-header: account=abc; cch=rotating-hash\n"
+        "Follow the repository instructions."
+    )
+
+    expected = "\n".join(
+        ["You are a coding assistant.", "Follow the repository instructions."]
+    )
+    assert canonicalize_system_prompt(text) == expected
+
+
+def test_canonicalize_system_prompt_is_idempotent_for_clean_input():
+    text = "You are a coding assistant.\nCurrent time: 2026-05-12T01:00:00Z"
+
+    assert canonicalize_system_prompt(text) == text
+
+
+def test_canonicalize_system_prompt_does_not_strip_user_visible_timestamp():
+    text = "Current time: 2026-05-12T01:00:00Z\nUse it when answering."
+
+    assert canonicalize_system_prompt(text) == text
+
+
+def test_canonicalize_system_prompt_handles_empty_and_none():
+    assert canonicalize_system_prompt("") == ""
+    assert canonicalize_system_prompt(None) is None
--- a/tests/test_responses_api.py
+++ b/tests/test_responses_api.py
@@ -416,6 +416,30 @@ class TestResponsesEndpoint:
         assert messages[1]["role"] == "user"
         assert messages[1]["content"] == "Hi"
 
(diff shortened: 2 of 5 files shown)


janhilgard (approved):
Clean implementation. The `prompt_canonicalize.py` module is well-isolated with an extensible `_STRIPPERS` tuple, and `canonicalize_system_messages()` correctly avoids mutation (copies only changed messages, returns the original list when nothing changes).

### Review

- Regex `(?im)^x-anthropic-billing-header:[^\n]*(?:\n|$)` — multiline + case-insensitive + anchored. Correctly strips the full line including trailing newline.
- Only targets `system` role — user content is preserved (explicitly verified by the chat completion test).
- `None` passthrough is correct.
- Placement in `server.py` is right: after `_normalize_messages()` but before media extraction.
- Tests cover the unit helper (strip, idempotent, None/empty), Chat Completions path, and Responses path.

### Minor note

The Anthropic adapter in `anthropic_adapter.py` has its own inline regex (`re.sub(r"x-anthropic-billing-header:[^\n]*\n?", "", system_text)`) without `(?im)` flags or `^` anchor. Not blocking, but a future cleanup could have the adapter call `canonicalize_system_prompt()` instead of duplicating the pattern.

CI 9/9. LGTM.

waybarrios (approved):
all set for my end

PR #536, opened 2026-05-16
Preserve MLLM content part order

Thump604:
## Summary

This is the vllm-mlx analogue of the Qwen3.5 chat-template issue fixed in vllm-project/vllm#42660. vLLM fixed detection for templates that loop over `content` via `{% for item in content %}`; vllm-mlx has a different code path, but it was still rebuilding OpenAI multimodal content lists before template application.

Before this patch, `MLXMultimodalLM.chat` and `stream_chat` extracted text and media separately, then emitted all image/audio placeholders before a single combined text part. That loses the original request order for content such as `text -> image -> text`, which Qwen3.5-style templates can represent directly.

This patch:

- adds a shared message reconstruction helper for `chat` and `stream_chat`
- preserves OpenAI text/image/audio part order when building template messages
- keeps text-only user/system messages in the existing text-part shape
- keeps assistant messages as plain text
- keeps audio source collection on the existing `_collect_audio_inputs` path to avoid double collection

## Tests

- `python -m pytest tests/test_mllm.py tests/test_mllm_message_ordering.py tests/test_mllm_mtp_routing.py::test_batched_prepare_mllm_messages_converts_audio_url`
- `python -m py_compile vllm_mlx/models/mllm.py tests/test_mllm_message_ordering.py`
- `uvx ruff check vllm_mlx/models/mllm.py tests/test_mllm_message_ordering.py --select E,F,W --ignore E402,E501,E731,F811,F841`
- `uvx black --check vllm_mlx/models/mllm.py tests/test_mllm_message_ordering.py`
- `git diff --check`

## Not claimed

- This does not claim full Qwen3.5 vision qualification.
- This does not address unrelated model-loading, Metal stream, or dequantization failures.
- This does not change sampling, runtime defaults, or server routing.

### DIFF
--- a/tests/test_mllm_message_ordering.py
+++ b/tests/test_mllm_message_ordering.py
@@ -1,0 +1,110 @@
+from vllm_mlx.models.mllm import _build_mllm_chat_messages
+
+
+def test_mllm_chat_messages_preserve_text_only_shape():
+    messages = [
+        {"role": "system", "content": "You are concise."},
+        {"role": "user", "content": "Summarize the image."},
+        {"role": "assistant", "content": "Ready."},
+    ]
+
+    chat_messages = _build_mllm_chat_messages(
+        messages,
+        all_image_urls=[],
+        video_frame_counts={},
+    )
+
+    assert chat_messages == [
+        {
+            "role": "system",
+            "content": [
+                {
+                    "type": "text",
+                    "text": "You are concise.",
+                    "content": "You are concise.",
+                }
+            ],
+        },
+        {
+            "role": "user",
+            "content": [
+                {
+                    "type": "text",
+                    "text": "Summarize the image.",
+                    "content": "Summarize the image.",
+                }
+            ],
+        },
+        {"role": "assistant", "content": "Ready."},
+    ]
+
+
+def test_mllm_chat_messages_preserve_image_text_order():
+    all_image_urls = []
+    messages = [
+        {
+            "role": "user",
+            "content": [
+                {"type": "text", "text": "compare the left panel to "},
+                {"type": "image_url", "image_url": {"url": "/tmp/chart.png"}},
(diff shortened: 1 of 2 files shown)


janhilgard (approved):
Good refactor. Eliminates duplicated message-reconstruction logic between `chat()` and `stream_chat()` while fixing the content-part reordering bug. CI 9/9.

### What this fixes

The old code collected all image/audio placeholders and emitted them _before_ a single concatenated text part:
```
[image, image, audio, text("A B C")]
```
Templates like Qwen3.5 that iterate over `content` parts can represent interleaved `text → image → text` directly. Reordering broke that. Now the original part order is preserved:
```
[text("A"), image, text("B")]
```

### Code review

- `_build_mllm_chat_messages` is a clean shared entry point for both paths.
- Helper decomposition (`_normalize_content_part`, `_extract_media_url`, `_text_content_part`, `_append_ordered_mllm_content_part`) is well-scoped — each does one thing.
- Audio placeholder insertion without URL collection is correct: audio is already collected in the first pass via `_collect_audio_inputs`.
- Video frame expansion (`built_parts.extend({"type": "image"} for _ in range(video_frame_count))`) returns 0 to "consume" the budget — works correctly for single-video-per-message (the common case).
- Assistant messages remain plain text (not content arrays) — consistent with chat-template expectations.
- Tests cover the three key shapes: text-only, image+text interleaved, audio+text interleaved.

### Minor notes

1. `_append_ordered_mllm_content_part` returns `video_frame_count` (unchanged) for non-video types and `0` for video. This means if a message has two video parts, only the first gets frame expansion. This is fine for now (multi-video per message is exotic) but worth documenting.

2. The type annotations use `dict[str, str]` for `built_parts` items, but the actual items can have different value types (`{"type": "image"}` has no `text` key). A `dict[str, Any]` would be more precise, though this is cosmetic.

LGTM.

PR #537, opened 2026-05-16
fix(reasoning): respect enable_thinking=false from chat_template_kwargs

ursk:
_First contribution here — I'm running vllm-mlx on a Mac Studio M3 Ultra (96 GB) as a local Claude Code backend with Qwen 3.6 and Gemma 4. More fixes likely as I keep wiring this up._

## Summary

With `--reasoning-parser qwen3 --default-chat-template-kwargs '{"enable_thinking": false}'`, streamed `/v1/messages` responses route every model token into a `thinking` content block and leave `text` empty. Claude Code sees `result: ""` despite tokens being generated.

Cause: the reasoning-parser gate only consulted `request.enable_thinking`. The server-default surface (`--default-chat-template-kwargs`) lives in `chat_kwargs["chat_template_kwargs"]` and was never checked. With thinking disabled the template injects no `<think>`, so `BaseThinkingReasoningParser.extract_reasoning_streaming` falls into its implicit-mode default and classifies all deltas as reasoning.

Fix: new `_thinking_disabled(request, chat_kwargs)` helper consulted at the four reasoning-parser entry points (`/v1/messages` streaming + non-streaming, `/v1/chat/completions` streaming, `/v1/responses` streaming). Parser behavior when thinking is enabled is unchanged.

Reproducer:
```
# qwen-27b plist config -- reasoning parser + enable_thinking=false default
claude -p --model sonnet "Reply with HELLO."
# before: {"result": "", "output_tokens": 3, ...}
# after:  {"result": "HELLO", "output_tokens": 3, ...}
```

## Validation

- New tests in `tests/test_chat_template_kwargs.py` (+114 lines): 5 unit cases for `_thinking_disabled` and one `_stream_anthropic_messages` integration test asserting that with the parser eating everything as reasoning, the SSE body contains only `text_delta` events. The integration test **fails on the parent commit** with the exact buggy stream body and passes on this branch.
- `pytest tests/test_chat_template_kwargs.py tests/test_reasoning_parser.py tests/test_thinking_aware_processor.py tests/test_thinking_processor.py` — 169/169 pass.
- `ruff` and `black --check` clean on changed files.
- mypy not run locally (CI is 3.11, my local is 3.12); change adds one typed helper.
- End-to-end on the live Qwen 3.6-27B server: empty `result` → `"HELLO"`, stream goes from `thinking_delta` to `text_delta`.

## Merge readiness

- [x] I am not merging my own substantive PR.
- [ ] The latest CI run for this exact head commit is green. _(will confirm after push)_
- [x] If I rebased, resolved conflicts, or pushed after review, I waited for the new CI run to finish green. _(N/A on initial push)_
- [x] Actionable review comments are resolved in this PR or linked to follow-up issues before merge.
- [x] Any intentional follow-up debt is linked here and called out in the PR body. _(none)_

## Risk notes

Gate-only change at four entry points; when thinking is enabled, behavior is byte-identical to before. The implicit-mode default in `BaseThinkingReasoningParser` (`think_parser.py:185-189`) is intentionally left alone since other callers may legitimately rely on it for chat-template-injected `<think>`. No scheduler/memory/security surface touched.

### DIFF
--- a/tests/test_chat_template_kwargs.py
+++ b/tests/test_chat_template_kwargs.py
@@ -309,3 +309,76 @@ async def test_simple_engine_stream_generate_text_applies_chat_template_kwargs()
             ]
             is False
         )
+
+
+# Regression tests for the reasoning-parser bypass when enable_thinking=False.
+# Server started with --default-chat-template-kwargs '{"enable_thinking": false}'
+# plus a Qwen3 reasoning parser used to route every token into a `thinking`
+# block and leave `text` empty, so Claude Code saw `result: ""`.
+
+
+@pytest.mark.parametrize(
+    "enable_thinking, ctk, expected",
+    [
+        (None, None, False),  # nothing set
+        (False, None, True),  # request-level flag
+        (None, {"enable_thinking": False}, True),  # server default surface
+        (True, {"enable_thinking": True}, False),  # explicit True
+        (None, {}, False),  # empty kwargs
+    ],
+)
+def test_thinking_disabled_helper(enable_thinking, ctk, expected):
+    request = SimpleNamespace(enable_thinking=enable_thinking)
+    chat_kwargs = {"chat_template_kwargs": ctk} if ctk is not None else {}
+    assert srv._thinking_disabled(request, chat_kwargs) is expected
+
+
+@pytest.mark.anyio
+async def test_stream_anthropic_skips_reasoning_parser_when_thinking_disabled():
+    from vllm_mlx.reasoning import DeltaMessage
+
+    class _EatsEverythingAsReasoning:
+        # Mimics BaseThinkingReasoningParser's implicit-mode default.
+        def reset_state(self):
+            pass
+
+        def extract_reasoning_streaming(self, prev, cur, delta):
+            return DeltaMessage(reasoning=delta)
+
+    async def fake_stream_chat(messages, **kwargs):
+        for piece in ("HEL", "LO"):
+            yield SimpleNamespace(new_text=piece, prompt_tokens=4, completion_tokens=1)
+
+    engine = MagicMock(stream_chat=fake_stream_chat)
+    msgs = [{"role": "user", "content": "Say HELLO"}]
+    openai_request = srv.ChatCompletionRequest(
+        model="test-model", messages=[srv.Message(**msgs[0])], max_tokens=8
+    )
+    anthropic_request = srv.AnthropicRequest(
+        model="test-model", max_tokens=8, messages=msgs
+    )
+    prepared = srv.PreparedChatInvocation(
+        messages=msgs,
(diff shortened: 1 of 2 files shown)


janhilgard (approved):
## Review Summary

Clean, well-scoped fix for a real bug affecting Claude Code users with `--default-chat-template-kwargs '{"enable_thinking": false}'`. The problem statement, root cause, and fix are all clearly documented. CI is green across all 9 checks.

### What the PR does

When a server is configured with `--reasoning-parser qwen3 --default-chat-template-kwargs '{"enable_thinking": false}'`, the chat template does not inject `<think>` into the prompt. However, `BaseThinkingReasoningParser.extract_reasoning_streaming` defaults to "implicit thinking mode" when it sees no `<think>` or `</think>` tags — classifying all model output as reasoning and leaving the text content block empty.

The fix introduces a `_thinking_disabled(request, chat_kwargs)` helper that checks both the request-level `enable_thinking` flag and the resolved `chat_template_kwargs`, and uses it to bypass the reasoning parser at all four streaming/non-streaming entry points.

### Code quality

- The helper function is clean, well-documented, and uses `getattr` for safe attribute access.
- Tests are thorough: 5 parametrized unit cases for the helper plus an integration test with a mock parser that reproduces the exact bug.
- The integration test properly saves/restores module-level globals in a finally block.
- Behavior when thinking IS enabled is unchanged (gate-only change).

### Minor observations (non-blocking)

1. **Precedence edge case in `_thinking_disabled`**: If a client explicitly sends `enable_thinking: true` at the request level but the server default has `chat_template_kwargs: {"enable_thinking": false}`, the function falls through the first check (request-level True is not False) and then returns True from the `ctk` check — incorrectly treating thinking as disabled. In practice this is unlikely because:
   - Clients using `enable_thinking: true` would typically also set it in their `chat_template_kwargs`
   - The `_prepare_chat_invocation` code (lines 662-687) actively reconciles these two surfaces when JSON logits processors are involved

   If you wanted to be defensive, adding `if getattr(request, "enable_thinking", None) is True: return False` before the `ctk` check would handle this, but it is not a regression introduced by this PR — the pre-existing `request.enable_thinking is not False` check at the old `stream_chat_completion` call site had the same gap.

2. **Non-streaming `/v1/responses` path (line 2168)**: This call site uses `extract_reasoning` (non-streaming), NOT `extract_reasoning_streaming`. The non-streaming `extract_reasoning` method correctly returns `(None, model_output)` when no think tags are found, so it does not suffer from the implicit-mode bug. This is why it is correctly excluded from the fix. Just noting this for reviewers who might wonder about the asymmetry.

### Verdict

Approving. The fix is correct, well-tested, narrowly scoped, and solves a real problem. The precedence edge case noted above is pre-existing and does not affect the primary use case (server-default `enable_thinking=false` without per-request override).

waybarrios:
Just checked. Pretty good.

PR #539, opened 2026-05-16
Fail invalid response_format output before returning success

### LINKED ISSUE #538: response_format can return 200 with invalid JSON content
### Repro
Non-streaming `/v1/chat/completions` with `response_format={"type":"json_object"}` can reach the server response-format post-processing path with model text that contains no valid JSON after reasoning/tool parsing.

Exact upstream code path compared: `vllm_mlx/server.py` on `main` at `6d55631`, inside `create_chat_completion`, after `parse_json_output(json_input, response_format)`.

### Observed behavior
When `parse_json_output` returns `is_valid=False`, the server logs `JSON validation failed` but still continues to build a successful chat completion response. That can return HTTP 200 with non-JSON text in `message.content` even though `response_format` requested JSON.

### Expected behavior
A request that asks for structured JSON should not receive HTTP 200 with non-JSON content. If the final output cannot be parsed or validated, the server should fail the request with a clear error.

### Minimal patch/design ask
Add a server-boundary helper that canonicalizes valid parsed JSON and raises `HTTPException(422)` when the final output does not satisfy `response_format`. Add unit coverage for the valid and invalid boundary cases.

### Not claimed
This is not claiming constrained decoding always fails. It is not changing model sampling. It is not changing streaming behavior in this patch. It only addresses the non-streaming server boundary so invalid structured output cannot be returned as a successful response.

Thump604:
## Summary
- add a shared response-format boundary helper that returns canonical JSON only after `parse_json_output` validates it
- have the non-streaming server boundary raise HTTP 422 instead of returning HTTP 200 with non-JSON `message.content` when `response_format` output is invalid
- force EOS once the JSON logits processor has already produced a complete JSON value
- keep `response_format` constraints in direct control of generation instead of hiding them behind thinking-state handling
- keep streaming `response_format` JSON/JSON-schema deltas on the assistant `content` channel when a thinking parser classifies direct JSON as reasoning
- add structured-output, constrained-decoding, and streaming content-channel tests for the hard-fail, EOS, thinking, and streaming interactions

Fixes #538.

## Local repro
The existing upstream path in `vllm_mlx/server.py` calls `parse_json_output(json_input, response_format)`. When `is_valid=False`, it logs `JSON validation failed` and then continues to return a normal `ChatCompletionResponse`.

The first focused unit repro covers the shared helper: `apply_response_format_or_error("I will reason first...", {"type": "json_object"})` raises `InvalidResponseFormatOutput`, and the server translates that into HTTP 422 at the API boundary.

A second local Runtime repro showed constrained JSON output could parse correctly but still continue until the output cap because `JSONSchemaLogitsProcessor` did not force EOS once the suffix was already complete JSON.

A third local Runtime repro showed thinking-enabled `response_format` could allow direct JSON emissions to bypass the JSON processor when the processor was wrapped behind thinking-state handling. The upstream branch had the same `_ThinkingAwareLogitsProcessor` path in both OpenAI and Anthropic preparation.

A fourth local Runtime repro showed a long streaming `response_format={"type":"json_schema"}` request on a Qwen thinking mode emitted valid JSON through the streaming reasoning channel instead of assistant content. The local certification artifact is `/opt/ai-runtime/run/qwen-response-format-streaming-certification/20260517T052855Z/summary.json`: HTTP 200, `finish_reason=stop`, JSON/schema valid, `content_length=2243`, `reasoning_length=0` after the content-channel fix.

- repo: `waybarrios/vllm-mlx`
- base: `main` first checked at `6d55631`; PR branch rebased onto `c2d3aec`; streaming follow-up compared against `upstream/main` at `c2d3aec4be51f3b7bc31b7b5d489ced41b1f8fbf`
- paths:
  - `vllm_mlx/server.py` non-streaming `create_chat_completion` response-format post-processing
  - `vllm_mlx/server.py` OpenAI and Anthropic response-format logits-processor attachment
  - `vllm_mlx/server.py::stream_chat_completion` reasoning-parser delta handling and streaming response-format fence stripping
  - `vllm_mlx/api/tool_calling.py` structured-output helpers
  - `vllm_mlx/constrained/json_schema_processor.py` EOS and mask handling

## Observed behavior
Invalid structured output could be returned as a successful response after only logging a warning.

A complete JSON value could remain open for more generation instead of stopping cleanly.

When thinking was explicitly enabled with `response_format`, JSON constraints could wait behind thinking-state handling and fail to own generation from the first token.

In streaming, a thinking parser could classify valid direct JSON as `reasoning_content`, leaving `delta.content` empty even though the request contract asked for `response_format` final assistant content.

## Expected behavior
If `response_format` is requested and final non-streaming output does not parse or validate, the API should fail the request instead of returning invalid content as success.

Once constrained decoding has produced a complete JSON value, EOS should be the only allowed continuation.

For `response_format`, the JSON processor should constrain generated content directly. This patch disables thinking template flags for that request path rather than wrapping the JSON processor behind thinking detection.

For streaming `response_format`, JSON/JSON-schema deltas that a reasoning parser emits as reasoning should be promoted back to assistant content unless ordinary content is already present.

## Tests
- `UV_PYTHON=/opt/homebrew/bin/python3.12 uv run --extra dev python -m pytest -q tests/test_constrained_decoding.py tests/test_thinking_aware_processor.py tests/test_structured_output.py`
- `UV_PYTHON=/opt/homebrew/bin/python3.12 uv run --extra dev python -m pytest tests/test_mcp_security.py tests/test_structured_output.py tests/test_reasoning_parser.py tests/test_tool_parsers.py tests/test_streaming_json_encoder.py tests/test_native_tool_format.py tests/test_memory_cache.py tests/test_prefix_cache.py tests/test_mllm_cache.py tests/test_api_models.py tests/test_api_utils.py tests/test_request.py tests/test_anthropic_models.py tests/test_anthropic_adapter.py tests/test_harmony_parsers.py tests/test_endpoint_model_policies.py tests/test_gemma4_openai_format.py tests/test_gemma4_streaming_edge.py tests/test_gemma4_tool_parser.py tests/test_minimax_tool_calling.py tests/test_qwen3_xml_parser.py tests/test_qwen3_xml_registration.py tests/test_qwen3_xml_streaming_chunks.py tests/test_streaming_fence_stripper.py tests/test_streaming_think_router.py tests/test_streaming_tool_filter.py tests/test_tool_call_promotion.py -q -k "not Integration and not InjectJson and not TestMLXMultimodalLMCache"`
- `python -m py_compile vllm_mlx/server.py vllm_mlx/api/tool_calling.py vllm_mlx/constrained/json_schema_processor.py tests/test_structured_output.py tests/test_constrained_decoding.py tests/test_thinking_aware_processor.py`
- `UV_PYTHON=/opt/homebrew/bin/python3.12 uv run --extra dev ruff check vllm_mlx/server.py vllm_mlx/api/tool_calling.py vllm_mlx/constrained/json_schema_processor.py tests/test_structured_output.py tests/test_constrained_decoding.py tests/test_thinking_aware_processor.py`
- `UV_PYTHON=/opt/homebrew/bin/python3.12 uv run --extra dev black --check vllm_mlx/server.py vllm_mlx/api/tool_calling.py vllm_mlx/constrained/json_schema_processor.py tests/test_structured_output.py tests/test_constrained_decoding.py tests/test_thinking_aware_processor.py`
- `.venv/bin/python -m pytest tests/test_server.py::TestStreamChatCompletion -q`
- `.venv/bin/ruff check vllm_mlx/server.py tests/test_server.py`
- `git diff --check`

Outbound local gates for the streaming follow-up:
- agent-slop-lint touched-file before/after comparison on `vllm_mlx/server.py` and `tests/test_server.py`: 75 violations before, 75 after; no regression
- Desloppify touched-file before/after comparison on `vllm_mlx/server.py` and `tests/test_server.py`: 56 open issues before, 56 after; objective 77.6 before and after; no regression
- `bin/lint-upstream-claims --root /tmp/outbound-vllm-mlx-539.VScyyJ vllm_mlx/server.py tests/test_server.py`

## Not claimed
- does not change model sampling defaults
- does not claim `tools` plus `response_format` semantics
- does not claim model-quality results
- does not claim every model/parser emits direct JSON through the same streaming path

### DIFF
--- a/tests/test_constrained_decoding.py
+++ b/tests/test_constrained_decoding.py
@@ -246,6 +246,22 @@ class TestProcessorMask:
         result = processor(tokens, logits)
         assert result.shape == logits.shape
 
+    def test_complete_json_forces_eos(self):
+        """Once the suffix is complete JSON, only EOS should remain allowed."""
+        import mlx.core as mx
+        import numpy as np
+
+        processor, tok = self._make_processor()
+        processor._prompt_len = 0
+        tokens = mx.array(tok.encode('{"a":1}'), dtype=mx.int32)
+        logits = mx.zeros((tok.vocab_size,))
+
+        masked = processor(tokens, logits)
+        arr = np.array(masked)
+        finite = {i for i, value in enumerate(arr) if np.isfinite(value)}
+
+        assert finite == {tok.eos_token_id}
+
 
 # ---------------------------------------------------------------------------
 # Anthropic adapter — response_format propagation.
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -2140,6 +2140,83 @@ class TestStreamChatCompletion:
         assert payloads[2]["choices"][0]["finish_reason"] == "stop"
 
     @pytest.mark.anyio
+    async def test_response_format_stream_promotes_reasoning_json_to_content(
+        self, monkeypatch
+    ):
+        """response_format JSON belongs on the streaming content channel."""
+        from vllm_mlx.engine.base import GenerationOutput
+        from vllm_mlx.reasoning import DeltaMessage
+        from vllm_mlx.server import (
+            ChatCompletionRequest,
+            Message,
+            stream_chat_completion,
+        )
+        import vllm_mlx.server as server
+
+        class FakeEngine:
+            model_name = "fake-engine"
+
+            async def stream_chat(self, messages, **kwargs):
+                yield GenerationOutput(
+                    text="",
+                    new_text='{"ok": true}',
+                    finished=True,
+                    finish_reason="stop",
+                    prompt_tokens=4,
+                    completion_tokens=3,
+                )
+
+        class ReasoningOnlyParser:
+            def reset_state(self):
+                pass
+
+            def extract_reasoning_streaming(
+                self, previous_text, current_text, delta_text
+            ):
+                return DeltaMessage(reasoning=delta_text)
+
+        monkeypatch.setattr(server, "_model_name", "served-model")
+        monkeypatch.setattr(server, "_reasoning_parser", ReasoningOnlyParser())
+        monkeypatch.setattr(server, "_enable_auto_tool_choice", False)
+        monkeypatch.setattr(server, "_tool_call_parser", None)
+        monkeypatch.setattr(server, "_tool_parser_instance", None)
+
+        request = ChatCompletionRequest(
+            model="served-model",
+            messages=[Message(role="user", content="return json")],
+            stream=True,
+            response_format={
+                "type": "json_schema",
+                "json_schema": {
+                    "name": "ok",
+                    "schema": {
+                        "type": "object",
+                        "properties": {"ok": {"type": "boolean"}},
+                        "required": ["ok"],
+                    },
+                },
+            },
+        )
+
+        chunks = [
+            chunk
+            async for chunk in stream_chat_completion(
+                FakeEngine(), request.messages, request
+            )
+        ]
+
+        payloads = [
+            json.loads(chunk.removeprefix("data: ").strip())
+            for chunk in chunks
+            if chunk != "data: [DONE]\n\n"
+        ]
+
+        delta = payloads[1]["choices"][0]["delta"]
+        assert json.loads(delta["content"]) == {"ok": True}
+        assert "reasoning_content" not in delta
+        assert payloads[1]["choices"][0]["finish_reason"] == "stop"
+
+    @pytest.mark.anyio
     async def test_streaming_chat_no_stream_thread_error_after_residency_preload(
         self, monkeypatch, caplog
     ):
--- a/tests/test_structured_output.py
+++ b/tests/test_structured_output.py
@@ -6,14 +6,17 @@ Tests the JSON parsing, validation, and response_format handling.
 """
 
 import json
+
 import pytest
+from vllm_mlx.api.models import ResponseFormat, ResponseFormatJsonSchema
 from vllm_mlx.api.tool_calling import (
-    validate_json_schema,
+    InvalidResponseFormatOutput,
+    apply_response_format_or_error,
+    build_json_system_prompt,
     extract_json_from_text,
     parse_json_output,
-    build_json_system_prompt,
+    validate_json_schema,
 )
-from vllm_mlx.api.models import ResponseFormat, ResponseFormatJsonSchema
 
 
 class TestValidateJsonSchema:
@@ -399,6 +402,27 @@ class TestParseJsonOutput:
         assert is_valid is True
 
 
+class TestServerResponseFormatBoundary:
+    """Tests for the server boundary after response_format parsing."""
+
+    def test_invalid_response_format_output_fails_before_success_response(self):
+        with pytest.raises(InvalidResponseFormatOutput) as excinfo:
+            apply_response_format_or_error(
+                "I will reason first and then maybe return JSON.",
+                {"type": "json_object"},
+            )
+
+        assert "Failed to extract valid JSON" in excinfo.value.message
+
+    def test_valid_response_format_output_returns_canonical_json(self):
+        content = apply_response_format_or_error(
+            'prefix {"ok": true, "count": 2} suffix',
+            {"type": "json_object"},
+        )
+
+        assert json.loads(content) == {"ok": True, "count": 2}
+
+
 class TestBuildJsonSystemPrompt:
     """Tests for build_json_system_prompt function."""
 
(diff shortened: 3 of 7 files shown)


janhilgard (approved):
## Review Summary

This PR addresses four distinct but related issues with `response_format` handling:

1. **Hard-fail on invalid output** (HTTP 422 instead of silent 200 with bad content)
2. **Force EOS after complete JSON** (stop generation instead of running to max_tokens)
3. **Disable thinking when response_format is active** (prevent constraint bypass)
4. **Promote streaming JSON from reasoning to content channel**

All four are genuine bugs with clear user-facing impact. The PR is well-structured, thorough, and includes good test coverage.

### Correctness

**1. `apply_response_format_or_error` and `InvalidResponseFormatOutput` (tool_calling.py)**

The new shared boundary helper correctly delegates to `parse_json_output` and raises on invalid output. The three-way return logic is correct:
- `is_valid and parsed_json is not None` -> canonical JSON string
- `is_valid and parsed_json is None` -> pass through (text format)
- `not is_valid` -> raise

One minor observation: when `is_valid=True` and `parsed_json is None`, this returns the raw `text` unchanged. This path triggers for `format_type == "text"` in `parse_json_output`, which is correct -- text format has no JSON to canonicalize.

**2. `_complete_json_eos_logits` and EOS forcing (json_schema_processor.py)**

The EOS-forcing logic after complete JSON is correct and addresses a real performance issue. The existing code already had an EOS guard that *prevented* EOS when JSON was incomplete (line ~791 on main), but it never *forced* EOS when JSON *was* complete. This fills that gap.

The extracted helper functions (`_eos_logits`, `_eos_logits_or_original`, `_complete_json_eos_logits`) are clean factorings of the existing disabled-path code. The refactoring of the disabled path to use `_eos_logits_or_original` is correct and reduces duplication.

**3. `_attach_response_format_logits_processor` (server.py)**

This replaces the previous dual-path logic (thinking-aware wrapper vs. thinking suppression) with always-suppress. The key insight is correct: wrapping the JSON processor behind `_ThinkingAwareLogitsProcessor` means the constraint is inactive during `<think>` generation, but models can emit JSON directly without a `<think>` block, bypassing the constraint entirely.

The old code had:
```python
if _reasoning_parser and effective_thinking:
    # Wrap in ThinkingAwareLogitsProcessor
else:
    # Suppress thinking
```

The new code always suppresses thinking and attaches the processor directly. This is a sound design decision -- if the user asked for `response_format`, they want JSON, not reasoning tokens.

**4. `_promote_streaming_response_format_delta` (server.py)**

Clean and correct. When a reasoning parser classifies direct JSON as reasoning (because there's no `</think>` marker), this swaps it to the content channel for `json_object` and `json_schema` requests. The guard `if content or not reasoning` ensures it only acts when content is empty and reasoning is present.

### Edge Cases

- **`response_format={"type": "text"}`**: `_response_format_type` returns `"text"`, which is not in the promotion set `("json_object", "json_schema")` -- correct, no promotion happens.
- **`response_format=None`**: `_response_format_type` returns `None` -- correct, no promotion.
- **Streaming with both content and reasoning present**: The `if content or not reasoning` guard correctly short-circuits -- no promotion when content already exists.
- **Non-streaming path**: Uses `_apply_response_format_or_raise` which wraps the error as HTTP 422 -- correct status code for "syntactically valid request, semantically invalid output".

### Performance

- The `_complete_json_eos_logits` check runs on every `__call__` invocation of the logits processor. The `_suffix_is_complete_json` method uses cached brace/bracket depth as a fast pre-check (returns False in O(1) for ~99% of calls when brackets are unbalanced), so the actual `json.loads` only fires when depths are zero. This is efficient.
- The module-level helper functions (`_eos_logits`, etc.) avoid method dispatch overhead vs. inline code -- negligible but marginally better.

### Code Style

- Clean separation of concerns: validation logic in `tool_calling.py`, HTTP boundary in `server.py`, generation constraint in `json_schema_processor.py`.
- The extracted helper functions in `json_schema_processor.py` are well-named and single-purpose.
- The `_strip_backslash_before_unicode` call moved into `_apply_response_format_or_raise` after `apply_response_format_or_error` returns -- this is correct, as the strip operates on the serialized JSON string.
- Tests cover all four behavioral changes with focused unit tests.

### Minor Observations (non-blocking)

1. `_apply_response_format_or_raise` catches `InvalidResponseFormatOutput` and re-raises as `HTTPException`, but then the caller in `create_chat_completion` catches that same `HTTPException` to log and re-raise. This double-catch pattern works but adds a layer of indirection. It is acceptable since `_apply_response_format_or_raise` is a reusable boundary function for both OpenAI and Anthropic paths.

2. The `_complete_json_eos_logits` function takes `is_complete_json` and `build_allow_mask` as parameters rather than being a method on the class. This is a reasonable design choice for testability, though it does mean the function signature is somewhat wide (5 parameters).

### CI Status

All 9 CI checks pass: lint, type-check, test-matrix (3.10, 3.11, 3.12, 3.13), test-apple-silicon (3.11, 3.13), tests.

### Verdict

Approved. The four fixes are correct, well-tested, and address real user-facing issues. The design choice to always suppress thinking for `response_format` requests is sound and eliminates a class of bypass bugs.

PR #540, opened 2026-05-16
Fail fast on occupied SimpleEngine serialized route

### LINKED ISSUE #495: Track admission-control invariant for serialized TextModel-direct routes

Thump604:
## Summary
- add a small `EngineBusy` exception for serialized engine admission failures
- make `SimpleEngine` fail fast by default when its serialized MLX lock is already occupied, with `VLLM_MLX_SIMPLE_ENGINE_LOCK_ADMISSION=wait` as an explicit compatibility mode
- track active serialized request metadata and expose lock/admission stats in `get_stats()`
- translate non-streaming completion/chat `EngineBusy` failures into retryable HTTP 503 responses with `error=text_generation_busy`

Fixes #495.

## Local repro
`SimpleEngine` serializes MLX work through `_generation_lock`. Before this patch, a second request entering the serialized route waited behind that lock, which can create request pileup under concurrent traffic. The focused regression starts one serialized request, then verifies a second serialized request fails immediately with `EngineBusy` and includes the active holder in the error text.

- repo: `waybarrios/vllm-mlx`
- base: current `main` at `c2d3aec`
- paths: `vllm_mlx/engine/simple.py` `_run_blocking_serialized` / `stream_generate`, `vllm_mlx/server.py` non-streaming completion/chat calls, and `vllm_mlx/engine/base.py`

## Observed behavior
The serialized SimpleEngine path had a lock, but no default fail-fast admission result and no active holder metadata for diagnosing the request currently occupying the route.

## Expected behavior
When the route is already occupied, a second request should get a retryable busy result instead of silently queueing behind the serialized MLX lock. Operators should be able to see the active holder, waiters, and busy rejection count.

## Tests
- `UV_PYTHON=/opt/homebrew/bin/python3.12 uv run --extra dev python -m pytest -q tests/test_simple_engine.py`
- `python -m py_compile vllm_mlx/engine/base.py vllm_mlx/engine/simple.py vllm_mlx/server.py tests/test_simple_engine.py`
- `uv run --with ruff ruff check vllm_mlx/engine/base.py vllm_mlx/engine/simple.py vllm_mlx/server.py tests/test_simple_engine.py --select E,F,W --ignore E402,E501,E731,F811,F841`
- `black --check vllm_mlx/engine/base.py vllm_mlx/engine/simple.py vllm_mlx/server.py tests/test_simple_engine.py`
- `git diff --check`

## Not claimed
- does not claim this is continuous batching
- does not change model sampling
- does not remove the serialized lock; it changes default admission behavior when that lock is occupied
- does not address streaming HTTP response semantics beyond the engine-side busy invariant

### DIFF
--- a/tests/test_simple_engine.py
+++ b/tests/test_simple_engine.py
@@ -104,6 +104,7 @@ class TestSimpleEngineConcurrency:
             engine = SimpleEngine("test-model")
             engine._model = mock_model
             engine._loaded = True
+            engine._generation_lock_admission = "wait"
 
             # Launch multiple concurrent generate calls
             tasks = [
@@ -128,6 +129,7 @@ class TestSimpleEngineConcurrency:
             engine = SimpleEngine("test-model")
             engine._model = mock_llm_model
             engine._loaded = True
+            engine._generation_lock_admission = "wait"
 
             # Launch multiple concurrent chat calls
             tasks = [
@@ -144,6 +146,76 @@ class TestSimpleEngineConcurrency:
                 f"Expected max concurrent to be 1, but got {mock_llm_model._max_concurrent}. "
                 "The lock is not working correctly."
             )
+
+    @pytest.mark.anyio
+    async def test_default_admission_rejects_second_serialized_request(self):
+        """Default SimpleEngine admission fails fast instead of queueing."""
+        from vllm_mlx.engine.base import EngineBusy
+        from vllm_mlx.engine.simple import SimpleEngine
+
+        with patch("vllm_mlx.engine.simple.is_mllm_model", return_value=False):
+            engine = SimpleEngine("test-model")
+            engine._loaded = True
+
+            started = threading.Event()
+            release = threading.Event()
+
+            def slow_call():
+                started.set()
+                release.wait(timeout=1.0)
+                return "ok"
+
+            first = asyncio.create_task(
+                engine._run_blocking_serialized(
+                    slow_call,
+                    request_id="first-serialized-request",
+                )
+            )
+            await asyncio.to_thread(started.wait, 1.0)
+
+            with pytest.raises(EngineBusy) as excinfo:
+                await engine._run_blocking_serialized(
+                    lambda: "late",
+                    request_id="second-serialized-request",
+                )
+
+            assert "text_generation_busy" == excinfo.value.code
+            assert "active=first-serialized-request" in str(excinfo.value)
+            assert engine._generation_busy_rejections == 1
+
+            release.set()
+            await first
+
+    @pytest.mark.anyio
+    async def test_blocking_serialized_tracks_active_request(self):
+        """Busy errors include the current blocking serialized holder."""
+        from vllm_mlx.engine.simple import SimpleEngine
+
+        with patch("vllm_mlx.engine.simple.is_mllm_model", return_value=False):
+            engine = SimpleEngine("test-model")
+            engine._loaded = True
+
+            def slow_call():
+                import time
+
+                time.sleep(0.05)
+                return "ok"
+
+            first = asyncio.create_task(
+                engine._run_blocking_serialized(
+                    slow_call,
+                    request_id="probe-active-holder",
+                )
+            )
+            for _ in range(100):
+                if engine._generation_lock.locked():
+                    break
+                await asyncio.sleep(0.001)
+
+            assert "probe-active-holder" in engine._generation_lock_holder_summary()
+
+            await first
+            assert engine._generation_lock_holder_summary() == "none"
 
     async def test_chat_with_tools_aggregates_streaming_path(self, mock_llm_model):
         """Tool-enabled non-stream chat should use the streaming path."""
@@ -1142,6 +1214,7 @@ class TestSimpleEngineConcurrency:
             engine = SimpleEngine("test-model")
             engine._model = mock_model
             engine._loaded = True
+            engine._generation_lock_admission = "wait"
 
             # Test that stream_generate acquires the lock
             # by checking if it blocks when lock is already held
@@ -2027,6 +2100,7 @@ class TestSimpleEngineConcurrency:
             engine = SimpleEngine("test-model")
             engine._model = mock_llm_model
             engine._loaded = True
+            engine._generation_lock_admission = "wait"
 
             task1 = asyncio.create_task(
                 engine.chat(
--- a/vllm_mlx/engine/base.py
+++ b/vllm_mlx/engine/base.py
@@ -37,6 +37,12 @@ class GenerationOutput:
     mtp_accepted: int = 0
 
 
+class EngineBusy(RuntimeError):
+    """Raised when a serialized engine route is already serving a request."""
+
+    code = "text_generation_busy"
+
+
 @contextmanager
 def suspend_cancellation():
     """Temporarily clear task cancellation so cleanup can finish deterministically."""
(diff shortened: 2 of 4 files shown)


janhilgard (approved):
## Code Review

Overall this is a well-scoped and cleanly implemented PR. The fail-fast admission control for SimpleEngine's serialized route is a sensible default — silently queueing behind a single-threaded MLX lock under concurrent traffic is a real footgun. The code is well-structured, tests cover the critical contention scenario, CI is green, and the PR description is thorough.

### Positive highlights

- **`EngineBusy` exception** in `base.py` is minimal and clean — a `code` class attribute for machine-readable error identity is the right pattern.
- **`_acquire_generation_slot` as an async context manager** is a good abstraction that centralizes admission logic.
- **Environment variable opt-in for `wait` mode** (`VLLM_MLX_SIMPLE_ENGINE_LOCK_ADMISSION`) preserves backward compatibility for users who rely on the queuing behavior.
- **`get_stats()` exposure** of lock state, admission mode, busy rejections, active requests, and waiters is excellent for operational debugging.
- **HTTP 503 with structured error body** is the correct status code for a retryable busy condition.
- **Tests** are well-written: the `test_default_admission_rejects_second_serialized_request` test properly uses threading events to deterministically hold the lock, and `test_blocking_serialized_tracks_active_request` verifies observability.
- **Existing tests updated** to set `engine._generation_lock_admission = "wait"` so they continue to verify the serialization behavior.

### Minor observations (non-blocking)

1. **`stream_chat` MLLM text-only fallback (line ~931)** still acquires `self._generation_lock` directly instead of going through `_acquire_generation_slot`. This is a pre-existing path not modified by this PR, so it's not a regression, but it does mean that code path bypasses fail-fast admission. Worth tracking for a follow-up to ensure all serialized paths get consistent admission behavior.

2. **`chat()` doesn't pass `request_id` to `_run_blocking_serialized`**, so those calls will get the fallback `f"simple-{id(func):x}"` identifier. This is fine functionally (the fallback is deterministic), but for operational debugging the hex address of the function object is less informative than a proper request ID. Could be improved in a follow-up.

3. **Streaming endpoints** (`stream_completion`, `stream_chat_completion`) don't catch `EngineBusy` at the HTTP layer. The PR description explicitly notes "does not address streaming HTTP response semantics beyond the engine-side busy invariant" — this is a reasonable scope boundary since streaming SSE connections have already started the response when the error would surface, so it's a different problem. Just noting for completeness.

4. **`_generation_waiters` counter**: There's a subtle edge case in `_acquire_generation_slot` — between the `locked()` check and `self._generation_waiters += 1`, another coroutine could release the lock. Since this is single-threaded asyncio (no preemption between awaits), this is actually safe: the `locked()` check and the `+= 1` happen in the same synchronous block before any `await`. Just wanted to confirm I traced the concurrency model correctly — it's fine.

5. **`stop()` doesn't reset** `_active_requests`, `_generation_waiters`, or `_generation_busy_rejections`. This is arguably fine since stop/start creates a fresh engine in practice, but if `stop()` is meant to be a clean teardown, resetting these counters would be more defensive.

### Verdict

The PR is well-implemented, correctly scoped, and all CI checks pass. The observations above are all minor and non-blocking — appropriate for follow-up work rather than blocking this PR.

**Approved.**

Timotej-Labsky:
Heads-up on a bug in the merged version: `__init__` reads and validates `VLLM_MLX_SIMPLE_ENGINE_LOCK_ADMISSION`, then unconditionally re-assigns `fail_fast` on the next line — so `wait` mode is silently impossible and the validation/warning is dead code:

```python
if self._generation_lock_admission not in {"fail_fast", "wait"}:
    logger.warning(...)
self._generation_lock_admission = "fail_fast"   # <- always runs, clobbers "wait"
```

The fix is moving the fallback inside the invalid-value branch. I have a one-line fix + regression test ready on a branch ([TimotejLabsky/vllm-mlx@fix/admission-env-respected](https://github.com/TimotejLabsky/vllm-mlx/tree/fix/admission-env-respected)) but PR creation appears restricted to collaborators right now — happy to open it if that's lifted.

Two related observations from running this in production (we carry a port of #540 in our fork, see [PATCHES.md #15](https://github.com/TimotejLabsky/vllm-mlx/blob/main/PATCHES.md)):
- Agent clients (e.g. opencode via LiteLLM) fire a title request and the main completion concurrently against the same model; with `fail_fast` one of every such pair 503s, so `wait` actually working matters — and is arguably the safer default.
- Streaming routes can't turn a mid-stream `EngineBusy` into a clean 503 (200 + SSE headers already sent); a pre-stream admission probe before building the `StreamingResponse` closes that gap.

Thump604:
Thanks for catching this. I reproduced the `wait` clobber on current `main` with a focused regression: `VLLM_MLX_SIMPLE_ENGINE_LOCK_ADMISSION=wait` still reported `generation_lock.admission=fail_fast` because the fallback assignment ran unconditionally after validation.

Opened the narrow fix here: #615. It moves the fallback assignment into the invalid-value branch and adds the regression test. I left the streaming pre-admission probe out of that PR so the scope stays limited to restoring the existing env contract.

PR #541, opened 2026-05-16
feat: multi-slot LRU for system-KV cache + hit-ratio counters (follow-up to #523)

vinayvobbili:
Follow-up to #523. cc @Thump604 — touches the `SimpleEngine` stream/cache paths you own; would value your read.

## Problem

#523 introduced a single-slot system-prompt KV snapshot. Works on the validation pattern it shipped with (2 turns, same system prefix: 100s → 7s). Regresses to zero benefit the moment a client uses more than one system prefix in a session.

The case we hit: Claude Code's main-agent + sub-agent dispatch pattern interleaves two distinct system prefixes within one session:

- Main agent: ~28K tokens (system prompt + full toolset)
- Sub-agents (Explore / Plan / general-purpose): ~6–7K tokens

Every sub-agent dispatch evicts the main snapshot; every return to main re-prefills 28K cold. On Qwen3-Coder-30B-A3B that's ~140s TTFT per turn, which trips the server's 300s request timeout before the end-of-stream STORE runs, so the snapshot never lands. Cache stays empty across the entire session.

### 1. Multi-slot LRU snapshot

Replace the 3 single-slot instance vars (`_system_kv_snapshot`, `_system_kv_hash`, `_system_kv_token_count`) with an `OrderedDict[str, tuple[list, int]]` LRU keyed by system-prefix hash.

- Capacity via `VLLM_MLX_SYSTEM_KV_SLOTS` env var, default **4** (covers main + 2–3 sub-agent variants with headroom). `=1` restores #523 behavior exactly.
- LRU `move_to_end()` / `popitem(last=False)` happen inside the serialized worker (`_run_blocking_serialized`), so the existing `_generation_lock` already prevents concurrent mutation.
- Gate read uses `dict.get()` returning an immutable `(snapshot, count)` tuple under the GIL — a concurrent store can't desync the snapshot from the hash that decided HIT.
- `EVICTED` log line on overflow; `mx.clear_cache()` after eviction to reclaim Metal heap.
- Applied to both `stream_chat` (the path Claude Code uses) and the older non-stream text path.

### 2. Counters + `/v1/cache/stats` exposure

Add `hits` / `misses` / `stores` / `evictions` counters to `SimpleEngine`, incremented only from inside the serialized worker (single writer, no locks). Exposed via a new `get_cache_stats()` method; the existing `/v1/cache/stats` route already picks it up via `hasattr(_engine, "get_cache_stats")`.

```json
{
  "engine_cache": {
    "system_kv_cache": {
      "capacity": 4,
      "in_use": 2,
      "counters": {
        "hits": 3, "misses": 2, "stores": 2, "evictions": 0,
        "hit_ratio": 0.6
      }
    }
  }
}
```

Rationale for bundling: #523's gap wasn't only the snapshot shape — it was that the cache could be 0% effective in production with no signal beyond log greps. Counters add ~10 LOC and let the next regression surface by metric, not by user complaint.

## Validation

Smoke runs `A, A, B, A, B` with two distinct ~70K-token system prompts on Qwen3-Coder-30B-A3B-Instruct-8bit:

| Step    | System | TTFT      | Cache event |
| ------- | ------ | --------- | ----------- |
| A1 cold | A      | 144.27s   | MISS → STORED 70008 tokens (6.9 GB) |
| A2 warm | A      | **0.85s** | HIT 70008 |
| B1 cold | B      | 133.33s   | MISS → STORED 67008 tokens (6.6 GB) |
| A3 warm | A      | **0.86s** | HIT 70008 ← only possible with multi-slot |
| B2 warm | B      | **0.81s** | HIT 67008 |

**~170× TTFT speedup on warm prefixes.** Under single-slot, B1 would evict A and A3 would be another 144s cold MISS. With the LRU at capacity 4, both prefixes coexist; round-trip alternation costs the same as repeated same-prefix traffic.

Post-run `/v1/cache/stats`:

```json
{"capacity": 4, "in_use": 2,
 "counters": {"hits": 3, "misses": 2, "stores": 2, "evictions": 0, "hit_ratio": 0.6}}
```

Memory: 2 snapshots × ~6.7 GB ≈ 13.5 GB on Metal heap. Collocated embeddings engine on the same Mac continued serving without issue.

## What this does **not** change

- System-prefix detection logic from #523 — untouched.
- Startup probe that disables snapshotting for sliding-window models (RotatingKVCache aliasing) — untouched.
- Decode-path codepath — identical.
- Single-slot behavior is recoverable: `VLLM_MLX_SYSTEM_KV_SLOTS=1`.

## Files touched

`vllm_mlx/engine/simple.py` only. ~170 lines of diff including counters; the cache-shape change itself is ~50 lines.

## Open questions

- Default capacity of **4** — defensible for Claude Code's main + 3 sub-agent variants. Env var keeps it tunable. Push higher?
- Should `clear_runtime_caches` (DELETE `/v1/cache`) also clear the system-KV LRU? Leaning yes for consistent "reset everything" semantics — happy to add if you want it in this PR.
- `hit_ratio` recomputed on every `/v1/cache/stats` read. Cheap today; could cache + invalidate on counter mutation if anyone scrapes high-frequency. Not worth complexity yet.

### DIFF
--- a/tests/test_simple_engine.py
+++ b/tests/test_simple_engine.py
@@ -790,22 +790,229 @@ class TestSimpleEngineConcurrency:
         assert chunks and chunks[0].text == "ok"
 
     @pytest.mark.anyio
+    async def test_stream_generate_text_skips_cache_path_when_text_model_unsafe(self):
+        """Same snapshot-safety contract as the LLM path, but for MLLM text
+        routing: ``_stream_generate_text`` must not enter the system-KV cache
+        branch when ``start()``'s probe of the derived ``_text_model`` returned
+        non-KVCache entries (e.g. a sliding-window text head). The gate must
+        fall back to the uncached path — no encode of the system prefix, no
+        snapshot store, no LRU mutation."""
+        from collections import OrderedDict
+
+        from vllm_mlx.engine.simple import SimpleEngine
+
+        tokenizer = MagicMock()
+        tokenizer.apply_chat_template.return_value = (
+            "<|im_start|>system\nYou are helpful.<|im_end|>\n"
+            "<|im_start|>user\nhello<|im_end|>\n"
+            "<|im_start|>assistant\n"
+        )
+        tokenizer.bos_token = None
+        tokenizer.encode = MagicMock(return_value=[1, 2, 3])
+        tokenizer.decode = MagicMock(return_value="")
+        tokenizer.eos_token_id = 99
+
+        text_model = MagicMock()
+        text_model.mtp = None
+
+        engine = SimpleEngine("test-model", force_mllm=True)
+        engine._loaded = True
+        engine._text_model = text_model
+        engine._text_tokenizer = tokenizer
+        # Probe at start() decided this text model is NOT snapshot-safe.
+        engine._supports_system_kv_cache = False
+        engine._system_kv_cache = OrderedDict()
+        # Seed counters to verify they don't move on the uncached path.
+        engine._system_kv_cache_stats = {
+            "hits": 0,
+            "misses": 0,
+            "stores": 0,
+            "evictions": 0,
+        }
+
+        def fake_stream_generate(*args, **kw):
+            # _run_all iterates this synchronously (`for resp in ...`), so the
+            # mock must be a sync generator, not an async one.
+            yield SimpleNamespace(
+                text="ok",
+                new_text="ok",
+                prompt_tokens=3,
+                completion_tokens=1,
+                finished=True,
+                finish_reason="stop",
+                token=99,
+            )
+
+        with (
+            patch("vllm_mlx.engine.simple._bind_worker_generation_streams"),
+            patch(
+                "mlx_lm.models.cache.make_prompt_cache",
+                return_value=["backbone-cache"],
+            ),
+            patch("mlx_lm.sample_utils.make_sampler", return_value=MagicMock()),
+            patch(
+                "mlx_lm.sample_utils.make_logits_processors",
+                return_value=[],
+            ),
+            patch("mlx_lm.stream_generate", side_effect=fake_stream_generate),
+        ):
+            chunks = [
+                c
+                async for c in engine._stream_generate_text(
+                    messages=[
+                        {"role": "system", "content": "You are helpful."},
+                        {"role": "user", "content": "hello"},
+                    ],
+                    max_tokens=4,
+                    temperature=0.7,
+                    top_p=0.95,
+                )
+            ]
+
+        assert chunks, "expected uncached fallback to emit at least one chunk"
+        # System-prefix tokenization must NOT have been invoked: the only
+        # encode() call is the full-prompt one for the uncached path. The
+        # cache branch would have issued a second encode() for the system
+        # prefix before storing.
+        assert tokenizer.encode.call_count <= 1, (
+            "snapshot path ran despite _supports_system_kv_cache=False: "
+            f"encode called {tokenizer.encode.call_count} times"
+        )
+        # And nothing should have landed in the LRU or moved counters.
+        assert len(engine._system_kv_cache) == 0
+        assert engine._system_kv_cache_stats == {
+            "hits": 0,
+            "misses": 0,
+            "stores": 0,
+            "evictions": 0,
+        }
+
+    @pytest.mark.anyio
+    async def test_stream_generate_text_skips_cache_path_under_bounded_kv(self):
+        """Bounded-KV snapshot-safety contract for the MLLM text route.
+
+        When ``_max_kv_size > 0`` the runtime cache branch builds the prompt
+        cache via ``make_prompt_cache(model, max_kv_size=N)``. For models
+        without a custom ``make_cache``, ``mlx_lm.models.cache.make_prompt_cache``
+        returns ``RotatingKVCache`` whose ``.state`` aliases buffers that
+        ``update_and_fetch`` mutates in place — snapshot capture would corrupt
+        the cached prefix on the next decode.
+
+        The startup probe is now wired with the same ``max_kv_size`` arg as
+        the runtime constructor, so under bounded KV the probe sees
+        ``RotatingKVCache`` and sets ``_supports_system_kv_cache=False``.
+        This test locks in the runtime side of that chain: with the flag set
+        False (the correct post-probe state for ``_max_kv_size > 0``),
+        ``_stream_generate_text`` must fall back to the uncached path — no
+        encode of the system prefix, no LRU store, no counter movement.
+
+        Regression for PR #541 review (Thump604, 2026-05-17 16:04 UTC).
+        """
+        from collections import OrderedDict
+
+        from vllm_mlx.engine.simple import SimpleEngine
+
+        tokenizer = MagicMock()
+        tokenizer.apply_chat_template.return_value = (
+            "<|im_start|>system\nYou are helpful.<|im_end|>\n"
+            "<|im_start|>user\nhello<|im_end|>\n"
+            "<|im_start|>assistant\n"
+        )
+        tokenizer.bos_token = None
+        tokenizer.encode = MagicMock(return_value=[1, 2, 3])
+        tokenizer.decode = MagicMock(return_value="")
+        tokenizer.eos_token_id = 99
+
+        text_model = MagicMock()
+        text_model.mtp = None
+
+        engine = SimpleEngine("test-model", force_mllm=True)
+        engine._loaded = True
+        engine._text_model = text_model
+        engine._text_tokenizer = tokenizer
+        # Bounded-KV serving: the startup probe (called with the same
+        # max_kv_size as runtime) would have built RotatingKVCache and
+        # flipped this flag to False.
+        engine._max_kv_size = 2048
+        engine._supports_system_kv_cache = False
+        engine._system_kv_cache = OrderedDict()
+        engine._system_kv_cache_stats = {
+            "hits": 0,
+            "misses": 0,
+            "stores": 0,
+            "evictions": 0,
+        }
+
+        def fake_stream_generate(*args, **kw):
+            yield SimpleNamespace(
+                text="ok",
+                new_text="ok",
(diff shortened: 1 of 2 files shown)


Thump604 (changes requested):
Thanks for the follow-up. The multi-slot shape and the A/A/B/A/B validation make sense, and the CI result is clean. I would hold merge for two cache-semantics gaps that are small to fix but important for this path.

1. `_stream_generate_text` can still enter the system-KV cache branch without a snapshot-safety probe. `start()` only sets `_supports_system_kv_cache` for the non-MLLM `_model` path, but text-only MLLM requests route into `_stream_generate_text`, which then checks only `has_system and self._text_model is not None` before snapshotting. That bypasses the same RotatingKVCache safety gate that `stream_chat` uses. Please either probe the derived `_text_model` prompt cache and gate this path on that result, or skip system-KV caching in `_stream_generate_text` unless the text model is proven snapshot-safe. A focused regression should cover `_supports_system_kv_cache=False` on the TextModel-direct path and prove it falls back cold rather than storing/restoring snapshots.

2. `clear_runtime_caches()` does not clear the new LRU. With multi-slot snapshots this can retain multiple multi-GB system-prefix entries after DELETE `/v1/cache`, while `/v1/cache/stats` still reports them. I agree with your open-question instinct: the reset endpoint should clear `self._system_kv_cache`, reset the counters, call `mx.clear_cache()` after dropping those references, and return that in the clear result. Please add a small test so this does not regress.

I did a local focused check of the current PR head:

`UV_PYTHON=/opt/homebrew/bin/python3.12 uv run --extra dev python -m pytest -q tests/test_simple_engine.py`

Result: `38 passed, 1 warning`.

I am not objecting to the multi-slot approach or the default of 4 here. These two points are about keeping the existing snapshot-safety contract true across the TextModel-direct route and making the operational cache reset actually release the new cache state.

vinayvobbili:
Thanks @Thump604 — both points addressed in `492ea1c`.

**1. `_stream_generate_text` snapshot-safety gate**

`start()` now runs the same `make_prompt_cache` / `KVCache` probe against the derived `_text_model` right after `build_text_model` returns, and sets `_supports_system_kv_cache` for the MLLM text route too. The cache branch at line ~1641 (`if has_system and self._text_model is not None and self._supports_system_kv_cache:`) now gates on that flag, so a sliding-window text head falls back to the uncached `mlx_lm.stream_generate` path instead of snapshotting a `RotatingKVCache` and silently desynchronizing on restore.

Regression: `test_stream_generate_text_skips_cache_path_when_text_model_unsafe` mocks the engine with `_supports_system_kv_cache=False`, exercises `_stream_generate_text` on a system+user prompt, and asserts:
- the system prefix is **not** re-tokenized (the cache branch would issue a second `encode()` for the prefix before storing)
- the LRU stays empty
- counters don't move

**2. `clear_runtime_caches` drops the LRU + counters**

Reset now clears `self._system_kv_cache`, zeroes all four counters, calls `mx.clear_cache()` to release the Metal heap snapshots, and reports `{"dropped_entries": N}` in the result. No lock acquisition — OrderedDict ops are atomic under the GIL, so a concurrent worker that already captured a snapshot tuple from `.get()` finishes safely against its own reference; any request arriving after clear hits MISS and repopulates from scratch.

Two regressions:
- `test_clear_runtime_caches_drops_lru_and_resets_counters` seeds a 2-entry LRU + non-zero counters, calls clear, asserts the LRU is empty, counters are zero, and the result includes `dropped_entries: 2`
- `test_clear_runtime_caches_no_op_when_lru_empty_and_counters_zero` preserves the existing `return None` contract for the idle/non-MLLM case

Local CI run will surface; will reply with the green check when it lands.

vinayvobbili:
CI is green on 6cce3dc — all 9 checks pass (lint, type-check, test-matrix 3.10/3.11/3.12/3.13, test-apple-silicon 3.11/3.13, tests). The two intermediate fixups (492ea1c → 49f0af0 → 6cce3dc) were mechanical follow-ups to my own initial commit: a black line-length wrap on the long `if` condition, a missed positional arg in the new gate test, and an `async def` mock that needed to be sync since `_run_all` iterates the generator with a sync `for` loop. The substance of the two review fixes is unchanged from 492ea1c. Ready for re-review.

Thump604 (changes requested):
Thanks for the follow-up. I rechecked the current head and the reset/counter path looks good; local focused validation also passes (`tests/test_simple_engine.py`: 41 passed).

I still think one bounded-KV case needs fixing before this is merge-ready. The new TextModel safety probe uses `make_prompt_cache(self._text_model)` during `start()`, but `_stream_generate_text` later restores/prefills with `make_prompt_cache(model, max_kv_size=self._max_kv_size or None)`. In `mlx_lm.models.cache.make_prompt_cache`, models without a custom `make_cache` get `KVCache` with no `max_kv_size` and `RotatingKVCache` when `max_kv_size` is supplied. That means the startup probe can mark the TextModel path snapshot-safe while the runtime cache branch creates a rotating cache under bounded-KV serving.

The pure LLM path already avoids this class of mismatch by blocking the system-KV cache branch when `_max_kv_size > 0`; the derived TextModel path should have the same protection or probe with the exact constructor arguments the runtime branch will use. Please add a focused regression with `_max_kv_size > 0` that proves `_stream_generate_text` does not snapshot/restore an unsafe bounded cache, or that the probe and runtime cache constructor are identical for the path being enabled.

vinayvobbili:
@Thump604 thanks. Addressed in 65392b8 by taking the "probe and runtime cache constructor are identical" option:

```python
probe_cache = make_prompt_cache(
    self._text_model, max_kv_size=self._max_kv_size or None
)
```

Single-line change, same `max_kv_size or None` kwarg the runtime branch (`_stream_generate_text`'s MISS-path `make_prompt_cache` call) already uses, so the probe sees the same `RotatingKVCache` the runtime would build under bounded-KV. The flag flips to `False` and the gate at `has_system and self._text_model is not None and self._supports_system_kv_cache` correctly skips. No behavior change when `_max_kv_size == 0`.

Regression: `test_stream_generate_text_skips_cache_path_under_bounded_kv` seeds `_max_kv_size=2048` + the post-probe `_supports_system_kv_cache=False` and asserts the runtime side — no system-prefix encode, no LRU mutation, no counter movement. Docstring narrates the chain so the next reader doesn't have to re-derive it.

Local focused run on an M3 Max (`PYTHONPATH=.` against a vllm-mlx-installed venv):

```
tests/test_simple_engine.py ..........................................   [100%]
42 passed in 2.88s
```

Same width / lint pattern as the existing TextModel probe block; black clean.

Thump604 (approved):
Thanks. I rechecked the latest head after `65392b8`.

The bounded-KV concern I raised is addressed: the TextModel probe now calls `make_prompt_cache` with `max_kv_size=self._max_kv_size or None`, matching the runtime constructor, and `_stream_generate_text` gates the cache branch on `_supports_system_kv_cache`. The new bounded-KV regression covers the runtime fallback path.

I also reran focused validation locally in a detached worktree:

`UV_PYTHON=/opt/homebrew/bin/python3.12 uv run --extra dev python -m pytest -q tests/test_simple_engine.py`

Result: `42 passed, 1 warning`.

CI is green. I do not see remaining blockers from my prior review.

janhilgard (approved):
Clean follow-up to #523. The OrderedDict LRU is the right primitive here — `move_to_end` / `popitem(last=False)` gives O(1) eviction with standard Python semantics, and gating reads via `dict.get()` returning an immutable tuple preserves the TOCTOU safety contract from the single-slot design.

I verified the concurrency model: all LRU mutations in `stream_chat` happen inside `_run_blocking_serialized` (worker thread under `_generation_lock`), and the HIT-path `move_to_end` + counter bump in `_stream_generate_text` happen synchronously on the event loop between `await` points — so no interleaving with other coroutines. The `clear_runtime_caches` GIL-atomicity claim is also correct; a concurrent worker's captured tuple reference survives the `.clear()`.

The TextModel probe chain (start → `make_prompt_cache(self._text_model, max_kv_size=...)` → `_supports_system_kv_cache` flag → gate in `_stream_generate_text`) correctly mirrors the pure-LLM path, including the bounded-KV fix in 65392b8.

Two minor nits (non-blocking):

1. **`mx.clear_cache()` after every STORE**: Lines ~519 and ~587 call `mx.clear_cache()` unconditionally after the store+evict loop, even when no eviction happened (LRU was below capacity). The comment says "Eviction dropped MLX array refs; reclaim Metal heap" but the call runs regardless. Consider guarding with `if evicted:` to avoid clearing the Metal allocator's reuse pool unnecessarily on the common non-eviction path.

2. **Squash before merge**: 6 commits (1 main + 5 fixups) — please squash into a single commit per project convention.

Validation numbers (170× TTFT speedup, A/A/B/A/B pattern holding both prefixes warm) are compelling and directly address the Claude Code main+sub-agent dispatch scenario.

vinayvobbili:
@janhilgard thanks for the review and the careful read of the TOCTOU + concurrency model. Both nits addressed:

**Nit 1 — `mx.clear_cache()` after every STORE.** Guarded with `if evicted_count` so the call only runs when the LRU was actually at capacity and dropped MLX array refs. With default capacity 4 and typical Claude Code traffic (main + 2–3 sub-agent variants) eviction is the exception, so the common path no longer flushes the Metal allocator's reuse pool. Applied to both `stream_chat` and `_stream_generate_text` store sites.

**Nit 2 — Squash before merge.** Squashed all 6 fixup commits into a single feat commit (`4ce8e11`). PR is now one commit on top of `c2d3aec`. Body covers the LRU shape, `clear_runtime_caches` LRU drop, TextModel snapshot-safety probe (incl. bounded-KV fix from the previous round), counters, and validation numbers.

Local focused run on an M3 Max after both fixes (`PYTHONPATH=.` against a vllm-mlx-installed venv):

```
tests/test_simple_engine.py ..........................................   [100%]
42 passed in 2.88s
```

Black clean. CI rerunning now on the squashed head.

waybarrios (approved):
Just read it carefully. 

It's all good!

Issue #546, opened 2026-05-18
Bug: strict json_schema decode wedges (no token progress, >5min) on Qwen3-Coder-30B-A3B


vinayvobbili:
### Environment

- `vllm-mlx 0.3.0` (pip)
- Mac Studio (Apple Silicon, 96 GB)
- Model: `Qwen3-Coder-30B-A3B-Instruct-8bit` (sparse MoE, 3B active per token)
- Launch args:
  ```
  vllm_mlx_server.py \
    --tool-call-parser qwen3_coder \
    --model Qwen3-Coder-30B-A3B-Instruct-8bit \
    --port 8003 --host 0.0.0.0 \
    --served-model-name qwen3-coder-30b-a3b \
    --max-request-tokens 131072 \
    --default-temperature 0.3
  ```
- Plus our local multi-slot LRU patch (follow-up to #523). Happy to retest against stock to confirm independence.

### Symptom

A `/v1/chat/completions` request with `response_format={"type":"json_schema","json_schema":{...},"strict":true}` for a moderately complex Pydantic-derived schema gets stuck mid-decode for 5+ minutes. `[disconnect_guard]` polls keep counting up while no token-progress logs appear. The request eventually 504s via the upstream proxy. Every subsequent request queued behind it also fails to progress until the server is restarted.

### Schema (the one that wedged)

```json
{
  "title": "NoveltyLLMResponse",
  "type": "object",
  "$defs": {
    "VulnerableProductMention": {
      "type": "object",
      "properties": {
        "product": {"type": "string"},
        "vendor": {"anyOf": [{"type":"string"},{"type":"null"}]},
        "version_constraint": {"anyOf": [{"type":"string"},{"type":"null"}]}
      },
      "required": ["product","vendor","version_constraint"],
      "additionalProperties": false
    }
  },
  "properties": {
    "novelty_score": {"type":"integer","minimum":1,"maximum":10},
    "novelty_label": {"type":"string","enum":["Seen Before","Familiar","Mostly New","Net New"]},
    "summary": {"type":"string"},
    "recommendation": {"type":"string"},
    "whats_new_reasons": {"type":"array","items":{"type":"string"}},
    "whats_familiar_reasons": {"type":"array","items":{"type":"string"}},
    "vulnerable_products": {"type":"array","items":{"$ref":"#/$defs/VulnerableProductMention"}}
  },
[...]

PR #547, opened 2026-05-19
Bound constrained JSON whitespace progress

Thump604:
## Summary

Bounds non-progress whitespace in `JSONSchemaLogitsProcessor` so constrained JSON decoding cannot keep selecting pure whitespace tokens indefinitely before making JSON structural/content progress.

This is a narrow follow-up for #546. I am not claiming the reported Qwen3-Coder wedge is fully root-caused to whitespace generation, and I am not claiming a sparse-MoE-specific failure. The verified local code path is that the constrained decoder currently permits unlimited pure JSON whitespace at legal positions such as the root and after `{`; in non-streaming calls that can look like no useful output progress while the request holds the serving slot.

## Code path compared

- Local code path: `vllm_mlx/constrained/json_schema_processor.py`, specifically `JSONSchemaLogitsProcessor.__call__`, `_get_json_context`, `_filter_at_key_context`, and the `lm-format-enforcer` allowed-token set.
- Upstream base checked: `waybarrios/vllm-mlx` `upstream/main` at `7e30484`.
- Reported release checked for processor construction/call behavior: `v0.3.0` tag. The issue schema did not hang processor construction or individual processor calls in either `v0.3.0` or current main.

## Observed behavior

Using the schema from #546 with `Qwen/Qwen3-Coder-30B-A3B-Instruct` tokenizer data, the allowed-token set includes many pure whitespace tokens at JSON positions before useful JSON content is produced. On current main before this patch, the processor leaves whitespace tokens finite after long whitespace suffixes.

## Expected behavior

Whitespace remains legal during normal JSON generation, but after a long trailing run of pure whitespace outside a JSON string, the next token should make structural/content progress. This patch removes pure-whitespace tokens only after the trailing whitespace run crosses the bounded threshold, preserving non-whitespace tokens returned by the enforcer.

## Validation

Passed:

- `python -m pytest tests/test_json_schema_processor_hardening.py tests/test_constrained_decoding.py -q` — 46 passed
- `python -m pytest tests/test_json_schema_processor_hardening.py tests/test_constrained_decoding.py tests/test_structured_output.py tests/test_api_models.py -k 'not InjectJsonInstruction' -q` — 167 passed, 2 skipped, 3 deselected
- `uvx ruff check vllm_mlx/constrained/json_schema_processor.py tests/test_json_schema_processor_hardening.py --select E,F,W --ignore E402,E501,E731,F811,F841`
- `python -m black --check vllm_mlx/constrained/json_schema_processor.py tests/test_json_schema_processor_hardening.py`
- `git diff --check`

Full local `pytest -q` did not collect in my Runtime venv because the installed `mlx_lm` package imports an overlay-only `get_generation_stream` symbol from `vllm_mlx.mlx_streams`. That failure appears before test execution in server/scheduler import paths and is not touched by this patch.

## Merge readiness

- [x] I am not merging my own substantive PR.
- [ ] The latest CI run for this exact head commit is green.
- [ ] If I rebased, resolved conflicts, or pushed after review, I waited for the new CI run to finish green.
- [x] Actionable review comments are resolved in this PR or linked to follow-up issues before merge.
- [x] Any intentional follow-up debt is linked here and called out in the PR body.

## Risk notes

This changes constrained JSON token masking only after a long trailing whitespace run outside a JSON string. It does not change schema simplification, parser construction, server request handling, streaming, or unconstrained generation.

### DIFF
--- a/tests/test_json_schema_processor_hardening.py
+++ b/tests/test_json_schema_processor_hardening.py
@@ -342,6 +342,70 @@ class TestCallSemantics:
         masked = proc(tokens, logits)
         assert masked.shape == (1, tok.vocab_size)
 
+    def test_long_leading_whitespace_cannot_continue_without_json_progress(self):
+        """Bound legal-but-non-progress whitespace in constrained JSON mode.
+
+        ``lm-format-enforcer`` permits arbitrary whitespace after JSON
+        structural positions such as ``{``.  That is spec-correct, but a model
+        can then spend a long non-streaming request emitting only whitespace
+        while the client sees no useful JSON progress.  Once a suffix has a
+        long trailing whitespace run outside a string, the processor should
+        force a structural/content token instead of allowing more pure
+        whitespace.
+        """
+        import mlx.core as mx
+        import numpy as np
+
+        proc, tok = _make_processor(
+            {
+                "type": "object",
+                "properties": {"a": {"type": "string"}},
+                "required": ["a"],
+                "additionalProperties": False,
+            }
+        )
+        prompt = tok.encode(" ")
+        proc(mx.array(prompt, dtype=mx.int32), mx.zeros((tok.vocab_size,)))
+
+        full = prompt + tok.encode(" " * 300)
+        masked = proc(mx.array(full, dtype=mx.int32), mx.zeros((tok.vocab_size,)))
+        arr = np.array(masked)
+
+        assert arr[tok._tok_to_id[" "]] == -np.inf
+        assert arr[tok._tok_to_id["\n"]] == -np.inf
+        assert arr[tok._tok_to_id["\t"]] == -np.inf
+        assert arr[tok._tok_to_id["{"]] == 0.0
+
+    def test_long_structural_whitespace_cannot_continue_without_json_progress(self):
+        """The same whitespace-progress bound applies after JSON structure."""
+        import mlx.core as mx
+        import numpy as np
+
+        proc, tok = _make_processor(
+            {
+                "type": "object",
+                "properties": {"a": {"type": "string"}},
+                "required": ["a"],
+                "additionalProperties": False,
+            }
+        )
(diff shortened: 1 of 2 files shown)


janhilgard (approved):
## Code Review

CI all green. Single commit, clean diff (+101/-0), purely additive.

### What it does

Adds a whitespace progress bound to `JSONSchemaLogitsProcessor`: after 256 consecutive trailing whitespace characters outside a JSON string, pure-whitespace tokens are removed from the allowed set, forcing structural/content progress. This prevents constrained JSON decoding from spending indefinitely on whitespace-only tokens.

### Assessment

**Architecture:** Clean and narrow. The new method `_filter_nonprogress_whitespace_tokens` is called only when `_json_ctx_in_string` is False, correctly preserving whitespace inside string values. The 256-char threshold is generous — legitimate JSON formatting rarely exceeds a few dozen whitespace characters.

**Safety:** The `return filtered if filtered else allowed` fallback ensures the allowed set is never empty, which would cause the enforcer stuck-detection to trigger. This is the right defensive pattern.

**Performance:** The filter iterates over the allowed token list and calls `_decode_token_cached` per token. Since `_decode_token_cached` is cached and the filter only activates after 256 trailing whitespace chars (rare path), the overhead is negligible in practice.

**Tests:** Two test cases cover both leading whitespace (before any JSON structure) and post-structural whitespace (after `{`). Both verify that whitespace tokens are masked to `-inf` while structural tokens remain at `0.0`.

### Minor notes (non-blocking)

1. `_MAX_NONPROGRESS_WHITESPACE_CHARS = 256` is module-level — if this ever needs to be configurable per-request, it would need to move to the instance. Fine as-is for now.

2. The `_decode_suffix` call in the filter decodes the full suffix on every invocation. This is already cached by suffix length in the parent class, so no concern.

LGTM.

PR #548, opened 2026-05-19
Route constrained non-stream chat through stream path

Thump604:
## Summary
- route non-stream LLM chat with request-local logits processors through the existing stream aggregation path
- keep ordinary non-stream LLM chat on the existing blocking `model.chat()` path
- add a regression test that ensures constrained-decoding kwargs reach `stream_generate` instead of `model.chat`

## Why
`response_format` uses request-local logits processors. In simple LLM mode without tools, non-stream chat previously called blocking `model.chat()`, so strict-schema decode could hold the serialized generation lock without token-boundary progress or cooperative cancellation. That matches the non-stream constrained JSON wedge shape reported in #546.

## Validation
- Red first: `uv run --python 3.12 pytest tests/test_simple_engine.py -k 'llm_nonstream_with_logits_processors_uses_stream_path' -q` failed with `AssertionError: blocking chat path should not be used`
- `uv run --python 3.12 pytest tests/test_simple_engine.py -k 'llm_nonstream_with_logits_processors_uses_stream_path' -q` = `1 passed, 38 deselected`
- `uv run --python 3.12 pytest tests/test_simple_engine.py tests/test_constrained_decoding.py tests/test_structured_output.py -q` = `112 passed, 2 skipped`
- `uv run --python 3.12 ruff check vllm_mlx/engine/simple.py tests/test_simple_engine.py` = pass
- `uv run --python 3.12 black --check vllm_mlx/engine/simple.py tests/test_simple_engine.py` = pass
- `git diff --check` = pass
- `/opt/ai-runtime/bin/upstream-local-repro-preflight --local-root /opt/ai-runtime/worktrees/vllm-mlx/issue-546-disconnect-root-cause --upstream-root /opt/ai-runtime/worktrees/vllm-mlx/upstream-main-clean --code-path vllm_mlx/engine/simple.py --code-path tests/test_simple_engine.py --no-model-artifact` = pass

## Not claimed
- This does not claim sparse-MoE behavior is the cause.
- This does not change ordinary non-stream chat without request-local logits processors.
- This does not replace #547's whitespace-progress hardening.
- This does not prove the reporter's exact MLX model no longer wedges because the exact MLX weights were not available locally.

Refs #546.

### DIFF
--- a/tests/test_simple_engine.py
+++ b/tests/test_simple_engine.py
@@ -1189,6 +1189,62 @@ class TestSimpleEngineConcurrency:
         assert output.finish_reason == "stop"
 
     @pytest.mark.anyio
+    async def test_llm_nonstream_with_logits_processors_uses_stream_path(self):
+        """Constrained non-stream chat must not call the blocking chat API.
+
+        ``response_format`` is implemented by request-local logits processors.
+        If a non-stream request goes through the blocking model.chat() path,
+        the server cannot observe token progress or cancel at token boundaries
+        when a client/proxy disconnects.  Aggregating stream_chat keeps the
+        constrained and unconstrained chat paths on the same cancellable stream
+        implementation.
+        """
+        from types import SimpleNamespace
+
+        from vllm_mlx.engine.simple import SimpleEngine
+
+        captured_stream_kwargs = {}
+
+        class FakeTokenizer:
+            bos_token = None
+
+            def apply_chat_template(self, messages, **kwargs):
+                return "<|im_start|>user\nhello"
+
+            def encode(self, text, **kwargs):
+                return [1, 2, 3]
+
+        class FakeModel:
+            tokenizer = FakeTokenizer()
+
+            def chat(self, **kwargs):
+                raise AssertionError("blocking chat path should not be used")
+
+            def stream_generate(self, **kwargs):
+                captured_stream_kwargs.update(kwargs)
+                yield SimpleNamespace(
+                    text="{}",
+                    finish_reason="stop",
+                    finished=True,
+                    prompt_tokens=3,
+                )
+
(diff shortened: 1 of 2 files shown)


janhilgard (approved):
## Code Review

CI all green. Minimal diff (+64/-0), purely additive.

### What it does

When a non-streaming LLM chat request carries `logits_processors` (from `response_format` / constrained JSON), routes it through `aggregate_stream_chat()` instead of blocking `model.chat()`. This gives token-boundary progress visibility and cancellation for constrained decoding, which can otherwise hold the generation lock indefinitely on a slow/degenerate schema.

### Routing chain after this change

1. `tools` + LLM → `aggregate_stream_chat()` (existing)
2. MLLM text-only → `aggregate_stream_chat()` (existing)
3. **`logits_processors` + LLM → `aggregate_stream_chat()`** (new)
4. MLLM media → blocking path (existing, unchanged)
5. LLM no tools/processors → blocking `model.chat()` (existing, unchanged)

### Assessment

**Correctness:** The condition `kwargs.get("logits_processors") and not self._is_mllm` is narrow and precise — only triggers for LLM (not MLLM) with active logits processors. `aggregate_stream_chat()` passes all kwargs through to `stream_chat`, which passes them to `stream_generate`, so processors reach the generation loop correctly.

**No behavior change for unconstrained requests:** Plain non-streaming chat without `response_format` still goes through the blocking `model.chat()` path, preserving existing behavior.

**Test:** The regression test verifies that `model.chat()` raises `AssertionError` if called (proving the stream path is used), that `stream_generate` receives the sentinel processor, and that output is correct.

**Complements #547:** This routes constrained requests through the stream path for cancellability; #547 bounds whitespace progress within the processor itself. Both are needed — #547 prevents the wedge, this PR makes it cancellable if it occurs.

LGTM.

Issue #542, opened 2026-05-17
[bug] --models-config (registry mode) returns model=None in OpenAI responses, pydantic ValidationError 500s every chat/completion


inmanbobby71:
### Description

When `vllm-mlx serve` is run with `--models-config <yaml>` (multi-model registry mode), every `/v1/chat/completions` and `/v1/completions` request returns HTTP 500 with a pydantic `ValidationError`. Streaming, non-streaming, and Anthropic-compat endpoints all fail. The single-model mode (`vllm-mlx serve <model>` without `--models-config`) is unaffected.

### Root cause

`load_model_registry()` at `server.py:3024` explicitly sets the module-level global `_model_name = None`. The response-builder sites then pass this directly to pydantic models that require a non-None string for the `model` field. The single-model path (`load_model`, `server.py:2908`) sets `_model_name = served_model_name or model_name`, which is why it doesn't exhibit the bug.

Line `server.py:2031` already uses the correct fallback pattern (`model=_model_name or request.model`), so the maintainer intent is clear — 11 other response sites just missed the same fix.

### Reproduction

```bash
# minimal models.yaml
cat > models.yaml <<EOF
manager:
  memory_budget_gb: 8
  contention_policy:
    strategy: wait_then_fail

models:
  - name: mlx-community/Qwen2.5-3B-Instruct-4bit
    source: mlx-community/Qwen2.5-3B-Instruct-4bit
    preload: true
EOF

vllm-mlx serve --models-config models.yaml --host 127.0.0.1 --port 8000 &
sleep 30  # wait for preload

curl -s -X POST http://127.0.0.1:8000/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{"model":"mlx-community/Qwen2.5-3B-Instruct-4bit","messages":[{"role":"user","content":"hi"}],"max_tokens":5}'
```

Expected: a chat completion response.
Actual: `Internal Server Error` with the daemon log showing:

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for ChatCompletionResponse
model
  Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]
```

### Affected sites (all in server.py)

| Line | Context |
|---|---|
| 4463 | `CompletionResponse(model=_model_name, ...)` — non-streaming /v1/completions |
| 4652 | `ChatCompletionResponse(model=_model_name, ...)` — non-streaming /v1/chat/completions |
[...]

Issue #529, opened 2026-05-12
use --model-config deploy llm model and chat error


youngtseup:
1. exec "vllm-mlx serve --models-config ./models.yml --host 0.0.0.0 --port 8080".
2. call api "/v1/chat/completions".
3. http 500 Internal Server Error.

models.yml:
```
manager:
  memory_budget_gb: 70
models:
  - path: /Users/root/qwen36_mlx/qwen36-35b-a3b-4bit
    name: qwen36-35b-a3b-4bit

  - path: /Users/root/qwen36_mlx/qwen36-27b-8bit
    name: qwen36-27b-8bit
```

error logs:
```
INFO:vllm_mlx.models.llm:Loading model: /Users/root/qwen36_mlx/qwen36-27b-8bit
INFO:vllm_mlx.models.llm:Qwen3 detected: setting eos_token to <|im_end|>
INFO:vllm_mlx.utils.tokenizer:[MTP] Config has num_nextn_predict_layers=1 but MTP weights not found, skipping MTP.
INFO:vllm_mlx.models.llm:Model loaded successfully: /Users/root/qwen36_mlx/qwen36-27b-8bit
INFO:vllm_mlx.engine.simple:SimpleEngine loaded: /Users/root/qwen36_mlx/qwen36-27b-8bit (MLLM=False)
INFO:vllm_mlx.server:[disconnect_guard] poll #10 disconnected=False elapsed=5.0s
INFO:vllm_mlx.server:[disconnect_guard] poll #20 disconnected=False elapsed=10.0s
INFO:vllm_mlx.server:[disconnect_guard] poll #30 disconnected=False elapsed=15.1s
INFO:vllm_mlx.server:[disconnect_guard] poll #40 disconnected=False elapsed=20.1s
INFO:vllm_mlx.server:[disconnect_guard] poll #50 disconnected=False elapsed=25.1s
INFO:vllm_mlx.server:[disconnect_guard] poll #60 disconnected=False elapsed=30.1s
INFO:vllm_mlx.server:[disconnect_guard] poll #70 disconnected=False elapsed=35.1s
INFO:vllm_mlx.server:[disconnect_guard] poll #80 disconnected=False elapsed=40.2s
INFO:vllm_mlx.server:Chat completion: 954 tokens in 41.81s (22.8 tok/s)
INFO:     192.168.100.74:1936 - "POST /v1/chat/completions HTTP/1.1" 500 Internal Server Error
[...]

PR #549, opened 2026-05-19
Use request model in registry responses

Thump604:
## Summary
- use the request model as the response `model` field when registry mode has no process-wide `_model_name`
- preserve the existing served-name response behavior for single-model mode
- add regression coverage for registry-mode response construction and a source guard for completion/message endpoint response builders

## Why
In `--models-config` registry mode, `load_model_registry()` leaves `_model_name = None` because there is no single served model for the process. Several OpenAI/Anthropic response builders still passed `_model_name` directly into response payloads. Pydantic response models require `model` to be a string, so non-stream responses can fail with HTTP 500; dict/SSE response sites can emit `model: null`.

## Validation
- Red first: `uv run --python 3.12 pytest tests/test_lifecycle_server.py -k 'registry_mode_completion_response_uses_request_model or registry_mode_response_builders_do_not_use_bare_global_model_name' -q` failed with `CompletionResponse(model=None)` and direct `_model_name` endpoint response fields
- `uv run --python 3.12 pytest tests/test_lifecycle_server.py -k 'registry_mode_completion_response_uses_request_model or registry_mode_response_builders_do_not_use_bare_global_model_name' -q` = `2 passed`
- `uv run --python 3.12 pytest tests/test_lifecycle_server.py -k 'ResponseModelFieldUsesServedName' -q` = `5 passed`
- `uv run --python 3.12 pytest tests/test_lifecycle_server.py tests/test_server.py -q` = `185 passed, 3 deselected`
- `uv run --python 3.12 ruff check vllm_mlx/server.py tests/test_lifecycle_server.py` = pass
- `uv run --python 3.12 black --check vllm_mlx/server.py tests/test_lifecycle_server.py` = pass
- `git diff --check` = pass
- `/opt/ai-runtime/bin/upstream-local-repro-preflight --local-root /opt/ai-runtime/worktrees/vllm-mlx/issue-542-registry-model-field --upstream-root /opt/ai-runtime/worktrees/vllm-mlx/upstream-main-clean-542 --code-path vllm_mlx/server.py --code-path tests/test_lifecycle_server.py --no-model-artifact` = pass

## Not claimed
- This does not change model validation or routing.
- This does not change `/v1/status` or admin/cancel metadata semantics.
- This does not change single-model served-name responses.

Refs #542.
Refs #529.

### DIFF
--- a/tests/test_lifecycle_server.py
+++ b/tests/test_lifecycle_server.py
@@ -3726,3 +3726,64 @@ class TestResponseModelFieldUsesServedName:
         body = json.loads(response.body.decode())
         assert body["model"] == served_name
         assert body["model"] != "user-sent-model-name"
+
+    @pytest.mark.anyio
+    async def test_registry_mode_completion_response_uses_request_model(
+        self, monkeypatch
+    ):
+        import vllm_mlx.server as srv
+        from vllm_mlx.engine.base import GenerationOutput
+
+        class FakeEngine:
+            preserve_native_tool_format = False
+            is_mllm = False
+
+            async def generate(self, **kwargs):
+                return GenerationOutput(
+                    text="hello",
+                    completion_tokens=1,
+                    prompt_tokens=1,
+                )
+
+        async def fake_acquire(
+            raw_request,
+            *,
+            total_timeout=None,
+            deadline=None,
+            count_activity=True,
+            model=None,
+        ):
+            return FakeEngine()
+
+        async def fake_release(*, count_activity=True):
+            pass
+
+        monkeypatch.setattr(srv, "_validate_model_name", lambda _m: None)
+        monkeypatch.setattr(srv, "_acquire_default_engine_for_request", fake_acquire)
+        monkeypatch.setattr(srv, "_release_default_engine", fake_release)
+        monkeypatch.setattr(srv, "_model_name", None)
+        monkeypatch.setattr(srv, "_default_max_tokens", 32)
+
+        class FakeRawRequest:
+            async def is_disconnected(self):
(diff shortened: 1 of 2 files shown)


janhilgard (approved):
## Code Review

CI all green. Clean diff (+79/-13), mechanical substitution with a clear helper.

### What it does

In `--models-config` registry mode, `_model_name` is `None` (no single process-wide model). All response builders previously passed `_model_name` directly into response payloads, causing:
- Pydantic `model: str` validation failures → HTTP 500 on non-streaming responses
- `"model": null` in SSE/dict streaming responses

The fix introduces `_response_model_name(request_model)` which returns `_model_name or request_model` — single-model mode is unchanged (returns the served name), registry mode falls back to the request's model field.

### Assessment

**Correctness:** All 13 substitution sites are mechanically correct — every `model=_model_name` in completion/chat/anthropic response construction is replaced with `model=_response_model_name(request.model)`. The helper is trivial (`return _model_name or request_model`), no edge cases.

**Backward compatibility:** In single-model mode, `_model_name` is always set, so `_response_model_name` returns it — behavior is identical to before.

**Source guard test:** `test_registry_mode_response_builders_do_not_use_bare_global_model_name` scans the source text of server.py from `# Completion Endpoints` onwards and asserts that `model=_model_name,` and `"model": _model_name,` patterns don't appear. This prevents regressions when new endpoints are added. Clever approach.

**Registry mode test:** `test_registry_mode_completion_response_uses_request_model` sets `_model_name = None` and verifies the response uses the request model name. Good coverage.

LGTM.

Issue #545, opened 2026-05-18
[bug] XLM-RoBERTa reranker models fail to load: classifier.weight KeyError + missing position_id offset


inmanbobby71:
## Summary

XLM-RoBERTa-based reranker models cannot be loaded via `--rerank-model` in vllm-mlx 0.3.0. Affects every modern strong reranker I'm aware of: `BAAI/bge-reranker-v2-m3`, `jinaai/jina-reranker-v2-base-multilingual`, the Qwen3-Reranker family — all use the `XLMRobertaForSequenceClassification` architecture. Two distinct bugs in `rerank_forward.py` combine to make these models unusable: one is fatal at first request, the other is a silent correctness bug that surfaces only after the first is patched.

Companion to #542 (registry-mode `_model_name=None`). I hit this while attempting to swap `cross-encoder/ms-marco-MiniLM-L6-v2` → `BAAI/bge-reranker-v2-m3` on an M4 Max daemon and rolled back the same evening.

## Bug 1 — Two-layer classification head not handled (FATAL)

`XLMRobertaClassificationHead` is a `Dense → tanh → out_proj` pattern, so the safetensors weights expose `classifier.dense.weight`, `classifier.dense.bias`, `classifier.out_proj.weight`, `classifier.out_proj.bias`. vllm-mlx's `classifier_forward()` (around `rerank_forward.py:83`) reads a flat `classifier.weight` / `classifier.bias` — the BERT-style single-layer pattern from `ms-marco-MiniLM`.

**Symptom:** first `POST /v1/rerank` after daemon start returns:

```json
{"detail": "'classifier.weight'"}
```

…and the rerank path is dead until the model is swapped back.

## Bug 2 — RoBERTa-family position ID offset missing (SILENT CORRECTNESS)

RoBERTa-family models require `position_ids = arange(padding_idx + 1, seq_len + padding_idx + 1)` with `padding_idx = 1` (see HF Transformers `create_position_ids_from_input_ids`). vllm-mlx uses plain `arange(seq_len)` directly. Once Bug 1 is patched, the model loads and serves requests — but the position embeddings are off-by-(padding_idx+1) and the resulting scores are garbage. No exception is raised; the rerank scores just stop being meaningful. This is the worse of the two bugs because nothing tells you it's broken.

## Reproduction

On a working vllm-mlx 0.3.0 daemon (registry mode or single-model mode, same result):

[...]

PR #550, opened 2026-05-19
Support XLM-RoBERTa reranker heads

Thump604:
## Summary
- support XLM-RoBERTa/RoBERTa two-layer sequence-classification heads in the rerank forward path
- use RoBERTa-family position IDs with the padding offset instead of BERT-style `arange(seq_len)`
- keep existing flat BERT classifier heads working

## Why
XLM-RoBERTa rerankers such as BGE/Jina v2 expose `classifier.dense.*` plus `classifier.out_proj.*`, not flat `classifier.weight` / `classifier.bias`. The old forward path raised `KeyError: 'classifier.weight'` on those models. RoBERTa-family models also reserve the pad position and offset real token positions, so plain `arange(seq_len)` gives the wrong position embeddings.

## Validation
- Red first: `uv run --python 3.12 pytest tests/test_rerank.py -k 'xlm_roberta_two_layer_head or roberta_position_ids_use_padding_offset' -q` failed with `KeyError: 'classifier.weight'` and missing `_position_ids_for_config`
- `uv run --python 3.12 pytest tests/test_rerank.py -k 'xlm_roberta_two_layer_head or roberta_position_ids_use_padding_offset' -q` = `2 passed`
- `uv run --python 3.12 pytest tests/test_rerank.py -q` = `41 passed`
- `uv run --python 3.12 ruff check vllm_mlx/rerank_forward.py tests/test_rerank.py` = pass
- `uv run --python 3.12 black --check vllm_mlx/rerank_forward.py tests/test_rerank.py` = pass
- `git diff --check` = pass
- `/opt/ai-runtime/bin/upstream-local-repro-preflight --local-root /opt/ai-runtime/worktrees/vllm-mlx/issue-545-xlm-roberta-rerank --upstream-root /opt/ai-runtime/worktrees/vllm-mlx/upstream-main-clean-545 --code-path vllm_mlx/rerank_forward.py --code-path tests/test_rerank.py --no-model-artifact` = pass

## Not claimed
- This does not add a new reranker API surface.
- This does not claim a full live BGE/Jina/Qwen3-Reranker quality benchmark was run.
- This does not change non-RoBERTa BERT position IDs or flat classifier heads.

Refs #545.

### DIFF
--- a/tests/test_rerank.py
+++ b/tests/test_rerank.py
@@ -425,6 +425,59 @@ class TestClassifierForward:
                 config,
             )
 
+    def test_classifier_forward_supports_xlm_roberta_two_layer_head(self):
+        """XLM-RoBERTa rerankers use dense -> tanh -> out_proj classifier heads."""
+        import mlx.core as mx
+
+        from vllm_mlx.rerank_forward import classifier_forward
+
+        config = {
+            "model_type": "xlm-roberta",
+            "hidden_size": 8,
+            "num_attention_heads": 2,
+            "intermediate_size": 16,
+            "num_hidden_layers": 1,
+            "num_labels": 1,
+            "vocab_size": 20,
+            "max_position_embeddings": 32,
+            "type_vocab_size": 2,
+            "layer_norm_eps": 1e-12,
+            "hidden_act": "gelu",
+            "pad_token_id": 1,
+        }
+        weights = _rename_bert_prefix(_make_bert_weights(config), "xlm-roberta")
+        del weights["classifier.weight"]
+        del weights["classifier.bias"]
+        weights["classifier.dense.weight"] = mx.random.normal((8, 8)) * 0.02
+        weights["classifier.dense.bias"] = mx.zeros((8,))
+        weights["classifier.out_proj.weight"] = mx.random.normal((1, 8)) * 0.02
+        weights["classifier.out_proj.bias"] = mx.zeros((1,))
+
+        logits = classifier_forward(
+            mx.array([[1, 2, 3]]),
+            mx.array([[1, 1, 1]]),
+            weights,
+            config,
+        )
+        mx.eval(logits)
+
+        assert logits.shape == (1, 1)
+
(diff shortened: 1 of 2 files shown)


janhilgard (approved):
## Code Review

CI all green. Clean diff (+99/-4), well-scoped.

### What it does

1. **Two-layer classifier head**: XLM-RoBERTa rerankers (BGE, Jina v2) use `classifier.dense` + `classifier.out_proj` instead of flat `classifier.weight`/`classifier.bias`. The new `_classification_head_forward` detects which layout is present via key existence and dispatches accordingly.

2. **RoBERTa position IDs**: RoBERTa-family models reserve the pad token position and offset real token positions (`padding_idx + 1` based). The new `_position_ids_for_config` correctly computes `cumsum(mask) * mask + padding_idx` for RoBERTa/XLM-RoBERTa, while keeping plain `arange(seq_len)` for BERT.

### Assessment

**Correctness:** The position ID computation matches the HuggingFace `RobertaEmbeddings.create_position_ids_from_input_ids` implementation. The two-layer head (dense → tanh → out_proj) matches the `RobertaClassificationHead` from transformers.

**Backward compatibility:** Existing BERT flat classifier heads are unaffected — `_classification_head_forward` falls through to the original `classifier.weight`/`classifier.bias` path when `classifier.dense.weight` is absent. BERT position IDs still use `arange(seq_len)`.

**Model type detection:** Checking for `roberta`, `xlm-roberta`, and `xlm_roberta` covers the known variants. The lowercase normalization handles case mismatches.

**Tests:** Two focused unit tests — one for the two-layer head shape, one for the position ID offset computation with padding. The `_rename_bert_prefix` helper is a clean reuse pattern.

### One minor note (non-blocking)

The `_classification_head_forward` hardcodes `tanh` activation between dense and out_proj. This matches the standard `RobertaClassificationHead`, but some fine-tuned variants might use different activations. Fine as-is — if such a model appears, it can be extended then.

LGTM.

Issue #535, opened 2026-05-15
Bug: Vision capabilities/tool calling in vllm-mlx vs mlx-vlm?


harish0201:
Hi!

I'm having a really confusing issue that is ongoing, and I'm unsure on how to trouble-shoot it. 

I'm running into this weird issue where when I use vllm-mlx, the LLM that I'm using is unable to see the image, but when I'm using mlx-vlm the LLM can parse through the image. In both the cases, I'm using openweb-ui with native tool calling enabled for Qwen3.6-35B-A3B-oQ4 quant retaining the multimodal capabilities.

This is how I'm running the two methods, tags within <> are me trying to show the first few lines.
```
vllm-mlx serve 
      --port 9008 --reasoning-parser qwen3
      --enable-auto-tool-choice --tool-call-parser auto
      --use-paged-cache --gpu-memory-utilization 0.9
      --ssd-cache-dir /Users/$USER/.vllm
       Qwen3.6-35B-A3B-oQ4
```
 
output with vllm-mlx:
```
<Thinking process>
The user has attached an image file called "Flowchart.png" and wants me to explain it. I need to look at this image first. However, I don't have a direct image viewing tool available. Let me check what tools I have available.

Looking at my available tools, I don't see a direct way to view or process images. But I can use the execute_code tool to potentially analyze the image if I had image processing capabilities, but that would require libraries that may not be available in Pyodide. ...... goes on
</Thinking process>
<response>
I notice you've attached an image file called "Flowchart.png", but I don't have the ability to directly view or analyze images in this conversation.However, I can still help you! Here are a few options:1. Describe the flowchart to me - If you tell me what steps, processes, or stages are shown in the flowchart, ........ goes on
</response>
```

With mlx-vlm:
```
mlx_vlm.server
      --port 9009
      --model Qwen3.6-35B-A3B-oQ4
```

output with mlx-vlm, seems to have no thinking (probably I didn't enable it explicitly?)
```
<response>
This image shows a detailed workflow pipeline. Let me break it down step by step:## .... goes on
</response>
```

Any ideas on what is happening? In both the cases I'm using openweb-ui to interact with the served model.

waybarrios:
Thanks for the detailed report.

[...]

PR #520, opened 2026-05-09
detect MLLM via config.json with substring matcher as fallback

waybarrios:
## Summary

Fixes #516. `is_mllm_model` previously routed any path whose lowercased form contained one of the `MLLM_PATTERNS` substrings (e.g. `vl-`, `qwen3.5-`, `gemma-3`) through the multimodal loader, regardless of what the model actually was. A user with a text-only Qwen3 stored under a path like `~/dl-vl-experiments/qwen3-32b/` would hit `mlx_vlm.load()` and crash with `ValueError: Model type qwen3 not supported`.

This PR keeps two validations and gives priority to the model's own metadata when it is available on disk:

1. **`config.json` inspection** (new, primary): when the input resolves to a local directory with a readable `config.json`, the model's `architectures` field and modality config keys (`vision_config`, `audio_config`, `vision_tower`, `image_token_id`, etc.) decide.
2. **Legacy `MLLM_PATTERNS` substring match** (kept as fallback): used when the input is a HuggingFace repo ID, when no config is reachable, or when the file fails to parse.

`config.json` wins when present because the model declares what it actually is. Filesystem layout never overrides that.

Both helpers are exposed as `_check_legacy_string_patterns(model_name)` and `_try_read_config_json(name_or_path)` / `_config_indicates_vlm(config)`, which keeps each validation independently testable.

## Changes

- `vllm_mlx/api/utils.py`:
  - `_VLM_CONFIG_KEYS`: keys that indicate multimodality (vision/audio).
  - `_VLM_ARCHITECTURE_KEYWORDS`: substrings inside `architectures` entries that identify VLMs (covers both `*ForConditionalGeneration` and the few VLMs that use `*ForCausalLM` like `Phi3V`, `Molmo`, `CogVLM`, `InternVL`, `MultiModalityCausalLM`, etc.).
  - `_try_read_config_json`: bounded read (1 MiB cap, swallows OS / parse errors, returns `None` for non-directories or repo IDs).
  - `_config_indicates_vlm`: defensive type checks, no false positives on `T5ForConditionalGeneration` and similar text seq2seq models.
  - `_check_legacy_string_patterns`: extracted from previous `is_mllm_model` body.
  - `is_mllm_model`: orchestrates the two with config-priority.
- `tests/test_api_utils.py`: 13 new regression tests in `TestIsMllmModelConfigPriority`.

## Test plan

[...]

PR #551, opened 2026-05-19
Keep MLLM media stream on owner thread

Thump604:
## Summary

Keep media-bearing SimpleEngine MLLM requests on the current `mlx_vlm.stream_chat()` thread instead of routing them through the generic blocking worker when an auxiliary TextModel route exists.

## Local repro

Issue #535 reports this startup/request shape:

- model is a Qwen3.6 VLM-capable local quant
- `--mllm` / VLM auto-detection enables SimpleEngine MLLM routing
- text-only routing builds an auxiliary TextModel
- a media-bearing chat request logs `Media request → MLLM path`
- the request then fails with `There is no Stream(gpu, 1) in current thread.`

The focused regression added here models that exact route split: `_is_mllm=True`, `_text_model` present, media content present, and `mlx_vlm.stream_chat()` requiring its owner/current thread. It fails on upstream main because the media route calls `_run_blocking_serialized()` and moves `stream_chat()` to a generic worker.

- `vllm_mlx/engine/simple.py::SimpleEngine.stream_chat`
- current upstream main `7e304840713c2abce8ea1fb40cb960a740c01ce2`

## Change

- Preserve the existing TextModel fast path for text-only MLLM requests.
- Keep media-bearing MLLM requests on the current `mlx_vlm.stream_chat()` path under the existing generation lock.
- Keep the existing text-only fallback behavior when no TextModel route exists.

## Validation

- Red/green focused regression: `tests/test_simple_engine_mllm_media_thread.py`
- `uv run --python 3.12 python -m pytest tests/test_simple_engine.py tests/test_simple_engine_mllm_media_thread.py -q` — 39 passed
- `uv run --python 3.12 python -m pytest tests/test_api_utils.py -q` — 93 passed
- `uv run --python 3.12 python -m pytest tests/test_server.py -k 'mllm or stream_thread or residency_preload' -q` — 5 passed
- `uv run --python 3.12 ruff check vllm_mlx/engine/simple.py tests/test_simple_engine_mllm_media_thread.py`
- `uv run --python 3.12 black --check vllm_mlx/engine/simple.py tests/test_simple_engine_mllm_media_thread.py`
- `git diff --check`
- Runtime upstream-local-repro preflight against clean upstream main: `/opt/ai-runtime/run/upstream-local-repro-preflight/20260519T225531Z-upstream-local-repro-preflight.json`

## What this does not claim

- This does not claim full live verification against the reporter's exact 35B quantized weights.
- This does not change model detection; config.json-based VLM detection is already on current main via #520.
- This does not change the separate TextModel owner-thread PR #543.


janhilgard (approved):
## Code Review

CI all green. Minimal diff (+61/-4), single logical change.

### What it does

Fixes the condition at line 764 in `SimpleEngine.stream_chat` so that media-bearing MLLM requests stay on the current thread instead of being routed through `_run_blocking_serialized` (worker thread), which breaks MLX stream ownership (`Stream(gpu, N)` mismatch).

### Before/After

```python
# Before (broken for media + text_model):
if self._text_model is None and not has_media_content(messages):
    # current thread ← only text-only without TextModel

# After (correct):
if has_media_content(messages) or self._text_model is None:
    # current thread ← media always, or text-only without TextModel
```

### Routing analysis

The full routing chain in `stream_chat` is:

1. **L729-749**: `_text_model` exists AND text-only → `_stream_generate_text` (TextModel fast path)
2. **L752+, L764 (this change)**: MLLM path:
   - `has_media_content` OR `_text_model is None` → current thread under `_generation_lock` ✅
   - else → `_run_blocking_serialized` (only reachable for text-only MLLM without TextModel, which is actually dead code after check at L729-749, but harmless)

The condition change is logically correct: media requests must always stay on the owner thread for MLX stream safety. Text-only requests with a TextModel are already handled by the early return at L749.

### Test

The regression test accurately models the failure: `FakeMllmModel.stream_chat` asserts `threading.get_ident()` matches the owner thread and raises the exact error from #535. The test also asserts `_run_blocking_serialized` is never called for media requests.

LGTM.

waybarrios:
The routing fix itself is right, media requests do need to stay on the owner thread. The problem is that #540 landed after this PR was opened and they now collide: #540 added fail-fast admission via `_acquire_generation_slot`, and the current-thread branch here takes the raw lock instead:

```python
# this PR (simple.py, current-thread branch)
async with self._generation_lock:   # raw lock: second media request queues
    for chunk in self._model.stream_chat(...):
        ...
```

Two consequences. First, two concurrent media requests queue on the lock instead of the second one getting the 503 `text_generation_busy` that #540 introduced. Second, `_active_requests` is never populated on this path, so when a text request fails fast while a media request holds the lock, the EngineBusy error reports `active=none`, which makes it useless for debugging. Repro sketch:

```python
import asyncio

async def main():
    engine = SimpleEngine("mlx-community/Qwen2.5-VL-3B-Instruct-4bit", force_mllm=True)
    await engine.start()
    media_msg = [{"role": "user", "content": [
        {"type": "image_url", "image_url": {"url": "..."}},
        {"type": "text", "text": "describe this"},
    ]}]

    async def consume():
        async for _ in engine.stream_chat(media_msg):
            pass

    # expected since #540: second request raises EngineBusy (503 text_generation_busy)
    # actual with this PR: it silently queues behind the first
    await asyncio.gather(consume(), consume())

asyncio.run(main())
```

The fix should be small: switch that branch to `async with self._acquire_generation_slot(request_id):` (or add the same fail-fast check before taking the lock). It would also be good if the test asserted the fail-fast rejection for a second concurrent media request, not just thread identity. Happy to merge after that.

waybarrios:
I refreshed this branch on current main and addressed the admission feedback. Media and MLLM draft streams now stay on the owner thread and use fail-fast admission, while native video remains on the worker so it does not block the event loop.

I also moved the regressions into `tests/test_simple_engine.py`, with coverage for thread ownership, concurrent media rejection, draft routing, and native-video responsiveness.

waybarrios (approved):
The owner-thread routing, fail-fast admission, and native-video worker path are handled correctly. The regression tests cover each route.

Issue #508, opened 2026-05-07
Feature request: adaptive idle polling and optional weight unload for the serve loop


jroth1111:
## Summary

Two related quality-of-life requests for `vllm-mlx serve` so the
OpenAI/Anthropic-compatible HTTP server can stay running indefinitely on
Apple Silicon without consuming appreciable CPU or memory between
requests:

1. **Light: adaptive idle polling.** Make `EngineConfig.step_interval`
   adaptive (or event-driven) so the async loop does not poll at 1 kHz
   when the scheduler is empty.
2. **Heavy: lazy load + idle unload.** Optional `--lazy-load` and
   `--idle-unload-seconds N` flags on `serve` to release model weights
   after a configurable idle timeout and reload on next request.

Goal is *minimal* idle footprint, not literally zero.

### 1 ms polling loop

`vllm_mlx/engine_core.py`:

```python
@dataclass
class EngineConfig:
    ...
    step_interval: float = 0.001  # 1ms between steps

while self._running:
    if self.scheduler.has_requests():
        output = await loop.run_in_executor(_executor, self.scheduler.step)
        ...
    else:
        await asyncio.sleep(step_interval)
```

With no requests in flight, the loop wakes ~1000×/s. `step_interval` is
not exposed as a CLI flag.

### Eager load, no unload

`server.load_model(...)` builds the global `BatchedEngine` before
`uvicorn.run(...)`. Lifespan startup calls `await _engine.start()`,
which loads weights through `load_model_with_fallback(...)` on the
dedicated `mlx-step` executor. `BatchedEngine.stop()` already tears
the engine down cleanly — but only on process exit.

`vllm_mlx/worker.py` carries the inherited vLLM `sleep`/`wake_up` stubs
that no-op with a comment that sleep mode is "not applicable for MLX
(unified memory)". That is correct for the vLLM page-GPU-to-CPU
semantics, but does not address the user-facing goal of releasing
weight RAM during long idle.

### Light tier (small PR, no API surface)

Replace the fixed `asyncio.sleep(step_interval)` idle branch with either
an asyncio `Event` woken by `add_request(...)`, or a two-speed interval
(`step_interval` while requests are running, `idle_step_interval` such
as 100 ms or 1 s when empty). Either should drop idle CPU by ~50–500×
without changing first-token latency.

### Heavy tier

```bash
vllm-mlx serve MODEL \
  --lazy-load \
  --idle-unload-seconds 1800
```

[...]

PR #552, opened 2026-05-20
Reduce EngineCore idle polling

Thump604:
## Summary

Implements the light-tier idle polling part of #508.

- Adds `EngineConfig.idle_step_interval` with a 100ms default for the empty-scheduler path.
- Adds a request event that `add_request()` sets so the idle loop wakes immediately when new work arrives.
- Keeps the active generation path unchanged: scheduler steps still run as before when requests are present.

## Observed behavior

The current empty-scheduler branch sleeps on the active `step_interval` value, defaulting to 1ms. That keeps the serve loop waking roughly 1000 times per second even when there is no queued work.

## Expected behavior

When the scheduler is empty, the loop should wait on a lower-frequency idle interval while still waking promptly on a new request.

## Minimal patch shape

This is intentionally limited to `EngineCore` idle-loop behavior and regression tests. It does not add a CLI flag or new public API surface beyond the `EngineConfig` field.

## Validation

- `uv run --python 3.12 --extra dev pytest tests/test_engine_core_idle_polling.py tests/test_engine_core_thread_streams.py tests/test_batching.py::TestEngineAsync::test_engine_lifecycle tests/test_batching.py::TestEngineAsync::test_engine_context_manager tests/test_batching.py::TestEngineAsync::test_stream_outputs_consumer_break_after_finished_does_not_abort -q` -> 7 passed
- `uv run --python 3.12 --extra dev ruff check vllm_mlx/engine_core.py tests/test_engine_core_idle_polling.py --select E,F,W --ignore E402,E501,E731,F811,F841` -> passed
- `uv run --python 3.12 --extra dev black --check --target-version py312 vllm_mlx/engine_core.py tests/test_engine_core_idle_polling.py` -> passed
- `python3 /opt/ai-runtime/bin/lint-upstream-claims --root /opt/ai-runtime/worktrees/vllm-mlx/issue-508-adaptive-idle-polling vllm_mlx/engine_core.py tests/test_engine_core_idle_polling.py` -> passed
- `git diff --check` -> passed

## Not claimed

- Does not implement lazy load, idle unload, `/health` state transitions, or memory release semantics.
- Does not change model loading, decoding, scheduling policy, streaming semantics, or request output behavior.
- Does not claim a measured CPU reduction; it only changes the empty-scheduler wait behavior and proves request wakeup semantics in unit tests.

### DIFF
--- a/tests/test_engine_core_idle_polling.py
+++ b/tests/test_engine_core_idle_polling.py
@@ -1,0 +1,68 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Regression tests for EngineCore idle polling behavior."""
+
+import pytest
+
+
+class _IdleScheduler:
+    def has_requests(self):
+        return False
+
+    def _close_batch_generator(self):
+        pass
+
+
+class _RequestCapturingScheduler:
+    def __init__(self):
+        self.requests = []
+
+    def add_request(self, request):
+        self.requests.append(request)
+
+
+@pytest.mark.anyio
+async def test_engine_loop_uses_idle_interval_when_scheduler_is_empty(monkeypatch):
+    """An empty scheduler should not keep polling at the active 1ms interval."""
+    import vllm_mlx.engine_core as engine_core
+    from vllm_mlx.engine_core import EngineConfig, EngineCore
+
+    engine = object.__new__(EngineCore)
+    engine.config = EngineConfig(step_interval=0.001, idle_step_interval=0.25)
+    engine.scheduler = _IdleScheduler()
+    engine._running = True
+    engine._request_event = None
+    engine._steps_executed = 0
+
+    sleeps = []
+
+    async def fake_sleep(delay):
+        sleeps.append(delay)
+        engine._running = False
+
+    monkeypatch.setattr(engine_core.asyncio, "sleep", fake_sleep)
+
+    await engine._engine_loop()
+
+    assert sleeps == [0.25]
+
+
+@pytest.mark.anyio
+async def test_add_request_wakes_idle_engine_loop():
+    """Adding work should wake the idle loop instead of waiting for timeout."""
+    import asyncio
+
+    from vllm_mlx.engine_core import EngineConfig, EngineCore
+
+    engine = object.__new__(EngineCore)
+    engine.config = EngineConfig()
+    engine.scheduler = _RequestCapturingScheduler()
+    engine._output_collectors = {}
+    engine._stream_states = {}
+    engine._finished_events = {}
+    engine._request_event = asyncio.Event()
+
+    request_id = await engine.add_request("hello", request_id="req-1")
+
+    assert request_id == "req-1"
+    assert len(engine.scheduler.requests) == 1
+    assert engine._request_event.is_set()
--- a/tests/test_engine_core_idle_polling.py
+++ b/tests/test_engine_core_idle_polling.py
@@ -1,6 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
 """Regression tests for EngineCore idle polling behavior."""
 
+import asyncio
+from types import SimpleNamespace
+
 import pytest
 
 
@@ -20,14 +23,26 @@ class _RequestCapturingScheduler:
         self.requests.append(request)
 
 
+class _ParkingEvent(asyncio.Event):
+    """Signals once the engine has reached its event-backed idle wait."""
+
+    def __init__(self):
+        super().__init__()
+        self.wait_started = asyncio.Event()
+
+    async def wait(self):
+        self.wait_started.set()
+        return await super().wait()
+
+
 @pytest.mark.anyio
 async def test_engine_loop_uses_idle_interval_when_scheduler_is_empty(monkeypatch):
-    """An empty scheduler should not keep polling at the active 1ms interval."""
+    """An empty scheduler should use its configured idle wait."""
     import vllm_mlx.engine_core as engine_core
     from vllm_mlx.engine_core import EngineConfig, EngineCore
 
     engine = object.__new__(EngineCore)
-    engine.config = EngineConfig(step_interval=0.001, idle_step_interval=0.25)
+    engine.config = EngineConfig(step_interval=0.25)
     engine.scheduler = _IdleScheduler()
     engine._running = True
     engine._request_event = None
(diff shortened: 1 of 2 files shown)


janhilgard (approved):
## Code Review

CI all green. Clean diff (+108/-3), narrowly scoped to idle polling behavior.

### What it does

Replaces the tight 1ms `asyncio.sleep(step_interval)` idle loop with a 100ms `asyncio.Event.wait(timeout)` pattern. When the scheduler has no requests, the engine loop waits up to 100ms but wakes immediately when `add_request()` signals new work via the event.

### Assessment

**Correctness:** The `_wait_for_idle_or_request` helper correctly handles all edge cases:
- `request_event is None` → falls back to `asyncio.sleep` (backward compat for engines constructed without event)
- `timeout <= 0` → yields with `asyncio.sleep(0)`
- Normal path → `asyncio.wait_for(event.wait(), timeout)` with `TimeoutError` suppression
- Event is cleared in `finally` block to avoid stale signals

**Latency impact:** Zero impact on active generation — the idle interval only applies when `scheduler.has_requests()` returns False. New requests wake the loop immediately via `_set_request_event`, so request latency is bounded by event loop scheduling, not the 100ms timeout.

**Defensive `getattr`:** The code uses `getattr(self, "_request_event", None)` at all call sites. This handles the case where an EngineCore was constructed before this change (e.g., deserialized or subclassed without `__init__`). Slightly defensive but not wrong.

**Resource savings:** Reduces idle wakeups from ~1000/s to ~10/s with instant wake on new work. Meaningful for battery/thermal on Apple Silicon when the server is idle.

**Tests:** Two focused tests — one verifying the idle interval is used (not the active 1ms), one verifying `add_request` sets the event for immediate wake.

### Minor notes (non-blocking)

1. The `step_interval` variable is no longer read in `_engine_loop` after this change (the active path uses `asyncio.sleep(0)`, not `asyncio.sleep(step_interval)`). The `step_interval` field on `EngineConfig` is now effectively unused. Consider deprecating or documenting this in a follow-up.

2. The `_clear_request_event` call at the top of the `has_requests()` branch is correct — it prevents a stale event from being set during processing, ensuring the next idle wait actually blocks.

LGTM.

waybarrios:
The event logic looks right, I went hunting for a lost-wakeup race and didn't find one (`asyncio.Event.wait()` returns immediately if the event is already set, so the idle-to-active transition is safe). Two things before this can merge though.

First, the tests never exercise the new event-driven path. This test forces the fallback:

```python
# test_engine_loop_uses_idle_interval_when_scheduler_is_empty
engine._request_event = None   # forces the plain-sleep fallback,
                               # so _wait_for_idle_or_request's event wait is never covered
```

and `test_add_request_wakes_idle_engine_loop` only checks `engine._request_event.is_set()` without the loop running, so nothing verifies the loop actually wakes early. Something like this would cover the real behavior:

```python
async def test_idle_loop_wakes_on_add_request():
    engine = EngineCore(..., idle_step_interval=5.0)  # long idle sleep on purpose
    await engine.start()
    await asyncio.sleep(0.05)            # loop is now parked in the idle wait
    t0 = time.monotonic()
    await engine.add_request(request)    # should wake it immediately
    _ = await get_first_output(engine, request_id)
    assert time.monotonic() - t0 < 1.0   # would be ~5s if the wakeup were lost
```

Second, this changes `_engine_loop`, which is hot path, so it needs measured numbers per the repo's PR rules: TTFT and decode tok/s for main vs this branch on the same hardware (batch=1 and batch=8), plus the idle CPU% delta since that's the headline claim. The PR body currently opts out of measurement, which doesn't fly for this file.

One minor note while you're in there: `abort_request` doesn't set the event, so an abort that empties the scheduler leaves the loop sleeping for up to 100ms (it used to notice in about 1ms). Probably harmless, but worth a code comment so nobody trips on it later.

vllm_mlx/engine_core.py
```diff
@@ -36,13 +36,43 @@ def _is_stream_thread_error(error: Exception) -> bool:
...
+
+
 @dataclass
 class EngineConfig:
     """Configuration for the engine."""
 
     model_name: str = ""
     scheduler_config: Optional[SchedulerConfig] = None
     step_interval: float = 0.001  # 1ms between steps
+    idle_step_interval: float = 0.1  # 100ms idle wait when scheduler is empty
```
waybarrios: `EngineConfig` is publicly exported and its dataclass constructor accepts positional arguments. Inserting this field before `stream_interval` and `gpu_memory_utilization` silently remaps existing calls. For example, `EngineConfig("m", None, .001, 4, .8)` used to set `stream_interval=4` and `gpu_memory_utilization=.8`; it now sets `idle_step_interval=4` and `stream_interval=.8`. Append the new field after the existing fields, and add a compatibility test for the prior positional signature.

vllm_mlx/engine_core.py
```diff
@@ -222,8 +254,8 @@ def _close_batch_generator_on_worker() -> None:
             _bind_worker_streams_once()
             self.scheduler._close_batch_generator()
 
-        step_interval = self.config.step_interval
         stream_interval = self.config.stream_interval
+        idle_step_interval = self.config.idle_step_interval
```
waybarrios: `step_interval` becomes dead configuration here. Before this change it controlled the empty-scheduler sleep; after this change an explicit `EngineConfig(step_interval=...)` is ignored and replaced by the new 100 ms default. Either keep using `step_interval` as the idle timeout and update its default/comment, or provide a compatible migration. Do not leave an exported configuration value silently unused.

tests/test_engine_core_idle_polling.py
```diff
@@ -0,0 +1,68 @@
...
+async def test_engine_loop_uses_idle_interval_when_scheduler_is_empty(monkeypatch):
+    """An empty scheduler should not keep polling at the active 1ms interval."""
+    import vllm_mlx.engine_core as engine_core
+    from vllm_mlx.engine_core import EngineConfig, EngineCore
+
+    engine = object.__new__(EngineCore)
+    engine.config = EngineConfig(step_interval=0.001, idle_step_interval=0.25)
+    engine.scheduler = _IdleScheduler()
+    engine._running = True
+    engine._request_event = None
```
waybarrios, line 48: The headline wakeup path is not exercised: setting `_request_event=None` forces the fallback sleep, while the second test only asserts that `add_request()` sets the flag. I replaced `_wait_for_idle_or_request()` with unconditional `asyncio.sleep(timeout)` and both tests still passed. Run the real loop with a long timeout, wait until it is parked on the Event, call `add_request()`, and assert that the scheduler steps promptly.

waybarrios (changes requested):
Requesting changes for two public configuration compatibility regressions and one test gap. The Event wakeup implementation itself worked in an isolated idle-to-request transition test, but the current suite can still pass if the Event is ignored. The previously requested TTFT, decode throughput, and idle CPU measurements are also still missing for this hot-loop change.

waybarrios:
@Thump604, please update the PR to address the requested changes, then rerun the tests and performance validation. In particular, preserve `EngineConfig` compatibility, keep `step_interval` functional, and add a regression test that exercises the real idle-to-request Event wakeup. Please also include the requested TTFT, decode throughput, and idle CPU measurements.

Thump604:
Updated the branch on current `main` and addressed the three requested code/test changes in `a2117cf`:

- preserved the existing positional `EngineConfig` constructor by avoiding a field insertion;
- kept `step_interval` functional as the idle Event timeout (default 100 ms);
- added a real long-timeout loop regression that parks on the Event, calls `add_request()`, and proves `scheduler.step()` runs promptly.

Focused validation: `tests/test_engine_core_idle_polling.py` + `tests/test_engine_core_thread_streams.py` — 5 passed; py_compile and diff-check clean.

The requested TTFT, B=1/B=8 decode-throughput, and idle-CPU measurements are still pending. I am not claiming performance validation from the unit tests and will post the matched same-hardware results separately.

Issue #555, opened 2026-05-21
[Bug] gpt-oss-120b-MXFP4-Q8 crashes every request: "There is no Stream(gpu, N) in current thread"


CBribiescas:
## Summary

`vllm-mlx serve mlx-community/gpt-oss-120b-MXFP4-Q8` cannot generate a single token. Every `POST /v1/chat/completions` returns 500. The traceback:

```
File ".../vllm_mlx/engine/simple.py", line 354, in run_bound
    for response in stream_generate(model, tokenizer, prompt, **kwargs):
File ".../mlx_lm/generate.py", line 716, in stream_generate
File ".../mlx_lm/generate.py", line 442, in generate_step
    mx.eval([c.state for c in prompt_cache])
RuntimeError: There is no Stream(gpu, N) in current thread.
```

The stream number `N` varies (0, 1, 2, …) depending on how many GPU ops the main thread executed before the first request — same root cause. Flags tried, none recover: `--tool-call-parser gpt-oss`, `--max-num-seqs 1`, `--enable-auto-tool-choice`, combinations of the above.

`Qwen3-Coder-30B-A3B-Instruct-MLX-4bit` on the same `vllm-mlx serve` works (also MoE, so this isn't "MoE in general").

## Environment

| | |
|---|---|
| `vllm-mlx` | 0.3.0 (PyPI) |
| `mlx-lm` | 0.31.3 |
| Python | 3.11 (conda) |
| OS | macOS Tahoe (Darwin 25.5.0) |
| Host | Apple M5 Pro Max, 128 GB unified memory |
| Model | `mlx-community/gpt-oss-120b-MXFP4-Q8` (MoE, 128 local experts, MXFP4 + group-size 32 affine quant) |

## Reproducer

Single Python script, no server needed (subprocess-isolates two scenarios to avoid MLX state cross-talk):

```bash
pip install vllm-mlx==0.3.0 mlx-lm==0.31.3
python repro.py /path/to/mlx-community/gpt-oss-120b-MXFP4-Q8
```

Expected output:

```
=== outer summary (subprocess-isolated) ===
  A (load main, gen worker): FAIL
  B (load worker, gen worker): OK
  >>> BUG REPRODUCED: load-on-main + generate-on-worker crashes;
                       load+generate on same worker thread succeeds.
```

What each scenario does:

| Scenario | Load thread | Generate thread | Mirrors vllm-mlx? | Result |
|---|---|---|---|---|
[...]

PR #556, opened 2026-05-21
fix(engine): pin MLX load + inference to one worker thread (fixes gpt-oss-120b crash)

CBribiescas:
## Summary

`vllm-mlx serve mlx-community/gpt-oss-120b-MXFP4-Q8` cannot generate a single token — every `POST /v1/chat/completions` returns 500 with `RuntimeError: There is no Stream(gpu, N) in current thread.` This PR gives `SimpleEngine` a dedicated single-thread MLX worker (`ThreadPoolExecutor(max_workers=1)`) and pins load, blocking inference, and streaming (both LLM and MLLM paths) to it. Load thread == inference thread == one MLX stream namespace, so models whose `nn.Module.__init__` touches the GPU (e.g. gpt-oss-120b's `QuantizedSwitchLinear`) no longer leave the model parameter arrays bound to a stream the inference worker can't reach.

Issue: #555 (root-cause walkthrough + reproducer).
Companion docs PR: #557 (M5 docs update, split out per review).

## Root cause (one paragraph)

`mlx_lm/models/switch_layers.py:QuantizedSwitchLinear.__init__` calls `mx.quantize(mx.random.uniform(shape=(num_experts, output_dims, input_dims)), mode="mxfp4")` before saved-weight load. That GPU op runs on whichever thread `MLXLanguageModel(...)` is constructed on — historically the event-loop thread — and promotes that thread's default stream so the model parameter `mx.array`s carry pending ops queued against it. `_run_blocking_serialized` then routes `model.chat` through `asyncio.to_thread`, landing on the default `ThreadPoolExecutor` (a different worker each call). `_bind_worker_generation_streams()` patches `mlx_lm.generate.generation_stream` for the worker, but can't re-thread the parameter arrays' stream affinity. `mx.eval([c.state for c in prompt_cache])` at `mlx_lm/generate.py:442` then can't find the loader's stream → `RuntimeError`. The comment at the old `vllm_mlx/engine/simple.py:239-242` explicitly predicted this class of issue ("Keep model load on the event-loop thread so default LLM stream_generate() runs on the same thread that owns model-associated streams") — but the streaming path doesn't actually run on the event-loop thread for all routes. This PR makes the invariant true unconditionally by owning one worker.

## Commits

1. **`180e3e7` engine: pin MLX load + inference to one worker thread**
   - `SimpleEngine.__init__` allocates `self._mlx_executor = ThreadPoolExecutor(max_workers=1, thread_name_prefix="mlx-worker")` and `self._mlx_thread_id: int | None = None`.
   - **`prepare_for_start`** self-routes: if the caller is on the MLX worker (re-entry from an inference path), runs inline; otherwise `self._mlx_executor.submit(self._mlx_marked_prepare).result()`. `_mlx_marked_prepare` records the worker thread id on first run. Callers on any thread — including `ResidencyManager._prepare_engine_start` in `lifecycle.py`, which calls `engine.prepare_for_start()` directly on the event-loop thread — end up with load running on the same single worker with **no `lifecycle.py` changes needed**.
   - **`_run_blocking_serialized`** routes through the engine's executor.
   - **`stream_generate`** body drives the sync iterator on the same executor via `_stream_in_mlx_executor`, piping chunks back to the event loop through `queue.Queue` + `await loop.run_in_executor(None, q.get)`. Event loop is no longer blocked synchronously during streaming.
   - **`stop`** shuts the executor down (`wait=False, cancel_futures=True`) and rebuilds it so re-`start` works in tests.

2. **`ad6364d` engine: extend MLLM `stream_chat` path through MLX worker too**
   - Same Stream(gpu, N) bug as the LLM path, different file. Hit live during a production migration when `gemma-4-E4B` (multimodal — loads through `mllm.py`) reproduced the error at `mllm.py:2161` via `mlx_vlm.generate.stream_generate`.
   - Routes the `_text_model is None and not has_media_content(messages)` branch in `stream_chat` through `_stream_in_mlx_executor` with the same pattern as the LLM path.

3. **`93effbf` engine: propagate cancel + drop redundant `wrap_future`** (review feedback)
   - **Cancel propagation in `_stream_in_mlx_executor`.** A consumer-side `threading.Event` is set in the async generator's `finally`; the producer checks it between chunks and exits early. Without this, a client disconnect leaves the producer generating tokens to completion on the single MLX worker, blocking all subsequent requests for the lifetime of the orphaned generation. **Verified live**: client disconnect after 2 chunks of an 800-token request → follow-up `'pong'` returns in **0.7 s** instead of waiting for the orphan to drain.
   - **Drop redundant `asyncio.wrap_future`.** `loop.run_in_executor` already returns an `asyncio.Future`; `asyncio.shield(task)` and `await task` work on it directly. Simplified to `task = loop.run_in_executor(self._mlx_executor, run_bound)`.

4. **`1e98201` test(simple_engine): update MLLM-stream-thread test for executor-pinning fix**
   - The existing regression `test_mllm_nonstream_text_only_without_text_model_keeps_stream_thread_owner` bypassed `prepare_for_start` and constructed `FakeMllmModel` on the test thread, asserting `stream_chat` ran on that thread. Under the new pinning behavior the model's owner is the MLX worker thread. Updated the test to construct `FakeMllmModel` via `prepare_for_start` under a patched `MLXMultimodalLM`, mirroring real load behavior. Same invariant — model owner thread == `stream_chat` thread — with the right owner.

`BatchedEngine` was not inspected; if it has the same architecture it would need the same shape — happy to send as a follow-up.

## Verification

Host: Apple **M5 Pro Max**, 128 GB unified memory, macOS Tahoe (Darwin 25.5.0).

- **Live smoke** against the actual misbehaving model:
  - `mlx-community/gpt-oss-120b-MXFP4-Q8` — chat: 200 OK with real text + `finish_reason`. Streaming: full SSE stream with `data: [DONE]` terminal. Three sequential warm-cache requests in <1 s each. No `Stream(gpu, N)` errors anywhere in server log.
  - `lmstudio-community/Qwen3-Coder-30B-A3B-Instruct-MLX-4bit` — `"pong"` with `finish_reason="stop"`. No regression.
  - `lmstudio-community/gemma-4-E4B-it-MLX-4bit` (MLLM-loaded) — `"pong"` with `finish_reason="stop"` via the MLLM text-only fallback path.
- **Cancel propagation** (commit 3): client disconnect mid-stream → follow-up request returns in 0.7 s; without the fix, follow-up would queue behind the orphaned generation.
- **Project test suite** on this branch:
  - Full fast suite (`pytest --ignore=test_lifecycle_cli.py`): **2070 passed, 0 failed**.
  - Targeted: `test_simple_engine` (38) + `test_simple_engine_cancel_serialization` (3) + `test_engine_core_stream_safety` (1) + `test_engine_core_thread_streams` (2) + `test_server::test_streaming_chat_no_stream_thread_error_after_residency_preload` (1) → **45/45 pass**.
  - `test_lifecycle_cli.py` has 3 pre-existing failures on `main` unrelated to this change (`'SimpleNamespace' object has no attribute 'models_config'` in `vllm_mlx/cli.py:37`); flagging FYI for a separate fix.

## Notes

- **Cancellation semantics in `_run_blocking_serialized`** preserved (`asyncio.shield(task) + on_cancel`), now operating directly on the `asyncio.Future` from `loop.run_in_executor`.
- **Streaming back-pressure**: the queue is unbounded; bounded only by `max_tokens` total memory (each chunk is a small `SimpleNamespace`). If you'd prefer bounded with backpressure, easy follow-up.
- **Alternative considered**: fix upstream in `mlx-lm` by lazy-init'ing `QuantizedSwitchLinear` so `__init__` doesn't touch the GPU. That would fix this for every consumer of mlx-lm, not just vllm-mlx, but it's a much wider surface. Happy to defer to your call.

## Test plan

- [ ] CI: full fast suite green
- [ ] Manual: `vllm-mlx serve` a gpt-oss-120b checkpoint, hit `/v1/chat/completions` with both `stream=false` and `stream=true`
- [ ] Manual: same against a non-MoE model (Qwen3-Coder-30B) and a multimodal model loaded as text-only (gemma-4-E4B-IT) to confirm both fixed paths and non-regression
- [ ] Manual: client-disconnect mid-stream → next request not blocked

### DIFF
--- a/tests/test_simple_engine.py
+++ b/tests/test_simple_engine.py
@@ -1152,11 +1152,16 @@ class TestSimpleEngineConcurrency:
     async def test_mllm_nonstream_text_only_without_text_model_keeps_stream_thread_owner(
         self,
     ):
-        """MLLM text-only non-stream path must keep stream_chat on model thread.
+        """MLLM text-only non-stream path must keep stream_chat on the same
+        thread as model construction (the SimpleEngine MLX worker).
 
-        Regression: aggregate_stream_chat -> stream_chat used _run_blocking_serialized,
-        which moved mlx_vlm stream generation to a worker thread and could raise
-        "There is no Stream(gpu, N) in current thread".
+        Regression: aggregate_stream_chat -> stream_chat used
+        _run_blocking_serialized, which moved mlx_vlm stream generation to
+        a different worker thread per call and could raise "There is no
+        Stream(gpu, N) in current thread". Now both load and inference are
+        pinned to ``self._mlx_executor`` (a single-thread executor), so the
+        model's "owner thread" is the executor thread and stream_chat must
+        run on that same thread.
         """
         from types import SimpleNamespace
 
@@ -1164,7 +1169,12 @@ class TestSimpleEngineConcurrency:
 
         class FakeMllmModel:
             def __init__(self):
+                # Captured on whichever thread SimpleEngine constructs us,
+                # which under the executor-pinning fix is the MLX worker.
                 self._owner_thread = threading.get_ident()
+
+            def load(self):
+                pass  # SimpleEngine.prepare_for_start calls this; no-op here
 
             def stream_chat(self, **kwargs):
                 if threading.get_ident() != self._owner_thread:
@@ -1175,15 +1185,23 @@ class TestSimpleEngineConcurrency:
                     prompt_tokens=3,
                 )
 
-        engine = SimpleEngine("test-model", force_mllm=True, mtp=False)
-        engine._loaded = True
-        engine._text_model = None
-        engine._model = FakeMllmModel()
+        # Patch the MLLM model class so SimpleEngine constructs FakeMllmModel
+        # via prepare_for_start on the MLX worker thread (matching prod load
+        # behavior), then the same executor runs stream_chat — single-thread
+        # invariant holds.
+        with patch(
+            "vllm_mlx.models.mllm.MLXMultimodalLM",
+            lambda *a, **k: FakeMllmModel(),
+        ):
+            engine = SimpleEngine("test-model", force_mllm=True, mtp=False)
+            engine.prepare_for_start()  # constructs FakeMllmModel on MLX worker
+            engine._loaded = True
+            engine._text_model = None
 
-        output = await engine.chat(
-            messages=[{"role": "user", "content": "Count: one, two, three"}],
-            max_tokens=16,
-        )
+            output = await engine.chat(
+                messages=[{"role": "user", "content": "Count: one, two, three"}],
+                max_tokens=16,
+            )
 
         assert output.text == "one, two, three"
         assert output.finish_reason == "stop"
--- a/vllm_mlx/engine/simple.py
+++ b/vllm_mlx/engine/simple.py
@@ -460,15 +460,23 @@ class SimpleEngine(BaseEngine):
         Mirrors the threading constraint enforced by ``prepare_for_start``:
         the model was loaded on the MLX worker, so iteration must happen on
         the same worker for MLX thread-local stream ownership to hold.
+
+        Cancellation: the consumer sets ``cancel_event`` in its ``finally``
+        block. The producer checks the event between chunks and breaks out
+        early so a disconnected client doesn't keep the single-threaded
+        ``_mlx_executor`` busy generating tokens nothing will consume.
         """
         import queue as _queue
 
         q: _queue.Queue = _queue.Queue()  # unbounded; max_tokens caps memory
+        cancel_event = threading.Event()
 
         def producer():
             try:
                 _bind_worker_generation_streams()
                 for chunk in factory():
+                    if cancel_event.is_set():
+                        break
                     q.put(("c", chunk))
             except BaseException as exc:
                 q.put(("e", exc))
@@ -486,8 +494,13 @@ class SimpleEngine(BaseEngine):
                     raise payload
                 yield payload
         finally:
-            # If the consumer broke early, the producer keeps pushing to an
-            # unbounded queue; it will exit on its own when the model
+            # Signal the producer to stop ASAP if the consumer is bailing
+            # (client disconnect, exception, or early break). Without this
+            # the producer continues generating tokens to completion on the
+            # single MLX worker thread and blocks all subsequent requests.
+            cancel_event.set()
+            # Producer will still emit the trailing ("d", None) sentinel after
+            # observing the cancel between chunks; no need to wait for fut.
             # iterator finishes or raises. No further action required here.
             if not fut.done():
                 pass
@@ -509,9 +522,10 @@ class SimpleEngine(BaseEngine):
             # asyncio.to_thread uses the default thread pool, landing on a
             # different worker per call and breaking model-parameter stream
             # affinity for models that touch the GPU during __init__.
+            # loop.run_in_executor already returns an asyncio.Future, so
+            # asyncio.shield/ensure_future work on it directly.
             loop = asyncio.get_running_loop()
-            future = loop.run_in_executor(self._mlx_executor, run_bound)
-            task = asyncio.ensure_future(asyncio.wrap_future(future))
+            task = loop.run_in_executor(self._mlx_executor, run_bound)
             try:
                 return await asyncio.shield(task)
             except asyncio.CancelledError:
--- a/vllm_mlx/engine/simple.py
+++ b/vllm_mlx/engine/simple.py
@@ -498,12 +498,10 @@ class SimpleEngine(BaseEngine):
             # (client disconnect, exception, or early break). Without this
             # the producer continues generating tokens to completion on the
             # single MLX worker thread and blocks all subsequent requests.
+            # The producer still emits the trailing ("d", None) sentinel
+            # after observing the cancel between chunks; no need to wait
+            # on the future explicitly.
             cancel_event.set()
-            # Producer will still emit the trailing ("d", None) sentinel after
-            # observing the cancel between chunks; no need to wait for fut.
-            # iterator finishes or raises. No further action required here.
-            if not fut.done():
-                pass
 
     async def _run_blocking_serialized(self, func, /, *args, on_cancel=None, **kwargs):
         """Run a blocking MLX operation under the generation lock.
--- a/vllm_mlx/engine/simple.py
+++ b/vllm_mlx/engine/simple.py
@@ -932,20 +932,25 @@ class SimpleEngine(BaseEngine):
             accumulated_text = ""
             token_count = 0
 
-            # Text-only fallback when no TextModel exists: keep execution on the
-            # current thread. Routing through to_thread can break mlx_vlm stream
-            # ownership on some models (Stream(gpu, N) mismatch).
+            # Text-only fallback when no TextModel exists: drive the sync
+            # iterator on the engine's dedicated MLX worker (same thread as
+            # load()), bridging chunks back via _stream_in_mlx_executor.
+            # mlx_vlm.stream_generate has the same Stream(gpu, N) thread-affinity
+            # bug as mlx_lm.stream_generate when run on a different thread than
+            # load() — see vllm-mlx PR #556 for the equivalent text-LLM fix.
             if self._text_model is None and not has_media_content(messages):
                 local_kwargs = mllm_call_kwargs()
 
                 async with self._generation_lock:
-                    _bind_worker_generation_streams()
-                    for chunk in self._model.stream_chat(
+                    stream_chat_kwargs = dict(
                         messages=messages,
                         max_tokens=max_tokens,
                         temperature=temperature,
                         tools=template_tools,
                         **local_kwargs,
+                    )
+                    async for chunk in self._stream_in_mlx_executor(
+                        lambda: self._model.stream_chat(**stream_chat_kwargs)
                     ):
                         token_count += 1
                         new_text = chunk.text if hasattr(chunk, "text") else str(chunk)
--- a/vllm_mlx/engine/simple.py
+++ b/vllm_mlx/engine/simple.py
@@ -13,6 +13,7 @@ import os
 import threading
 import time
 from collections.abc import AsyncIterator
+from concurrent.futures import ThreadPoolExecutor
 from typing import Any
 
 import mlx.core as mx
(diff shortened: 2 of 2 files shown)


janhilgard (commented):
## Code Review

CI: lint, type-check, test-matrix all pass. Apple Silicon runners still pending.

### Root cause analysis

Excellent diagnosis. `QuantizedSwitchLinear.__init__` calls `mx.quantize(mx.random.uniform(...))` which binds model parameter arrays to the loading thread's MLX stream. When `_run_blocking_serialized` routes inference through `asyncio.to_thread` (a different worker each call), the inference worker can't access those stream-bound arrays → `RuntimeError: There is no Stream(gpu, N) in current thread`.

### Core approach: dedicated MLX worker thread

Sound design. Pinning load + inference to a single `ThreadPoolExecutor(max_workers=1)` guarantees stream affinity. The approach is clean and the implementation touches only `SimpleEngine` internals.

### Issues found

**1. MLLM `stream_chat` inline path is not pinned to the MLX worker**

Lines 764-795 (current main) run `self._model.stream_chat()` synchronously on the event loop thread under `_generation_lock`:

```python
if self._text_model is None and not has_media_content(messages):
        _bind_worker_generation_streams()
        for chunk in self._model.stream_chat(...):  # ← event loop thread, not MLX worker
```

After this PR, model load happens on the MLX worker. If a future MLLM model has the same GPU-touching `__init__` pattern as gpt-oss-120b, this path would hit the same `Stream(gpu, N)` error. The fix should route this path through `_stream_in_mlx_executor` as well for consistency.

The `_run_blocking_serialized` MLLM path (lines 797-812) IS fixed by this PR since it now uses `_mlx_executor`. So only the inline streaming path is a gap.

**2. `_stream_in_mlx_executor` — producer doesn't stop on consumer cancellation**

```python
# If the consumer broke early, the producer keeps pushing to an
# unbounded queue; it will exit on its own when the model
# iterator finishes or raises. No further action required here.
if not fut.done():
    pass
```

When a client disconnects mid-stream, the producer keeps generating tokens to completion on the MLX worker thread. For long generations (`max_tokens=32768`), the MLX worker is occupied for minutes after the client is gone — and since `max_workers=1`, ALL other requests are queued behind it.

This is worse than the previous architecture where `asyncio.to_thread` would at least not block the _next_ request. Consider adding a `threading.Event` that the consumer sets on early exit, and the producer checks between chunks:

```python
cancel_event = threading.Event()

def producer():
    try:
        _bind_worker_generation_streams()
        for chunk in factory():
            if cancel_event.is_set():
                break
            q.put(("c", chunk))
    except BaseException as exc:
        q.put(("e", exc))
    finally:
        q.put(("d", None))

# In finally:
cancel_event.set()
```

**3. Redundant `asyncio.wrap_future` in `_run_blocking_serialized`**

```python
future = loop.run_in_executor(self._mlx_executor, run_bound)
task = asyncio.ensure_future(asyncio.wrap_future(future))
```

`loop.run_in_executor` already returns an `asyncio.Future`. `asyncio.wrap_future` on an already-asyncio Future is a no-op. This should simply be:

```python
task = loop.run_in_executor(self._mlx_executor, run_bound)
```

The `asyncio.shield(task)` and `task.add_done_callback` work on `asyncio.Future` directly.

**4. Two unrelated commits should be split**

The M5 docs update (14 files) is orthogonal to the engine fix. As the author suggests, it should be a separate PR to keep review focused.

### What works well

- `prepare_for_start` re-entry guard (`threading.get_ident() == self._mlx_thread_id`) is correct and handles `ResidencyManager` calling from the event loop thread
- `stop()` rebuilds the executor for test re-use
- The queue-based bridge in `_stream_in_mlx_executor` is a clean async/sync boundary pattern
- Cancellation semantics in `_run_blocking_serialized` are preserved (`asyncio.shield` + `on_cancel` callback)

### Verdict

The core fix is correct for the reported issue (LLM path with gpt-oss-120b). Point 2 (no cancellation propagation to producer) is the most important issue — a single cancelled request can block all subsequent requests for the lifetime of its generation. Points 1 and 3 are secondary.

janhilgard (changes requested):
Requesting changes for one blocking issue.

### Blocking: producer must stop on consumer cancellation

`_stream_in_mlx_executor` has no cancellation propagation. When a client disconnects mid-stream, the producer continues generating tokens to completion on the MLX worker. Since `max_workers=1`, this blocks ALL subsequent requests until the orphaned generation finishes (potentially minutes at `max_tokens=32768`).

Fix: add a `threading.Event` the consumer sets on early exit:

```python
async def _stream_in_mlx_executor(self, factory):
    import queue as _queue

    q: _queue.Queue = _queue.Queue()
    cancel = threading.Event()

    def producer():
        try:
            _bind_worker_generation_streams()
            for chunk in factory():
                if cancel.is_set():
                    break
                q.put(("c", chunk))
        except BaseException as exc:
            q.put(("e", exc))
        finally:
            q.put(("d", None))

    fut = self._mlx_executor.submit(producer)
    loop = asyncio.get_running_loop()
    try:
        while True:
            tag, payload = await loop.run_in_executor(None, q.get)
            if tag == "d":
                return
            if tag == "e":
                raise payload
            yield payload
    finally:
        cancel.set()
```

### Non-blocking notes (from previous review, still valid)

- Redundant `asyncio.wrap_future` — `loop.run_in_executor` already returns an asyncio Future
- MLLM `stream_chat` inline path (lines 764-795) not pinned to MLX worker — gap for future MLLM models with GPU-touching `__init__`
- M5 docs commit should be a separate PR

CBribiescas:
Thanks for the careful review @janhilgard. Pushed 3 commits addressing all three code items; replies inline.

### `735fb83` — MLLM `stream_chat` inline path (review item #1)

You're right that the inline path at the old lines 764–795 was a gap. I'd already hit this locally — during a production migration to drop Ollama for `gemma-4-E4B` (a multimodal model that loads through `mllm.py`), the same `Stream(gpu, N) in current thread` error fired from `mlx_vlm.generate.stream_generate` at `mllm.py:2161`. **Same root cause as the LLM path this PR originally fixed**: model loaded on one thread, sync iterator driven from another. I patched it locally with the same `_stream_in_mlx_executor` pattern at the time but hadn't pushed because the original PR was scoped to gpt-oss-120b — your review made it clear it belongs in this PR rather than a follow-up, so it's in now.

### `bf53f75` — cancel propagation (blocking, review item #2)

Implemented exactly as you suggested: `threading.Event` set by the consumer's `finally`, checked by the producer between chunks. Verified live on the actual stack:

- Client opens a streaming completion, reads 2 chunks (`max_tokens=800` request), then disconnects.
- Immediately fires a second non-streaming request to the same single-thread executor.
- **Before the fix:** the second request would queue behind the orphaned 800-token generation (multiple seconds blocked).
- **After the fix:** second request returns `'pong'` in **0.7 s** — the producer observed `cancel_event` between chunks and exited, freeing the worker.

Also bundled the second part of this commit:

### `bf53f75` — drop redundant `asyncio.wrap_future` (review item #3)

Correct, `loop.run_in_executor` already returns an `asyncio.Future`. Simplified to `task = loop.run_in_executor(self._mlx_executor, run_bound)`. `asyncio.shield(task)` and the `await task` in the cancel-handler both work directly.

### `3db91c4` — test update

The existing `test_mllm_nonstream_text_only_without_text_model_keeps_stream_thread_owner` regression test bypassed `prepare_for_start` and constructed `FakeMllmModel` on the test thread, then asserted that `stream_chat` ran on that same test thread. Under the new executor-pinning behavior the model's owner is the MLX worker thread, not the caller. Updated the test to construct `FakeMllmModel` via `prepare_for_start` under a patched `MLXMultimodalLM`, mirroring real load behavior. The test still asserts the same invariant — model owner thread == stream_chat thread — just with the correct owner now.

### Test results on the rebased branch

- `test_simple_engine` (38) + `test_simple_engine_cancel_serialization` (3) + `test_engine_core_stream_safety` (1) + `test_engine_core_thread_streams` (2) + `test_server::test_streaming_chat_no_stream_thread_error_after_residency_preload` (1) → **45/45 pass**
- Full fast suite (`pytest --ignore=test_lifecycle_cli.py`) → **2070 passed, 0 failed**

### Review item #4 — split M5 docs into separate PR

Will do this in a follow-up. Hold off rebase-then-force-push until you've had a chance to look at the other three changes, in case the rebase obscures the diff during re-review.

CBribiescas:
Update on review item #4: M5 docs commit is now its own PR — #557. Rebased this branch to drop `e0ac67f`; the force-push leaves this PR with only the 4 engine/test commits. Both PRs now reviewable independently.

janhilgard (approved):
All four review items addressed. CI fully green (lint, type-check, test-matrix 3.10–3.13, Apple Silicon 3.11+3.13).

Verified in the updated diff:

1. **MLLM `stream_chat` text-only fallback** (`stream_chat` lines 911+) now routes through `_stream_in_mlx_executor`, matching the LLM `stream_generate` fix. ✓

2. **Cancel propagation** — `threading.Event` checked between chunks in the producer, set in the consumer's `finally`. The live verification (0.7 s follow-up after mid-stream disconnect vs blocked-to-completion before) is convincing. ✓

3. **Redundant `asyncio.wrap_future`** dropped — `loop.run_in_executor` returns an `asyncio.Future` directly, `asyncio.shield(task)` works on it. ✓

4. **M5 docs** split into #557. ✓

5. **Test** correctly mirrors production load behavior: `FakeMllmModel` constructed via `prepare_for_start` on the MLX worker, asserting the same invariant (owner thread == `stream_chat` thread). ✓

One minor nit (non-blocking): the `if not fut.done(): pass` at `_stream_in_mlx_executor` line ~462 is dead code left from the pre-cancel-event version. Can be dropped in a follow-up or squash.

Thump604:
@CBribiescas this looks approved and CI-green, but GitHub currently reports it as conflicting against `main` (`mergeable=CONFLICTING`, `mergeStateStatus=DIRTY`). Could you rebase/refresh the branch when you have a chance so Wayner can evaluate it for merge without conflict handling? I am not asking for any scope change here.

CBribiescas:
@Thump604 rebased onto current `main` (force-push, --force-with-lease). One textual conflict in `simple.py` from the `_should_route_text_through_text_model` helper added in #507 — resolved by keeping both methods (they live in the same area but don't overlap semantically). All 45 targeted tests pass on the rebased branch; full fast suite still green.

Also addressed @janhilgard's nit (`0fd493e`): dropped the dead `if not fut.done(): pass` block in `_stream_in_mlx_executor` — leftover from the pre-cancel-event version.

Branch now 5 commits ahead of `main`. Ready for merge whenever Wayner gets to it.

CBribiescas:
Closing as superseded — `vllm-mlx==0.4.0rc1` resolves the original Stream(gpu, N) crash via your stream-rebind approach in \`vllm_mlx/mlx_streams.py\`, which is a smaller and more targeted fix than the engine-wide executor pinning this PR proposed. Verified the crash is gone on `gpt-oss-120b-MXFP4-Q8` with 0.4.0rc1.

Thanks @janhilgard for the careful review and the cancel-propagation catch — that pattern (Event between producer and consumer) is worth keeping in mind if executor-pinning ever shows up in a future PR.

The fork branch will stay on `CBribiescas/vllm-mlx` in case anything's useful for reference.

PR #561, opened 2026-05-21
feat: live inference KV cache quantization (QuaRot) and linear-attention state quantization

janhilgard:
## Summary

Two related cache-quantization features that reduce the memory footprint of Apple Silicon inference, with a focus on Qwen3.5/3.6 hybrid models. Each feature is independent and gated behind its own flag, so they can be reviewed and adopted separately. All new flags default off — fully backward compatible.

---

## Feature 1 — Live inference KV cache quantization (QuaRot)

`BatchQuantizedKVCache` quantizes the full-attention KV cache during generation.

- **Asymmetric K/V bits** — `--kv-k-bits` / `--kv-v-bits` accept 2/3/4/6/8-bit (keys generally need higher precision than values).
- **QuaRot Hadamard rotation** — `--kv-hadamard` applies `mx.hadamard_transform` before quantization. The transform is orthogonal (preserves dot products: `H(q)·H(k) = q·k`) and self-inverse, so the attention result is mathematically unchanged while outlier channels are spread uniformly across the head dimension.
- Operates directly on quantized data via `mx.quantized_matmul` — no dequantization at inference.
- Prefix cache compatible — `extract()` un-rotates, `merge()` re-rotates.

### Quality — cosine similarity vs FP16 reference (synthetic activations with outlier channels)

| Config | Plain affine | QuaRot | Gap closed |
|--------|-------------|--------|-----------|
| K=8 V=4 | 0.98689 | **0.99692** | 77% |
| K=8 V=3 | 0.96658 | **0.98716** | 62% |
| K=3 V=3 | 0.812   | **0.989**   | 94% |

QuaRot makes aggressive low-bit configs viable: plain K=3 V=3 collapses to 0.81 cosine similarity, QuaRot K=3 V=3 recovers 0.989 — at ~78% KV cache memory savings vs an FP16 baseline.

---

## Feature 2 — Linear-attention (GatedDeltaNet) state quantization

Qwen3.5/3.6 hybrid models alternate full-attention layers with GatedDeltaNet linear-attention layers in a 3+1 pattern, so **3/4 of all layers are linear attention** and their `ArraysCache` state (a recurrent matrix + a conv state) was previously never quantized:

| Model | Layers | Full-attn | GatedDeltaNet (linear) |
|-------|-------:|----------:|-----------------------:|
| Qwen3.6-27B | 64 | 16 | **48** |
| Qwen3.6-35B-A3B | 40 | 10 | **30** |

### 2a. `--linear-state-quantization` — prefix cache (recommended, low risk)

Quantizes the GatedDeltaNet `ArraysCache` when it is stored in the prefix cache: a one-time quantize on store, dequantize on fetch — no quantized matmul, so no rotation is needed and plain affine `mx.quantize` is used. A new `_QuantizedArraysCacheWrapper` records per-slot dtype so the recurrent matrix (float32) and conv state (bfloat16) each round-trip to their original precision. Default 8-bit; `--linear-state-quantization-bits` accepts 4/8.

**Per prefix-cache entry, 8-bit:**

| Model | Linear state — float32 | → 8-bit | Reduction |
|-------|----------------------:|--------:|----------:|
| Qwen3.6-27B | 153.9 MB | 41.7 MB | **3.69×** (−112 MB/entry) |
| Qwen3.6-35B-A3B | 64.4 MB | 17.5 MB | **3.68×** (−47 MB/entry) |

The prefix cache is memory-bounded, so this lets the same budget hold ~3.7× more linear state — higher hit rate rather than a fixed RAM saving.

### 2b. `--linear-state-bf16` — live recurrent state (opt-in, experimental)

Stores the GatedDeltaNet recurrent matrix in a 16-bit dtype during live inference, halving its footprint. Compute stays float32 (the patch upcasts on read; the Metal kernel templates on `state.dtype` and accumulates in float32 regardless), only the persisted tensor between decode steps is 16-bit. The dtype is selected by the `VLLM_MLX_LINEAR_STATE_DTYPE` env var (`bf16` | `fp16`, default `bf16`).

**A/B quality study** — 30 prompts, greedy decoding (`temperature=0`), float32 baseline vs 16-bit:

| Storage dtype | Token-exact match vs float32 | Correctness (12 deterministic prompts) | Mantissa bits |
|---------------|-----------------------------:|:--------------------------------------:|--------------:|
| bfloat16 | 13% (4/30) | 12/12 | 7 |
| **float16** | **27% (8/30)** | **12/12** | 10 |
| float32 | 100% (baseline) | 12/12 | 23 |

float16 (10-bit mantissa) tracks the float32 recurrent state roughly twice as closely as bfloat16 and is the recommended dtype if this feature is enabled. Both 16-bit dtypes preserve correctness on deterministic tasks but change the greedy trajectory on the majority of prompts — the recurrent state accumulates across decode steps, so any sub-float32 precision eventually flips a token. This flag is therefore opt-in and off by default; `--linear-state-quantization` (the one-time prefix-cache roundtrip) is the low-risk path.

---

## Tests

New test files (56 new tests; full suite green, `black` clean):

| File | Tests | Coverage |
|------|------:|----------|
| `tests/test_batch_quantized_kv_cache.py` | 30 | asymmetric SDPA, QuaRot quality assertions, prefix cache roundtrip |
| `tests/test_linear_state_quantization.py` | 19 | ArraysCache quantize/dequantize roundtrip, dtype preservation, hybrid 3+1 mix, memory accounting, store/fetch integration |
| `tests/test_gated_delta_bf16_patch.py` | 7 | patch applies + idempotent, bf16 and fp16 storage, numerical equivalence, guards |

## Production status

Both features are running in production on Qwen3.6-27B and Qwen3.6-35B-A3B — correct reasoning, no throughput regression. `--kv-hadamard` runs at K=3 V=3; `--linear-state-quantization` at 8-bit; `--linear-state-bf16` with `VLLM_MLX_LINEAR_STATE_DTYPE=fp16`.


Thump604 (changes requested):
Thanks for pushing this. The direction is valuable, but I do not think this is merge-ready yet. Two issues look concrete from the current diff.

1. `max_kv_size` appears to bypass the new live quantized cache. In `install_quantized_kv_cache()`, `_patched_make_cache()` returns `BatchRotatingKVCache(max_kv_size, left_padding)` for every layer whenever `max_kv_size is not None`. That means `--kv-cache-quantization` plus a bounded KV config does not use `BatchQuantizedKVCache` at all on that path. Runtime commonly uses bounded KV/context configs, so this should either be supported with a bounded quantized cache path, or explicitly rejected/documented with a regression proving the behavior. A test should cover `kv_cache_quantization=True` with `max_kv_size` set.

2. The benchmark path does not appear to expose or forward `kv_hadamard`. The serve parser/config gets `--kv-hadamard`, but the bench parser only adds `--kv-k-bits` / `--kv-v-bits`, and `run_benchmark()` forwards the asymmetric bits without `kv_hadamard`. Since the PR body presents QuaRot performance/quality as a key result, the upstream CLI should be able to reproduce that path, or the PR should narrow the claim to serve-only and add the missing benchmark wiring in a follow-up.

I am not objecting to the feature direction. I am asking that the bounded-KV behavior and reproducible QuaRot benchmark surface be tightened before merge.

janhilgard:
Thanks for the thorough review. Both issues addressed in c8e7c2a:

**1. `max_kv_size` + quantization interaction**

The `max_kv_size` fallback path (models without `make_cache()`) now emits an explicit `WARNING` log when quantization is enabled, documenting that bounded cache layers remain unquantized:

```
max_kv_size=1024 is set but model has no make_cache() method.
All layers will use unquantized BatchRotatingKVCache — KV cache quantization
(K=8, V=4) will NOT be applied. Consider removing --max-kv-size to enable quantization.
```

Similarly, when `to_batch_cache()` encounters a `RotatingKVCache` layer from `model.make_cache()`, it logs a warning about the unquantized conversion.

Two new tests cover this:
- `test_make_cache_with_max_kv_size_logs_warning` — verifies fallback path emits warning and returns `BatchRotatingKVCache`
- `test_make_cache_model_with_make_cache_uses_quantization` — verifies models with `make_cache()` still get `BatchQuantizedKVCache` for KVCache layers (regardless of `max_kv_size`)

Note: Models currently used in production (Qwen3.5/3.6) have `make_cache()` methods, so they take the `to_batch_cache()` path where KVCache layers are properly quantized. The `max_kv_size` parameter is ignored by upstream `mlx_lm` when `make_cache()` exists — this is mlx_lm's design, not introduced by this PR. A `BatchQuantizedRotatingKVCache` combining both bounded eviction and quantization would be a good follow-up.

**2. Bench parser `--kv-hadamard`**

Added `--kv-hadamard` flag to the bench parser and forwarded through `run_benchmark()` → `SchedulerConfig(kv_hadamard=args.kv_hadamard)`. QuaRot benchmarks are now reproducible via:

```bash
vllm-mlx bench model --kv-cache-quantization --kv-k-bits 3 --kv-v-bits 3 --kv-hadamard
```

Thump604 (approved):
Thanks, this addresses my requested changes.

The bench path now exposes/forwards `--kv-hadamard`, so the QuaRot benchmark path is reproducible from the CLI.

For `max_kv_size`, the current patch now makes the behavior explicit and tested: models without `make_cache()` fall back to unquantized `BatchRotatingKVCache` with a warning, while models with `make_cache()` still quantize `KVCache` layers through `BatchQuantizedKVCache`. That is acceptable for this PR. A bounded `BatchQuantizedRotatingKVCache` remains a real follow-up if we want full bounded-cache quantization coverage, but I do not think it has to block this broader feature surface.

I am approving with that boundary understood.

janhilgard:
Force-pushed a squashed update that folds the SimpleEngine wiring (previously opened as #571, now closed) into this PR's single commit. Combined scope:

**BatchedEngine (already in the original PR):**
- `BatchQuantizedKVCache` with asymmetric K/V bit support
- `asymmetric_quantized_sdpa` for separate K/V bit matmuls
- Monkey-patches for `_make_cache`, `_merge_caches`, and SDPA routing
- `--kv-k-bits` / `--kv-v-bits` CLI flags
- Scheduler / MLLMScheduler / MemoryCache config updates

**SimpleEngine wiring (newly folded in):**
- `install_quantized_kv_cache()` gains a `simple_mode=True` path that additionally patches `mlx_lm.models.cache.make_prompt_cache` to wrap `KVCache` layers in `BatchQuantizedKVCache` (B=1, `left_padding=[0]`). Non-`KVCache` layers (`ArraysCache`, `RotatingKVCache`) pass through with a single warning.
- The SDPA patch also retrofits any already-imported `mlx_lm.models.<arch>` modules whose `scaled_dot_product_attention` binding still points at the original function (`from .base import scaled_dot_product_attention` captures the function object at module import time).
- `SimpleEngine.__init__` accepts the matching `kv_cache_quantization*` / `kv_hadamard` kwargs and calls `install_quantized_kv_cache(simple_mode=True)` **before** model load so freshly imported model modules pick up the patched binding directly.
- `server.load_model()` forwards the new params; `cli.py` prints the resulting KV cache config in the `Mode: Simple` summary alongside the existing `Mode: Continuous batching` summary.

Validated end-to-end on MiniMax-M2.7 in Simple mode (continuous-batching scheduler has a separate hang on `minimax_m2`, so Simple is the only path available for that model today). With `--kv-cache-quantization --kv-cache-quantization-bits 8 --kv-k-bits 3 --kv-v-bits 3 --kv-hadamard` the startup logs report `Installed BatchQuantizedKVCache: K=3-bit, V=3-bit, group_size=64, QuaRot=ON, mode=simple+batched` and `KV cache quantization: 3-bit, group_size=64 +QuaRot`, and `/v1/chat/completions` returns model output through the quantized SDPA path.

`asymmetric_quantized_sdpa` is slower than the fused `mx.fast.scaled_dot_product_attention` used by bf16 `KVCache`, so enabling KV quantization for SimpleEngine is a memory-vs-throughput trade-off rather than a free win. Aggressive bit widths (e.g. K=3 V=3) can also degrade output quality on some models. The choice is left to the operator per `--kv-*` flag values — this change only fixes the wiring so the flags actually apply in Simple mode.

janhilgard:
Closing for the same reason as #569: production experience (2026-05-24) showed the asymmetric_quantized_sdpa runtime path is 4-8x slower than mlx's fused mx.fast.scaled_dot_product_attention. With 3-bit K/V + Hadamard on Qwen3.6-35B-A3B we measured 8-17 t/s vs 70 t/s for bf16 KV cache. We have reverted production to bf16 (no --kv-cache-quantization) and cannot honestly defend the throughput numbers in this PR's description without a fused-kernel quantized SDPA in mx.fast (which does not exist today). The BatchQuantizedKVCache implementation itself is correct and could become useful once an upstream fused kernel lands — feel free to revive then.

PR #562, opened 2026-05-22
fix(gpt-oss): plumb harmony tool calls all the way through to response

CBribiescas:
## Summary

Fixes a three-layered bug that prevented gpt-oss-120b (and other harmony-format models) from returning structured `tool_calls` when both `--tool-call-parser gpt-oss` and `--reasoning-parser gpt_oss` were set. Each layer independently blocked tool calls; all three need to land for tools to work end-to-end.

Live verified on `mlx-community/gpt-oss-120b-MXFP4-Q8` on Apple M5 Pro Max: gpt-oss now emits `finish_reason: "tool_calls"` with properly structured `tool_calls[]`, and `reasoning_content` populated from the analysis channel.

Related: this is downstream of #555 (the Stream(gpu,N) crash, which 0.4.0rc1 already fixes via your own approach) — same model, distinct issue. With both fixes in place, gpt-oss-120b is finally usable for tool-using workloads.

### 1. `clean_output_text` strips harmony tokens before the parser

`api/utils.py:clean_output_text` runs on every chat-completion output. For gpt-oss output it invokes `_clean_gpt_oss_output`, which when no `<|channel|>final` block is present runs a generic regex that strips ALL channel/structural tokens — including the `<|channel|>commentary to=functions.*<|call|>` blocks that `HarmonyToolParser` needs.

After stripping, the parser sees only bare arg-JSON glued to surrounding analysis text and extracts nothing.

**Fix:** add an early-bypass in `clean_output_text` that returns the raw text untouched when a commentary-to-functions block is present. The chat-completion handler runs `clean_output_text` a second time on the parser's residue (`result.content`), where the bypass no longer matches and normal stripping applies.

### 2. `HarmonyToolParser` required a `<|call|>` terminator the model often doesn't emit

gpt-oss-120b commonly stops after emitting the arg-JSON without echoing the trailing `<|call|>` token — either because vllm-mlx consumed `<|call|>` as a stop sequence without forwarding it, or because the model self-terminated mid-block. In our live trace the raw output ended at `}\n` with no terminator and `finish_reason: "stop"`.

The existing `_COMMENTARY_BLOCK_PATTERN` required `<|call|>` to match, so legitimate but un-terminated tool calls extracted nothing.

**Fix:** relax the terminator to accept `<|call|>`, `<|end|>`, `<|return|>`, `<|start|>`, OR end-of-string (`\Z`). The existing `json.loads`-or-keep-raw logic in the parser body provides the false-positive guard.

`test_tool_parser_incomplete_call` was deliberately asserting the old strict behavior; updated to `test_tool_parser_incomplete_call_with_valid_args` to reflect that a valid-JSON commentary block without a terminator IS now extracted.

### 3. `_extract_reasoning_and_tool_calls` blanks the input when no `final` channel is present

When the reasoning parser is enabled, server.py runs it FIRST on the raw output to split out reasoning. For gpt-oss the reasoning parser returns `(reasoning, content) = (analysis_text, None)` when the model jumps directly from `<|channel|>analysis` to `<|channel|>commentary to=functions.*` — no `<|channel|>final` is ever emitted because the model is calling a tool, not responding.

The chat handler then ran:

```python
elif reasoning_text is not None:
    text_for_tool_parse = ""
```

setting the tool-parser input to the empty string and hiding every tool call. Since the tool parser's regex anchors on `<|channel|>commentary to=functions.` (which doesn't overlap the analysis block), it's safe to pass the full raw output and let the parser find the commentary block itself.

**Fix:** pass `output_text` through to the tool parser when `cleaned_reasoning_text is None and reasoning_text is not None`.

## Verification

**Host:** Apple M5 Pro Max, 128 GB unified memory, macOS Tahoe.

### Live tested on three models (two parser families)

All servers run with `--enable-auto-tool-choice` and the appropriate `--tool-call-parser` / `--reasoning-parser` flags.

| Model | Parser | Test | Result |
|---|---|---|---|
| `mlx-community/gpt-oss-120b-MXFP4-Q8` | `gpt-oss` + `gpt_oss` reasoning | single-arg `read_file({"path":"/etc/hosts"})` | ✅ `finish_reason: "tool_calls"` |
| `mlx-community/gpt-oss-20b-MXFP4-Q8` | `gpt-oss` + `gpt_oss` reasoning | single-arg `read_file(...)` | ✅ `finish_reason: "tool_calls"`, `reasoning_content` populated |
| `mlx-community/gpt-oss-20b-MXFP4-Q8` | `gpt-oss` + `gpt_oss` reasoning | **multi-arg** `send_message({"channel":"general","message":"Hello","priority":"high"})` | ✅ all 3 args correct, enum value respected |
| `mlx-community/gpt-oss-20b-MXFP4-Q8` | `gpt-oss` + `gpt_oss` reasoning | plain chat (no tools) — non-regression | ✅ `content: "pong"`, `finish_reason: "stop"` |
| `lmstudio-community/Qwen3-Coder-30B-A3B-Instruct-MLX-4bit` | `qwen3_coder` (different family) | single-arg `read_file(...)` | ✅ `finish_reason: "tool_calls"` — **non-regression on a different parser family** |

### Reproducer (single-prompt minimum)

```bash
vllm-mlx serve mlx-community/gpt-oss-20b-MXFP4-Q8 \
  --port 8103 \
  --tool-call-parser gpt-oss --enable-auto-tool-choice \
  --reasoning-parser gpt_oss

curl http://127.0.0.1:8103/v1/chat/completions \
  -H 'Content-Type: application/json' -d '{
    "model": "/path/to/gpt-oss-20b-MXFP4-Q8",
    "messages": [{"role":"user","content":"Use the read_file tool on /etc/hosts. Do not explain."}],
    "tools": [{"type":"function","function":{"name":"read_file",
      "parameters":{"type":"object","properties":{"path":{"type":"string"}},
      "required":["path"]}}}],
    "tool_choice": "auto", "max_tokens": 256
  }'
```

**Before this PR (on 0.4.0rc1):**
```json
"message": {
  "content": "The user wants to read /etc/hosts. We have a tool read_file. We need to call it.{\n  \"path\": \"/etc/hosts\"\n}",
  "tool_calls": null
}
```
finish_reason: `"stop"`. The bare arg JSON was glued to the analysis text in `content`; no `tool_calls` field.

**After this PR:**
```json
"message": {
  "content": null,
  "reasoning_content": "The user wants to read /etc/hosts. We have a tool read_file. We need to call it.",
  "tool_calls": [{
    "id": "call_96762d58",
    "type": "function",
    "function": {"name": "read_file", "arguments": "{\"path\": \"/etc/hosts\"}"}
  }]
}
```
finish_reason: `"tool_calls"`. ✓

### Unit tests (on v0.4.0rc1 base + this PR)

- `tests/test_harmony_parsers.py` — **41 passed** (1 deliberately retargeted from the old strict-terminator assertion to reflect the new behavior).
- `tests/test_simple_engine.py` + `test_simple_engine_cancel_serialization` + `test_engine_core_stream_safety` + `test_engine_core_thread_streams` — **44 passed**.
- Streaming (no tools) — 52 SSE chunks delivered cleanly ✓.

**FYI separately:** `tests/test_ssd_cache.py` has 4 pre-existing failures on clean v0.4.0rc1 (verified by stashing this PR's changes and re-running) — `TestSpillPath::test_evict_lru_calls_ssd_spill`, `TestMemoryCacheSSDCheck::test_check_ssd_returns_none_on_ram_hit`, and both `TestIntegrationSpillAndFetch` tests. Not touched by this PR but flagging for a separate look.

## Test plan

- [ ] CI: full fast suite green (except the 4 pre-existing `test_ssd_cache` failures noted above)
- [ ] Manual: `vllm-mlx serve` a gpt-oss-120b checkpoint with `--tool-call-parser gpt-oss --enable-auto-tool-choice --reasoning-parser gpt_oss`, send the reproducer above, confirm `tool_calls` populated
- [ ] Manual: same server, send a plain chat (no tools) — confirm `content` populated, `tool_calls: null`, `finish_reason: "stop"`

### DIFF
--- a/tests/test_api_utils.py
+++ b/tests/test_api_utils.py
@@ -98,6 +98,24 @@ class TestCleanOutputText:
         assert "42" in result
         assert "<|im_start|>" not in result
 
+    def test_gpt_oss_commentary_final_returns_final_text(self):
+        text = (
+            "<|channel|>commentary to=functions.get_weather"
+            '<|message|>{"loc": "SF"}'
+            "<|channel|>final<|message|>Done<|return|>"
+        )
+        assert clean_output_text(text) == "Done"
+        assert "<|channel|>" not in clean_output_text(text)
+        assert "<|message|>" not in clean_output_text(text)
+
+    def test_gpt_oss_commentary_call_returns_args_json(self):
+        text = (
+            "<|channel|>commentary to=functions.get_weather"
+            '<|message|>{"loc": "SF"}<|call|>'
+        )
+        assert clean_output_text(text) == '{"loc": "SF"}'
+        assert "<|call|>" not in clean_output_text(text)
+
 
 class TestSpecialTokensPattern:
     """Tests for the special tokens regex pattern."""
--- a/tests/test_gpt_oss_tool_parsing.py
+++ b/tests/test_gpt_oss_tool_parsing.py
@@ -1,0 +1,144 @@
+# SPDX-License-Identifier: Apache-2.0
+"""
+Tests for GPT-OSS harmony-format tool call parsing at the server boundary.
+
+Tests exercise _extract_reasoning_and_tool_calls directly: when the model emits
+an analysis (reasoning) channel followed by a commentary channel, the analysis
+block must be stripped before the text is handed to the tool parser so reasoning
+prose is never run through the generic fallback (which can extract spurious tool
+calls from JSON-shaped text) and never duplicates into response content.
+
+Usage:
+    pytest tests/test_gpt_oss_tool_parsing.py -v
+"""
+
+from types import SimpleNamespace
+
+
+def test_tools_branch_strips_analysis_block_before_parser(monkeypatch):
+    import vllm_mlx.server as server
+
+    class FakeReasoningParser:
+        def extract_reasoning(self, text):
+            return (
+                'I will use {"name": "get_weather", "arguments": {"city": "SF"}}',
+                None,
+            )
+
+    seen = []
+
+    def fake_parse(text, request, **_):
+        seen.append(text)
+        return text, None
+
+    raw = (
+        '<|channel|>analysis<|message|>I will use {"name": "get_weather", '
+        '"arguments": {"city": "SF"}}<|end|>'
+        "<|channel|>commentary to=functions.get_weather"
+        '<|message|>{"city": "S'
+    )
+    request = SimpleNamespace(tools=[{"type": "function"}])
+    monkeypatch.setattr(server, "_reasoning_parser", FakeReasoningParser())
+    monkeypatch.setattr(server, "_parse_tool_calls_with_parser", fake_parse)
+
+    reasoning, cleaned, tool_calls = server._extract_reasoning_and_tool_calls(
+        raw, request
+    )
+
+    assert reasoning is not None and reasoning != ""
+    assert tool_calls is None
+    assert len(seen) == 1
+    assert "<|channel|>analysis" not in seen[0]
+    assert seen[0] == (
+        "<|end|><|channel|>commentary to=functions.get_weather" '<|message|>{"city": "S'
+    )
+
+
+def test_tools_branch_analysis_then_commentary_call_passes_through(monkeypatch):
+    """Happy path: analysis channel followed by a commentary channel with a
+    <|call|> terminator. The call must still be passed through to the parser
+    (commentary block preserved), but the analysis block must be stripped."""
+    import vllm_mlx.server as server
+
+    class FakeReasoningParser:
+        def extract_reasoning(self, text):
+            return "analysis content", None
+
+    seen = []
+
+    def fake_parse(text, request, **_):
+        seen.append(text)
+        return None, ["call_extracted"]
+
+    raw = (
+        "<|channel|>analysis<|message|>thinking..."
+        "<|channel|>commentary to=functions.read_file"
+        '<|message|>{"path":"/etc/hosts"}<|call|>'
+    )
+    request = SimpleNamespace(tools=[{"type": "function"}])
+    monkeypatch.setattr(server, "_reasoning_parser", FakeReasoningParser())
+    monkeypatch.setattr(server, "_parse_tool_calls_with_parser", fake_parse)
+
+    reasoning, cleaned, tool_calls = server._extract_reasoning_and_tool_calls(
+        raw, request
+    )
+
+    assert reasoning == "analysis content"
+    assert tool_calls == ["call_extracted"]
+    assert len(seen) == 1
+    assert "<|channel|>analysis" not in seen[0]
+    assert seen[0] == (
+        "<|channel|>commentary to=functions.read_file"
+        '<|message|>{"path":"/etc/hosts"}<|call|>'
+    )
+
+
+# ============================================================================
+# Streaming gate activation for gpt-oss harmony commentary
+# ============================================================================
+
+
+def test_streaming_marker_possible_for_commentary():
+    """The streaming marker gate must activate for a gpt-oss commentary block.
+
+    _STREAMING_TOOL_MARKERS now includes "<|channel|>commentary" and "<|call|>",
+    otherwise _streaming_tool_markup_possible would never fire for harmony
+    tool calls and the streaming parser path would stay dormant.
+    """
+    import vllm_mlx.server as server
+
+    commentary = (
+        "<|channel|>commentary to=functions.get_weather"
+        '<|message|>{"city": "SF"}<|call|>'
+    )
+    assert server._streaming_tool_markup_possible(commentary) is True
+
+    # A commentary block that starts before the current delta also activates
+    # the gate (boundary-window scan).
+    assert (
+        server._streaming_tool_markup_possible_after_delta(
+            "prefix text ", "<|channel|>commentary to=functions.get_weather"
+        )
+        is True
+    )
+    # The terminator alone is a marker too (e.g. split across chunks).
+    assert (
+        server._streaming_tool_markup_possible_after_delta(
+            "<|channel|>commentary to=functions.get_weather"
+            '<|message|>{"city": "SF"}',
+            "<|call|>",
+        )
+        is True
+    )
+
+
+def test_streaming_marker_false_for_plain_text():
+    """Plain content must not activate the harmony tool-markup gate."""
+    import vllm_mlx.server as server
+
+    plain = "The weather in San Francisco is 72 degrees and sunny."
+    assert server._streaming_tool_markup_possible(plain) is False
+    assert server._streaming_tool_markup_possible_after_delta("", "72 degrees") is False
+    # Final-channel content (no commentary) is not tool markup either.
+    final_only = "<|channel|>final<|message|>The answer is 42.<|return|>"
+    assert server._streaming_tool_markup_possible(final_only) is False
--- a/tests/test_harmony_parsers.py
+++ b/tests/test_harmony_parsers.py
@@ -632,12 +632,23 @@ class TestHarmonyToolDefinitionConverter:
 class TestHarmonyEdgeCases:
     """Edge case tests for Harmony parsers."""
 
-    def test_tool_parser_incomplete_call(self):
-        """Incomplete tool call (missing <|call|>) is not parsed."""
+    def test_tool_parser_incomplete_call_with_valid_args(self):
+        """Tool call without trailing <|call|> IS parsed when args are valid JSON.
+
+        gpt-oss-120b commonly produces commentary blocks without echoing the
+        <|call|> terminator — either because vllm-mlx consumed it as a stop
+        token, or because the model emitted a stop without the terminator.
+        We extract the call as long as the function name + valid JSON args
+        are present at end-of-text.
+        """
+        import json as _json
+
         parser = HarmonyToolParser()
         text = "<|channel|>commentary to=functions.func\n" '<|message|>{"arg": "value"}'
         result = parser.extract_tool_calls(text)
-        assert not result.tools_called
+        assert result.tools_called
+        assert result.tool_calls[0]["name"] == "func"
+        assert _json.loads(result.tool_calls[0]["arguments"]) == {"arg": "value"}
 
     def test_tool_parser_unicode_content(self):
         """Handle unicode in tool arguments."""
--- a/tests/test_harmony_parsers.py
+++ b/tests/test_harmony_parsers.py
@@ -650,6 +650,55 @@ class TestHarmonyEdgeCases:
         assert result.tool_calls[0]["name"] == "func"
         assert _json.loads(result.tool_calls[0]["arguments"]) == {"arg": "value"}
 
+    def test_tool_parser_truncated_at_eos_does_not_extract(self):
+        """Truncated args at end-of-string MUST NOT become a tool call.
+
+        When the regex match terminates at \\Z (no explicit <|call|>,
+        <|end|>, <|return|>, or <|start|>) AND the captured arg-text is not
+        valid JSON, this is almost certainly a truncated generation
+        (max_tokens hit mid-block, OOM, etc). Synthesizing a tool call with
+        broken arguments would surface to the API consumer as a malformed
+        function call — much worse than treating the partial output as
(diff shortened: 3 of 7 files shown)


Thump604 (changes requested):
Requesting one fix before merge.

The `\Z` fallback is useful for the observed gpt-oss case where the model emits a complete JSON argument object but the `<|call|>` marker is not present in decoded text. However, the current parser still appends a tool call on `json.JSONDecodeError` by preserving raw arguments. With `\Z` as a valid terminator, a genuinely truncated end-of-output block such as:

```text
<|channel|>commentary to=functions.read_file
<|message|>{"path":"/etc/hosts"
```

matches `_COMMENTARY_BLOCK_PATTERN` and becomes a structured tool call with invalid/raw arguments. That changes a length/truncation or malformed-generation case into `finish_reason="tool_calls"` with a broken function payload.

I think the fix is narrow: keep the existing raw-argument fallback for explicit terminators like `<|call|>`, `<|end|>`, `<|return|>`, or `<|start|>`, but when the match terminates by end-of-string, require `json.loads(args_str)` to succeed before emitting the tool call. Add both tests: valid JSON at EOS extracts, invalid/truncated JSON at EOS does not.

waybarrios (changes requested):
The terminator and EOS `json.loads` fix looks right, and the with-tools path is clean. One regression to fix first: the `clean_output_text` bypass in `api/utils.py` fires for every caller of that shared helper, not just the two-pass tool path. On the no-tools branch of `_extract_reasoning_and_tool_calls` (`server.py:2723-2724`) the parser is skipped, so `cleaned_text` stays raw and the re-clean at `server.py:4771` (chat completions) and `:5190` (Anthropic) bypasses again, leaking harmony tokens (`<|channel|>`, `<|message|>`, `<|call|>`) into the response content. Low frequency, it needs the model to emit a `to=functions` block with no tools defined, but it is a regression vs `main`, which returned an empty string there. Gating `text_for_tool_parse = output_text` on `request.tools`, or scoping the bypass inside `_parse_tool_calls_with_parser`, closes it.

CBribiescas:
Pushed the fix in `98d4f83`:

- Gated `text_for_tool_parse = output_text` on `request.tools` (server.py:2704-2715). When reasoning is extracted with no `final` channel and no tools are defined, fall back to `""` — restores pre-PR behavior on that path so the `clean_output_text` bypass can't preserve `<|channel|>` / `<|message|>` / `<|call|>` tokens in response content.
- Two regression tests in `TestHelperFunctions`: `test_extract_reasoning_no_tools_drops_raw_harmony_commentary` (primary guard) and `test_extract_reasoning_with_tools_preserves_raw_for_parser` (guards against over-correction).

157 server + harmony + tool-parser tests pass locally; black clean. Ready for another look.

CBribiescas:

Thump604:
Reopened because the final head appears to address my prior blocker. The EOS handling now requires valid JSON before emitting a tool call at `\Z`, while explicit terminators keep the raw-argument fallback. The later `98d4f83` change also appears to address Wayner's no-tools raw Harmony token leak by gating raw forwarding on `request.tools`. CI is green on the updated head.

From my side, the stale changes-requested item is resolved. @waybarrios / @janhilgard, please take the final maintainer re-review from here because Wayner's prior changes-requested review is still the remaining review state.

Thump604 (approved):
Approving from my side after re-reviewing the final head. The blocker I raised on EOS-truncated Harmony arguments is addressed: valid JSON at `\Z` still extracts, invalid/truncated JSON at `\Z` does not, and explicit terminators keep the raw-argument fallback. The later no-tools gating fix also appears to address Wayner's raw Harmony token leak concern.

Wayner's prior changes-requested review is still the remaining maintainer state, so I am leaving final merge judgment to @waybarrios / @janhilgard.

janhilgard (approved):
LGTM — surgical layered fix with an exemplary root-cause writeup. This has been sitting a month with green CI and clean mergeability; worth landing.

## Root-cause quality

The PR body identifies **three independently-blocking layers**, each of which would still hide tool calls even if the others were fixed:

1. `clean_output_text` regex stripping harmony structural tokens before the parser sees them
2. `HarmonyToolParser` requiring a `<|call|>` terminator the model often doesn't emit
3. `_extract_reasoning_and_tool_calls` blanking the input when no `<|channel|>final` block exists

This is exactly the kind of layered bug that gets "fixed" one layer at a time by successive PRs, each of which appears to have no effect because the other layers still block. The full stack-of-three writeup means a maintainer can verify the causal chain, not just spot-check the diff.

## Layer-by-layer review

### Layer 1: `_COMMENTARY_TOOL_RE` bypass in `clean_output_text`

- Bypass triggers only when `<|channel|>commentary to=functions.` is present → won't affect any output that doesn't have a tool call in it. Safe.
- Comment correctly notes the handler runs `clean_output_text` **again** on `result.content` (parser residue) — where the bypass won't match and normal stripping applies. This is the important invariant: the bypass doesn't leak harmony tokens into the final response content.
- Pattern is a compiled re at module scope. Efficient.

### Layer 2: relaxed terminator alternation

- Change from `<|call|>` to `<|call\||<|end\||<|return\||<|start\||\Z` correctly enumerates the terminators the model can plausibly emit plus end-of-string for truncated/stop-token-consumed outputs.
- The named group `?P<terminator>` capture + `matched_at_eos = match.group("terminator") == ""` gives downstream code the ability to distinguish clean-terminated from truncated calls — useful for logging/telemetry even if not used in the current strict path.
- Comment explicitly notes "end-of-string treated more strictly than explicit terminators — truncated mid-block output must not become a structured tool call with malformed args." Combined with the existing `json.loads`-or-keep-raw guard, this is the right defensive posture.
- Test rename from `test_tool_parser_incomplete_call` to `test_tool_parser_incomplete_call_with_valid_args` correctly documents the semantic change: incomplete-but-JSON-valid IS now extracted; incomplete-and-JSON-invalid still isn't.

### Layer 3: gate on `request.tools` before passing raw output

The critical addition:
```python
if request is not None and getattr(request, "tools", None):
    text_for_tool_parse = output_text
else:
    text_for_tool_parse = ""
```

This is the right gate. The comment explains it well: "without tools the parser is skipped below, and clean_output_text at the response boundary intentionally preserves commentary blocks, so raw harmony tokens would leak into response content." Passing raw output only when tools are declared prevents accidental leaking on plain-chat requests to a gpt-oss endpoint.

## Verification quality

Live testing on **three models across two parser families**:
- gpt-oss-120b + gpt-oss-20b (harmony) — 3 test shapes (single-arg, multi-arg with enum, plain chat non-regression)
- Qwen3-Coder-30B (qwen3_coder parser) — non-regression on a **different parser family**

The Qwen non-regression is important: it proves the `_extract_reasoning_and_tool_calls` change doesn't affect codepaths that go through other reasoning parsers. Without this, one could reasonably ask whether the layer-3 change breaks Qwen tool calling. The answer is no, and it's tested.

## Tests

- `test_harmony_parsers.py` +31 lines: covers the incomplete-call-with-valid-args case (new semantics) plus terminator alternation coverage.
- `test_server.py` +65 lines: exercises the reasoning-extraction → tool-parse handoff (layer 3), which is where the highest-value regression risk lives.

Right-sized coverage for a targeted fix.

## Mergeability

`mergeable: MERGEABLE, mergeStateStatus: CLEAN`. Still applies against current main. Nothing needed on the contributor's side.

## CI

All 9 checks green.

## Personal context

I don't currently deploy any gpt-oss variant (retired GPT-OSS-20B from our stack April 2026 for memory reasons), so I can't provide independent live confirmation. But the root-cause chain is well-documented and the tests match the claimed behavior. If the maintainer wants a second live confirmation before merging, worth pinging someone who's actively running gpt-oss on vllm-mlx.

Ship it. Month-old PRs on external contributions bleed context — CBribiescas has done the diligence, tests pass, mergeable is clean.

alikiran:
Additional GPT-OSS 20B parser variant reproduced on released vllm-mlx 0.4.0.

## Environment

- vllm-mlx: 0.4.0
- model: mlx-community/gpt-oss-20b-MXFP4-Q8
- macOS / Apple M4 Max
- server flags:

```text
--enable-auto-tool-choice
--tool-call-parser gpt-oss
--reasoning-parser gpt_oss
```

- client: Pi using the OpenAI Chat Completions API
- tool: ffgrep

## Minimal prompt

```text
Use ffgrep to search for TextualDisplayView, then report the first matching file.
```

The model emits a valid recipient and argument payload, but places the recipient on the analysis channel:

```text
<|channel|>analysis<|message|>We need to run ffgrep. Let us search.
<|end|><|start|>assistant<|channel|>analysis to=functions.ffgrep code<|message|>
{"path":"","pattern":"TextualDisplayView"}<|call|>
```

The current HarmonyToolParser regex only recognizes:

```text
<|channel|>commentary to=functions.<name> ...
```

No tool call is extracted. The control-token cleanup then collapses the raw output into visible content resembling:

```text
assistantcode{"path":"","pattern":"TextualDisplayView"}
```

The response has finish_reason "stop" instead of "tool_calls".

This is adjacent to the three propagation fixes in this PR, but the current PR regex still appears to require the commentary channel, so this output variant remains unhandled.

## Locally verified compatibility normalization

Before passing output to HarmonyToolParser, normalize this narrowly scoped recipient form:

```text
<|start|>assistant<|channel|>(analysis|commentary)
to=functions.<name> <constraint><|message|>
```

to:

```text
<|channel|>commentary to=functions.<name>
<|constrain|><constraint><|message|>
```

Also allow hyphens in function names, for example fff-multi-grep.

After this normalization:

- Pi receives a structured ffgrep tool call
- ffgrep executes successfully
- the model consumes the tool result and returns a normal final answer
- a five-test compatibility suite passes
- no regression was observed for the canonical commentary form

LM Studio successfully handles tool calling with the same model weights, which further points to the serving/parser boundary rather than the model or client tool implementation.

Would you prefer this handled in #562, or tracked as a separate parser-normalization issue?

Thump604:
Thanks for the concrete reproduction. I would track this as a separate, narrow parser-normalization issue/PR rather than broaden #562.

#562 is already clean and approved around the canonical Harmony form (`commentary to=functions.*`) and fixes three propagation defects on that path. The newly observed `analysis to=functions.* code` form is adjacent, but it introduces a distinct compatibility decision: which noncanonical recipient/channel variants should be normalized before `HarmonyToolParser` runs. Keeping it separate makes both changes easier to validate and review.

A focused follow-up should include fixtures for:

- the reported `analysis to=functions.ffgrep code` form,
- the canonical `commentary to=functions.*` form as a non-regression,
- hyphenated function names,
- malformed/truncated argument JSON,
- analysis text that mentions `to=functions.*` but is not a recipient block,
- structured `tool_calls`, `finish_reason=tool_calls`, and a successful post-tool continuation.

The normalization should be scoped to Harmony parser input, not global output cleanup. Your LM Studio comparison and successful local normalization make this a strong follow-up candidate. Please link it here when opened; #562 should remain merge-ready rather than absorb another parser dialect.

waybarrios:
Following up on the review discussion. I applied the fixes for the points that came up, and the branch now has them in. Here is what changed:

- **`clean_output_text` is back to always cleaning.** The raw harmony control tokens no longer leak into responses on any path (batched, simple, Responses, Anthropic, no-tools chat); the tool path still receives the raw output it needs via `text_for_tool_parse`.

- **The commentary parser no longer crosses channel boundaries.** Args stop at the next channel, and a block only becomes a tool call when the JSON is valid (at end-of-string or a channel boundary), so it can't glue following-channel text onto a call.

- **Reasoning stays out of the tool parser.** The analysis channel is stripped before parsing, so reasoning isn't duplicated into the response and can't be mistaken for a spurious tool call.

- **Streaming now works like non-streaming.** The harmony markers activate the streaming gate, the parser emits the call when a commentary block completes (no `<|call|>` needed), and the end-of-stream fallback returns `finish_reason="tool_calls"`.

Regression tests were added for each case, and the harmony, gpt-oss and tool-parser suites pass locally. Thanks for the solid root-cause writeup on the original PR, it made addressing these quick.

waybarrios (approved):
Everything looks good now. The fixes for the review points are in, the regression tests pass, and all CI checks are green, including apple-silicon. Approving.

PR #564, opened 2026-05-22
fix(gemma4): match mlx-vlm 0.5.0 Attention signature (shared_kv, 3-tuple)

CBribiescas:
## Summary

Updates `vllm_mlx/patches/gemma4_mllm.py` to match the `mlx-vlm` 0.5.0 (released 2026-05-21) Gemma 4 `Attention.__call__` signature. Without this, every Gemma 4 request under `--continuous-batching` fails with:

```
TypeError: patch_gemma4_attention_for_batching.<locals>._patched_call()
got an unexpected keyword argument 'shared_kv'
```

## Root cause

`mlx_vlm.models.gemma4.language.Attention.__call__` upstream changed in two ways the fork's monkey-patch never tracked:

1. **Added kwargs** `shared_kv: Optional[tuple]` and `offset: Optional[Any]`.
2. **Changed return shape** from a single `mx.array` to a 3-tuple `(output, (keys, values), offset)`. `DecoderLayer` now does `h, shared_kv, offset = self.self_attn(...)` and the old patch's single-return crashed the unpack.

The original reason for the patch — added in commit `98c682d` ("fix: Gemma 4 BatchKVCache, reasoning parser, and MLLM stop tokens") — was to snapshot `cache.offset` before `update_and_fetch` mutates it in place (BatchKVCache's offset is an `mx.array` and `+=` is in-place; without the snapshot, RoPE on queries sees a post-update offset and applies the wrong position). That behavior is preserved.

## What this PR changes

- **`vllm_mlx/patches/gemma4_mllm.py`** — extend `_patched_call` signature to accept `shared_kv` and `offset` kwargs, take the `shared_kv` branch when supplied (skipping `k_proj/v_proj/k_norm/v_norm/cache.update_and_fetch` entirely), and return the 3-tuple shape that mlx-vlm 0.5.0's `DecoderLayer` expects.
- Replace the `self.is_kv_shared_layer + cache.state` lookup the old patch used with the new `shared_kv` parameter — mlx-vlm 0.5.0 threads shared KV through `DecoderLayer` explicitly rather than recovering it from the cache. Behaviorally equivalent on KV-shared layers.
- **`tests/test_gemma4_mllm_patch.py`** (new) — three regression tests:
  1. `test_patched_call_accepts_shared_kv_and_offset_kwargs` — patch accepts new kwargs, returns 3-tuple.
  2. `test_patched_call_snapshots_offset_before_update` — `cache.offset` is read pre-`update_and_fetch`; RoPE on queries sees the pre-update value (the original BatchKVCache fix is preserved).
  3. `test_patched_call_uses_shared_kv_when_provided` — when `shared_kv` is set, `k_proj` is NOT called and the inner `(k, v)` arrays pass through identity-preserved.

## Where `shared_kv` actually fires

Upstream comment in `mlx_vlm/models/gemma4/language.py` calls it the "KV sharing (2B/4B models)" path. KV sharing is on when `text_config.num_kv_shared_layers > 0`:

| Model | num_hidden_layers | num_kv_shared_layers | Fires shared_kv branch? |
|---|---:|---:|---|
| gemma-4-e2b | (small) | > 0 | yes |
| gemma-4-e4b-it | 42 | 18 | yes (layers 24–41 per request) |
| gemma-4-26b-a4b-it | 30 | 0 | no — uses `attention_k_eq_v=true` + MoE |
| gemma-4-31b-it | 60 | 0 | no — dense, `attention_k_eq_v=true` |

So this PR's new code primarily affects E2B/E4B; on 26B/31B it just ensures the signature/return-shape change doesn't regress.

## Verification

**Host:** Apple M5 Pro Max, 128 GB unified memory.

### Unit tests
```
tests/test_gemma4_mllm_patch.py — 3 passed (new)
```

Lint clean: `ruff check vllm_mlx/patches/gemma4_mllm.py tests/test_gemma4_mllm_patch.py` + `black --check` both pass.

### Runtime verification

`vllm-mlx serve <model> --continuous-batching --enable-prefix-cache --cache-memory-mb 50 --ssd-cache-dir /tmp/test --ssd-cache-max-gb 5`, then 9 distinct ≥200-token prompts.

| Model | KV-shared layers | Pre-fix | Post-fix |
|---|---|---|---|
| `gemma-4-E4B-it-MLX-4bit` (text-multimodal, mllm=True) | 18 / 42 | every request fails: `TypeError: ... unexpected keyword argument 'shared_kv'`; `entry_count: 0` | ✅ all requests succeed; prefix cache stores (`entry_count > 0`) |
| `mlx-community/gemma-4-26b-a4b-it-4bit` (26B MoE, k_eq_v=true) | 0 / 30 | n/a (model wasn't running) | ✅ loads, all 9 requests succeed @ ~15 tok/s; no `TypeError`, no `Stream(gpu, N)`, no `Abort` — confirms non-regression on the non-shared-kv branch + new `use_k_eq_v=True` path + MoE blocks |

## Test plan

- [ ] CI: `tests/test_gemma4_mllm_patch.py` green on all matrix Pythons.
- [ ] Manual: `vllm-mlx serve mlx-community/gemma-4-e4b-it-4bit --continuous-batching`, send any chat request, confirm 200 with valid content (pre-fix this returns 200 with `finish_reason: error` + `content: null`).
- [ ] Manual: same on `mlx-community/gemma-4-26b-a4b-it-4bit`, confirm non-regression.

### DIFF
--- a/tests/test_gemma4_mllm_patch.py
+++ b/tests/test_gemma4_mllm_patch.py
@@ -101,7 +101,12 @@ def test_patched_call_accepts_shared_kv_and_offset_kwargs(monkeypatch):
 
 
 def test_patched_call_snapshots_offset_before_update(monkeypatch):
-    """RoPE on queries must see the pre-update offset, even with BatchKVCache."""
+    """RoPE on queries must see the pre-update offset, even when BatchKVCache
+    mutates cache.offset in place via __iadd__ (which mx.array supports).
+    Re-bind (self.offset = self.offset + 1) is NOT the bug we're guarding —
+    use a real in-place += so the test would catch a regression that strips
+    the defensive copy.
+    """
     Attention = _install_fake_gemma4_modules(monkeypatch)
     from vllm_mlx.patches.gemma4_mllm import patch_gemma4_attention_for_batching
 
@@ -119,12 +124,14 @@ def test_patched_call_snapshots_offset_before_update(monkeypatch):
             return x
 
     class MutatingBatchCache:
-        # Mimics BatchKVCache: offset is mx.array, update_and_fetch advances it.
+        # mx.array.__iadd__ is in-place — same mutation pattern as real
+        # BatchKVCache. A naive non-copying snapshot survives a rebind but
+        # NOT an in-place +=; this is the actual regression to defend.
         def __init__(self):
             self.offset = mx.array([5])
 
         def update_and_fetch(self, keys, values):
-            self.offset = self.offset + 1
+            self.offset += 1
             return keys, values
 
     attn = Attention()
--- a/tests/test_gemma4_mllm_patch.py
+++ b/tests/test_gemma4_mllm_patch.py
@@ -1,0 +1,247 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Regression tests for the Gemma 4 MLLM attention patch.
+
+Two distinct properties this test guards against:
+
+1. **Signature compatibility with mlx-vlm 0.5.0+.** The patched call must
+   accept ``shared_kv: Optional[tuple]`` and ``offset: Optional[Any]`` and
+   return a 3-tuple ``(output, (keys, values), offset)``. The pre-fix
+   version of this patch only accepted ``(x, mask, cache)`` and returned a
+   single ``mx.array`` — every Gemma 4 request under
+   ``--continuous-batching`` then crashed with::
+
+       TypeError: patch_gemma4_attention_for_batching.<locals>._patched_call()
+       got an unexpected keyword argument 'shared_kv'
+
+2. **Offset snapshot before update_and_fetch (the original reason).**
+   ``BatchKVCache`` holds offset as an ``mx.array``; the patched function
+   must snapshot it BEFORE ``cache.update_and_fetch`` mutates it in place,
+   so the RoPE applied to the queries afterwards still sees the pre-update
+   value.
+"""
+
+import sys
+import types
+from typing import Any
+
+import mlx.core as mx
+
+
+def _install_fake_gemma4_modules(monkeypatch):
+    """Replace mlx_vlm.models.gemma4 / .base in sys.modules with fakes whose
+    ``Attention`` class accepts arbitrary kwargs and records its call.
+
+    Lets us patch + invoke the wrapper without pulling in real mlx-vlm
+    weights or compilation. The fakes are scoped via monkeypatch so they
+    disappear after the test.
+    """
+
+    base_module = types.ModuleType("mlx_vlm.models.base")
+
+    def fake_sdpa(queries, keys, values, cache=None, scale=1.0, mask=None):
+        # Return the queries unchanged — the test only cares about the
+        # return-shape contract, not numeric correctness.
+        return queries
+
+    base_module.scaled_dot_product_attention = fake_sdpa
+
+    language_module = types.ModuleType("mlx_vlm.models.gemma4.language")
+
+    class _IdentityNorm:
+        def __call__(self, x):
+            return x
+
+    class DummyRope:
+        def __call__(self, x, offset=0):
+            return x
+
+    class FakeGemma4Attention:
+        """Minimal stand-in for mlx_vlm.models.gemma4.language.Attention.
+
+        We do not redefine ``__call__`` here — the patch will inject the
+        replacement. The orig call (used by the patch's _orig_call capture)
+        is a simple identity so the patch installs cleanly.
+        """
+
+        n_heads = 8
+        n_kv_heads = 8
+        head_dim = 64
+        scale = 1.0
+        use_k_eq_v = False
+        is_kv_shared_layer = False
+
+        def __init__(self):
+            self.q_proj = lambda x: mx.zeros(
+                (x.shape[0], x.shape[1], self.n_heads * self.head_dim), dtype=x.dtype
+            )
+            self.k_proj = lambda x: mx.zeros(
+                (x.shape[0], x.shape[1], self.n_kv_heads * self.head_dim), dtype=x.dtype
+            )
+            self.v_proj = lambda x: mx.zeros(
+                (x.shape[0], x.shape[1], self.n_kv_heads * self.head_dim), dtype=x.dtype
+            )
+            self.q_norm = _IdentityNorm()
+            self.k_norm = _IdentityNorm()
+            self.v_norm = _IdentityNorm()
+            self.rope = DummyRope()
+            self.o_proj = lambda x: x
+
+        def __call__(self, *args, **kwargs):  # noqa: D401
+            raise AssertionError(
+                "Unpatched __call__ should never run — the patch should "
+                "have replaced it before the test calls it."
+            )
+
+    language_module.Attention = FakeGemma4Attention
+
+    # Stitch the package hierarchy so `from mlx_vlm.models.gemma4.language
+    # import Attention` resolves.
+    gemma4_module = types.ModuleType("mlx_vlm.models.gemma4")
+    gemma4_module.language = language_module
+    models_module = types.ModuleType("mlx_vlm.models")
+    models_module.base = base_module
+    models_module.gemma4 = gemma4_module
+    mlx_vlm_module = types.ModuleType("mlx_vlm")
+    mlx_vlm_module.models = models_module
+
+    monkeypatch.setitem(sys.modules, "mlx_vlm", mlx_vlm_module)
+    monkeypatch.setitem(sys.modules, "mlx_vlm.models", models_module)
+    monkeypatch.setitem(sys.modules, "mlx_vlm.models.base", base_module)
+    monkeypatch.setitem(sys.modules, "mlx_vlm.models.gemma4", gemma4_module)
+    monkeypatch.setitem(sys.modules, "mlx_vlm.models.gemma4.language", language_module)
+
+    return FakeGemma4Attention
+
+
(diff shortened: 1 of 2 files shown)


janhilgard (changes requested):
Thanks for the thorough writeup and the model-coverage table. The core fix — extending `_patched_call` to the mlx-vlm 0.5.0 signature (`shared_kv`/`offset` kwargs + 3-tuple return) — is correct and necessary; verified against `mlx_vlm-0.5.0/mlx_vlm/models/gemma4/language.py:199-243`, the `DecoderLayer` unpack at line 319, and the model loop at lines 532-536. Two issues should be addressed before merge.

## 1. Blocking — `test_patched_call_snapshots_offset_before_update` does not guard the regression it claims to

The test's `MutatingBatchCache.update_and_fetch` does:

```python
self.offset = self.offset + 1   # re-assignment via __add__
```

That is a **rebind**, not an in-place mutation. The actual bug this patch guards against (per the module docstring) is `mx.array.__iadd__` being **in-place**. With a rebind, the original array object is never mutated, so a naive non-copying snapshot (`offset = cache.offset`) survives unchanged — the test passes even with a completely broken snapshot.

Empirically confirmed:

```
PR test 2 cache (self.offset = self.offset + 1):
  naive non-copy snapshot after update = [5]   -> test 2 PASSES even with a broken snapshot
Faithful cache (self.offset += 1, the real bug):
  naive non-copy snapshot after update = [6]   -> a correct test WOULD catch the bug
```

Fix: make the fake cache use `self.offset += 1` so it reproduces the in-place `__iadd__` the patch actually defends against.

## 2. Important — the patch is now ~95% a duplicate of upstream 0.5.0; reason #1 is obsolete

Upstream mlx-vlm 0.5.0 line 223 already does:

```python
offset = mx.array(cache.offset) if cache is not None else 0
```

`mx.array(existing_mx_array)` is a defensive copy — after an in-place `+=` on the original, the copy is unchanged (verified). So **upstream 0.5.0 already fixes the BatchKVCache in-place-offset bug** that was the original reason for this patch. The updated docstring still presents reason #1 ("BatchKVCache compatibility — the original reason for this patch") as a live, current reason, which is now misleading.

After this PR, `_patched_call` differs from upstream `Attention.__call__` in only two ways:
- `if offset is None: offset = _snapshot_cache_offset(cache)` instead of upstream's unconditional `offset = mx.array(cache.offset)`
- the mask-trimming block (lines 152-154), which upstream lacks

So the only genuine vllm-mlx-specific behavior left is the mask trim. Please either (a) confirm the mask trim is still required for BatchKVCache continuous batching and make that the documented reason the patch exists, or (b) consider dropping the patch entirely if upstream 0.5.0 plus mask handling elsewhere is sufficient. Maintaining a near-verbatim copy of upstream's attention as a monkey-patch is exactly what caused this breakage — every upstream change to Gemma 4 attention silently desyncs it.

## Minor

- **Semantic divergence (else branch):** upstream unconditionally recomputes `offset = mx.array(cache.offset)`; the patch only does so `if offset is None`. Behavior matches today because non-shared layers always receive `offset=None` from the model loop, but mirroring upstream verbatim avoids future drift.
- **Dead code:** `_orig_call = Gemma4Attention.__call__` (line 94) is captured but never used inside `_patched_call`.
- **Test fidelity:** all three tests run against fake `mlx_vlm` modules, so they exercise the patch's control flow but never the real 0.5.0 `Attention`/`DecoderLayer` — they would not catch a real upstream signature drift, which is the exact failure mode this PR fixes. Not blocking (the manual runtime table partly covers it), but a thin integration test importing the real `DecoderLayer` would have more teeth.

## Positives

- shared_kv branch and 3-tuple return match upstream exactly.
- The `offset=None` path in the shared_kv branch is never hit in practice — the model loop always threads `offset` together with `shared_kv` via `intermediates[prev_idx]`. Same as upstream; not a bug.
- Mask trimming preserved from the original patch (not lost in the rewrite).
- CI green across the matrix; solid manual verification on real E4B and 26B models.

Item 1 is blocking. Item 2 at minimum needs the docstring corrected; ideally a decision on whether the patch should remain.

emby:
Hit this exact TypeError on mlx-community/gemma-4-e4b-it-4bit with --continuous-batching against mlx-vlm 0.5.0. Worked around it locally with a slightly different approach (using inspect.signature to detect old vs new contract and dispatching to either), and ran a broader benchmark to make sure the fix didn't just silence the crash but produced sensible RoPE positions in batched decode.

Numbers from MacBook M3 Max 128 GB:

| concurrency | balanced (tok/s) | TTFT (ms) |
|-------------|------------------|-----------|
| 1           | 79.5             | 34        |
| 5           | 198.4            | 107       |
| 10          | 225.6            | 213       |
| 20          | 365.6            | 415       |

Outputs were coherent across the sweep, which is the indirect signal that the offset snapshot is still correct under batching after the signature change. Metal peak: 8.2 GB.

Sharing as confirmation that the underlying bug is reproducible and that fixing the signature is sufficient — I don't think the inspect-based fallback in my local patch is worth carrying upstream, especially since d4d41aa correctly observes that upstream mlx-vlm now handles the offset snapshot itself. Looking forward to #564 landing so I can drop my local patch.

janhilgard (approved):
All review items from the previous CHANGES_REQUESTED are addressed in d4d41aa — verified locally.

**Issue 1 (blocking) — fixed.** `MutatingBatchCache.update_and_fetch` now uses `self.offset += 1` (in-place `__iadd__`), faithfully reproducing the real bug. Confirmed the test now has teeth: temporarily reverting the patch to a non-copying `offset = cache.offset` makes `test_patched_call_snapshots_offset_before_update` fail with `assert 6 == 5`, where before it passed regardless.

**Issue 2 (important) — fixed.** The patch is now scoped to its one genuinely vllm-mlx-specific behavior: the oversized-mask trim. `_snapshot_cache_offset` is gone; the else branch mirrors upstream 0.5.0 verbatim (`offset = mx.array(cache.offset) if cache is not None else 0`). The docstring is now accurate — it states the offset-snapshot bug is handled upstream and gives a concrete justification for the mask trim (sliding-window layers cap keys at window=512 while the mask is built once for the full prompt).

**Issues 3 & 4 (minor) — fixed.** Else branch matches upstream verbatim; dead `_orig_call` removed.

Tests pass (3/3), `black --check` clean. The signature, 3-tuple return, and shared_kv branch all match `mlx_vlm-0.5.0/models/gemma4/language.py` exactly. Independent reproduction from @emby corroborates the fix.

LGTM.

waybarrios:
The 0.5.0 fix looks correct and I see it is already approved, so this is just one follow-up item that did not come up in the thread, plus a small test note. Not relitigating anything already settled.

The patch now returns a 3-tuple and is applied unconditionally in `MLLMBatchGenerator.__init__`, which is right for mlx-vlm 0.5.0. The catch is that `pyproject.toml` still declares the floor as `mlx-vlm>=0.4.3`, and on 0.4.x the stock attention returns a single array while the stock `DecoderLayer` does:

```python
# mlx_vlm/models/gemma4/language.py (0.4.x)
h = self.self_attn(h, mask, cache)
h = self.post_attention_layernorm(h)
```

So on any mlx-vlm in [0.4.3, 0.5.0) the patched `__call__` binds a 3-tuple to `h` and `post_attention_layernorm` raises on the first Gemma 4 request. A fresh `pip install` pulls 0.5.0 so it is fine, but anyone with a pinned 0.4.x environment who upgrades vllm-mlx without bumping mlx-vlm hits a cryptic crash. The minimal fix aligns the declared floor with the version the code now needs:

```diff
# pyproject.toml
-    "mlx-vlm>=0.4.3",  # 0.4.3+ required for Gemma 4 support
+    "mlx-vlm>=0.5.0",  # 0.5.0+ required: Gemma 4 Attention returns a 3-tuple
```

This also gives a home to the forward-drift concern from the review (every upstream change to Gemma 4 attention silently desyncs the copy). A signature guard turns both the backward break and the forward drift into a loud skip instead of a silent crash:

```python
import inspect
...
    if getattr(Gemma4Attention, "_batch_patched", False):
        return True

    if "shared_kv" not in inspect.signature(Gemma4Attention.__call__).parameters:
        logger.warning(
            "[Gemma4 patch] Installed mlx-vlm predates the 0.5.0 Attention "
            "signature (no shared_kv); skipping patch. Upgrade to "
            "mlx-vlm>=0.5.0 for Gemma 4 continuous batching."
        )
        return False
```

Separately, on the integration test deferred to a follow-up: worth making it cover the mask trim specifically. That trim is now the patch's sole documented reason to exist, yet all three current tests pass `mask=None` and the fake `scaled_dot_product_attention` ignores the mask, so a regression in `mask[..., -keys.shape[-2]:]` would stay green. A focused check:

```python
def test_patched_call_trims_oversized_mask(monkeypatch):
    Attention = _install_fake_gemma4_modules(monkeypatch)
    captured = {}

    def recording_sdpa(queries, keys, values, cache=None, scale=1.0, mask=None):
        captured["mask"], captured["keys"] = mask, keys
        return queries

    sys.modules["mlx_vlm.models.base"].scaled_dot_product_attention = recording_sdpa
    from vllm_mlx.patches.gemma4_mllm import patch_gemma4_attention_for_batching
    patch_gemma4_attention_for_batching()

    attn = Attention()
    L = 4
    x = mx.zeros((1, L, Attention.n_heads * Attention.head_dim))
    wide_mask = mx.zeros((1, 1, L, L + 6))  # wider than keys
    attn(x, mask=wide_mask, cache=None)
    assert captured["mask"].shape[-1] == captured["keys"].shape[-2] == L
```

PR #569, opened 2026-05-23
feat: TurboQuant — R1 Hadamard rotation for outlier-free MoE weight quantization

janhilgard:
## Summary

Adds support for **TurboQuant**, a QuaRot-style R1 (residual stream Hadamard) rotation fused offline into model weights. Enables 4-bit weight quantization on hybrid-attention Qwen3.5/3.6 models (dense and MoE) without quality degradation.

R1 fusion is mathematically transparent — the rotation cancels through the forward pass — so the model produces equivalent-up-to-quantization-noise logits at inference but with quantization-friendly weight distributions (outlier channels spread by the Hadamard rotation).

## What's included

| File | Purpose |
|------|---------|
| `scripts/turboquant_rotations.py` | Hadamard primitives, RMSNorm fold (handles Qwen3.5/3.6 `(1 + raw)` offset), R1/R2/R4 rotation helpers, toy-block forward equivalence test |
| `scripts/turboquant_quantize.py` | End-to-end BF16 → TurboQuant safetensors conversion (per-shard, low memory). Supports both dense (Qwen3.6-27B) and MoE (Qwen3.6-35B-A3B) architectures with sensitive-layer 8-bit logic matching `mixed_4_8`. Handles fused `experts.gate_up_proj` split + GatedDeltaNet + 3D `switch_mlp` weights |
| `scripts/TURBOQUANT.md` | Usage docs + recipe explanation |
| `vllm_mlx/utils/tokenizer.py` | Loader hook detecting `quantization.type == "turboquant_v1"` |
| `vllm_mlx/patches/turboquant_r4.py` | Opt-in monkey-patch for online R4 Hadamard — only useful when paired with activation quantization (NOT applied by default) |
| `vllm_mlx/mllm_batch_generator.py` | Registers R4 patch when `VLLM_MLX_TURBOQUANT_R4` env var is set |
| `tests/test_turboquant_r4_patch.py` | Unit tests for R4 patch idempotency + scope filtering |

## Recipe (default, weight-only)

- **R1** (residual stream Hadamard) — fused offline into `embed_tokens`, `lm_head`, ALL in/out projections.
- **R2** (head-space rotation) — SKIPPED for Qwen3.5/3.6 because gated attention breaks offline R2 cancellation. Applicable to standard-attention models (MiniMax-M2 etc.) — future work.
- **R4** (online Hadamard before down_proj) — **NOT applied by default**. Per [QuaRot paper](https://arxiv.org/abs/2404.00456) + [AMD Quark docs](https://quark.docs.amd.com/release-0.9/pytorch/tutorial_quarot.html): R4 is only needed for activation quantization. Opt-in via `--enable-r4` + `VLLM_MLX_TURBOQUANT_R4=all`.
- **Quantization**: matches `mixed_4_8` — 4-bit affine base on dense MLP / routed experts / attention, 8-bit on sensitive layers (first 1/8, last 1/8, every 3rd middle for `down_proj` + `v_proj`), 8-bit on `embed_tokens` + `lm_head`. Router gate (`mlp.gate`) and norms kept in fp16.
- **RMSNorm fold**: `(1 + raw_norm)` (Qwen3.5/3.6 offset semantics) folded into downstream projections; norm stored as 1.0 (identity).

### Qwen3.6-35B-A3B (MoE, 256 routed experts, 40 layers)
- **Disk size**: 22 GB (compared to baseline `mixed_4_8` at 24 GB)
- **Quality**: 12/12 coherent answers on a deterministic prompt suite (math, code, reasoning, tool-call style)
- **Architecture coverage exercised**: GatedDeltaNet (linear_attn) + full attention hybrid + MoE SparseMoeBlock with router + shared_expert + 256 routed experts via SwitchGLU

### Qwen3.6-27B Heretic (dense, 64 layers)
- **Disk size**: 19 GB (compared to baseline `mixed_4_8` at 22 GB)
- **Quality**: 12/12 coherent answers on the same deterministic prompt suite
- **Architecture coverage exercised**: dense Qwen3NextMLP (`gate_proj`/`up_proj`/`down_proj`) + GatedDeltaNet + full attention hybrid

## Test plan

- [x] Toy-block forward equivalence test (R1, R4, R1+R4): all < 1e-6 max abs diff in fp32.
- [x] Hadamard math validation: H(H(x)) ≤ 1e-6, R4 cancellation ≤ 1e-4.
- [x] End-to-end conversion of Qwen3.6-35B-A3B BF16 → 22 GB TurboQuant.
- [x] End-to-end conversion of Qwen3.6-27B Heretic BF16 → 19 GB TurboQuant.
- [x] Model load via `vllm_mlx serve` for both architectures: correct `QuantizedLinear` / `QuantizedSwitchLinear` / `QuantizedEmbedding` bit assignments verified.
- [x] 12 deterministic prompt quality test on both models.
- [x] Unit tests for R4 patch idempotency + scope filtering.
- [x] CI: lint, type-check, test-apple-silicon (3.11, 3.13), test-matrix (3.10–3.13) all green.

## Validated on

- Apple M3 Ultra 256 GB.
- mlx 0.31.2, mlx_lm 0.31.3, mlx_vlm 0.5.0.
- BF16 source shards SHA256-verified against HuggingFace LFS pointers before conversion (silent shard corruption from aria2c resume can produce broken-but-converting weights).

## Notes

- Conversion is fast: ~15 s for Qwen3.6-35B-A3B, ~30 s for Qwen3.6-27B Heretic on M3 Ultra.
- Sensitive-layer 8-bit logic matches the PROD `mixed_4_8` predicate, so the recipe is a strict superset (same bit allocation + R1 rotation on top).

## Future work

- R3 (online Q/K Hadamard) for KV cache outlier suppression.
- Lloyd-Max codebook quantization on rotated experts (JANGTQ-style).
- R2 head-space rotation for standard-attention models like MiniMax-M2.
- Throughput benchmarks under controlled load (in isolation from other servers competing for the same GPU).

### DIFF
--- a/scripts/TURBOQUANT.md
+++ b/scripts/TURBOQUANT.md
@@ -1,0 +1,73 @@
+# TurboQuant — Hadamard-rotated weight quantization
+
+QuaRot-style R1 (residual-stream Hadamard) rotation fused offline into model
+weights, enabling 4-bit quantization on hybrid-attention MoE models like
+Qwen3.5/3.6 and MiniMax-M2 without quality degradation.
+
+## What's included
+
+- `turboquant_rotations.py` — math primitives (Hadamard helpers, RMSNorm fold,
+  R1/R2/R4 fusion).
+- `turboquant_quantize.py` — main conversion script for Qwen3.5/3.6 family
+  (BF16 source). Supports both dense (Qwen3.6-27B) and MoE (Qwen3.6-35B-A3B).
+- `minimax_m2_convert.py` — conversion script for MiniMax-M2 family
+  (FP8 block-128 source). Handles `mx.from_fp8` dequantization + per-block
+  scaling + per-expert stacking (256 separate `block_sparse_moe.experts.<n>.w1/w2/w3`
+  → batched `switch_mlp.{gate,down,up}_proj` per layer).
+- `vllm_mlx/utils/tokenizer.py` — loader hook detecting `quantization.type ==
+  "turboquant_v1"` and installing (optional) R4 runtime patch.
+- `vllm_mlx/patches/turboquant_r4.py` — opt-in monkey-patch for online R4
+  Hadamard (only useful with activation quantization).
+
+## Recipe (default)
+
+- `R1` (residual stream Hadamard): fused offline into embed, lm_head, and ALL
+  in/out projections (q/k/v/o_proj, gate/up/down_proj per expert, etc).
+- `R2` (head-space rotation): SKIPPED for Qwen3.5/3.6 because of their gated
+  attention quirk (`output = o_proj(attn_out * sigmoid(query_gate))`) —
+  element-wise gate breaks R2's offline cancellation. Applicable to MiniMax-M2
+  and standard-attention models.
+- `R4` (online Hadamard before down_proj): **NOT applied by default**. Per the
+  QuaRot paper, R4 is only needed for activation quantization. For weight-only
+  4-bit/8-bit affine schemes, R4 is redundant. The implementation is included
+  as opt-in (`--enable-r4` + `VLLM_MLX_TURBOQUANT_R4=all`) for users who pair
+  this with NVFP4/MXFP8 activation quantization downstream.
+- RMSNorm fold: `(1 + raw_norm_weight)` (Qwen3.5/3.6 offset semantics) folded
+  into downstream projections. Norm weights stored as 1.0 so runtime
+  `weight * normalized = normalized` is identity.
+- Quantization: 4-bit affine on routed `switch_mlp.*` experts; 8-bit affine
+  on attention, linear_attn projections, `shared_expert`, embed_tokens,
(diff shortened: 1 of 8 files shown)


janhilgard:
Closing: the original validation (TurboQuant R1-only at +8% throughput vs mixed_4_8) was performed when combined with 3-bit KV cache quantization + Hadamard. Subsequent production experience (2026-05-24) showed the asymmetric_quantized_sdpa runtime path dominates the cost, dragging actual throughput to roughly an order of magnitude below the mixed_4_8 baseline (8 t/s vs 70 t/s on Qwen3.6-35B-A3B in our test). We have reverted production to mixed_4_8 and cannot honestly defend the throughput claim in this PR's description without a clean isolated re-benchmark (TurboQuant model with plain 8-bit KV, no Hadamard). The R1 Hadamard math itself remains technically sound — if anyone wants to pursue this, the conversion script in /Users/janhilgard/oss/quarot/turboquant_quantize.py and the runtime patch in this PR are a working starting point.

PR #571, opened 2026-05-24
feat(simple-engine): wire live KV cache quantization through SimpleEngine

janhilgard:
## Summary

`SimpleEngine` previously ignored `--kv-cache-quantization`, `--kv-k-bits`, `--kv-v-bits`, and `--kv-hadamard` because the live KV quantization machinery introduced in #561 only patches the `BatchedEngine` cache entrypoints (`mlx_lm.generate._make_cache` and `_merge_caches`). The SimpleEngine generation path runs through `mlx_lm.generate.stream_generate -> generate_step -> cache.make_prompt_cache(model)`, so the per-layer caches always came out as plain bf16 `KVCache` regardless of the CLI flags.

This PR makes the same flags effective for the `SimpleEngine` path so models served in non-batched mode can also use `BatchQuantizedKVCache` with asymmetric K/V bits and optional QuaRot R3 Hadamard rotation.

Note: this is stacked on #561 — it depends on `BatchQuantizedKVCache` and `asymmetric_quantized_sdpa` introduced there.

## Changes

- `utils/quantized_kv_cache.install_quantized_kv_cache()` gains `simple_mode: bool = False`. When set, it additionally patches `mlx_lm.models.cache.make_prompt_cache` to wrap returned `KVCache` layers in `BatchQuantizedKVCache` (B=1, `left_padding=[0]`). Non-`KVCache` layers (`ArraysCache`, `RotatingKVCache`) pass through unchanged with a single warning so hybrid / sliding-window models do not regress.
- The SDPA patch also retrofits any already-imported `mlx_lm.models.<arch>` modules whose `scaled_dot_product_attention` binding still points at the original function. Together with installing the patch *before* model load in `SimpleEngine.start()`, this guarantees the model's attention path routes through `asymmetric_quantized_sdpa` even though `from .base import scaled_dot_product_attention` captures the function object at import time.
- `SimpleEngine.__init__` accepts the matching `kv_cache_quantization*` / `kv_hadamard` kwargs and calls `install_quantized_kv_cache(simple_mode=True)` before the backing model is constructed. The MLLM path is left as future work — `mlx_vlm` has its own cache machinery.
- `server.load_model()` forwards the new params to `SimpleEngine`.
- `cli.py` passes the parsed flag values through in the non-batched branch and prints the resulting KV cache config (`3-bit, group_size=64 +QuaRot`) in the existing `Mode: Simple` summary block.

No changes to `BatchedEngine`, `MLLMScheduler`, or the existing `BatchedEngine` test surface.

## Test plan

- [x] `python -m black vllm_mlx/` — formatted
- [x] `python -m py_compile` on each modified file
- [x] Manual end-to-end: `vllm-mlx serve <minimax-m2.7> --kv-cache-quantization --kv-cache-quantization-bits 8 --kv-k-bits 3 --kv-v-bits 3 --kv-hadamard` produces `Installed BatchQuantizedKVCache: K=3-bit, V=3-bit, group_size=64, QuaRot=ON, mode=simple+batched` at startup and `KV cache quantization: 3-bit, group_size=64 +QuaRot` in the `Mode: Simple` summary.
- [x] Functional: `/v1/chat/completions` returns model output through the quantized SDPA path (verified on MiniMax-M2.7 with several bit-width configurations).
- [x] Probe-driven `_supports_system_kv_cache` correctly falls back to the uncached `stream_chat` branch when the prompt cache contains `BatchQuantizedKVCache` entries (logs `System KV cache snapshot disabled: model returned non-KVCache entries ([BatchQuantizedKVCache])`).
- [x] Existing BatchedEngine path unchanged — `simple_mode=False` is the default and the previously-shipped `Mode: Continuous batching` summary still prints the same KV cache line.

## Notes on performance / quality

`asymmetric_quantized_sdpa` is slower than the fused `mx.fast.scaled_dot_product_attention` used by bf16 `KVCache`, so enabling KV quantization for SimpleEngine is a memory-vs-throughput trade-off rather than a free win. Aggressive bit widths (e.g. K=3 V=3) can also degrade output quality on some models. The choice is left to the operator per `--kv-*` flag values — this PR only fixes the wiring so the flags actually apply in Simple mode.


janhilgard:
Closing — same feature as #561 and should not have been opened as a separate stacked PR. Folding the SimpleEngine wiring into #561 as a single squashed commit.

PR #572, opened 2026-05-24
feat(simple-engine): track system KV cache hit/miss counters and expose in /v1/status

janhilgard:
## Summary

SimpleEngine's \`_system_kv_snapshot\` mechanism already caches the system-prompt KV state across requests with an identical system prefix and is enabled by default for any model whose prompt cache is a plain \`KVCache\` (logged as \`System KV cache HIT / MISS / STORED\` at INFO level). The throughput impact is substantial — typical repeated calls to the same system prompt show 4–5× prefill speedup — but no metric was reported to monitoring: \`/v1/status\` only forwarded the BatchedEngine cache stats, so the SimpleEngine \`cache\` field always rendered as \`null\` in dashboards.

This PR adds hit/miss/tokens-saved counters and a stable \`system_kv_cache\` field in \`/v1/status\` so dashboards can render a prompt-cache panel for SimpleEngine the same way they already do for BatchedEngine prefix cache.

## Changes

- \`SimpleEngine\` tracks \`_system_kv_hits\`, \`_system_kv_misses\`, and \`_system_kv_tokens_saved\`. Both stream paths (\`stream_chat\` and the text-model routing path) increment hits on hash+token-count match and misses otherwise. Counters reset alongside the existing snapshot reset in \`stop()\`.
- \`SimpleEngine.get_stats()\` always emits \`system_kv_cache\` (previously only when a snapshot existed) so dashboards have a stable schema to render the panel before the first request lands. Schema:
  - \`enabled\` (mirrors \`_supports_system_kv_cache\`)
  - \`tokens\`, \`hash\`, \`memory_mb\`
  - \`hits\`, \`misses\`, \`hit_rate\`, \`tokens_saved\`
- \`GET /v1/status\` adds a top-level \`system_kv_cache\` field forwarding the SimpleEngine stat (null for BatchedEngine, which has its own \`cache\` block).

No behavioral change to the cache logic itself — purely observability.

## Test plan

- [x] \`python -m black vllm_mlx/\` — formatted
- [x] \`python -m py_compile\` on modified files
- [x] Manual end-to-end: three identical \`/v1/chat/completions\` calls with the same system prompt against MiniMax-M2.7 in Simple mode show one MISS followed by two HITs in the server log, and \`/v1/status\` reports \`system_kv_cache: { hits: 2, misses: 1, hit_rate: 0.667, tokens_saved: …, tokens: …, memory_mb: … }\`.
- [x] BatchedEngine \`/v1/status\` still returns \`system_kv_cache: null\` and the existing \`cache\` block is unchanged.

### DIFF
--- a/vllm_mlx/engine/simple.py
+++ b/vllm_mlx/engine/simple.py
@@ -189,6 +189,12 @@ class SimpleEngine(BaseEngine):
         self._system_kv_snapshot = None  # List of (keys, values) per backbone layer
         self._system_kv_hash = None  # Hash of system prefix text
         self._system_kv_token_count = 0  # Tokens in cached prefix
+        # Hit/miss counters surfaced via get_stats() / GET /v1/status so
+        # dashboards can monitor SimpleEngine prompt-cache effectiveness the
+        # same way they monitor BatchedEngine prefix cache.
+        self._system_kv_hits = 0
+        self._system_kv_misses = 0
+        self._system_kv_tokens_saved = 0
         # True only when the model's prompt cache is composed entirely of
         # plain ``KVCache`` entries. Sliding-window models (gemma3_text,
         # olmo3, recurrent_gemma) return ``RotatingKVCache`` whose ``.state``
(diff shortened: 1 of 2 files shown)


janhilgard:
Closing: we no longer use SimpleEngine in our deployment (the only model that needed it, MiniMax-M2.7, has been retired). Without an in-flight use case to validate against, this observability improvement does not warrant adding to the upstream review queue. Patch is small (38 lines, 2 files) — happy to reopen / re-submit if anyone runs into the same blind spot in the future.

Issue #570, opened 2026-05-23
[Bug]: Qwen3.6-35B-A3B dequantize shape mismatch (works in mlx-vlm)


worldfar2016:
### Environment

python 3.12.9 (main, Feb 12 2025, 15:09:19) [Clang 19.1.6 ]
mlx device Device(gpu, 0)
ProductName:            macOS
ProductVersion:         26.5
BuildVersion:           25F71

### Model and quantization

unsloth/Qwen3.6-35B-A3B-UD-MLX-4bit

### Reproduction

```shell
## Server command


vllm-mlx serve unsloth/Qwen3.6-35B-A3B-UD-MLX-4bit \
  --host 127.0.0.1 --port 1234 \
  --enable-auto-tool-choice --tool-call-parser auto


## Client request


curl -X POST http://127.0.0.1:1234/v1/chat/completions \
  -H 'Content-Type: application/json' \
  -d '{
    "model": "default",
    "messages": [{"role":"user","content":"Hello"}],
    "max_tokens": 100,
    "stream": true
  }'
```

### Expected behavior

The model should load and respond normally, as it does in mlx-vlm:

$ python -m mlx_vlm.chat --model /path/to/unsloth/Qwen3.6-35B-A3B-UD-MLX-4bit
Model loaded successfully!
You: Hello
Assistant: [normal response, fluent multilingual output]

mlx-vlm works flawlessly with this model, confirming the model files and 
underlying MLX quantization support are not the issue.

### Actual behavior

```shell
Model loads successfully (no error during weight loading), but inference 
fails at the first dequantize call:

INFO:vllm_mlx.models.mllm:MLLM loaded successfully: .../Qwen3.6-35B-A3B-UD-MLX-4bit
INFO:vllm_mlx.models.mllm:Native video pipeline enabled (temporal 3D conv + M-RoPE)
INFO:vllm_mlx.text_model_from_vlm:Transferred 1597 weight arrays from vlm language_model
WARNING:vllm_mlx.text_model_from_vlm:No MTP weights found in Qwen3.6-35B-A3B-UD-MLX-4bit
INFO:vllm_mlx.engine.simple:SimpleEngine loaded: ... (MLLM=True, routing=per-request)

[Request comes in]

INFO:vllm_mlx.engine.simple:Text-only request → LLM path (MTP=False)
INFO:vllm_mlx.engine.simple:System KV cache MISS: will prefill 1543 system tokens, 21 suffix tokens
[...]

PR #573, opened 2026-05-24
Guard MLLM text prefix cache on hybrid models

Thump604:
## Summary

This narrows the Qwen3.6 text-only MLLM failure reported in #570 to the manual system-prefix cache path and makes that path respect the existing cache-safety flag.

Qwen3.6 text models use hybrid cache entries (`ArraysCache` for linear layers plus `KVCache` for attention layers). The pure-LLM `stream_chat` cache path already skips manual prefix snapshots when the cache class is not snapshot-safe. The MLLM TextModel route did not apply that guard, so a text-only request with a system/tool prefix could enter the manual cache branch before normal generation.

## What Changed

- Skip the manual system-prefix cache branch in `_stream_generate_text()` when the engine reports a non-snapshot-safe cache class.
- Add a regression proving MLLM TextModel routing does not call `make_prompt_cache()` for the manual cache path when `_supports_system_kv_cache` is false.

## Evidence

- Red test before fix:
  - `tests/test_simple_engine.py::TestSimpleEngineConcurrency::test_stream_generate_text_skips_system_cache_when_text_model_not_safe`
  - Failed because `_stream_generate_text()` called `make_prompt_cache()` despite `_supports_system_kv_cache=false`.
- After fix:
  - `.venv/bin/python -m pytest tests/test_simple_engine.py -q`
  - `39 passed`
  - `.venv/bin/python -m pytest tests/test_server.py::TestChatCompletionStreamingModeSwitching::test_nonstream_mllm_endpoint_text_only_does_not_raise_stream_thread_error -q`
  - `1 passed`
  - `.venv/bin/python -m ruff check vllm_mlx/engine/simple.py tests/test_simple_engine.py`
  - `All checks passed!`
  - `git diff --check`
  - passed

## What Is Not Claimed

- This does not claim a full Qwen3.6 model-quality certification.
- This does not claim every possible Qwen3.6 quantized artifact has been runtime-tested locally.
- This does not change MLLM media routing, native MTP behavior, CB, or the model loader.

Refs #570.

### DIFF
--- a/tests/test_simple_engine.py
+++ b/tests/test_simple_engine.py
@@ -1689,6 +1689,69 @@ class TestSimpleEngineConcurrency:
         ]
 
     @pytest.mark.anyio
+    async def test_stream_generate_text_skips_system_cache_when_text_model_not_safe(
+        self,
+    ):
+        """MLLM TextModel routing must not snapshot hybrid ArraysCache state."""
+        from types import SimpleNamespace
+
+        from vllm_mlx.engine.simple import SimpleEngine
+
+        tokenizer = MagicMock()
+        tokenizer.apply_chat_template.return_value = (
+            "<|im_start|>system\nYou are helpful.<|im_end|>\n"
+            "<|im_start|>user\nhello<|im_end|>\n"
+            "<|im_start|>assistant\n"
+        )
+        tokenizer.bos_token = None
+        tokenizer.eos_token_id = 42
+        tokenizer.encode = MagicMock(
+            side_effect=[
+                list(range(10)),  # full prompt
+                list(range(5)),  # system prefix
+            ]
+        )
+
+        captured_prompts = []
+
+        def fake_stream_generate(model, tokenizer, prompt, **kwargs):
+            captured_prompts.append(prompt)
+            yield SimpleNamespace(text="Hello", finish_reason="stop")
+
+        def fail_if_manual_cache_path_runs(*args, **kwargs):
+            raise AssertionError("manual system-cache path must be skipped")
+
+        engine = SimpleEngine("test-model", force_mllm=True, mtp=False)
+        engine._loaded = True
+        engine._text_model = MagicMock()
+        engine._text_model.mtp = None
+        engine._text_tokenizer = tokenizer
+        engine._supports_system_kv_cache = False
+
+        with (
+            patch("mlx_lm.stream_generate", side_effect=fake_stream_generate),
+            patch(
+                "mlx_lm.models.cache.make_prompt_cache",
+                side_effect=fail_if_manual_cache_path_runs,
+            ),
+        ):
+            outputs = [
+                chunk
+                async for chunk in engine._stream_generate_text(
+                    messages=[
+                        {"role": "system", "content": "You are helpful."},
+                        {"role": "user", "content": "hello"},
+                    ],
+                    max_tokens=16,
+                    temperature=0.3,
+                    top_p=0.8,
+                )
+            ]
+
+        assert outputs[-1].text == "Hello"
+        assert captured_prompts == [tokenizer.apply_chat_template.return_value]
+
+    @pytest.mark.anyio
     async def test_stream_generate_text_disables_mtp_when_logits_processors_active(
         self,
     ):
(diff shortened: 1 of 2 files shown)


janhilgard:
Reviewed the change end-to-end against #570 and the existing `stream_chat` guard. Quick notes from a fellow contributor.

**Correctness**: The repro in #570 shows the failure surfaces immediately after `System KV cache MISS: will prefill 1543 system tokens` — i.e. inside the manual system-prefix branch this PR guards. Qwen3.6's hybrid cache (mixed `ArraysCache` + `KVCache`) is exactly what the `_supports_system_kv_cache` probe (lines 282-304) was designed to catch, and the probe already turns the flag off for those models. Skipping the manual branch when the flag is off is the right narrow fix and avoids reading the wrong layer's quant scales/biases via the snapshot/restore roundtrip.

**Parity with `stream_chat`**: The new block mirrors the existing one in `stream_chat` (~line 1025-1040) exactly — same `cache_blocking_controls = []` list, same `"non_kv_cache_class"` reason string, same log shape. This is the right thing to copy and keeps the two paths from drifting. The list-of-reasons pattern also leaves room for the additional gates the `stream_chat` side already checks (e.g. `max_kv_size`) to be added to the text route later if and when they apply.

**Test**: I like the negative assertion — patching `mlx_lm.models.cache.make_prompt_cache` to raise if entered is the cleanest possible proof that the manual branch is bypassed. Asserting `captured_prompts == [apply_chat_template.return_value]` confirms the uncached path actually receives the full prompt text rather than a sliced suffix. Strong red-before-green.

**Scope**: Explicit "What Is Not Claimed" section is appreciated. To make the residual story explicit for future readers: the underlying ArraysCache (and any non-plain-KVCache) snapshot/restore path remains broken for hybrid models — this PR turns the trip wire off but does not fix the trap. That probably deserves a follow-up issue if there isn't one yet, so anyone digging into why the manual cache snapshot was disabled for these models has a pointer to the real bug rather than just to the guard.

**Minor naming nit (non-blocking)**: `cache_blocking_controls` is a slightly awkward phrase, but it already exists in `stream_chat`, so changing it here in isolation would just add noise. If someone wants to rename it, both call sites should move together.

Nothing else surfaced. CI is green on all matrix entries.

janhilgard (approved):
Approving based on the detailed review above. To summarize the justification:

- **Correct narrow fix** for the failure path identified in #570 — the dequantize shape mismatch fires immediately after `System KV cache MISS`, which is exactly the manual snapshot branch this guard disables for hybrid-cache models.
- **Mirrors the existing `stream_chat` guard verbatim** (same `cache_blocking_controls` list shape, same `non_kv_cache_class` reason, same log format), so the two paths stay in sync rather than diverging.
- **Strong regression test** — patching `make_prompt_cache` to raise is the cleanest possible negative assertion, and the prompt-equality check confirms the uncached path receives the full prompt.
- **CI green** on all 9 matrix entries (lint, type-check, test-apple-silicon ×2, test-matrix ×4, tests).
- **Honest scope** — the PR explicitly does not claim to fix the underlying hybrid snapshot/restore corruption; it disables the trip wire and leaves the deeper bug to a follow-up. Appropriate for a guard fix.

The only follow-up I'd encourage is filing a separate issue for the underlying `ArraysCache` snapshot/restore correctness on hybrid models, so the guard does not silently outlive the memory of why it exists.

waybarrios:
Heads up. this guard already landed on `main` via #541 (commit 1656c15, May 28), which added the exact same `has_system and self._text_model is not None and self._supports_system_kv_cache` check to `_stream_generate_text`. This PR went up and got approved on May 24, right before #541 merged, so the hybrid-cache text path from #570 is already covered. The change here just rewrites that guard into the equivalent `cache_blocking_controls` form (`not cache_blocking_controls` is the same as `_supports_system_kv_cache`), so all that's really left is the INFO skip-log plus the regression test, and it also drops the "Snapshot-safety" comment along the way. Nothing's broken, it just doesn't fix a live bug anymore. I'd either close it as superseded by #541, or rebase on `main` and keep just the regression test, that's the part worth saving.

waybarrios:
Closing as superseded by #541 (1656c15), see https://github.com/waybarrios/vllm-mlx/pull/573#issuecomment-4684392805 for the details. If you want to rescue the regression test as a standalone PR on top of current main, happy to take that.

PR #576, opened 2026-05-24
Fix hybrid cache snapshot aliasing

### LINKED ISSUE #575: System-prefix cache snapshots alias hybrid ArraysCache state containers
## Problem

The manual system-prefix cache path stores prompt-cache layer state directly from `c.state` and restores it later. That is safe for plain `KVCache` tuples, but hybrid models can include `ArraysCache`, whose `state` property returns the cache's mutable `cache` list.

If the saved snapshot keeps that list by reference, later suffix/decode cache updates can mutate the stored system-prefix snapshot. A later cache hit can then restore a snapshot that no longer represents only the system prefix.

## Local repro

A focused regression test demonstrates the aliasing without requiring a large model:

```sh
.venv/bin/python -m pytest tests/test_simple_engine.py::TestSimpleEngineConcurrency::test_stream_chat_system_cache_copies_arrays_cache_state -q
```

The failing-before behavior is:

- prefill writes `[[1, 2, 3]]` into an `ArraysCache`-like state list
- the system-prefix snapshot stores that list directly
- decode mutates the live cache list to `[[9, 9, 9]]`
- the stored snapshot observes `[[9, 9, 9]]` too

## Code path compared

The affected path is the manual system-prefix cache branch in `vllm_mlx/engine/simple.py`:

- `SimpleEngine.stream_chat()` snapshot store/restore
- `SimpleEngine._stream_generate_text()` snapshot store/restore for MLLM TextModel routing
- `mlx_lm.models.cache.ArraysCache.state`, which returns a mutable list container

This is separate from the narrow guard in #573. That guard disables the trip wire for unsupported cache classes. This issue tracks the underlying snapshot/restore aliasing behavior for hybrid `ArraysCache` state.

## Expected behavior

System-prefix snapshots should be isolated from later request decode state. Mutable cache state containers should be copied at snapshot and restore boundaries. Unsupported cache classes, such as rotating/sliding-window cache classes that need additional cursor metadata, should remain guarded off.

## What is not claimed

- This is not a model-quality claim.
- This does not claim all cache classes are snapshot-safe.
- This does not claim runtime validation on every Qwen3.6 artifact.
- This does not change MTP, SpecPrefill, continuous batching, KV quantization, or constrained decoding behavior.

## Proposed narrow fix

Add explicit snapshot/restore helpers that copy mutable cache state containers while reusing immutable MLX arrays, allow `KVCache` and `ArraysCache` through the system-prefix cache eligibility probe, and keep unsupported cache classes guarded off.

Thump604:
## Summary

Fixes #575.

This follows the guard in #573. The guard keeps unsupported cache classes off the manual system-prefix cache branch; this patch fixes the underlying hybrid-cache snapshot aliasing case so `ArraysCache` can be snapshotted safely.

## Problem

`ArraysCache.state` returns the cache's mutable state list. The manual system-prefix cache path stored `c.state` directly, so later suffix/decode cache updates could mutate the saved system-prefix snapshot by reference.

That is not the same shape as plain `KVCache`, where `state` returns a tuple of arrays. Hybrid models that mix `ArraysCache` and `KVCache` need the state containers copied at snapshot and restore boundaries.

## Change

- Add explicit `SimpleEngine` snapshot/restore helpers for prompt-cache state.
- Copy mutable list/tuple state containers while reusing immutable MLX arrays.
- Evaluate nested cache-state arrays without assuming every state is a `(keys, values)` tuple.
- Treat `KVCache` and `ArraysCache` as system-prefix snapshot-safe.
- Keep unsupported cache classes, including rotating/sliding-window cache classes, guarded off.
- Probe MLLM TextModel cache support after the TextModel is built.

## Tests

```sh
.venv/bin/python -m pytest tests/test_simple_engine.py -q
.venv/bin/python -m pytest -q --ignore=tests/test_ssd_cache.py
.venv/bin/python -m ruff check vllm_mlx/engine/simple.py tests/test_simple_engine.py
.venv/bin/python -m black --check vllm_mlx/engine/simple.py tests/test_simple_engine.py
git diff --check
/opt/ai-runtime/bin/lint-upstream-claims --root /Users/David/code/vllm-mlx vllm_mlx/engine/simple.py tests/test_simple_engine.py
```

Local full-suite note: `tests/test_ssd_cache.py` still has inherited local failures that reproduce on clean `upstream/main`; this PR was verified with that file excluded.

## What is not claimed

- No model-quality claim.
- No claim that every cache class is snapshot-safe.
- No change to MTP, SpecPrefill, continuous batching, KV quantization, or constrained decoding semantics.
- No runtime certification claim for every Qwen3.6 artifact.

## Dependency note

This branch is stacked on #573 so the follow-up can be reviewed with the guard context visible. If #573 lands first, this PR should reduce to the snapshot/restore helper changes and focused tests.

### DIFF
--- a/tests/test_simple_engine.py
+++ b/tests/test_simple_engine.py
@@ -1689,6 +1689,69 @@ class TestSimpleEngineConcurrency:
         ]
 
     @pytest.mark.anyio
+    async def test_stream_generate_text_skips_system_cache_when_text_model_not_safe(
+        self,
+    ):
+        """MLLM TextModel routing must not snapshot hybrid ArraysCache state."""
+        from types import SimpleNamespace
+
+        from vllm_mlx.engine.simple import SimpleEngine
+
+        tokenizer = MagicMock()
+        tokenizer.apply_chat_template.return_value = (
+            "<|im_start|>system\nYou are helpful.<|im_end|>\n"
+            "<|im_start|>user\nhello<|im_end|>\n"
+            "<|im_start|>assistant\n"
+        )
+        tokenizer.bos_token = None
+        tokenizer.eos_token_id = 42
+        tokenizer.encode = MagicMock(
+            side_effect=[
+                list(range(10)),  # full prompt
+                list(range(5)),  # system prefix
+            ]
+        )
+
+        captured_prompts = []
+
+        def fake_stream_generate(model, tokenizer, prompt, **kwargs):
+            captured_prompts.append(prompt)
+            yield SimpleNamespace(text="Hello", finish_reason="stop")
+
+        def fail_if_manual_cache_path_runs(*args, **kwargs):
+            raise AssertionError("manual system-cache path must be skipped")
+
+        engine = SimpleEngine("test-model", force_mllm=True, mtp=False)
+        engine._loaded = True
+        engine._text_model = MagicMock()
+        engine._text_model.mtp = None
+        engine._text_tokenizer = tokenizer
+        engine._supports_system_kv_cache = False
+
+        with (
+            patch("mlx_lm.stream_generate", side_effect=fake_stream_generate),
+            patch(
+                "mlx_lm.models.cache.make_prompt_cache",
+                side_effect=fail_if_manual_cache_path_runs,
+            ),
+        ):
+            outputs = [
+                chunk
+                async for chunk in engine._stream_generate_text(
+                    messages=[
+                        {"role": "system", "content": "You are helpful."},
+                        {"role": "user", "content": "hello"},
+                    ],
+                    max_tokens=16,
+                    temperature=0.3,
+                    top_p=0.8,
+                )
+            ]
+
+        assert outputs[-1].text == "Hello"
+        assert captured_prompts == [tokenizer.apply_chat_template.return_value]
+
+    @pytest.mark.anyio
     async def test_stream_generate_text_disables_mtp_when_logits_processors_active(
         self,
     ):
--- a/tests/test_simple_engine.py
+++ b/tests/test_simple_engine.py
@@ -996,6 +996,109 @@ class TestSimpleEngineConcurrency:
         }
 
     @pytest.mark.anyio
+    async def test_stream_chat_system_cache_copies_arrays_cache_state(self):
+        """Hybrid cache snapshots must not alias ``ArraysCache.state`` lists."""
+        import mlx.core as mx
+
+        from vllm_mlx.engine.simple import SimpleEngine
+
+        def apply_chat_template_side_effect(messages, **kwargs):
+            user_content = ""
+            for m in reversed(messages):
+                if m.get("role") == "user":
+                    user_content = m.get("content") or ""
+                    break
+            return (
+                "<|im_start|>system\nYou are helpful.<|im_end|>\n"
+                f"<|im_start|>user\n{user_content}<|im_end|>\n"
+                "<|im_start|>assistant\n"
+            )
+
+        tokenizer = MagicMock()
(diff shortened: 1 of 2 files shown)


janhilgard (approved):
Reviewed against #573 (which I approved earlier) and issue #575. This is exactly the follow-up the #573 review was asking for — fix the underlying aliasing rather than relying forever on the guard.

**Approve rationale**

- **Core insight is correct.** `KVCache.state` returns an immutable tuple, `ArraysCache.state` returns the cache's mutable `self.cache` list. Copying the *containers* (tuple/list) is exactly the right granularity; the `mx.array` payload is functionally immutable in this code path because the cache writes happen via reassignment (`cache.cache[i] = new_array`), not in-place mutation. `_clone_cache_state` rebuilds the container without duplicating the arrays, which is the cheapest correct fix.
- **Both directions protected.** `_snapshot_prompt_cache` copies on capture and `_restore_prompt_cache` copies on restore. Without the second half, a later snapshot could still pick up decode-time writes via a fresh prompt cache aliased to the saved snapshot. The PR closes both paths.
- **Generic `mx.eval` walk.** Replacing `mx.eval([s for pair in snapshot for s in pair])` with `_iter_cache_state_arrays` is a real correctness improvement: the old form assumed every state was a 2-tuple of arrays, which holds for `KVCache` but not for `ArraysCache` (a list with variable shape). Duck-typing on `hasattr(value, "shape") and hasattr(value, "dtype")` is the right discriminator for mx-array payloads inside cache state.
- **MLLM TextModel probe added.** The previous `_supports_system_kv_cache` probe only ran on the non-MLLM path, so the parallel `_text_model` used by `_stream_generate_text` defaulted to `False` and never benefited from the snapshot path. Probing the text model after it's built (with route tag `"mllm_text"` in the log) is a real bug fix beyond the aliasing one.
- **Whitelist is appropriately narrow.** Only `KVCache` and `ArraysCache` are added; `RotatingKVCache` (and anything else) still falls through to the uncached path. The new `test_system_cache_probe_allows_arrays_cache_and_rejects_rotating` pins this contract.
- **Repro test is strong.** `test_stream_chat_system_cache_copies_arrays_cache_state` directly demonstrates the bug: write `[[1,2,3]]` into a `MutableListCache`, snapshot, mutate live cache to `[[9,9,9]]`, then assert the saved snapshot still reads `[[1,2,3]]` while the live cache reads `[[9,9,9]]`. That assertion would have failed before this PR.
- **CI is green** on lint, type-check, all four `test-matrix` Python versions, with `test-apple-silicon` still in flight.

**Layering with #573**

Both PRs are still useful together:

- #573 is the guard — narrow trip wire that keeps unsupported cache classes off the snapshot path entirely. With #576 in place the trip wire mostly stops firing for hybrid models, but it remains as defense-in-depth for any future cache class that lands without going through the snapshot-safe whitelist.
- #576 fixes the underlying behavior so `_supports_system_kv_cache` can stay `True` for hybrid models and they actually get the prefill amortization the snapshot path was designed to provide.

The PR description's "Dependency note" already commits to reducing #576 to just the snapshot/restore helpers if #573 lands first, which is the right cleanup posture — no objection from me to either landing order.

**Minor (non-blocking)**

- `_cache_class_is_system_snapshot_safe` has a fallback string-name path when the `mlx_lm.models.cache` import fails. In practice that import never fails because `mlx_lm` is a hard dep, so the fallback is dead code in any realistic call site. It's harmless and consistent with defensive style in adjacent helpers, just worth knowing it exists.
- Same `_stream_generate_text` guard test appears here as in #573. As you noted, that's expected from the stacking; nothing to do.

Approving.

Issue #578, opened 2026-05-24
Bug: SpecPrefill backbone policy is not threaded through SimpleEngine


Thump604:
## Summary

SpecPrefill has a keep percentage control, but the evenly spaced backbone coverage knob is not threaded end to end through the SimpleEngine SpecPrefill paths. That means callers cannot request deterministic backbone chunk coverage when using SpecPrefill sparse prefill.

## Exact upstream code path checked

Checked `waybarrios/vllm-mlx` `upstream/main` at `9c83c84f78143811d997802dcfc6649eb491eb7a`:

- `vllm_mlx/specprefill.py::select_chunks`
- `vllm_mlx/engine/simple.py::SimpleEngine.stream_generate`
- `vllm_mlx/engine/simple.py::SimpleEngine._stream_generate_specprefill`
- `vllm_mlx/engine/simple.py::SimpleEngine._stream_generate_text`
- `vllm_mlx/api/models.py`
- `vllm_mlx/server.py`
- `vllm_mlx/cli.py`
- `vllm_mlx/model_registry.py`
- `vllm_mlx/lifecycle.py`

## Local repro

Adding focused tests that pass `specprefill_backbone_pct=0.25` into the SimpleEngine SpecPrefill routes fails on upstream main because `select_chunks(...)` is called without `backbone_pct`.

Command:

```bash
python -m pytest -q tests/test_simple_engine.py -k 'specprefill_success_preserves_mtp_path or specprefill_threads_same_cancel_check_to_helpers'
```

Observed failing assertion from the added tests before the patch:

```text
KeyError: 'backbone_pct'
```

## Expected behavior

When SpecPrefill is enabled, the configured or per-request backbone percentage should be passed to `select_chunks(..., backbone_pct=...)` on both SimpleEngine SpecPrefill paths:

- non-MLLM `stream_generate` path
- text MLLM/TextModel route path

`select_chunks` should reserve the requested fraction of chunks for evenly spaced coverage while still meeting the requested keep budget.

## Minimal patch shape

- Add `backbone_pct` to `select_chunks` with default `0.0` to preserve current behavior.
- Add `specprefill_backbone_pct` to SimpleEngine config and per-request kwargs.
- Thread the field through API models, server request forwarding, CLI, lifecycle, and registry config.
- Add tests proving both SimpleEngine SpecPrefill paths pass the backbone policy to `select_chunks`.

## Not claimed

- This is not a performance claim.
- This is not a model-quality or qualification claim.
[...]

PR #579, opened 2026-05-24
fix: thread SpecPrefill backbone policy

Thump604:
## Summary

Threads the SpecPrefill backbone coverage policy through the SimpleEngine SpecPrefill surface.

Related issue: #578

## What changed

- Adds `backbone_pct` to `select_chunks(..., default=0.0)` so current behavior is preserved when unset.
- Adds `specprefill_backbone_pct` to SimpleEngine startup config and per-request kwargs.
- Threads the field through API models, server request forwarding, CLI, lifecycle, and registry config.
- Adds tests proving both SimpleEngine SpecPrefill paths pass `backbone_pct` to `select_chunks`.

## Exact upstream code path compared

Compared against `waybarrios/vllm-mlx` `upstream/main` at `9c83c84f78143811d997802dcfc6649eb491eb7a`:

- `vllm_mlx/specprefill.py::select_chunks`
- `vllm_mlx/engine/simple.py::SimpleEngine.stream_generate`
- `vllm_mlx/engine/simple.py::SimpleEngine._stream_generate_specprefill`
- `vllm_mlx/engine/simple.py::SimpleEngine._stream_generate_text`
- `vllm_mlx/api/models.py`
- `vllm_mlx/server.py`
- `vllm_mlx/cli.py`
- `vllm_mlx/model_registry.py`
- `vllm_mlx/lifecycle.py`

Runtime preflight artifact:

`/opt/ai-runtime/run/upstream-local-repro-preflight/20260524T235045Z-upstream-local-repro-preflight.json`

## Verification

```bash
python -m pytest -q tests/test_simple_engine.py tests/test_cli.py tests/test_model_registry.py tests/test_lifecycle_server.py
# 129 passed, 1 warning

python -m pytest -q --ignore=tests/test_ssd_cache.py
# 2068 passed, 11 skipped, 23 deselected, 1 warning

ruff check vllm_mlx/ tests/ --select E,F,W --ignore E402,E501,E731,F811,F841
# passed

black --check --target-version py312 vllm_mlx/ tests/
# passed

git diff --check
# passed
```

Full `pytest -q` was also run. It now gets past this patch's lifecycle/CLI coverage and reports only the existing SSD-cache test failures in `tests/test_ssd_cache.py` on this machine.

## Not claimed

- No performance claim.
- No model-quality or qualification claim.
- No BatchedEngine SpecPrefill claim.
- No default behavior change when `specprefill_backbone_pct` is unset.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -67,10 +67,16 @@ def test_serve_command_propagates_all_sampling_defaults(monkeypatch):
     from vllm_mlx import cli, server
     from vllm_mlx.utils import download
 
+    loaded = {}
+
     monkeypatch.setattr(
         download, "ensure_model_downloaded", lambda *args, **kwargs: "local-test-model"
     )
-    monkeypatch.setattr(server, "load_model", lambda *args, **kwargs: None)
+    monkeypatch.setattr(
+        server,
+        "load_model",
+        lambda *args, **kwargs: loaded.update({"args": args, "kwargs": kwargs}),
+    )
     monkeypatch.setattr("uvicorn.run", lambda *args, **kwargs: None)
 
     for attr in (
@@ -91,6 +97,7 @@ def test_serve_command_propagates_all_sampling_defaults(monkeypatch):
             default_min_p=0.0,
             default_presence_penalty=0.0,
             default_repetition_penalty=1.0,
+            specprefill_backbone_pct=0.25,
         )
     )
 
@@ -100,3 +107,4 @@ def test_serve_command_propagates_all_sampling_defaults(monkeypatch):
     assert server._default_min_p == 0.0
     assert server._default_presence_penalty == 0.0
     assert server._default_repetition_penalty == 1.0
+    assert loaded["kwargs"]["specprefill_backbone_pct"] == 0.25
--- a/tests/test_model_registry.py
+++ b/tests/test_model_registry.py
@@ -74,6 +74,7 @@ def _defaults() -> RegistryServeDefaults:
         specprefill_enabled=False,
         specprefill_threshold=8192,
         specprefill_keep_pct=0.3,
+        specprefill_backbone_pct=0.0,
         specprefill_draft_model=None,
         stream_interval=1,
         gpu_memory_utilization=0.9,
--- a/tests/test_simple_engine.py
+++ b/tests/test_simple_engine.py
@@ -1544,6 +1544,10 @@ class TestSimpleEngineConcurrency:
             captured["kwargs"] = kwargs
             yield SimpleNamespace(text="B", finish_reason="stop")
 
+        def fake_select_chunks(_importance, **kwargs):
+            captured["select_chunks_kwargs"] = kwargs
+            return mx.array([0, 1, 2], dtype=mx.int32)
+
         tokenizer = MagicMock()
         tokenizer.apply_chat_template.return_value = "<|im_start|>user\nhello"
         tokenizer.bos_token = None
@@ -1588,7 +1592,7 @@ class TestSimpleEngineConcurrency:
             ),
             patch(
                 "vllm_mlx.specprefill.select_chunks",
-                return_value=mx.array([0, 1, 2], dtype=mx.int32),
+                side_effect=fake_select_chunks,
             ),
             patch(
                 "vllm_mlx.specprefill.sparse_prefill",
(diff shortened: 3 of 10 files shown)


janhilgard (approved):
Reviewed end-to-end against issue #578 and the existing `specprefill_keep_pct` threading pattern. The bug is real, the scope is correctly bounded, and the implementation mirrors the existing analog cleanly.

**Approve rationale**

- **`select_chunks` algorithm is sound.** The new path keeps `keep_pct` as the effective floor for both chunk count and token count (via the new `target_tokens = max(1, ceil(M * keep_pct))` budget check), allocates `backbone_n = ceil(n_chunks * backbone_pct)` evenly-spaced chunk indices via `round(i * (n_chunks - 1) / max(1, backbone_n - 1))`, deduplicates against the score-ranked top-`top_n` set, and pads with additional top-scoring chunks if the union under-shoots either budget. Default `backbone_pct=0.0` short-circuits to the original score-only behavior with the `if backbone_pct > 0 else 0` guard, so existing callers are byte-equivalent.
- **Per-request override semantics are correct.** The new resolver uses `specprefill_backbone_pct if specprefill_backbone_pct is not None else self._specprefill_backbone_pct` rather than the truthiness pattern (`X or self._X`) used by `effective_keep` immediately above it. That means a caller can explicitly send `specprefill_backbone_pct=0.0` to override an engine-level non-zero default, which is the semantics you want for a "disable backbone for this request" knob. Leaving the existing `keep_pct` truthiness untouched is the right scoping call (out of scope here).
- **Threading is complete.** Mirroring `specprefill_keep_pct` end-to-end: `cli.py` argparse + status print + both forwarding sites; `lifecycle.ModelSpec`; all three `model_registry` dataclasses (`RegistryServeDefaults` / `RegisteredModel` / `ResolvedModelConfig`) + YAML parsing + per-model resolver + `_instantiate_model`; `api/models.{ChatCompletionRequest,CompletionRequest}`; `server.{_prepare_chat_completion_invocation,_build_engine,load_model,create_completion,stream_completion}`; `engine/simple.SimpleEngine.{__init__,stream_generate,_stream_generate_specprefill,_stream_generate_text,get_stats}`. Both SimpleEngine SpecPrefill call sites (`_stream_generate_specprefill` and the `_run_specprefill` nested helper inside `_stream_generate_text`) now pass `backbone_pct=effective_backbone` to `select_chunks`. The `getattr(request, "specprefill_backbone_pct", None)` defensive reads in the server handlers are slightly more verbose than the matching `request.specprefill_keep_pct` reads but they're harmless and protect against older request schemas.
- **Test coverage is the right shape.** Both new assertions verify the threading hypothesis directly — replacing the previous `return_value=...` mock on `vllm_mlx.specprefill.select_chunks` with a `side_effect` that captures kwargs lets the test assert `captured["select_chunks_kwargs"]["backbone_pct"] == 0.25` for each path. `test_cli.py` adds an `loaded["kwargs"]["specprefill_backbone_pct"] == 0.25` end-to-end check via `serve_command`. `test_model_registry.py` extends the `_defaults()` fixture.
- **CI is green** on all nine matrix entries (lint, type-check, test-apple-silicon ×2, test-matrix ×4, tests).
- **Honest scope.** "Not claimed" section accurately states no perf claim, no BatchedEngine claim, no default behavior change. That matches the diff.

**Minor follow-ups (non-blocking)**

1. With `backbone_pct > keep_pct` (e.g., `keep_pct=0.1`, `backbone_pct=0.5`) the selected chunk count can exceed the nominal `keep_pct` budget — `keep_n = ceil(n_chunks * keep_pct)` becomes a floor in the fallback loop, not a ceiling, when backbone alone fills more. That's a reasonable interpretation of "reserve the requested fraction of chunks for evenly spaced coverage **while still meeting the requested keep budget**" — backbone is treated as a hard reservation rather than a portion of the budget — but it's worth a one-line docstring note so operators don't get surprised when `backbone_pct=0.5` with `keep_pct=0.1` returns more tokens than expected.
2. The new `effective_backbone` resolver uses `is not None` semantics; the adjacent `effective_keep` uses `or` truthiness. Both call sites would read more uniformly if `keep_pct` were migrated to the same `is not None` style at some point, but that's a separate cleanup and not appropriate to fold into this PR.

Approving.

waybarrios (commented):
I carefully checked - I green-lighted it

Issue #568, opened 2026-05-23
Bug: gpt-oss chat-template diverges from training format → ~25% CC quality loss vs ollama


CBribiescas:
First — apologies for the volume of issues/PRs from me lately (#556, #562, #563, #564, now this). I'm running "prod" on the fork and reporting as I benchmark. After this I think the large pieces are in place.

## TL;DR

For `gpt-oss-120b` on multi-turn tool-using workloads, vllm-mlx scores ~25 percentage points lower than ollama on the **same model weights**, **same OpenAI chat-completions payload**, **same harness**. CC1 of a local Claude-Code-style agentic benchmark: vllm-mlx **0.400** vs ollama **0.650** (20 tasks).

Root cause: the prompt rendered for the model diverges in three concrete ways from the wire format the gpt-oss weights were trained on. Ollama ships its own Go chat template that matches; vllm-mlx faithfully uses the Jinja `chat_template` from the gpt-oss `tokenizer_config.json` — and that bundled Jinja is, in three concrete ways, off-spec.

## The three structural deltas

Sending an identical 5-message conversation (system, user, assistant+`tool_calls`, tool, user) through ollama (via `/v1/chat/completions`) and through vllm-mlx, then extracting both rendered prompts:

| element | vllm-mlx (bundled harmony Jinja) | ollama (own Go template, matches training) |
|---|---|---|
| Developer-block section order | `# Instructions` then `# Tools` | `# Tools` then `# Instructions` |
| Tool-call header | `<\|start\|>assistant to=functions.X<\|channel\|>commentary json<\|message\|>{...}<\|call\|>` | `<\|start\|>assistant<\|channel\|>commentary to=functions.X <\|constrain\|>json<\|message\|>{...}<\|call\|>` |
| Tool result | `<\|start\|>functions.X to=assistant<\|channel\|>commentary<\|message\|>"json-quoted"<\|end\|>` | `<\|start\|>functions.X to=assistant<\|message\|>raw content<\|end\|>` |

Single-turn reproduction: on the same multi-turn prompt asking to fix a bug in `foo.py`, ollama emits `write_file({"path":"foo.py","content":"def foo():\n    return 1..."})` as its very next tool call. vllm-mlx emits freeform reasoning text ending in a stray `{"path":"foo.py"}` JSON fragment that the harmony tool-call parser can't extract → **zero tool calls returned**, the agent treats the turn as terminal, the verifier sees the filesystem still broken → `failure_mode=logical`.

[...]

PR #580, opened 2026-05-25
fix(llama-tool-parser): recognize Llama 3.1+ / 3.3 tool-call formats

CBribiescas:
First — apologies for another PR; I know I've been busy in the issue tracker lately. After this one, the only thing I'm still investigating locally is the gpt-oss harmony rendering side from #568 (waiting on maintainer direction there per @Thump604's reply). Putting this up because it's a narrow, well-scoped fix in a different parser.

## The bug

`LlamaToolParser` only matches the legacy `<function=name>{"json"}</function>` XML format. But Llama 3.1+ and Llama 4 emit calls with the `<|python_tag|>{"name":..., "parameters":...}` JSON envelope per [Meta's model card](https://github.com/meta-llama/llama-models/blob/main/models/llama3_1/MODEL_CARD.md), and Llama 3.3 emits the same JSON shape **bare**, without any marker:

```
{"type": "function", "name": "read_file", "parameters": {"path": "foo.py"}}
```

Both currently leak through to `message.content` untouched and downstream callers see `tool_calls: []`, treating the assistant turn as a clean stop.

## Fix shape

Two extraction helpers added ahead of the existing XML matcher:

- `_extract_python_tag_calls` — scans for `<|python_tag|>` markers and uses `json.JSONDecoder().raw_decode()` to extract each following JSON object (so multiple tool calls in one response work).
- `_extract_bare_json_call` — only fires when the *whole stripped content* is a JSON object **with `name` AND (`parameters` OR `arguments`)**. Intentionally strict to avoid mis-parsing user-visible JSON the model might emit for unrelated reasons.

Existing `<function=...>{...}</function>` XML behavior is unchanged.

## Empirical signal

On a multi-turn Claude-Code-style agentic benchmark (20 tasks, T=0):

| variant | tasks_passed | avg_turns_taken | score |
|---|---:|---:|---:|
| Llama-3.3-70B-Instruct-4bit, pre-patch | 0/20 | 1.00 | 0.000 |
| Llama-3.3-70B-Instruct-4bit, post-patch | **4/20** | **3.40** | **0.200** |

Same model weights, same harness, same prompt. The avg_turns lift from 1.00 -> 3.40 is the signal: the model was emitting valid bare-JSON tool calls on every turn — they were just being silently dropped.

## Test coverage

8 new tests + 3 existing — all passing:

- python_tag JSON, with `parameters` and the `arguments` alias
- multiple python_tag calls in one response
- bare-JSON envelope, with and without the `type:function` field
- precedence: python_tag short-circuits the bare-JSON path (no double-count)
- negative cases: plain text and non-tool JSON must NOT be flagged

```
$ pytest tests/test_tool_parsers.py::TestLlamaToolParser -v
============================== 11 passed in 0.48s ==============================
```

## Relation to #568

Same class of bug as the gpt-oss rendering issue I reported in #568 — vllm-mlx code paths haven't kept up with formats the models actually emit at inference time. This one is on the *output* side (parsing) rather than the *input* side (prompt rendering), which is why it's a much cleaner, smaller patch: no template-format design choices, no `openai-harmony` vs in-tree decision, no maintainer-direction blocker. Just two regex/decoder helpers and tests.

Happy to address review feedback.


Thump604 (changes requested):
Thanks for narrowing this. The complete-response extraction path looks like the right direction, but I think one served path needs to be covered before merge.

Blocking items:

1. The streaming parser still only recognizes the legacy XML marker. `extract_tool_calls_streaming()` returns `{"content": delta_text}` whenever `"<function=" not in current_text`, so streamed Llama 3.1+/3.3/4 outputs using `<|python_tag|>{...}` or the bare JSON envelope will still be emitted as assistant content and never become `tool_calls`. Since the server calls this streaming parser for streamed chat completions, this leaves `stream=True` tool-use clients in the old failure mode. Please add streaming coverage for the new formats, or otherwise make the streaming path use equivalent detection once the tagged/bare JSON object is complete.

2. CI lint is currently failing because Black would reformat `vllm_mlx/tool_parsers/llama_tool_parser.py`.

Not claiming the non-stream extraction fix is wrong; this is about making the existing streaming parser path match the new non-stream behavior before merge.

CBribiescas:
Pushed the fix in `9222c4e`:

- Streaming parser now detects `<|python_tag|>` and bare-JSON envelopes alongside legacy `<function=`. 4 new tests in `TestLlamaToolParserStreaming` cover both new shapes plus regression guards for plain-text and legacy-XML paths.
- Black-formatted `vllm_mlx/tool_parsers/llama_tool_parser.py` and `tests/test_tool_parsers.py`.

16 Llama parser tests + 124 tool-parser suite green; no regression in `test_native_tool_format`. Ready for another look.

CBribiescas:

Thump604 (approved):
Thanks for the follow-up fix. I re-reviewed the final head after `9222c4e`: the streaming parser now covers the Llama python-tag and bare-JSON formats as well as the existing legacy XML path, with regression coverage for plain text and formatting fixed. CI is green on the updated head. Approving.

waybarrios:
Hey, thanks for this one. The non-streaming Llama 3.1+/3.3 parsing looks great and I checked it against current main, all good there.

I did hit one thing on the streaming side though. The server gate `_streaming_tool_markup_possible` (the `_STREAMING_TOOL_MARKERS` tuple in `server.py`) wasn't including `<|python_tag|>`, so streamed `<|python_tag|>{...}` responses never actually reached `extract_tool_calls_streaming`. They just fell through and got sent as plain content. The new streaming tests still pass because they call the parser directly, so they skip that gate.

I went ahead and pushed `8856470` that adds `<|python_tag|>` to the markers, so the python_tag streaming path works end to end now. I left bare-JSON (`{...}`, Llama 3.3) streaming alone on purpose, since matching a leading `{` felt too risky to gate safely, so that one only resolves in non-streaming for now.

Can you take a look and let me know if it matches what you had in mind? And if you want bare-JSON streaming covered too, happy to figure out a safe way to do it.

CBribiescas:
Thanks for digging into this, and my apologies for the very slow reply — work got busy and this dropped off my radar completely. It didn't help that I'd written the PR off as stalled about an hour before your comment landed, so I wasn't watching the thread when you came back to it. Entirely on me.

Your diagnosis is right, and the gap was worse than a missing marker. My streaming tests called `extract_tool_calls_streaming` directly, so they never went through `_streaming_tool_markup_possible` at all — the gate short-circuited before the parser ever saw `<|python_tag|>`, and green tests actively hid that. So the tests were the real defect; the marker was just where it surfaced.

I've read through `8856470` and it's exactly what I had in mind. Gating on the special token is the right call, and I'd keep the comment explaining why bare-JSON is deliberately excluded — that reasoning is not obvious six months from now.

On bare-JSON streaming: I think it can be gated safely, and the pattern already exists in the file. `_STREAMING_BARE_BRACKET_MARKER` / `_STREAMING_BARE_BRACKET_PARTIAL` solve this same class of problem for `[func({`. The Llama 3.3 equivalent would be roughly:

```python
_STREAMING_BARE_JSON_MARKER = re.compile(r'^\s*\{\s*"(?:name|type)"\s*:')
_STREAMING_BARE_JSON_PARTIAL = re.compile(r'^\s*\{\s*"?(?:n(?:a(?:m(?:e)?)?)?|t(?:y(?:p(?:e)?)?)?)?"?\s*:?$')
```

Two things make that meaningfully tighter than matching a leading `{`:

- **Anchored at position 0.** A Llama 3.3 bare-JSON tool call is the *entire* response — it never appears mid-prose. `^` therefore rules out the case that makes a bare `{` scary, which is a stray brace inside ordinary explanatory text.
- **The first key must be `name` or `type`,** not just any object open. `{"name":` is a far narrower target than `{`.

The partial does hold a leading `{` for a few characters until the first key is readable, then either promotes to a tool call or flushes as content — the same trade the bracket path already makes, and bounded by `_STREAMING_TOOL_MARKUP_SCAN_CHARS`.

Worth noting for the safety argument: every call site of the gate is behind `if tool_parser`, and the parser is only constructed when `tools and tool_choice != "none"`. So this can never fire on plain chat. The residual false positive is a request that supplies tools *and* legitimately asks for JSON output beginning `{"name": ...` — narrow, and genuinely ambiguous even to a human reader.

Happy to push that, plus streaming tests that drive the requests through the server gate instead of calling the parser directly — that's the hole that let the python_tag case through, and it'll keep letting things through until something covers it. Say the word and I'll add both here, or I'm equally happy for you to land this as-is and take bare-JSON in a follow-up.

Thump604:
Yes, please add the anchored bare-JSON streaming gate and drive both the `python_tag` and bare-JSON cases through the server gate rather than calling the parser directly. Your proposed `name|type` first-key restriction, bounded partial scan, and tools-enabled-only call site are the right safety boundary. Please also rebase the branch on current `main`, since GitHub now reports it conflicting. I will review that exact head once pushed.

waybarrios (approved):

PR #582, opened 2026-05-25
Validate OpenAI tool function names

### LINKED ISSUE #577: [Bug]: Tool name containing spaces is not parsed correctly with `--tool-call-parser gemma4`
### Environment

- Chip M2
- RAM: 16GB
- macOS version: 26.5
- Python version: 3.14.5
- vllm-mlx version: 0.3.0
- mlx-lm: 0.31.3
- mlx-vlm: 0.5.0
- mlx-audio: 0.4.3
- mlx-embeddings: 0.1.0

Example server configuration:

```
vllm-mlx serve mlx-community/gemma-4-e4b-it-6bit --port 8000 --host 0.0.0.0 --timeout 1200 --enable-auto-tool-choice --tool-call-parser gemma4 --reasoning-parser gemma4
```

### Model and quantization

mlx-community/gemma-4-e4b-it-6bit

### Reproduction

```shell
  -d '{
    "model": "mlx-community/gemma-4-e4b-it-6bit",
    "messages": [
      {
        "role": "user",
        "content": "Return some hobbies using the Answer Tool."
      }
    ],
    "tools": [
      {
        "type": "function",
        "function": {
          "name": "Answer Tool",
          "description": "Return hobbies",
          "parameters": {
            "type": "object",
            "properties": {
              "hobbies": {
                "type": "array",
                "items": { "type": "string" }
              }
            },
            "required": ["hobbies"]
          }
        }
      }
    ],
    "stream": false
  }'
```

### Expected behavior

It is unclear whether tool names with spaces are supported or not in the OpenAI-compatible API.

From my perspective:

* If tool names with spaces are **not supported**, it would be helpful if the server explicitly rejected them (e.g. validation error), instead of silently failing to parse.
* If they are **supported**, then this seems to be a bug in the `gemma4` tool-call parser.

Currently, the behavior is a silent fallback to raw text, which makes debugging difficult.

### Actual behavior

```shell
{"id":"chatcmpl-9db638af","object":"chat.completion","created":1779631670,"model":"mlx-community/gemma-4-e4b-it-6bit","choices":[{"index":0,"message":{"role":"assistant","content":"<|tool_call>call:Answer Tool{hobbies:[<|\"|>learning and processing information<|\"|>,<|\"|>generating creative text formats<|\"|>,<|\"|>answering questions<|\"|>,<|\"|>code generation<|\"|>]}<tool_call|>","reasoning_content":"The user is asking for my hobbies and explicitly instructs me to use the `Answer Tool` to answer.\nThe `Answer Tool` requires a `hobbies` argument, which is a list of strings.\nSince I am an AI, I don't have personal hobbies in the human sense. I can list activities related to my function or development.\n\nI will call the `Answer Tool` with a list of relevant \"hobbies\" for an AI."},"finish_reason":"stop"}],"usage":{"prompt_tokens":90,"completion_tokens":135,"total_tokens":225}}
```

### Additional context

Following example works fine:

```bash
  -d '{
    "model": "mlx-community/gemma-4-e4b-it-6bit",
    "messages": [
      {
        "role": "user",
        "content": "Return some hobbies using the Answer Tool."
      }
    ],
    "tools": [
      {
        "type": "function",
        "function": {
          "name": "answer_tool",
          "description": "Return hobbies",
          "parameters": {
            "type": "object",
            "properties": {
              "hobbies": {
                "type": "array",
                "items": { "type": "string" }
              }
            },
            "required": ["hobbies"]
          }
        }
      }
    ],
    "stream": false
  }'
```

Result:

```
{"id":"chatcmpl-2999d1fc","object":"chat.completion","created":1779631527,"model":"mlx-community/gemma-4-e4b-it-6bit","choices":[{"index":0,"message":{"role":"assistant","content":null,"reasoning_content":"The user is asking for my hobbies and explicitly instructs me to use the `answer_tool` to answer.\nThe `answer_tool` requires a `hobbies` argument, which is a list of strings.\nSince I am an AI, I don't have personal hobbies in the human sense. I should provide a list of activities related to my function or nature as an AI.\n\nI will call `answer_tool` with a list of relevant \"hobbies\".","tool_calls":[{"id":"call_d59f107d","type":"function","function":{"name":"answer_tool","arguments":"{\"hobbies\":[\"learning and processing information\",\"generating creative text formats\",\"answering questions\",\"translating languages\"]}"}}]},"finish_reason":"tool_calls"}],"usage":{"prompt_tokens":92,"completion_tokens":141,"total_tokens":233}}
```

Thump604:
## Summary

- validate OpenAI-compatible function tool names at the request model boundary
- reject names with spaces before template rendering or tool-parser fallback can leak raw tool text into `message.content`
- add regression coverage for accepted function-name characters and invalid spaced names from raw request payloads

Fixes #577.

## Tests

```sh
/Users/David/code/vllm-mlx/.venv/bin/python -m ruff check vllm_mlx/api/models.py tests/test_api_models.py
/Users/David/code/vllm-mlx/.venv/bin/python -m black --check --fast vllm_mlx/api/models.py tests/test_api_models.py
/Users/David/code/vllm-mlx/.venv/bin/python -m pytest tests/test_api_models.py tests/test_tool_choice_forced.py tests/test_tool_choice_none.py tests/test_gemma4_openai_format.py -q
/Users/David/code/vllm-mlx/.venv/bin/python -m pytest -q --ignore=tests/test_ssd_cache.py
git diff --check
/opt/ai-runtime/bin/lint-upstream-claims --root /Users/David/.config/superpowers/worktrees/vllm-mlx/gemma4-invalid-tool-name-validation vllm_mlx/api/models.py tests/test_api_models.py
```

## What is not claimed

- This does not add support for function names containing spaces.
- This does not change Gemma 4 parser behavior for valid function names.
- This is not a model-quality claim.

### DIFF
--- a/tests/test_api_models.py
+++ b/tests/test_api_models.py
@@ -8,6 +8,9 @@ These are pure Pydantic models with no MLX dependency.
 
 import time
 
+import pytest
+from pydantic import ValidationError
+
 from vllm_mlx.api.models import (
     AssistantMessage,
     AudioSeparationRequest,
@@ -196,6 +199,25 @@ class TestToolCalling:
         assert td.type == "function"
         assert td.function["name"] == "get_weather"
 
+    def test_tool_definition_allows_openai_function_name_characters(self):
+        td = ToolDefinition(
+            function={
+                "name": "get-weather_2",
+                "description": "Get weather",
+            }
+        )
+
+        assert td.function["name"] == "get-weather_2"
+
+    def test_tool_definition_rejects_function_name_with_spaces(self):
+        with pytest.raises(ValidationError, match="function.name"):
+            ToolDefinition(
+                function={
+                    "name": "Answer Tool",
+                    "description": "Return hobbies",
+                }
+            )
+
 
 class TestResponseFormat:
     """Tests for structured output models."""
@@ -273,6 +295,22 @@ class TestChatCompletion:
         assert req.stream_options.include_usage is True
         assert req.tools is not None
         assert req.timeout == 30.0
+
+    def test_request_rejects_tool_name_with_spaces(self):
+        with pytest.raises(ValidationError, match="function.name"):
+            ChatCompletionRequest(
+                model="test-model",
+                messages=[Message(role="user", content="Use the Answer Tool.")],
+                tools=[
+                    {
+                        "type": "function",
+                        "function": {
+                            "name": "Answer Tool",
+                            "description": "Return hobbies",
+                        },
+                    }
+                ],
+            )
 
     def test_mllm_request_params(self):
         req = ChatCompletionRequest(
--- a/vllm_mlx/api/models.py
+++ b/vllm_mlx/api/models.py
@@ -9,11 +9,12 @@ These models define the request and response schemas for:
 - MCP (Model Context Protocol) integration
 """
 
+import re
 import time
 import uuid
 from typing import Any
 
-from pydantic import AliasChoices, BaseModel, Field, model_serializer
+from pydantic import AliasChoices, BaseModel, Field, model_serializer, model_validator
 
 # =============================================================================
 # Content Types (for multimodal messages)
@@ -88,6 +89,9 @@ class Message(BaseModel):
 # =============================================================================
 
 
+_OPENAI_FUNCTION_NAME_RE = re.compile(r"[A-Za-z0-9_-]{1,64}")
+
+
 class FunctionCall(BaseModel):
     """A function call with name and arguments."""
 
@@ -108,6 +112,15 @@ class ToolDefinition(BaseModel):
 
     type: str = "function"
     function: dict
+
+    @model_validator(mode="after")
+    def _validate_openai_function_name(self):
+        if self.type != "function":
+            return self
+        name = self.function.get("name")
+        if not isinstance(name, str) or not _OPENAI_FUNCTION_NAME_RE.fullmatch(name):
+            raise ValueError("function.name must match ^[A-Za-z0-9_-]{1,64}$")
+        return self
 
 
 # =============================================================================


janhilgard (approved):
Reviewed against issue #577 and the OpenAI function-name spec.

**Approve rationale**

- **Right level of intervention.** The underlying gemma4 parser bug (silently swallowing the tool call into `message.content` when the name has a space) is real but separable; fixing it inside every tool-call parser would be a much wider change. Rejecting non-conforming names at the request-model boundary surfaces the problem to the client immediately with a structured `ValidationError` instead of returning a 200 with garbled raw text. The PR description honestly says "This does not add support for function names containing spaces" and "This does not change Gemma 4 parser behavior for valid function names" — that matches the diff exactly.

- **Regex matches the OpenAI spec.** `[A-Za-z0-9_-]{1,64}` + `fullmatch` is the documented constraint for OpenAI function tool names. Anthropic's stricter subset is also satisfied. The validator correctly only fires for `type == "function"` so future non-function tool types (e.g. `code_interpreter`) don't get accidentally rejected.

- **Edge cases handled.** Empty string fails `{1,64}` (min length 1). Missing `name` key returns `None` from `function.get("name")`, fails the `isinstance(name, str)` guard, raises a clear `ValueError`. Non-`str` types (int, list) hit the same guard. `model_validator(mode="after")` runs after field coercion, so callers passing a dict literal for `function` still get validated.

- **Test shape is appropriate.** Three tests cover the contract:
  - positive case (`get-weather_2` — exercising hyphen, digit, underscore) accepted
  - negative case at `ToolDefinition` level (the literal `Answer Tool` repro from #577)
  - negative case at `ChatCompletionRequest` level — proves the nested `tools=[{...}]` array dispatches through `ToolDefinition` validation end-to-end, which is the actual path the failing curl in #577 takes.

  Both negative tests pin the error message via `match="function.name"`, so the assertion is stable against pydantic's message-prefix formatting.

- **CI is green** on all nine matrix entries (lint, type-check, test-apple-silicon ×2, test-matrix ×4, tests).

**Minor follow-up (non-blocking)**

The error message currently only shows the regex pattern. Including the rejected value would make client-side debugging quicker, e.g.:

```python
raise ValueError(
    f"function.name {name!r} must match ^[A-Za-z0-9_-]{{1,64}}$"
)
```

Not worth a re-review round on its own; happy to fold it in if a follow-up touches this file.

Approving.

waybarrios (approved):
it's good!

PR #583, opened 2026-05-25
fix(qwen-tool-parser): extract mixed XML+function calls, strip truncated markup

janhilgard:
## Summary

Three related fixes for `<tool_call>` / `<function=` markup leaking into `content` against Qwen3.6 (and Qwen3.5) in non-streaming chat completions.

### 1. Parser — mixed-format extraction
The function-style pass was gated by `if not tool_calls:`. When the model emitted an XML call followed by a `<function=...>` call in the same response, only the XML call was extracted; the function-style markup was returned in `content`.

### 2. Parser — truncated tool-call markup
Qwen3.6 nests function-style calls inside `<tool_call>` wrappers:

```
<tool_call>
<function=convert_currency>
<parameter=amount>100</parameter>
...
</function>
</tool_call>
```

When generation hits `max_tokens` mid-call, the unclosed `<tool_call>\n<function=...` markup was left in `content`. Neither `XML_PATTERN` (requires `{json}`) nor `FUNCTION_PATTERN` (requires `</function>`) matches the truncated form.

A new `_strip_unclosed_markup()` finds the earliest unmatched `<tool_call>` / `<function=` / `[Calling tool:` marker at the tail and drops everything from there. Applied both on the success path (after extraction) and on the no-tool-call fallback. The no-tool-call path now always returns a string (empty if fully stripped) instead of `or None`, so callers can distinguish "parser processed this" from "parser declined".

### 3. Server fallback — preserve parser's cleaned content
`_parse_tool_calls_with_parser` previously fell straight through to the generic `parse_tool_calls` when the configured parser found no tools. The generic parser does not strip incomplete markup, so the chat handler received the raw model output and leaked the markup even though the configured parser had already cleaned it. The fallback now uses the generic parser only when it actually finds tool calls; otherwise it returns the configured parser's cleaned content.

## Test plan

- [x] 24/24 existing Qwen unit tests still pass.
- [x] 8 new regression tests in `TestQwenMixedFormatAndTruncation` covering mixed XML+function, nested function-in-tool_call, truncated second tool call, truncated `<tool_call>` only, truncated `<function=` only, truncated `[Calling tool:` only, complete-function-then-truncated-function, and the always-string `content` contract.
- [x] 156/156 tests pass across `tests/test_tool_parsers.py`, `tests/test_native_tool_format.py`, `tests/test_tool_choice_forced.py`, and `tests/test_tool_choice_none.py`.
- [x] Live stress test against Qwen3.6-35B-A3B (20 multi-tool requests, varying `max_tokens` 80–1500): 4/20 responses leaked truncated markup into `content` before the fix; 20/20 clean after the fix even when `finish_reason="length"` cut generation mid-tool-call.

### DIFF
--- a/tests/test_tool_parsers.py
+++ b/tests/test_tool_parsers.py
@@ -1295,6 +1295,160 @@ class TestQwenFunctionFormat:
         assert result.tool_calls[0]["name"] == "get_weather"
 
 
+class TestQwenMixedFormatAndTruncation:
+    """Regression tests for multi-format extraction and truncated output.
+
+    Bug 1: previously the function-style pass was guarded by ``if not tool_calls``,
+    so an XML tool call followed by a ``<function=...>`` tool call only yielded the
+    XML one — the function markup leaked into ``content``.
+
+    Bug 2: when generation hit ``max_tokens`` mid tool call (Qwen3.6 emits nested
+    ``<tool_call>\n<function=name>...</function>\n</tool_call>``), the truncated
+    second call's raw markup was returned in ``content``.
+    """
+
+    @pytest.fixture
+    def parser(self):
+        return QwenToolParser()
+
+    def test_xml_then_function_mixed(self, parser):
+        """XML call followed by function-style call — both should be extracted."""
+        text = (
+            '<tool_call>\n{"name": "get_weather", "arguments": {"city": "Prague"}}'
+            "\n</tool_call>\n"
+            "<function=convert_currency>\n"
+            "<parameter=amount>100</parameter>\n"
+            "<parameter=from_currency>EUR</parameter>\n"
+            "<parameter=to_currency>CZK</parameter>\n"
+            "</function>"
+        )
+        result = parser.extract_tool_calls(text)
+        assert result.tools_called
+        assert len(result.tool_calls) == 2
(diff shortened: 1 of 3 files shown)


Thump604 (approved):
Reviewed the parser/server fallback changes and ran the focused tool-parser verification locally on the PR checkout.

What I checked:

```bash
python -m pytest tests/test_tool_parsers.py tests/test_native_tool_format.py tests/test_tool_choice_forced.py tests/test_tool_choice_none.py -q
ruff check vllm_mlx/tool_parsers/qwen_tool_parser.py vllm_mlx/server.py tests/test_tool_parsers.py --select E,F,W --ignore E402,E501,E731,F811,F841
black --check vllm_mlx/tool_parsers/qwen_tool_parser.py vllm_mlx/server.py tests/test_tool_parsers.py
git diff --check upstream/main...HEAD
```

Results: 156 parser/tool tests passed; ruff, black check, and diff-check passed.

The patch is narrow and matches the reported failure shape: it extracts mixed XML/function Qwen tool calls, strips truncated tail markup, and preserves the configured parser's cleaned content when the generic fallback finds no calls. I do not see a merge blocker.

PR #585, opened 2026-05-28
fix(simple): expose request/token counters for /v1/status parity

janhilgard:
## Summary

`SimpleEngine.get_stats()` returned only `engine_type`, `model_name`, `uptime`, `is_mllm`, `loaded` — none of the running totals that `BatchedEngine` exposes through its `_mllm_scheduler`. As a result, `/v1/status` reported zeros for `total_requests_processed`, `total_prompt_tokens`, `total_completion_tokens`, and `generation_tps` on any SimpleEngine-backed service, breaking dashboards built on top of `/v1/status`.

This affects services that explicitly run without `--continuous-batching`, including the **Gemma 4 31B MTP path** (`--mllm-draft-model` + `--mllm-draft-kind mtp`), which requires SimpleEngine per the `cli.py` guard.

## Changes

- Add counters in `__init__`: `_total_requests_processed`, `_total_prompt_tokens`, `_total_completion_tokens`, `_num_running`, plus a bounded `deque _recent_completions` for rolling `generation_tps`.
- Add `_track_request_stream()`: a yield-through wrapper that records the final `prompt_tokens` / `completion_tokens` of an inner async generator on exit (including cancellation, via `try/finally`).
- Rename `stream_generate` / `stream_chat` bodies to `_stream_generate_impl` / `_stream_chat_impl`. Add thin public wrappers that route through `_track_request_stream` so every external invocation is tracked exactly once.
- Update the two internal fallback paths inside `_stream_chat_impl` to call `_stream_generate_impl` directly (avoids re-entrant double counting).
- Extend `get_stats()` to expose `running`, `num_running`, `num_waiting`, `num_requests_processed`, `total_prompt_tokens`, `total_completion_tokens`, and `batch_generator.generation_tps` (rolling over the last 20 completions). All keys match what `BatchedEngine` emits, so the server's `/v1/status` mapping in `server.py` works without changes.

## Test plan

Validated on Gemma 4 31B + MTP drafter (`google/gemma-4-31B-it-assistant`) with mlx-vlm 0.5.0 on Apple M3 Ultra:

- [x] Single request (`max_tokens=100`): `total_completion_tokens` increments from 0 → 102, `generation_tps` reports a realistic 26.7 tok/s under multi-server load.
- [x] Multiple sequential requests: counters accumulate correctly; `num_running` returns to 0 between requests.
- [x] `BatchedEngine` path is untouched — fields in `/v1/status` are still served from `_mllm_scheduler.get_stats()`.
- [x] Internal fallback in `_stream_chat_impl` calls `_stream_generate_impl` (not `self.stream_generate`) so re-entrant tracking does not double-count.
- [x] `black vllm_mlx/engine/simple.py` clean.

## Notes

`generation_tps` is computed across a rolling window of the last 20 completions, which matches the qualitative feel of `BatchedEngine`'s EWMA without bringing in an extra dependency. `prompt_tps` is intentionally left at 0.0 in this PR — SimpleEngine doesn't currently isolate prompt-processing time from generation time, and a fake value would mislead dashboards. Happy to follow up if the maintainers want a separate measurement there.

### DIFF
--- a/vllm_mlx/engine/simple.py
+++ b/vllm_mlx/engine/simple.py
@@ -7,14 +7,25 @@ performance when serving a single user at a time.
 """
 
 import asyncio
+import contextvars
 import hashlib
 import logging
 import os
 import threading
 import time
-from collections import OrderedDict
+import uuid
+from collections import OrderedDict, deque
 from collections.abc import AsyncIterator
 from typing import Any
+
+# Re-entrancy guard for SimpleEngine._track_request_stream so that
+# internal fallback paths inside _stream_chat_impl (which call back into
+# self.stream_generate) don't double-count a single external request.
+# contextvars propagates per-asyncio-task, so concurrent requests still
+# each get their own outermost tracking pass.
+_in_tracker: contextvars.ContextVar[bool] = contextvars.ContextVar(
+    "_simple_engine_in_tracker", default=False
+)
 
 import mlx.core as mx
 
@@ -160,6 +171,19 @@ class SimpleEngine(BaseEngine):
         self._mtp = mtp
         self._mtp_num_draft_tokens = mtp_num_draft_tokens
         self._prefill_step_size = prefill_step_size
+
+        # Request stats (parity with BatchedEngine for /v1/status monitoring).
+        # Without these, monitoring sees zero traffic for SimpleEngine-backed
+        # servers (e.g. Gemma 4 31B with --mllm-draft-model + MTP).
+        self._total_requests_processed: int = 0
+        self._total_prompt_tokens: int = 0
+        self._total_completion_tokens: int = 0
+        self._num_running: int = 0
+        # Rolling window of (completion_tokens, duration_s) for tps computation.
+        self._recent_completions: deque = deque(maxlen=20)
+        # Live per-request state, mirroring BatchedEngine's "requests" list
+        # in /v1/status (request_id, phase, ttft_s, tokens_per_second, ...).
+        self._active_requests: dict[str, dict[str, Any]] = {}
 
         # SpecPrefill config
         self._specprefill_enabled = specprefill_enabled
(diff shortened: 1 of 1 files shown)


Thump604 (changes requested):
Thanks for adding the SimpleEngine status parity path. I can reproduce the CI failure locally on the PR checkout with:

```bash
python -m pytest \
  tests/test_server.py::TestChatCompletionStreamingModeSwitching::test_nonstream_then_stream_chat_completion_keeps_stream_thread_valid \
  tests/test_server.py::TestChatCompletionStreamingModeSwitching::test_stream_then_nonstream_chat_completion_keeps_stream_thread_valid \
  tests/test_server.py::TestChatCompletionStreamingModeSwitching::test_nonstream_then_stream_parser_init_does_not_reintroduce_stream_error \
  -q
```

Observed result locally: the three tests fail with the same server log error as CI:

```text
Streaming error, ensuring terminal frame: <Token var=<ContextVar name='_simple_engine_in_tracker' ...>> was created in a different Context
```

The failing path is in `SimpleEngine._track_request_stream()`: it sets `_in_tracker` with `token = _in_tracker.set(True)` and resets that token in the async generator `finally`. Under the `/v1/chat/completions` streaming endpoint, the generator can be finalized from a different async context than the one that created the ContextVar token, so `ContextVar.reset(token)` raises during stream finalization. That reintroduces the exact class of terminal-frame streaming error these regression tests are meant to catch.

I think this needs a revision before merge. The minimal shape I would look for is: keep the request counter/status accounting, but avoid a ContextVar token whose reset crosses async-generator finalization contexts. For example, make the re-entrancy state request/engine-local rather than token-reset based, or otherwise ensure the set/reset happens entirely within the same context. The existing three tests above are a good acceptance gate for the fix.

janhilgard:
Thanks for the thorough repro and the pointer to the failing tests — that was an obvious miss on my part.

Pushed `3ac9ed2` (force-with-lease, rebased on current `main`). The fix is exactly what you suggested: keep the `_in_tracker` ContextVar as the re-entrancy detector, but replace `set(token)` / `reset(token)` with `set(True)` / `set(False)`. Set returns a fresh value in whatever context the generator is finalized in, instead of trying to roll back a Token whose origin context no longer exists.

I also reproduced the three `TestChatCompletionStreamingModeSwitching` failures locally and confirmed they pass with the updated branch.

Live-run validation on Qwen 3.6-27B + native MTP (mlx-vlm 0.5.0): streaming `/v1/chat/completions` now finishes cleanly (no `Token was created in a different Context` in logs), `/v1/status` shows `total_requests_processed`, `total_completion_tokens`, `generation_tps`, `requests[]`, and `memory_aware_cache` populated as expected, and `mtp.acceptance_rate` continues to report ~78% under greedy decoding.

Ready for another look when you have a minute.

Thump604 (approved):
Re-reviewed the updated branch after `3ac9ed2`.

The ContextVar-token issue from my previous review is addressed: `_track_request_stream()` now uses value set/unset rather than resetting a token across async generator finalization contexts.

Local verification on the PR checkout:

```bash
python -m pytest \
  tests/test_server.py::TestChatCompletionStreamingModeSwitching::test_nonstream_then_stream_chat_completion_keeps_stream_thread_valid \
  tests/test_server.py::TestChatCompletionStreamingModeSwitching::test_stream_then_nonstream_chat_completion_keeps_stream_thread_valid \
  tests/test_server.py::TestChatCompletionStreamingModeSwitching::test_nonstream_then_stream_parser_init_does_not_reintroduce_stream_error \
  -q
# 3 passed

python -m pytest tests/test_server.py -k 'status or StreamingModeSwitching' -q
# 6 passed

ruff check vllm_mlx/engine/simple.py tests/test_server.py
black --check vllm_mlx/engine/simple.py tests/test_server.py
git diff --check
# pass
```

The status-counter change is now narrow enough for this PR, and the streaming finalization regression I flagged is covered by the focused tests above.

PR #586, opened 2026-05-28
fix: update gemma4 attention patch for mlx-vlm 0.5.0 compatibility

mxl:
## Problem

`mlx-vlm 0.5.0` updated `Attention.__call__` in the Gemma 4 model to accept two new keyword arguments (`shared_kv`, `offset`) and return a tuple `(output, (keys, values), offset)` instead of just `mx.array`.

The existing `patch_gemma4_attention_for_batching` in `vllm_mlx/patches/gemma4_mllm.py` was written against the older API, so at runtime it raises:

```
TypeError: patch_gemma4_attention_for_batching.<locals>._patched_call() got an unexpected keyword argument 'shared_kv'
```

## Fix

- Add `shared_kv` and `offset` parameters to `_patched_call`
- When `shared_kv` is provided, use it directly (skip k/v projection) — matching mlx-vlm's KV-sharing logic for 2B/4B models
- Move the defensive `offset` snapshot inside the `else` branch (only needed when computing fresh k/v)
- Return `(output, (keys, values), offset)` to match the new mlx-vlm signature

### DIFF
--- a/vllm_mlx/patches/gemma4_mllm.py
+++ b/vllm_mlx/patches/gemma4_mllm.py
@@ -71,36 +71,41 @@ def patch_gemma4_attention_for_batching() -> bool:
         x: mx.array,
         mask: Optional[mx.array] = None,
         cache: Optional[Any] = None,
-    ) -> mx.array:
+        shared_kv: Optional[tuple] = None,
+        offset: Optional[Any] = None,
+    ) -> tuple:
         B, L, _ = x.shape
 
         queries = self.q_proj(x).reshape(B, L, self.n_heads, self.head_dim)
         queries = self.q_norm(queries)
 
-        # Snapshot offset BEFORE update_and_fetch can mutate it in-place.
-        # Preserves per-sequence mx.array offsets for correct batched RoPE.
-        offset = _snapshot_cache_offset(cache)
-
-        if self.is_kv_shared_layer and cache is not None:
-            state = cache.state
-            keys, values = state[0], state[1]
+        if shared_kv is not None:
+            keys, values = shared_kv
         else:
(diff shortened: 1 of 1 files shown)


Thump604 (changes requested):
Thanks for picking this up. The `mlx-vlm 0.5.0` Gemma 4 `Attention.__call__` signature break is real, but I would not merge this PR as-is because it duplicates the same patch surface as #564 and is currently the narrower/less-covered version.

Compared paths:

- #586 changes only `vllm_mlx/patches/gemma4_mllm.py` for `shared_kv`, `offset`, and the 3-tuple return.
- #564 covers the same `shared_kv` / `offset` / 3-tuple compatibility path, adds `tests/test_gemma4_mllm_patch.py`, and has already gone through review against the `mlx-vlm 0.5.0` Gemma 4 code path.

If maintainers prefer #586 as the branch to land, I think it needs the same missing closure items before merge:

1. focused tests for the new signature, 3-tuple return, and `shared_kv` pass-through;
2. either raise the declared `mlx-vlm` floor to `>=0.5.0` or add a loud signature guard, because the current dependency floor still permits 0.4.x where the stock Gemma 4 decoder expects the old single-array return;
3. an explicit decision against #564 so we do not merge two divergent edits to the same Gemma 4 monkey-patch.

Not claiming this patch is runtime-wrong; I did not run the model path from this PR. This is about merge safety: as it stands, #564 is the more complete reviewed path for the same defect, and #586 should either be consolidated with it or brought up to the same test/dependency standard.

Thump604:
Closing this as superseded by #564, which has already merged the mlx-vlm 0.5.0 Gemma 4 attention signature compatibility path with focused regression coverage.

Thanks for picking up the same breakage. The issue was real; we just do not need two divergent edits to the same Gemma 4 patch surface in the release queue.

PR #589, opened 2026-05-31
Require mlx-vlm 0.6.2 for Step 3.7 Flash support

Thump604:
## Summary
- Raise `mlx-vlm` floor to `0.6.2`, the latest patch release containing Step 3.7 Flash support plus follow-up fixes.
- Keep the `mlx-lm` floor at `0.31.3`, matching the current `mlx-vlm` runtime support floor used by this package set.
- Add/update dependency-floor regression coverage so future installs do not silently resolve to a Step 3.7-incompatible `mlx-vlm`.

## Evidence
- Step 3.7 Flash support landed in `Blaizzy/mlx-vlm#1245`, which is contained in `mlx-vlm` tags `v0.6.0`, `v0.6.1`, and `v0.6.2`.
- The related ByteLevel streaming detokenizer fix landed in `Blaizzy/mlx-vlm#1246`, also contained in `v0.6.0+`.
- `mlx-vlm v0.6.2` also includes relevant follow-up fixes after `v0.6.0`, including Gemma 4 long-text prefill, structured output with thinking enabled, Qwen3.5 vision model type acceptance, Gemma shared-KV loading, and APC single-request behavior.
- The previous lower floor `mlx-vlm>=0.4.3` can resolve to `0.4.x`, which lacks the Step 3.7 loader surface. The earlier `0.5.0` floor in this PR was also too low for the actual Step 3.7 release placement.
- Resolver dry-run from this branch selects `mlx-vlm-0.6.2` and `mlx-lm-0.31.3`.

## Tests
- `.venv/bin/python -m pytest tests/test_dependency_floors.py -q` — 2 passed
- `.venv/bin/python -m ruff check tests/test_dependency_floors.py` — pass
- `git diff --check` — pass
- `/opt/ai-runtime/venv-live/bin/python -m pip install --dry-run --ignore-installed -e .` — resolves `mlx-vlm-0.6.2` / `mlx-lm-0.31.3`

## Not claimed
- This does not certify Step 3.7 runtime quality or serving behavior.
- This does not start or load any model.
- This only prevents installs from resolving to dependency versions that cannot include the Step 3.7 loader and follow-up support surface.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,6 +82,7 @@ jobs:
             tests/test_request.py \
             tests/test_anthropic_models.py \
             tests/test_anthropic_adapter.py \
+            tests/test_dependency_floors.py \
             tests/test_harmony_parsers.py \
             tests/test_endpoint_model_policies.py \
             tests/test_gemma4_openai_format.py \
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,8 +29,8 @@ classifiers = [
 
 dependencies = [
     "mlx>=0.29.0",
-    "mlx-lm>=0.31.3",  # 0.31.3+ required by mlx-vlm 0.5.0
-    "mlx-vlm>=0.5.0",  # 0.5.0+ required for Step 3.7 Flash support
+    "mlx-lm>=0.31.3",  # 0.31.3+ required by current mlx-vlm runtime support
+    "mlx-vlm>=0.6.2",  # 0.6.2+ includes Step 3.7 Flash and follow-up fixes
     "transformers>=5.0.0",  # mlx-lm 0.30.5+ requires transformers 5.0 (rc3 bug fixed in stable)
     "tokenizers>=0.19.0",
     "huggingface-hub>=0.23.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,8 +29,8 @@ classifiers = [
 
 dependencies = [
     "mlx>=0.29.0",
-    "mlx-lm>=0.31.0",  # 0.31+ required for ArraysCache native batching (hybrid models)
-    "mlx-vlm>=0.4.3",  # 0.4.3+ required for Gemma 4 support
+    "mlx-lm>=0.31.3",  # 0.31.3+ required by mlx-vlm 0.5.0
+    "mlx-vlm>=0.5.0",  # 0.5.0+ required for Step 3.7 Flash support
     "transformers>=5.0.0",  # mlx-lm 0.30.5+ requires transformers 5.0 (rc3 bug fixed in stable)
     "tokenizers>=0.19.0",
     "huggingface-hub>=0.23.0",
--- a/tests/test_dependency_floors.py
+++ b/tests/test_dependency_floors.py
@@ -28,13 +28,13 @@ def _has_minimum(requirement: str, minimum: str) -> bool:
     return _version_tuple(version) >= _version_tuple(minimum)
 
 
-def test_mlx_vlm_floor_includes_step37_flash_support():
+def test_mlx_vlm_floor_includes_step37_flash_support_and_followups():
     dependencies = _project_dependencies()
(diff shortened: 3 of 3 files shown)


janhilgard (approved):
Reviewed the diff and traced it against actual CI behavior. Approving — the floor bump itself is well-justified and verifiable. Two substantive notes on the regression-test side that are worth folding into this PR (or a fast follow-up).

**Verified:**

- `mlx-vlm==0.6.2` and `mlx-lm==0.31.3` are both live on PyPI, and a fresh resolver does land on them with this floor.
- Blaizzy/mlx-vlm#1245 (Step 3.7 Flash loader) merged 2026-05-30 and Blaizzy/mlx-vlm#1246 (ByteLevel streaming detokenizer fix for Step 3.7) merged the same day — both are contained in `v0.6.0+` releases as the PR body claims.
- `0.6.2` also bundles the Qwen3.5 vision `model_type` acceptance fix, which I needed to monkey-patch locally during a recent Heretic-v2 conversion (the upstream model ships `vision_config.model_type = "qwen3_5_vision"` and `mlx-vlm 0.5.0` rejects it). The Gemma 4 long-text prefill follow-up and APC single-request fix are similarly meaningful for production deploys.
- Previous floor `mlx-vlm>=0.4.3` could legitimately resolve to `0.4.x` / `0.5.x`, neither of which has the Step 3.7 loader surface — so the bump is correct.
- Test logic correctly checks `>=`-pinned major versions; `_version_tuple` is fine for the simple `X.Y.Z` shape used by these deps.

**Non-blocking but worth addressing in this PR:**

1. **The new test isn't actually wired into CI.** `.github/workflows/ci.yml` invokes pytest with an explicit file list rather than discovery (e.g. `pytest tests/test_mcp_security.py tests/test_structured_output.py …` on the `test-matrix` jobs and a separate list on `test-apple-silicon`). `tests/test_dependency_floors.py` isn't in either list, so the "floor regression" it's meant to provide is silent — a future PR that lowers `mlx-vlm` back to `0.4.3` would not turn the test red. Adding `tests/test_dependency_floors.py` to the `test-matrix` step (it's a pure-Python no-MLX test, so it fits there) closes the loop.

2. **`import tomllib` is Python 3.11+ stdlib.** `pyproject.toml` keeps `requires-python = ">=3.10"` and the `test-matrix` job runs 3.10 too. The reason CI is green is exactly point (1): the test is never collected. If point (1) is fixed, the 3.10 job will fail on `ModuleNotFoundError: No module named 'tomllib'`. Easiest fix:

    ```python
    try:
        import tomllib  # Python 3.11+
    except ModuleNotFoundError:
        import tomli as tomllib  # 3.10 fallback
    ```

    `tomli` is already a transitive dep of recent pytest on 3.10 (visible in the CI install log: `Collecting tomli>=1 (from pytest)`), so no new dev-deps required. Alternatively, raise `requires-python = ">=3.11"` if you prefer to drop 3.10.

**Heads-up for downstream consumers (not blocking):** This is a noticeable floor jump (`0.4.3 → 0.6.2`). Anyone pinned in the middle (I'm on `0.5.0` locally with a customised path for Gemma 4 MTP assistant drafters) will need to revalidate. Probably fine for everyone — the cited follow-up fixes are mostly net-positive — but the version bump deserves a CHANGELOG line if you keep one.

LGTM otherwise.

Thump604:
Addressed Jan's two CI/test notes in f56c6be:

- added `tests/test_dependency_floors.py` to the explicit `test-matrix` pytest list, so the floor regression now runs in CI;
- added the Python 3.10 `tomli` fallback for `tomllib`.

Verification after the update:

- local focused dependency-floor test: `2 passed`;
- local explicit unit-test list under the Runtime venv: `1007 passed, 6 skipped, 17 deselected`;
- local Ruff target matching CI: pass;
- local Black check: pass;
- GitHub CI run 27063744128: all checks passed, including `test-matrix (3.10)`.

PR #594, opened 2026-06-06
fix(scheduler): normalize empty logits processor slots

### LINKED ISSUE #592: Engine loop crash: 'NoneType' object is not iterable at logits_processors[e] under continuous batching (pure-LLM path)
## Summary

Under `--continuous-batching` on the **pure-LLM path**, the engine loop crashes intermittently with `TypeError: 'NoneType' object is not iterable` when iterating `self.logits_processors[e]` inside `mlx_lm`'s `BatchGenerator._step`. The exception is **not** the `_is_stream_thread_error` case `engine_core._engine_loop` recovers from, so it is re-raised, logged as `Engine loop error`, and the **entire in-flight batch step is aborted** → affected requests return 500.

It is strongly correlated with **long generations** (the line immediately preceding the error is consistently a `disconnect_guard` poll at high elapsed, e.g. `elapsed=111.2s`) and with **long server residence** (many insert/remove slot cycles). On a Qwen3.6-27B-8bit deployment it accounts for ~**2.7%** of error-log lines (103,121 occurrences in one ~14h process lifetime) and correlates with ~**9.9%** HTTP 500 at our proxy.

## Traceback

```
ERROR:vllm_mlx.engine_core:Engine loop error: 'NoneType' object is not iterable
  File ".../vllm_mlx/engine_core.py", line 246, in _engine_loop
    output = await loop.run_in_executor(
  File ".../concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File ".../vllm_mlx/engine_core.py", line 174, in _step_on_worker
    output = self.scheduler.step()
  File ".../vllm_mlx/scheduler.py", line 2500, in step
  File ".../mlx_lm/generate.py", line 1855, in next
    return self._next()
  File ".../mlx_lm/generate.py", line 1775, in _next
    generation_responses = self._generation_batch.next()
  File ".../mlx_lm/generate.py", line 1415, in next
    tokens, logprobs = self._step()
  File ".../mlx_lm/generate.py", line 1346, in _step
    for processor in self.logits_processors[e]:
                     ~~~~~~~~~~~~~~~~~~~~~~^^^
TypeError: 'NoneType' object is not iterable
```

## Environment

| Component | Version |
|---|---|
| vllm-mlx | **0.3.0** (also affects **0.4.0rc1**, see below) |
| mlx-lm | 0.31.3 |
| mlx | >= 0.29 |
| Python | 3.12, macOS arm64 (M2 Ultra) |

**Serve flags:** `--continuous-batching --max-tokens 4096 --enable-auto-tool-choice --tool-call-parser qwen --default-chat-template-kwargs '{"enable_thinking": false}' --offline`
**Model:** `mlx-community/Qwen3.6-27B-8bit` (pure LLM, non-MLLM path).

### 0.4.0rc1 does not fix this

I diffed the 0.3.0 and 0.4.0rc1 source trees on the pure-LLM path:
- `engine_core.py`: **byte-identical**.
- `scheduler.py`: the only delta is one unrelated line (`list(request.output_token_ids)` → `request.output_token_ids`); the `logits_processors` slot management (`keep_idx` filtering, `extend` on merge, and the insert at ~L2054 `"logits_processors": [lp] if lp else [[]]`) is **identical**.
- Both pin `mlx-lm>=0.31.0` → resolve the same 0.31.3, where the iteration that throws lives.

So the crashing path is unchanged across 0.3.0 → 0.4.0rc1 for non-MLLM models.

## Root cause hypothesis

`scheduler.py` (~L2052) explicitly tries to avoid this — the insert passes `[lp] if lp else [[]]` with the comment *"Always pass logits_processors (even empty list) so that mlx_lm BatchGenerator never stores None per-sequence."* So the **caller already guarantees a list**.

That points to the per-slot bookkeeping **inside `mlx_lm`'s `BatchGenerator`**: after many insert/remove cycles under continuous batching (long residence + long generations), a slot's `self.logits_processors[e]` ends up `None` instead of `[]`. The unguarded iteration at `generate.py:1346` then raises.

## Suggested fix

Defensive guard at the iteration site (tolerate a `None` slot rather than abort the whole batch step):

```python
# mlx_lm/generate.py:1346  (_step)
for processor in (self.logits_processors[e] or []):
    ...
```

…and/or ensure slot release/realloc resets the entry to `[]`. The same unguarded pattern also exists in `vllm_mlx/scheduler.py:797` and is worth hardening symmetrically.

## Reproduction

Not trivial to trigger cold (a fresh process may run a long time before hitting it — it appears tied to slot reuse under sustained concurrent + long-gen traffic). Observed reliably in production under real concurrent load with tool-calling/constrained decoding enabled and generations in the ~100s range. Happy to run a targeted load script if you have a preferred harness.

## Related

- #155 — `mlx-lm >= 0.31.0` zip-unpacking crash (different symptom, same integration boundary).
- #546 — strict json_schema decode wedges on long Qwen generations (neighboring area: constrained decoding + long duration).

Thump604:
## Summary

Fixes a scheduler boundary case where stale `BatchGenerator` state can contain
`None` for a per-sequence `logits_processors` slot. New inserts already pass
empty lists for requests without processors; this patch normalizes existing
active or partial batch state before decode so an empty processor slot stays an
empty list instead of aborting the request.

## Scope

- Normalizes stale active and partial `BatchGenerator` `logits_processors`
  slots before `next()`.
- Applies the same normalization in the local MTP patched step before iterating
  processor slots.
- Adds a regression test proving stale `[None]` is normalized and does not abort
  the running request.

## Validation

- `.venv/bin/python -m pytest tests/test_batching.py::TestSchedulerBasic::test_step_normalizes_stale_none_logits_processor_slots -q`
- `.venv/bin/python -m pytest tests/test_batching.py tests/test_continuous_batching.py tests/test_memory_stability.py tests/test_mllm_continuous_batching.py -q`
- `.venv/bin/ruff check vllm_mlx/scheduler.py tests/test_batching.py`
- `git diff --check`

Full `pytest -q` was also run. It completed with the new batching regression
passing, but the suite still has unrelated existing failures in prefix-cache and
SSD-cache tests that use fake `mx`/SSD expectations.

Closes #592.

### DIFF
--- a/tests/test_batching.py
+++ b/tests/test_batching.py
@@ -694,6 +694,45 @@ class TestSchedulerBasic:
         assert scheduler.get_num_running() == 0
         assert not scheduler.has_requests()
 
+    def test_step_normalizes_stale_none_logits_processor_slots(
+        self, mock_model, mock_tokenizer
+    ):
+        """Scheduler should sanitize stale BatchGenerator per-slot None values."""
+
+        class FakeActiveBatch:
+            logits_processors = [None]
+
+        class FakeBatchGenerator:
+            def __init__(self):
+                self.active_batch = FakeActiveBatch()
+                self.called = False
+
+            def next(self):
+                self.called = True
+                assert self.active_batch.logits_processors == [[]]
+                return []
+
+        scheduler = Scheduler(
+            model=mock_model,
+            tokenizer=mock_tokenizer,
+        )
+        request = Request(
+            request_id="test-1",
+            prompt="Hello",
+            sampling_params=SamplingParams(max_tokens=2),
+        )
+        request.status = RequestStatus.RUNNING
+        scheduler.requests[request.request_id] = request
+        scheduler.running[request.request_id] = request
+        batch_generator = FakeBatchGenerator()
+        scheduler.batch_generator = batch_generator
+
+        output = scheduler.step()
+
+        assert batch_generator.called is True
+        assert output.finished_request_ids == set()
+        assert request.request_id in scheduler.running
+
 
 # Integration tests require actual MLX model
 @pytest.mark.integration
--- a/vllm_mlx/scheduler.py
+++ b/vllm_mlx/scheduler.py
@@ -42,6 +42,28 @@ CACHE_CORRUPTION_PATTERNS = [
 ]
 
 
+def _normalize_logits_processors(logits_processors):
+    """Normalize empty per-sequence processor slots to lists."""
+    if logits_processors is None:
+        return None
+    return [processors or [] for processors in logits_processors]
+
+
+def _sanitize_batch_generator_logits_processors(batch_generator) -> None:
+    """Sanitize stale BatchGenerator processor state before decode."""
+    active_batch = getattr(batch_generator, "active_batch", None)
+    if active_batch is not None and hasattr(active_batch, "logits_processors"):
+        active_batch.logits_processors = _normalize_logits_processors(
+            active_batch.logits_processors
+        )
+
+    partial = getattr(batch_generator, "_partial", None)
+    if isinstance(partial, dict) and "logits_processors" in partial:
+        partial["logits_processors"] = _normalize_logits_processors(
+            partial["logits_processors"]
+        )
+
+
 class SchedulingPolicy(Enum):
     """Scheduling policy for request ordering."""
 
@@ -786,6 +808,7 @@ def _install_mtp(
             logits = logits[:, -1, :]
 
         # --- Apply logits processors + sample primary ---
+        logits_processors = _normalize_logits_processors(logits_processors) or []
         if any(logits_processors):
             logger.debug(
                 f"[logits_proc] applying {sum(len(lp) for lp in logits_processors)} "
@@ -2497,6 +2520,7 @@ class Scheduler:
 
                 # Run generation step if we have running requests
                 if self.batch_generator is not None and self.running:
+                    _sanitize_batch_generator_logits_processors(self.batch_generator)
                     result = self.batch_generator.next()
                     output.has_work = True
 


janhilgard (approved):
Reviewed the diff against the actual production code paths. Approving — this is a real fix for a meaningful stability issue (~9.9% proxy 5xx in #592), well-scoped, and the sanitize call points are correct.

**Verified:**

- The `batch_generator._partial` and `batch_generator.active_batch` attributes the sanitize function reaches for are **not** on stock `mlx_lm.generate.BatchGenerator` — they're attached by the vllm-mlx MTP monkey-patch in `_apply_mtp_patch` (see `scheduler.py` ~L258 `batch_gen._partial = None` and ~L591 `self._partial = {"logits_processors": list(logits_processors), ...}` inside the patched `_process_prompts`). So the sanitize actually fires on the production-used BatchGenerator instances, and the defensive `getattr(..., None)` / `isinstance(partial, dict)` checks make it cleanly a no-op when those vehicles aren't present (non-MTP or future refactors).
- The `_mtp_step` change uses `_normalize_logits_processors(logits_processors) or []` — `[[]]` is truthy so the `or []` only fires when normalize returns `None`, which is the right semantics. Then `any(logits_processors)` further down treats `[[]]` as no-op (correct).
- Test reproduces the failure mode (`[None]` slot at decode time) and asserts both (a) the sanitize ran (`active_batch.logits_processors == [[]]`) and (b) the request stayed running rather than aborting. This matches the production crash shape in #592 (`TypeError: 'NoneType' object is not iterable` at `for processor in self.logits_processors[e]`).
- Idempotent: calling `_normalize_logits_processors` twice yields the same list. The list-comprehension allocates a new list each step, which is O(N) per scheduler step and negligible against the forward pass.
- All 9 CI checks pass on `9f8a656`.

**Non-blocking notes:**

1. The root cause is upstream `mlx_lm.generate.BatchGenerator` at L1066: when no sequence has processors, it sets `self.logits_processors = [None] * len(self.uids)`. After mixed-population slot recycling that `None` survives into `self.logits_processors[e]` and crashes at L1346. This PR is the right defensive layer for the production path; worth a follow-up upstream issue/PR against `ml-explore/mlx-lm` so the symptom can eventually be removed at the source.
2. The unit test mocks `active_batch`/`_partial` explicitly. If a future refactor of the MTP patch renames either attribute (e.g. `_partial` → `_pending_prefill`), the sanitize silently stops firing without a test failure. Not a blocker — getattr-based discovery is the right pattern here — but a small integration test that drives a real patched BatchGenerator through one decode cycle would catch attribute drift.
3. `_sanitize_batch_generator_logits_processors` ignores `batch_generator.logits_processors` on the BatchGenerator itself. On the stock mlx_lm path that's where the crash actually iterates from, so a third sanitize point there (`logits_processors = getattr(batch_generator, "logits_processors", None)`) would close the gap if you ever run continuous batching without the MTP patch installed. Same defensive shape, no risk of double-application since `_normalize_logits_processors([None])` → `[[]]` and a second call → `[[]]` unchanged.

None of those should hold this up.

waybarrios:
@Thump604 Checked your code - all is right!

Issue #590, opened 2026-06-02
Gemma 4 (gemma4_text): TextModel dispatch, logit_bias API, and MLLM fallback stalls


gprot42:
# Gemma 4 (MLLM / `gemma4_text`): fast TextModel path, stalls, and API gaps

## Environment

- **vllm-mlx**: 0.3.0 (PyPI latest as of 2026-06-02)
- **mlx-vlm**: ≥ 0.5.0 recommended (RoPE offset + thread-local MLX streams)
- **mlx-lm**: ≥ 0.31.0 (`mlx_lm.models.gemma4_text` exists)
- **Model**: `mlx-community/gemma-4-26B-A4B-it-heretic-4bit` (and other Gemma 4 MLLM checkpoints with `text_config.model_type == "gemma4_text"`)
- **Client**: Kilo Code / OpenAI-compatible agents

Reproducer project (local patches + analysis): community usage of Gemma 4 Heretic on Apple Silicon via vllm-mlx.

---

## Summary

Serving Gemma 4 multimodal weights through vllm-mlx 0.3.0 has three interacting problems:

1. **`text_model_from_vlm.py` always imports Qwen 3.5** → Gemma 4 fast path fails → silent **MLLM fallback** → empty-delta loops and broken tool-call streaming after turn 2.
2. **No OpenAI `logit_bias` passthrough** → cannot ban leaked Harmony control tokens (`<|channel>`, etc.) at sampling time.
3. **Gemma 4 + continuous batching** needs a small attention mask trim when batched masks are longer than sliding-window keys (vllm-mlx `main` already has this in `patches/gemma4_mllm.py`; mlx-vlm ≥ 0.5.0 fixed the older BatchKVCache RoPE bug).

Items (1) and (2) are **not** fixed in the PyPI 0.3.0 wheel. Item (3) is partially addressed on vllm-mlx GitHub `main` and mlx-vlm ≥ 0.5.0.

---

### Current upstream (vllm-mlx 0.3.0)

`vllm_mlx/text_model_from_vlm.py` always does:

```python
from mlx_lm.models.qwen3_5 import TextModel, TextModelArgs
args = TextModelArgs.from_dict(text_config)
```

For Gemma 4, `text_config` uses `model_type: "gemma4_text"` with different field shapes. This raises **`ZeroDivisionError`** in `TextModelArgs.__post_init__` (head_dim derivation).

### Observable startup

```
[...]

PR #596, opened 2026-06-06
feat(api): support chat logit_bias

Thump604:
## Summary

Adds OpenAI-compatible `logit_bias` support for chat completions by converting
the request map into a normal request-local logits processor.

## Scope

- Adds `ChatCompletionRequest.logit_bias`.
- Coerces OpenAI token-id string keys to integer token ids before calling
  `mlx_lm.sample_utils.make_logits_processors(logit_bias=...)`.
- Appends the generated processor to existing request-local processors so it can
  coexist with structured-output constraints and other decode controls.
- Keeps this separate from the Gemma 4 TextModel dispatch fix in #595.

This is a partial fix for #590. It does not change Gemma attention/mask handling
or model dependency floors.

## Validation

- `.venv/bin/python -m pytest tests/test_api_models.py::TestChatCompletion::test_request_accepts_logit_bias -q`
- `.venv/bin/python -m pytest tests/test_server.py::TestPromptCanonicalization::test_prepare_chat_completion_forwards_logit_bias -q`
- `.venv/bin/python -m pytest tests/test_api_models.py tests/test_server.py -q`
- `.venv/bin/ruff check vllm_mlx/api/models.py vllm_mlx/server.py tests/test_api_models.py tests/test_server.py`
- `git diff --check`

### DIFF
--- a/tests/test_api_models.py
+++ b/tests/test_api_models.py
@@ -296,6 +296,15 @@ class TestChatCompletion:
         assert req.tools is not None
         assert req.timeout == 30.0
 
+    def test_request_accepts_logit_bias(self):
+        req = ChatCompletionRequest(
+            model="test-model",
+            messages=[Message(role="user", content="Hello")],
+            logit_bias={"123": -100.0, "456": 2.5},
+        )
+
+        assert req.logit_bias == {"123": -100.0, "456": 2.5}
+
     def test_request_rejects_tool_name_with_spaces(self):
         with pytest.raises(ValidationError, match="function.name"):
             ChatCompletionRequest(
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -209,6 +209,38 @@ class TestPromptCanonicalization:
         )
         assert request.messages[0].content == system_text
 
+    def test_prepare_chat_completion_forwards_logit_bias(self):
+        from unittest.mock import patch
+
+        from vllm_mlx.server import (
+            ChatCompletionRequest,
+            Message,
+            _prepare_chat_completion_invocation,
+        )
+
+        engine = SimpleNamespace(is_mllm=False, preserve_native_tool_format=False)
+        processor = object()
+        calls = []
+
+        def fake_make_logits_processors(**kwargs):
+            calls.append(kwargs)
+            return [processor]
+
+        request = ChatCompletionRequest(
+            model="test-model",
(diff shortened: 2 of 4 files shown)


janhilgard (approved):
Reviewed the diff end-to-end. Approving — the change is small, well-scoped, and the integration with the existing processor pipeline is correct.

**Verified:**

- `mlx_lm.sample_utils.make_logits_processors` signature on a current install confirms `logit_bias: Dict[int, float] | None` — the string→int coercion at the API boundary is the right move, and matches OpenAI's wire format where keys arrive as strings.
- Processor ordering in `_prepare_chat_completion_invocation` is now logit_bias → json (response_format) → thinking-wrapper, which is consistent with how each layer composes (thinking wraps json as inner delegate; logit_bias is independent and additive).
- The updated thinking-processor block at the tail correctly strips only the trailing `json_logits_processor` before appending `thinking_proc`, so the logit_bias processor is preserved when both response_format and thinking are active. The `is`-identity check is the right way to detect "this is the one thinking wraps."
- `_coerce_logit_bias` returns HTTP 400 with a descriptive message on bad token-id strings — good error surface for clients that send float-looking strings or non-numeric junk.
- Tests exercise both the request-model acceptance and the wiring through `_prepare_chat_completion_invocation`, including verifying that the keys are int-coerced when the processor factory is called. Local CI (lint, type-check, tests, apple-silicon × 2, matrix × 4) is green on `3800f59`.

**Non-blocking follow-ups (out of scope for this PR, mentioning for visibility):**

1. `CompletionRequest` doesn't grow a `logit_bias` field in this PR — the legacy `/v1/completions` route stays unaffected. Worth a follow-up if you want full OpenAI parity, but the chat path is by far the dominant one and the scope split is sensible.
2. OpenAI clamps `logit_bias` values to `[-100, 100]`. The PR forwards floats unchanged; in practice `mlx_lm` will happily apply larger magnitudes, which effectively force/forbid tokens. Probably fine for an internal API, but consider clamping or warning at the boundary if this is meant to be a drop-in OpenAI replacement.
3. There's no validation that integer token ids are in vocab. Out-of-range entries should be silently no-ops in `make_logits_processors`, but it's worth a quick sanity check against `len(tokenizer)` at request time if you want a clean 400 instead of a possibly confusing acceptance.

None of those should hold this up.

PR #597, opened 2026-06-06
fix(qwen3-xml): parse bare <function=> without <tool_call> wrapper

CBribiescas:
> Note: just back from vacation — should be responsive on review feedback or follow-ups.

## Summary

Qwen3-Coder under load (large system prompt + many tools) sometimes drops the outer
`<tool_call></tool_call>` wrapper and emits only the inner
`<function=NAME><parameter=...>...</function>` block. The current parser is
wrapper-gated and leaks these tool calls into the response as raw text. Local
Claude Code sessions hitting Qwen3-Coder-30B-A3B-Instruct show this as
`<function=Agent>` printed as prose followed by hung 0-token subagents.

Same bug class as upstream vLLM PR #26345 ("[Bugfix]fix Qwen3 xml tool parser",
merged 2025-10-15), which was never backported here. This PR is broader: it
also covers the prose-misclassification and multi-sequential-bare-call cases
that #26345's eager-commit approach would regress.

Three coordinated changes in `StreamingXMLToolCallParser`:

1. **Buffer-wait for any partial tool tag.** `_looks_like_partial_tool_open` extends
   the wait-for-more-data branch (and the `tag_end < tag_end2` text-fragment branch)
   to cover `<function=`, `<parameter=`, and the closing variants — not just
   `<tool_call>`. Without this, partial `<function=Ag` chunks either leak as text
   or crash expat with `not well-formed (invalid token)`.

2. **Deferred commitment on bare `<function=Name>`.** When auto-opening a
   `<tool_call>` because the model dropped the wrapper, the function-name delta
   is held in `_pending_implicit_delta` until `<parameter=` opens (or `</function>`
   closes a parameterless call). If non-whitespace character data arrives outside
   any parameter, the `<function=...>` was prose — roll back via
   `_abandon_pending_implicit_tool_call` so no tool_call delta is ever emitted.
   This is what keeps the existing `test_streaming_plain_text_is_not_misclassified_as_tool_call`
   regression test green.

3. **Implicit-wrapper close on `</function>`.** When the surrounding `<tool_call>`
   was synthesised, also close it on `</function>` so the next bare `<function=...>`
   opens a fresh tool_call with an independent id/index instead of merging into
   the previous call (which would emit `}}` and `name=None` for the second call).

## Type of change

- Bug fix

## Surface touched

- MCP / tool calling

## Test plan

\`\`\`bash
pytest tests/test_qwen3_xml_parser.py \\
       tests/test_qwen3_xml_streaming_chunks.py \\
       tests/test_qwen3_xml_registration.py -q
\`\`\`

Two new parametrized tests added:

- `test_streaming_bare_function_without_tool_call_wrapper[1/5/7/24/256]`
- `test_streaming_two_sequential_bare_functions[1/5/7/24/256]`

## Test results

\`\`\`
tests/test_qwen3_xml_parser.py ........... (11 passed)
tests/test_qwen3_xml_streaming_chunks.py .............................. (32 passed, +10 new)
tests/test_qwen3_xml_registration.py .... (4 passed)
============================== 47 passed in 1.11s ==============================
\`\`\`

Real-session repro (Claude Code → ccr → LiteLLM → vllm-mlx Qwen3-Coder-30B-A3B-Instruct-MLX-4bit):

- Before: `<function=Agent>` printed as prose, 41-min/0-token runaway subagent.
- After: structured tool_calls emitted, Agent executes normally. (Server restart required to pick up the parser change.)

## Performance impact

N/A, parser-only change on the tool-call hot path. The added work is one
prefix-list iteration per partial-buffer resolve and one boolean check per
`_char_data` callback when an implicit wrapper is open. No allocations on
the common (wrapped) path.

## Output parity

Wrapped calls (`<tool_call>...<function=...>...</tool_call>`) produce
byte-identical streamed output before and after this patch — verified by the
unchanged-and-still-green existing test suite. Bare-function calls now produce
output where previously the parser produced unparseable text.

## Risk notes

- Parser-only; no scheduler / KV cache / sampler interaction.
- Three new internal state fields (`implicit_tool_call_wrapper`,
  `_pending_implicit_delta`, plus the helper methods); all reset by
  `reset_streaming_state`.
- The deferred-commitment design means a bare `<function=Name>` followed only
  by whitespace and then truncated (no further data) will NOT emit a tool_call.
  This matches the existing truncation contract from
  `test_streaming_truncated_at_end_of_parameter_recovers_tool_call`
  ("If the name was emitted, that's enough... no spurious extra calls").
- Upstream vLLM PR #26345 fixes the same bug class with a narrower (eager-commit)
  approach. This PR is a strict superset: handles prose disambiguation and
  multi-sequential-bare cases that #26345 would regress on.

### DIFF
--- a/tests/test_qwen3_xml_streaming_chunks.py
+++ b/tests/test_qwen3_xml_streaming_chunks.py
@@ -370,3 +370,99 @@ def test_object_parameter_with_single_quotes_split_mid_literal():
         if v["name"]
     ]
     assert streamed_calls == expected
+
+
+# ---------------------------------------------------------------------------
+# Bare <function=...> (no <tool_call> wrapper)
+#
+# Qwen3-Coder under heavy load (large system prompt + many tools) sometimes
+# drops the outer `<tool_call></tool_call>` wrapper and emits only the inner
+# `<function=NAME><parameter=...>...</function>` block. The streaming parser
+# must still recognise these as tool calls without misclassifying prose that
+# merely contains `<function=foo>` as text (covered by
+# `test_streaming_plain_text_is_not_misclassified_as_tool_call` above).
+# ---------------------------------------------------------------------------
+
+
+_BARE_SINGLE = (
+    "Thinking.\n<function=Agent>\n"
+    "<parameter=description>\nfind bug\n</parameter>\n"
+    "<parameter=prompt>\nlook at code\n</parameter>\n"
+    "<parameter=subagent_type>\nExplore\n</parameter>\n"
+    "</function>"
+)
+
+_BARE_TWO = (
+    "Step 1.\n<function=Read>\n<parameter=file_path>\n/a.py\n</parameter>\n</function>\n"
+    "Step 2.\n<function=Agent>\n<parameter=description>\nfind bug\n</parameter>\n"
+    "<parameter=prompt>\nlook\n</parameter>\n<parameter=subagent_type>\nExplore\n</parameter>\n"
+    "</function>"
+)
+
+
+@pytest.mark.parametrize("chunk_size", [1, 5, 7, 24, 256])
+def test_streaming_bare_function_without_tool_call_wrapper(chunk_size):
+    """A bare <function=...> block must stream into a structured tool call.
+
+    Pins down the contract that the model can drop the outer <tool_call>
+    wrapper (a documented Qwen3-Coder failure mode under heavy CC load)
+    and the streaming parser still emits a correct tool call with
+    name + arguments, regardless of how the input is chunked.
+    """
+    parser = Qwen3XMLToolParser(None)
+    expected = _expected_calls(_BARE_SINGLE, parser)
+    assert expected == [
+        {
+            "name": "Agent",
+            "arguments": {
+                "description": "find bug",
+                "prompt": "look at code",
+                "subagent_type": "Explore",
+            },
+        }
+    ]
+    parser = Qwen3XMLToolParser(None)
+    streamed = _stream(parser, _fixed_chunks(_BARE_SINGLE, chunk_size))
+    streamed_calls = [
+        {"name": tc["name"], "arguments": json.loads(tc["arguments"])}
+        for tc in streamed["tool_calls"]
+        if tc["name"]
+    ]
+    assert streamed_calls == expected
+    # No leaked tag fragments in user-visible content.
+    for fragment in ("<function=", "</function>", "<parameter=", "</parameter>"):
+        assert fragment not in streamed["content"], (
+            f"leaked tag fragment {fragment!r} in streamed content"
+        )
+
+
+@pytest.mark.parametrize("chunk_size", [1, 5, 7, 24, 256])
+def test_streaming_two_sequential_bare_functions(chunk_size):
+    """Two bare <function=...> blocks back-to-back must become two distinct
+    tool calls with independent indexes and ids.
+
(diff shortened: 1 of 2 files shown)


Thump604 (changes requested):
Thanks for the detailed patch and tests. I think this is targeting a real parser gap, but I am requesting changes for one streaming regression in the prose rollback path.

The new deferred implicit-call path prevents a false tool call, but it drops user-visible prose when `<function=...>` turns out not to be a tool call. Current test `test_streaming_plain_text_is_not_misclassified_as_tool_call` only checks that no tool calls are emitted; it does not assert that streamed content is preserved.

Local repro against this PR head:

```bash
PYTHONPATH=/tmp/vllm-mlx-pr597 /Users/David/code/vllm-mlx/.venv/bin/python - <<'PY'
from tests.test_qwen3_xml_streaming_chunks import _fixed_chunks
from vllm_mlx.tool_parsers.qwen3_xml_tool_parser import Qwen3XMLToolParser

text = "I considered using <function=foo> but decided against it. The answer is 42."
for size in [1, 2, 4, 8, 256]:
    parser = Qwen3XMLToolParser(None)
    accumulated = ""
    content = []
    calls = []
    for piece in _fixed_chunks(text, size):
        prev = accumulated
        accumulated += piece
        delta = parser.extract_tool_calls_streaming(prev, accumulated, piece)
        if not delta:
            continue
        if delta.get("content"):
            content.append(delta["content"])
        if delta.get("tool_calls"):
            calls.extend(delta["tool_calls"])
    joined = "".join(content)
    print(size, "calls", calls, "content_equal", joined == text, "content", repr(joined))
PY
```

Observed:

```text
1 calls [] content_equal False content 'I considered using ut decided against it. The answer is 42.'
2 calls [] content_equal False content 'I considered using t decided against it. The answer is 42.'
4 calls [] content_equal False content 'I considered using t decided against it. The answer is 42.'
8 calls [] content_equal False content 'I considered using cided against it. The answer is 42.'
256 calls [] content_equal False content 'I considered using '
```

So the parser is not emitting a false tool call, but it is losing normal assistant text. That should be covered by strengthening the existing prose test to assert streamed content equals the original text, then fixing `_abandon_pending_implicit_tool_call` / `_char_data` so the abandoned `<function=...>` fragment and following character data are emitted as content instead of swallowed.

Separately, CI lint is failing because Black would reformat `tests/test_qwen3_xml_streaming_chunks.py`; the rest of CI is green.

Focused checks I ran locally:

```bash
PYTHONPATH=/tmp/vllm-mlx-pr597 /Users/David/code/vllm-mlx/.venv/bin/python -m pytest \
  tests/test_qwen3_xml_parser.py \
  tests/test_qwen3_xml_streaming_chunks.py \
  tests/test_qwen3_xml_registration.py -q

/Users/David/code/vllm-mlx/.venv/bin/black --check \
  vllm_mlx/tool_parsers/qwen3_xml_tool_parser.py \
  tests/test_qwen3_xml_streaming_chunks.py
```

Result: focused tests passed (`47 passed`), Black reproduced the same formatting failure.

CBribiescas:

Thump604 (approved):
Thanks for addressing the regression. I re-reviewed the final head after `03164f2`: the abandoned `<function=>` prose path now restores the raw fragment and buffered text instead of swallowing user-visible content, and the added chunk-size coverage directly guards the issue I flagged. CI is green on the updated head. Approving.

PR #598, opened 2026-06-06
fix(step): add Step3p5 parser support

Thump604:
## Summary

Adds Step3p5 parser support for Step 3.7 Flash style serving:

- registers a `step3p5` reasoning parser for `<think>...</think>` and prompt-seeded `</think>` output
- registers a `step3p5` / `step` tool parser for Step XML function calls
- lets the CLI accept registered tool parsers from the parser registry instead of a hardcoded list
- forwards parser stop tokens through the MLLM streaming path so `</tool_call>` can stop generation

## Local repro / observed behavior

In local Runtime testing, Step 3.7 Flash loaded and direct text worked, but tool-bearing serving was not clean until Step-specific parser support was present:

- `--reasoning-parser step3p5` was needed to separate Step reasoning from final visible content.
- `--tool-call-parser step3p5` was needed for Step XML function calls such as:

```xml
<tool_call>
<function=get_weather>
<parameter=city>Austin</parameter>
<parameter=units>metric</parameter>
</function>
</tool_call>
```

Without the parser registration, the CLI did not accept `step3p5` as a tool parser. In the MLLM streaming path, parser stop tokens also did not reach `mlx_vlm.stream_generate`, so parser-declared stop tokens could not reliably halt generation.

Local Runtime evidence used for this patch shape:

- direct text smoke: `/opt/ai-runtime/run/step37-coding-readiness/20260606T184019Z-step3p5-direct-smoke/summary.json`
- forced tool smoke: `/opt/ai-runtime/run/step37-coding-readiness/20260606T192802Z-step3p5-tool-smoke-fallback-deduped/summary.json`

Compared against current `waybarrios/vllm-mlx` main at `015e080`:

- `vllm_mlx/reasoning/__init__.py` registry
- `vllm_mlx/tool_parsers/__init__.py` registry
- `vllm_mlx/tool_parsers/qwen_tool_parser.py` XML/function parser behavior
- `vllm_mlx/cli.py` parser choice handling
- `vllm_mlx/models/mllm.py` streaming generation kwargs
- `vllm_mlx/server.py` parser stop-token merge path

The Step tool parser intentionally reuses the existing XML/function parsing shape but registers it under Step-specific names so Step contracts do not rely on Qwen parser naming.

## Expected behavior

Launching with Step parser flags should be possible:

```bash
vllm-mlx serve <MODEL_PATH> \
  --mllm \
  --reasoning-parser step3p5 \
  --enable-auto-tool-choice \
  --tool-call-parser step3p5
```

Expected serving behavior:

- reasoning is separated from visible assistant content
- Step XML tool calls parse into OpenAI-compatible `tool_calls`
- consecutive duplicate Step tool calls are collapsed before returning response tool calls
- parser stop token `</tool_call>` reaches the MLLM streaming generator

## Minimal patch shape

- Add `Step3p5ReasoningParser`.
- Add `Step3p5ToolParser` as a Step-named wrapper around the existing XML/function parser shape.
- Switch CLI tool parser choices to the parser registry.
- Forward MLLM streaming `stop` kwargs and add a small stop-aware streaming helper.
- Add focused parser, CLI, MLLM stop, and server parser tests.

## What this does not claim

- Does not enable Step as a default or resident model.
- Does not claim tool-bearing coding-agent production readiness for every client.
- Does not change sampling, model loading, quantization, or routing defaults.
- Does not enable structured JSON / `response_format` for Step.

## Verification

```bash
uv run --extra dev python -m pytest tests/test_step3p5_reasoning_parser.py tests/test_step3p5_tool_parser.py tests/test_step3p5_server_tool_parser.py tests/test_cli.py::test_serve_parser_accepts_registered_step3p5_tool_parser tests/test_mllm.py::TestMLLMHelperFunctions::test_stream_chat_passes_stop_and_stops_on_stop_sequence -q
```

Result: `14 passed`.

```bash
uv run --extra dev python -m pytest tests/test_reasoning_parser.py tests/test_tool_parsers.py tests/test_qwen3_xml_registration.py tests/test_tool_choice_none.py tests/test_step3p5_reasoning_parser.py tests/test_step3p5_tool_parser.py tests/test_step3p5_server_tool_parser.py tests/test_cli.py -q
```

Result: `254 passed`.

```bash
uv run --extra dev ruff check vllm_mlx/reasoning/step3p5_parser.py vllm_mlx/tool_parsers/step3p5_tool_parser.py vllm_mlx/reasoning/__init__.py vllm_mlx/tool_parsers/__init__.py vllm_mlx/cli.py vllm_mlx/models/mllm.py tests/test_step3p5_reasoning_parser.py tests/test_step3p5_tool_parser.py tests/test_step3p5_server_tool_parser.py tests/test_cli.py tests/test_mllm.py
```

Result: `All checks passed`.

```bash
git diff --check
```

Result: clean.


janhilgard (approved):
Approving. Well-bounded feature PR with infrastructure improvements that benefit the rest of the parser ecosystem, not just Step 3.7 Flash. Tests cover the surface honestly, and CI is clean.

**Verified against the actual parser registry on a fresh local install:**

- `ToolParserManager.list_registered()` returns **33 names** today (auto, deepseek, deepseek_r1, deepseek_v3, glm4, glm47, gpt-oss, granite, granite3, gemma4, harmony, hermes, kimi, kimi_k2, llama, llama3, llama4, meetkai, minimax, minimax_m2, mistral, moonshot, nemotron, nemotron3, nous, qwen, qwen3, qwen3.5, qwen3_coder, qwen3_xml, xlam, functionary, generic). The previous hardcoded CLI list only exposed **17** of those, so several legitimate registered parsers (`deepseek_r1`, `glm4`, `granite3`, `kimi_k2`, `llama3`, `llama4`, `meetkai`, `minimax_m2`, `moonshot`, `nemotron3`, `nous`, `qwen3`, `qwen3.5`, `qwen3_xml`) were silently unreachable via `--tool-call-parser` even though they were registered. Switching to `_TOOL_PARSER_CHOICES = ToolParserManager.list_registered()` closes that gap as a side effect — the CLI now exposes exactly what the registry exposes. That alone is a meaningful tidy-up beyond Step 3.7.

- `extra_stop_tokens` integration is via the existing `get_parser_stop_tokens()` helper in `tool_parsers/__init__.py`. The Step parser declares `extra_stop_tokens = ["</tool_call>"]` at class level and that flows through `get_parser_stop_tokens()` into the user's stop list. The new MLLM `stream_chat` then pops `stop=` and forwards it via `_generation_kwargs_with_stop` into `mlx_vlm.stream_generate(..., stop=...)`, with `_until_stop_sequence` providing a substring-based guard layered on top. Belt-and-braces — even if `mlx_vlm.stream_generate` doesn't honor `stop=` natively today, the substring guard still terminates iteration when `</tool_call>` shows up. The test pins both: `seen["stop"] == ["</tool_call>"]` (kwarg landed) and `chunks == ["call", "</tool_call>", ""]` (post-guard terminates after the stop chunk yields).

- `Step3p5ToolParser(QwenToolParser)` with `@ToolParserManager.register_module(["step3p5", "step"])` reuses the existing XML/function parsing shape but registers under Step-specific aliases. That's the right move — coupling Step contracts to "qwen" by name has bitten communities before (operators conclude their parser preference depends on model lineage when it doesn't). The dedup override using `json.dumps(args, sort_keys=True)` for signature comparison is stable under argument-key reordering, which matters when the LLM emits the same logical call twice with the same fields in different orders. The "preserves distinct consecutive calls" test (`Austin` then `Dallas` survive) is the right counterweight to the "collapses repeated identical" test.

- `Step3p5ReasoningParser` handles the prompt-seeded `</think>` case explicitly. The Step chat template seeds `<think>` into the prompt, so the model only emits the closing tag, and the base `BaseThinkingReasoningParser` would otherwise treat the entire output as content. The `.removesuffix("\n").strip()` cleanup on both the reasoning and content sides matches what users actually see in tokenizer output. `_newline_after_end_delta` correctly distinguishes "just the newline after `</think>`" from real content delta — without that, the first byte of post-thinking content tends to be a stranded `\n` that some clients render as a blank message.

**Code-review notes:**

1. `_until_stop_sequence` yields the chunk that *contains* the stop sequence (i.e. the `</tool_call>` chunk reaches the client) before breaking. That matches typical "include the stop marker, do not include anything after it" semantics and matches the test's expected behavior. Worth a one-line comment to make that explicit so a future contributor doesn't try to "fix" it by stripping the stop chunk before yielding.

2. `extra_stop_tokens = ["</tool_call>"]` is a class-level mutable list. Subclasses or instances could mutate it in-place without realizing they're affecting the class state. Pythonic nit: `tuple(["</tool_call>"])` would be safer (the only consumer is `get_parser_stop_tokens()` which iterates, never appends, so tuple works fine). Not blocking.

3. The `monkeypatch.setitem(sys.modules, "mlx_vlm", fake_mlx_vlm)` setup in `test_stream_chat_passes_stop_and_stops_on_stop_sequence` patches four sibling modules. Functionally correct, but if the import path in `stream_chat` ever migrates (e.g. someone adds `from mlx_vlm.utils import X`), the fake registry will fail in a noisy way rather than silently. That's fine — noisy failure is what we want from test infrastructure — but worth knowing the test is tightly coupled to the current import surface.

4. Step dedup is *consecutive only*. `A A B A` becomes `A B A`, not `A B`. The PR body claims "consecutive duplicate Step tool calls are collapsed" so this matches intent. Some downstream clients might want full de-duplication (which would treat the second `A` as identical to the first), but that's a contract decision and arguably more aggressive than warranted. Keeping it consecutive-only is the safer default.

5. The CLI-choices-from-registry switch has a secondary effect worth a CHANGELOG note: `--tool-call-parser` argparse error messages will now list more options than before. If anyone is grepping error output in CI fixtures for a fixed parser list, those tests will need to be loosened. Low-risk surface, but mention-worthy.

6. `_chunk_text` returns `chunk.text if hasattr(chunk, "text") else str(chunk)`. For `mlx_vlm` `GenerationResult` chunks this is fine, but `str(chunk)` on an arbitrary object will produce something like `"<GenerationResult object at 0x...>"` which would never match a stop sequence anyway. Safe degradation, just won't help with stopping in a fake-chunk scenario.

7. All 9 CI checks green on `569b1c98`. The author also ran the full focused test suite (`254 passed`) locally per the PR body, which is a thorough sweep.

**Production relevance for me (not a Step 3.7 user yet):**

- The CLI registry switch and the MLLM stop-sequence forwarding are infrastructure improvements that benefit my Gemma 4 / Qwen 3.5 deploys too, since `extra_stop_tokens` is the same surface Gemma 4 uses for `<|tool_response>`. Reading the existing `get_parser_stop_tokens()` docstring it's clear this is exactly the contract the helper was designed for; #598 is the first PR to actually plug it into the MLLM streaming path. That fixes a long-standing minor: previously parser-declared stops only reached the LLM (text) path, not the MLLM path.

LGTM — this should ship.

Thump604:
Rebased onto v0.4.0 current main (`0dd1157`) and resolved the two additive conflicts by preserving both the new upstream Mistral registration/specprefill assertion and the Step parser additions. Focused parser/CLI/MLLM verification passes: 44 passed, 6 deselected; Ruff and diff checks pass. The outbound strict slop scan also caught and prompted a small split of the Step streaming normalizer; the rescanned new source is clean.

janhilgard:
Rebased onto current `main` (`94c008b`). `473e8ea` → `d372d23`, now `MERGEABLE/CLEAN`.

## The conflict

One hunk, in `vllm_mlx/reasoning/__init__.py::_register_builtin_parsers`. #644 landed `poolside_v1` in the same block where this PR adds `step3p5`:

```
<<<<<<< HEAD
    register_parser("poolside_v1", PoolsideV1ReasoningParser)
=======
    register_parser("step3p5", Step3p5ReasoningParser)
>>>>>>>
```

Independent additions to the same registry — kept both. The import lines above had already auto-merged cleanly, so only the two `register_parser` calls needed a decision. `vllm_mlx/tool_parsers/__init__.py` auto-merged with both `PoolsideV1ToolParser` and `Step3p5ToolParser` intact; I verified rather than assumed.

## Verification

The risk with a registry conflict is a resolution that compiles while silently dropping one family, so I checked the runtime result directly:

```
reasoning parsers: 9
  ✓ poolside_v1  → PoolsideV1ReasoningParser
  ✓ step3p5      → Step3p5ReasoningParser
tool parsers: 36
  ✓ poolside_v1  → PoolsideV1ToolParser
  ✓ step3p5      → Step3p5ToolParser
```

Both families resolve through the real registries after the rebase.

Tests on the rebased branch:

- `test_step3p5_reasoning_parser.py`, `test_step3p5_tool_parser.py`, `test_step3p5_server_tool_parser.py`, `test_poolside_v1_parsers.py`, `test_cli.py` → **26 passed**
- `test_mllm.py` + `test_cli.py` (this PR touches `models/mllm.py` and `cli.py`) → **32 passed, 6 deselected**
- `black --check vllm_mlx/ tests/` → clean, 222 files

Original authorship preserved; `Co-Authored-By` added for the rebase.

## Unrelated but worth knowing

While going through the assigned queue I found that **#653** merged cleanly onto post-#644 `main` while being semantically broken — its finalizer was still handed the module-level `_reasoning_parser` after #644 moved streaming to request-local instances. Clean merge, green CI (from before #644), silently non-functional. Fixed there.

This PR is not affected: the conflict was loud, and the parts that auto-merged are additive registry entries with no shared mutable state. Different failure mode entirely — just flagging it since anything with pre-#644 CI touching the parser layer deserves a second look.

CI is re-running on the rebase.

Thump604:
@waybarrios Jan rebased this onto current main, retained both Poolside and Step registry entries, verified both registrations at runtime, and all CI checks are green. This is ready for merge review.

waybarrios:
Just updated the branch and it seems ready to go .

PR #600, opened 2026-06-06
Honor chat_template_kwargs (enable_thinking) on the --mllm path

eejd:
## Problem
When a model is served with `--mllm`, `MLLMModel.chat()` pops only `enable_thinking` (default `True`)
and `tools` from kwargs and never reads `chat_template_kwargs`. So template kwargs passed via the
OpenAI request / `--default-chat-template-kwargs` (e.g. `{"enable_thinking": false}`) are silently
dropped — reasoning can't be disabled on the `--mllm` path and thinking leaks into the response. The
text path (`models/llm.py`) already handles this correctly via `template_kwargs.update(chat_template_kwargs)`.

## Change
`MLLMModel.chat()` (both the non-streaming and streaming generators) now pops `chat_template_kwargs`,
lets a nested `enable_thinking` override the default, and forwards the remaining keys into
`get_chat_template()` with a `TypeError`-safe retry that drops unsupported keys (mirrors `llm.py`).

## Validation
Serving `Qwen3.5-4B-MLX-4bit` with `--mllm --reasoning-parser qwen3 --default-chat-template-kwargs
'{"enable_thinking": false}'` now suppresses thinking. On an 11-test tool-use + quality battery the
score rose **0.67 → 0.92** (the three non-tool quality tests went 0.00 → 1.00) with tool-calling
unchanged; a bare control without the suppression flags stayed 0.67, confirming it's this fix.
Gemma-4-e4b similarly went 0.48 → 0.93.

### DIFF
--- a/vllm_mlx/models/mllm.py
+++ b/vllm_mlx/models/mllm.py
@@ -1895,6 +1895,14 @@ class MLXMultimodalLM:
         tools = kwargs.pop("tools", None)
         use_cache = kwargs.pop("use_cache", True)
         enable_thinking = kwargs.pop("enable_thinking", True)
+        # Honor chat_template_kwargs on the MLLM path (parity with the text path in
+        # llm.py). enable_thinking is commonly nested here via
+        # --default-chat-template-kwargs {"enable_thinking": false}; it was previously
+        # dropped for MLLM models, so reasoning could not be disabled and leaked into
+        # the response.
+        chat_template_kwargs = kwargs.pop("chat_template_kwargs", None) or {}
+        if "enable_thinking" in chat_template_kwargs:
+            enable_thinking = chat_template_kwargs.pop("enable_thinking")
 
         # Collect video and audio inputs from messages
         _msg_video_inputs = self._collect_video_inputs(messages)
@@ -1958,8 +1966,22 @@ class MLXMultimodalLM:
         template_extra_kwargs = {}
         if tools:
             template_extra_kwargs["tools"] = tools
+        # Forward any remaining chat_template_kwargs (parity with the text path)
+        template_extra_kwargs.update(chat_template_kwargs)
 
         try:
+            formatted_prompt = get_chat_template(
+                self.processor,
+                chat_messages,
+                add_generation_prompt=True,
+                enable_thinking=enable_thinking,
+                **template_extra_kwargs,
+            )
+        except TypeError:
+            # The processor's chat template doesn't accept some forwarded
+            # chat_template_kwargs — drop them and retry (mirror llm.py).
+            for key in chat_template_kwargs:
+                template_extra_kwargs.pop(key, None)
             formatted_prompt = get_chat_template(
                 self.processor,
                 chat_messages,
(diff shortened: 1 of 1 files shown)


janhilgard (approved):
Independent reproduction + validation on a different MLLM checkpoint than the ones in your test matrix. Approving — fix lands cleanly and behaves exactly as described.

**Repro before the patch (Gemma 4 12B-IT, mlx-vlm 0.6.2, `gemma4_unified` arch, `--mllm --tool-call-parser gemma4 --reasoning-parser gemma4`):**

Request `chat_template_kwargs={"enable_thinking": false}` returned:
```
reasoning_content: "" (empty, parser saw nothing to extract)
content: "<|channel>thought\n*   Task: Say \"hello\".\n    *   Constraint: Exactly 5 words.\n\n    *   \"Hello, how are you today?\" (5 words)\n    ..."
```
i.e. the raw `<|channel>thought` markers and the full thinking block leaked into `content`. That matches your problem statement — `chat_template_kwargs` was silently dropped on the `--mllm` path, the template defaulted to `enable_thinking=True`, the model emitted thinking with markers, and because the chat template was never actually told to skip thinking, the reasoning parser had nothing aligned to extract from on the response side either.

**After applying #600 locally and restarting the same 12B server:**
```
reasoning_content: ""
content: "Hello, how are you today?"
markers in content: NONE
```
Thinking suppressed at the template level, response clean, no marker leak.

**No regression on the default path** (no `chat_template_kwargs`, thinking left on, prompt "What is 17*23?"):
```
reasoning_content: 1179 chars (the worked-out multiplication)
content: short final answer containing "391"
markers in content: NONE
```
So the `enable_thinking` default path still routes through the reasoning parser cleanly.

**Code-review notes (all positive):**

- Both `chat()` and `stream_chat()` get the same change — important since the streaming path is the one most clients actually hit.
- The `chat_template_kwargs.enable_thinking` extraction *before* the explicit `enable_thinking` kwarg means a caller can pass either form and they compose correctly; nested wins over implicit default, explicit kwarg wins over both (caller order: kwargs.pop runs first, then the chat_template_kwargs override). That matches what the text path in `llm.py` does.
- `template_extra_kwargs.update(chat_template_kwargs)` correctly forwards *all* remaining template kwargs (not just `enable_thinking`), so e.g. `add_generation_prompt` overrides or future model-specific template flags will flow through.
- The `TypeError`-safe retry that drops unrecognized kwargs is the right behavior — processors with stricter chat templates won't crash a request, they just degrade to ignoring the extra hints, mirroring `llm.py`.

**Non-blocking nit:** `chat_template_kwargs` is `pop()`-ed but the dict is mutated by `.pop("enable_thinking", ...)` and then re-used in `template_extra_kwargs.update(chat_template_kwargs)`. That's fine (we *want* `enable_thinking` consumed before the forward), but worth a one-line comment so the next reader doesn't wonder if the mutation is a bug. Not a blocker.

CI hasn't run yet on this branch (`no checks reported on 'fix/mllm-chat-template-kwargs'`), worth keeping an eye on when it lands.

LGTM.

PR #601, opened 2026-06-06
Guard --mllm against continuous batching (silent empty output)

eejd:
## Problem
Serving a multimodal model with both `--mllm` and `--continuous-batching` (BatchedEngine) returns
**empty output** for every request, silently — while `--mllm` alone (SimpleEngine) works. The batched
path pre-formats the prompt and hands a string to the MLLM scheduler, which re-processes it with a
mismatched processor/token-injection context.

## Change
Until the batched-MLLM path is fixed, fail loudly: raise a clear `ValueError` when `force_mllm` and
`use_batching` are combined, at both engine-construction entry points (`_build_engine` for the
registry/residency path and `load_model` for the single-model serve path), mirroring the existing
`"MLLM draft models are supported only by SimpleEngine"` guard.

## Validation
`serve <vlm> --mllm --continuous-batching` now errors with a clear message ("use SimpleEngine by
omitting --continuous-batching") instead of serving empty.

(The deeper fix — making the batched-MLLM path actually generate — is a separate, larger change; this
PR is the safe guard.)

### DIFF
--- a/vllm_mlx/server.py
+++ b/vllm_mlx/server.py
@@ -1168,6 +1168,11 @@ def _get_cache_dir() -> str:
 
 def _build_engine(spec: ModelSpec) -> BaseEngine:
     """Construct an engine instance from a model spec without starting it."""
+    if spec.use_batching and spec.force_mllm:
+        raise ValueError(
+            "--mllm is not supported with continuous batching (it produces empty "
+            "output). Run the model with SimpleEngine by omitting --continuous-batching."
+        )
     if spec.use_batching:
         from .engine.batched import BatchedEngine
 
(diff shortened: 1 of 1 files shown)


waybarrios:
Verified this against the code and the guard is incomplete: it only fires for explicit `--mllm`, but BatchedEngine auto-detects multimodal models at [batched.py L189](https://github.com/waybarrios/vllm-mlx/blob/547a4fa496609ad9aac42efd3c59a2e445d26462/vllm_mlx/engine/batched.py#L188-L190):

```python
# BatchedEngine.__init__ — force_mllm is not the only way into the MLLM path
self._is_mllm = force_mllm or is_mllm_model(model_name)
```

So this still hits the silent-empty-output path with both guards in place:

```bash
# no --mllm flag, model auto-detected as MLLM, guard never fires
vllm-mlx serve mlx-community/Qwen2.5-VL-7B-Instruct-4bit --continuous-batching
```

The fix is to mirror what BatchedEngine actually checks, in both [_build_engine](https://github.com/waybarrios/vllm-mlx/blob/547a4fa496609ad9aac42efd3c59a2e445d26462/vllm_mlx/server.py#L1170-L1176) and [load_model](https://github.com/waybarrios/vllm-mlx/blob/547a4fa496609ad9aac42efd3c59a2e445d26462/vllm_mlx/server.py#L3072-L3078); `is_mllm_model` is already imported in server.py:

```python
# _build_engine
if spec.use_batching and (spec.force_mllm or is_mllm_model(spec.model_name)):
    raise ValueError(
        "MLLM models are not supported with continuous batching (silent empty "
        "output). Run with SimpleEngine by omitting --continuous-batching."
    )

# load_model
if use_batching and (force_mllm or is_mllm_model(model_name)):
    raise ValueError(...)  # same message
```

With that widened, this is mergeable.

PR #605, opened 2026-06-10
fix(ssd-cache): spill native QuantizedKVCache + handle bfloat16 buffer

CBribiescas:
Recreates #602 (branch was deleted) on top of latest `main`.

### Problem
With `--kv-cache-quantization` enabled, KV layers are `QuantizedKVCache` whose keys/values are tuples of `(packed, scales, biases)`. The SSD-spill serializer didn't recognize this layout, so spilling a quantized KV cache failed. Separately, numpy's PEP 3118 buffer protocol can't read `bfloat16`, which broke serialization of bf16 tensors.

### Fix
- Detect quantized layers (`_is_quantized_layer`) and route them through a serializer that handles the `(packed, scales, biases)` tuple.
- Cast `bfloat16` to a numpy-readable dtype before serialize.
- Test coverage in `tests/test_ssd_cache.py`.

Supersedes the closed #602; rebased clean on current main.

### DIFF
--- a/tests/test_ssd_cache.py
+++ b/tests/test_ssd_cache.py
@@ -897,7 +897,7 @@ class TestMemoryCacheSSDCheck:
 
     def test_check_ssd_returns_candidate_on_miss(self, tmp_path):
         model = MagicMock()
-        config = MemoryCacheConfig(max_memory_mb=1, max_entries=2)
+        config = MemoryCacheConfig(max_memory_mb=1, max_entries=2, min_prefix_tokens=1)
         cache = MemoryAwarePrefixCache(model, config)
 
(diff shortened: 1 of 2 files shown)


waybarrios:
Merged, thanks. One follow-up worth a small PR: the dequant path casts bf16 to fp16 before spilling but doesn't record the original dtype, so restore hands back fp16 where the model computed bf16. The plain-cache path already solves this with metadata:

```python
# existing pattern in _mx_to_numpy_safe
snapshot["keys_original_dtype"] = "bfloat16"  # restored on reload
```

Mirroring that for the dequantized layers would keep the round-trip dtype-faithful. While at it, `SERIALIZER_SUPPORT_MATRIX` could gain a `"QuantizedKVCache": "supported_via_dequant_on_spill"` entry so the diagnostics table matches reality.

PR #610, opened 2026-06-11
fix(engine): stop MLLM text route at the model's full config EOS set

ursk:
## Problem

On the SimpleEngine MLLM text route (`text-only -> mlx_lm TextModel`), generation never stops at the model's chat end-of-turn token. The route passes the raw HF processor tokenizer to `mlx_lm.stream_generate`, which wraps it with `eos_token_ids={tokenizer.eos_token_id}` — dropping turn terminators that chat models declare only in the config EOS list.

Gemma 4 is the clearest case: `config.json` declares `eos_token_id: [1, 106, 50]`, where `<turn|>`=106 terminates chat turns and `<|tool_response>`=50 terminates tool responses, while `tokenizer.eos_token` stays `<eos>`=1. The model emits 106, the sampler doesn't stop, the marker detokenizes into visible text, and generation free-runs until `max_tokens` — every text-route response ends with `finish_reason: "length"` and trailing garbage:

```
<|channel>thought
<channel|>Yes, there is a file named `asitop_powermetrics…` in /tmp.<turn|>
C'est parti ! On va s'occuper de ton projet de site web. …   ← free-running
```

The hard-coded Qwen3.5 `<|im_end|>` patch already in that block is this same bug, previously fixed per-model.

Two adjacent defects in the same path, included here because they surface in the same repro:

- **`usage.prompt_tokens` is 0** on the text route unless the system-KV-cache branch (requires a system message *and* ChatML `<|im_start|>` markers — never true for Gemma templates) or specprefill happens to run.
- **Reasoning markers leak into `content` when thinking is disabled.** The `allow_reasoning` gate (#537) skips the reasoning parser entirely under `enable_thinking=false`, but Gemma 4 opens `<|channel>thought` regardless, so raw channel markers end up in `message.content`.

## Fix

1. **`collect_eos_token_ids()`** (new, `utils/tokenizer.py`): unions `tokenizer.eos_token_id` / `.eos_token_ids` with the `config.json` + `generation_config.json` EOS lists. The text route now wraps its tokenizer in `mlx_lm.TokenizerWrapper` with that set. `MLLMScheduler._get_stop_tokens` (the batched path) is refactored onto the same helper.

   ⚠️ **Behavioral change on the batched path**: the old `_get_stop_tokens` read only `generation_config.json`; the shared helper also reads `config.json`. Models that list extra ids (e.g. pad/bos) in the config EOS set will now stop on tokens the batched path previously ignored — that union is the point of the fix. The Gemma 4 set `{1, 106, 50}` is pinned through `MLLMScheduler._get_stop_tokens` in `tests/test_collect_eos_token_ids.py`.

2. **Up-front prompt tokenization** in `_stream_generate_text` so `usage.prompt_tokens` is always reported; the system-KV and specprefill branches reuse the tokens instead of re-encoding.

3. **Refined `allow_reasoning` gate — non-streaming and streaming**:
   - Non-streaming: when thinking is disabled, run `extract_reasoning` iff the parser's explicit start/end markers are present in the output. With explicit markers, the implicit-thinking-swallows-content hazard #537 guards against cannot occur; without them, behavior is unchanged.
   - Streaming (chat completions, Anthropic messages, Responses): with thinking disabled the parser stays off until the model emits the parser's explicit marker in the accumulated raw stream; from that point deltas are routed through the parser. Parsed reasoning is suppressed (the request disabled thinking) and only cleaned content is emitted — this also keeps the Anthropic block choreography valid when the text block is already open. Implicit-thinking parsers without explicit marker attributes never latch, so #537 behavior is unchanged.

## Verification

Live against `gemma-4-31b-it` 4-bit MLX on SimpleEngine (M3 Ultra):

- Canonical 3-turn tool flow (`user → assistant(tool_call) → tool(long result)`): **before** — correct sentence, then `<turn|>` leak, then drift to the 256-token cap, `finish_reason: "length"`, `prompt_tokens: 0`; **after** — one-sentence answer, `finish_reason: "stop"`, 25 completion tokens, real `prompt_tokens`.
- Tool-call emission leg: `finish_reason: "tool_calls"` with well-formed `arguments`, `usage` populated.
- 12 hermetic unit tests for `collect_eos_token_ids` incl. the batched-path Gemma 4 pin (`tests/test_collect_eos_token_ids.py`).
- Marker-guard tests for the reasoning leak in both paths (`tests/test_chat_template_kwargs.py`): non-streaming extractor with/without markers, plus chat-completions and Anthropic streaming with Gemma 4 channel markers under `enable_thinking=false`. The streaming tests fail without the streaming commit.
- Rebased onto current `main` (post-#576); full suite: 2195 passed, 11 skipped.

### DIFF
--- a/tests/test_chat_template_kwargs.py
+++ b/tests/test_chat_template_kwargs.py
@@ -282,6 +282,10 @@ async def test_simple_engine_stream_generate_text_applies_chat_template_kwargs()
         engine._is_mllm = True
         engine._text_tokenizer = MagicMock()
         engine._text_tokenizer.apply_chat_template.return_value = "prompt"
+        # Realistic tokenizer attrs: the up-front usage tokenization reads
+        # bos_token and encode() on every request.
+        engine._text_tokenizer.bos_token = None
+        engine._text_tokenizer.encode.return_value = [1, 2, 3]
         engine._text_model = MagicMock()
         engine._text_model.model = MagicMock()
 
--- a/tests/test_chat_template_kwargs.py
+++ b/tests/test_chat_template_kwargs.py
@@ -386,3 +386,153 @@ async def test_stream_anthropic_skips_reasoning_parser_when_thinking_disabled():
     assert '"type": "thinking"' not in body
     assert "HEL" in body and "LO" in body
     assert '"type": "text_delta"' in body
+
+
+# Explicit-marker guard: thinking disabled must not leak raw reasoning
+# markers (Gemma 4 opens <|channel>thought regardless of the template
+# kwarg). Non-streaming parses iff explicit markers are present; the
+# streaming paths latch the parser on when a marker appears and emit only
+# the parsed content (reasoning is suppressed — the request disabled it).
+
+_GEMMA_OUTPUT = "<|channel>thought\nLet me think.<channel|>The answer is 42."
+_GEMMA_DELTAS = (
+    "<|channel>thought\n",
+    "Let me think.",
+    "<channel|>",
+    "The answer is 42.",
+)
+
+
+def _gemma_parser():
+    from vllm_mlx.reasoning.gemma4_parser import Gemma4ReasoningParser
+
+    return Gemma4ReasoningParser()
+
+
+def test_extract_reasoning_parses_explicit_markers_when_thinking_disabled():
+    saved = srv._reasoning_parser
+    srv._reasoning_parser = _gemma_parser()
+    try:
+        reasoning, content, tool_calls = srv._extract_reasoning_and_tool_calls(
+            _GEMMA_OUTPUT, None, allow_reasoning=False
+        )
+    finally:
+        srv._reasoning_parser = saved
+
+    assert tool_calls is None
+    assert content == "The answer is 42."
+    assert reasoning and "Let me think." in reasoning
+    assert "<|channel>" not in content and "<channel|>" not in content
+
+
+def test_extract_reasoning_stays_disabled_without_markers():
+    # No explicit markers: the parser must remain skipped so implicit-mode
+    # parsers can't swallow plain content into reasoning (PR #537 hazard).
+    saved = srv._reasoning_parser
+    srv._reasoning_parser = _gemma_parser()
+    try:
+        reasoning, content, _ = srv._extract_reasoning_and_tool_calls(
+            "Plain answer.", None, allow_reasoning=False
+        )
+    finally:
+        srv._reasoning_parser = saved
+
+    assert reasoning is None
+    assert content == "Plain answer."
+
+
+@pytest.mark.anyio
+async def test_stream_anthropic_strips_markers_when_thinking_disabled():
+    async def fake_stream_chat(messages, **kwargs):
+        for piece in _GEMMA_DELTAS:
+            yield SimpleNamespace(new_text=piece, prompt_tokens=4, completion_tokens=1)
+
+    engine = MagicMock(stream_chat=fake_stream_chat)
+    msgs = [{"role": "user", "content": "What is the answer?"}]
+    openai_request = srv.ChatCompletionRequest(
+        model="test-model", messages=[srv.Message(**msgs[0])], max_tokens=8
+    )
+    anthropic_request = srv.AnthropicRequest(
+        model="test-model", max_tokens=8, messages=msgs
+    )
+    prepared = srv.PreparedChatInvocation(
+        messages=msgs,
+        chat_kwargs={"chat_template_kwargs": {"enable_thinking": False}},
+        response_format=None,
+        json_logits_processor=None,
+    )
+
+    saved = (srv._reasoning_parser, srv._model_name)
+    srv._reasoning_parser, srv._model_name = _gemma_parser(), "test-model"
+    try:
+        body = "".join(
+            [
+                c
+                async for c in srv._stream_anthropic_messages(
+                    engine, openai_request, anthropic_request, prepared
+                )
+            ]
+        )
+    finally:
+        srv._reasoning_parser, srv._model_name = saved
+
+    # Raw markers and the thought text must not reach the text block.
+    assert "<|channel>" not in body and "<channel|>" not in body
+    assert "Let me think." not in body
+    # Thinking was disabled — no thinking block, only cleaned content.
+    assert "thinking_delta" not in body
+    assert "The answer is 42." in body
+    assert '"type": "text_delta"' in body
+
+
+@pytest.mark.anyio
+async def test_stream_chat_completion_strips_markers_when_thinking_disabled():
+    import json as _json
+
+    async def fake_stream_chat(messages, **kwargs):
+        for i, piece in enumerate(_GEMMA_DELTAS):
+            yield SimpleNamespace(
+                new_text=piece,
+                prompt_tokens=4,
+                completion_tokens=i + 1,
+                finished=i == len(_GEMMA_DELTAS) - 1,
+                finish_reason="stop" if i == len(_GEMMA_DELTAS) - 1 else None,
+            )
+
+    engine = MagicMock(stream_chat=fake_stream_chat)
+    request = srv.ChatCompletionRequest(
+        model="test-model",
+        messages=[srv.Message(role="user", content="What is the answer?")],
+        max_tokens=8,
+    )
+
+    saved = (srv._reasoning_parser, srv._model_name)
+    srv._reasoning_parser, srv._model_name = _gemma_parser(), "test-model"
+    try:
+        chunks = [
+            c
+            async for c in srv.stream_chat_completion(
+                engine,
+                request.messages,
+                request,
+                chat_template_kwargs={"enable_thinking": False},
+            )
+        ]
+    finally:
+        srv._reasoning_parser, srv._model_name = saved
+
+    body = "".join(chunks)
+    assert "<|channel>" not in body and "<channel|>" not in body
+
+    contents, reasonings = [], []
+    for line in body.splitlines():
+        if not line.startswith("data: ") or line == "data: [DONE]":
+            continue
+        delta = _json.loads(line[len("data: ") :])["choices"][0]["delta"]
+        if delta.get("content"):
+            contents.append(delta["content"])
+        if delta.get("reasoning"):
+            reasonings.append(delta["reasoning"])
+
+    assert "".join(contents) == "The answer is 42."
+    assert reasonings == []  # reasoning suppressed when thinking disabled
--- a/tests/test_collect_eos_token_ids.py
+++ b/tests/test_collect_eos_token_ids.py
@@ -1,0 +1,87 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for collect_eos_token_ids — the shared EOS/stop-token-set helper.
+
+Regression context: the SimpleEngine MLLM text route passed the raw HF
+processor tokenizer to mlx_lm.stream_generate, which defaults its stop set
+to {tokenizer.eos_token_id} — dropping turn terminators that chat models
+declare only in the config EOS list (Gemma 4: <turn|>=106,
+<|tool_response>=50). Generation then ran straight through end-of-turn
+until max_tokens on every request.
+"""
+
+import json
+
+from vllm_mlx.utils.tokenizer import collect_eos_token_ids
+
+
+class FakeTokenizer:
+    def __init__(self, eos_token_id=None, eos_token_ids=None, name_or_path=None):
+        if eos_token_id is not None:
+            self.eos_token_id = eos_token_id
+        if eos_token_ids is not None:
+            self.eos_token_ids = eos_token_ids
+        if name_or_path is not None:
+            self.name_or_path = name_or_path
+
+
+def _write_config(path, name, eos):
+    (path / name).write_text(json.dumps({"eos_token_id": eos}))
+
+
+def test_tokenizer_int_eos_only():
+    assert collect_eos_token_ids(FakeTokenizer(eos_token_id=2)) == {2}
+
+
+def test_tokenizer_list_eos():
+    assert collect_eos_token_ids(FakeTokenizer(eos_token_id=[2, 7])) == {2, 7}
+
+
+def test_wrapper_eos_token_ids_attr():
+    tok = FakeTokenizer(eos_token_id=2, eos_token_ids={2, 5})
+    assert collect_eos_token_ids(tok) == {2, 5}
+
+
+def test_no_eos_anywhere_returns_empty():
+    assert collect_eos_token_ids(FakeTokenizer()) == set()
+
+
+def test_gemma4_config_eos_list_unioned(tmp_path):
+    """The bug: <turn|>=106 / <|tool_response>=50 live only in the config."""
+    _write_config(tmp_path, "config.json", [1, 106, 50])
+    tok = FakeTokenizer(eos_token_id=1)
+    assert collect_eos_token_ids(tok, tmp_path) == {1, 106, 50}
+
+
+def test_generation_config_eos_unioned(tmp_path):
+    _write_config(tmp_path, "generation_config.json", [1, 106])
+    tok = FakeTokenizer(eos_token_id=1)
+    assert collect_eos_token_ids(tok, tmp_path) == {1, 106}
+
+
+def test_both_configs_unioned(tmp_path):
+    _write_config(tmp_path, "config.json", [1, 106])
+    _write_config(tmp_path, "generation_config.json", [1, 50])
+    tok = FakeTokenizer(eos_token_id=1)
+    assert collect_eos_token_ids(tok, tmp_path) == {1, 50, 106}
+
+
+def test_model_path_falls_back_to_name_or_path(tmp_path):
+    _write_config(tmp_path, "config.json", [1, 106, 50])
+    tok = FakeTokenizer(eos_token_id=1, name_or_path=str(tmp_path))
+    assert collect_eos_token_ids(tok) == {1, 106, 50}
+
+
+def test_scalar_config_eos(tmp_path):
+    _write_config(tmp_path, "generation_config.json", 7)
+    assert collect_eos_token_ids(FakeTokenizer(), tmp_path) == {7}
+
+
+def test_corrupt_config_ignored(tmp_path):
+    (tmp_path / "config.json").write_text("{not json")
+    tok = FakeTokenizer(eos_token_id=2)
+    assert collect_eos_token_ids(tok, tmp_path) == {2}
+
+
+def test_missing_model_path_ok():
+    tok = FakeTokenizer(eos_token_id=2)
+    assert collect_eos_token_ids(tok, "/nonexistent/path") == {2}
--- a/tests/test_collect_eos_token_ids.py
+++ b/tests/test_collect_eos_token_ids.py
@@ -85,3 +85,29 @@ def test_corrupt_config_ignored(tmp_path):
 def test_missing_model_path_ok():
     tok = FakeTokenizer(eos_token_id=2)
     assert collect_eos_token_ids(tok, "/nonexistent/path") == {2}
+
+
+def test_mllm_scheduler_batched_path_pins_gemma4_stop_set(tmp_path):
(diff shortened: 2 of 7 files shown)


waybarrios:
The fixes are right and the deep review came back clean, but #576 just landed on main and touches the same `_stream_generate_text` region, so this is now conflicting and needs a rebase. Two things worth addressing while you're in there.

The marker-detection guard for reasoning leakage only fires in `_extract_reasoning_and_tool_calls`, which is the non-streaming path. The streaming path gates on `_thinking_disabled` directly and never sees the new guard, so Gemma 4 streaming with thinking disabled still leaks markers:

```python
# server.py, streaming path: never calls _extract_reasoning_and_tool_calls
if not _thinking_disabled(request, chat_kwargs):
    ...
```

Not a regression (streaming was already broken), but the PR description claims the leak is fixed and right now that's only true for non-streaming. Either cover streaming too or scope the claim. There are also no tests for the reasoning-marker fix in either path.

Second, the MLLMScheduler refactor changes behavior slightly: the old `_get_stop_tokens` read only `generation_config.json`, while `collect_eos_token_ids` also reads `config.json`. That's the point of the fix, but models that list pad/bos ids in the config EOS set will now stop on tokens they previously ignored on the batched path. Worth a line in the description and ideally a test pinning the Gemma 4 token set.

ursk:
All three points addressed:

- **Rebased onto main** (post-#576). The conflict was in the text-route tokenizer setup and the `_stream_generate_text` preamble; resolution keeps #576's `_probe_system_kv_cache_support` call and `cache_blocking_controls` gate alongside this PR's `TokenizerWrapper` wrap and up-front tokenization. Full suite after rebase: 2195 passed, 11 skipped.
- **Streaming is now covered** (`a2d9a41`): all three streaming paths (chat completions, Anthropic messages, Responses) get the streaming counterpart of the marker guard — with thinking disabled the parser stays off until the parser's explicit marker appears in the accumulated raw stream, then deltas are routed through it. Parsed reasoning is suppressed (the request disabled thinking) and only cleaned content is emitted, which also keeps the Anthropic block sequence valid when the text block is already open. Implicit parsers without explicit marker attributes never latch, so the #537 hazard is untouched. Tests cover the guard in the non-streaming extractor and both the chat-completions and Anthropic streaming paths with Gemma 4 channel markers; the streaming tests fail without the fix commit.
- **Batched-path EOS change** (`2b931cf`): the Gemma 4 set `{1, 106, 50}` is now pinned through `MLLMScheduler._get_stop_tokens` (both processor shapes), and the PR description calls out the config.json union as a behavioral change for models that list extra ids in the config EOS set.

ursk:
Live E2E verification of the rebased branch is done — with one important caveat that turned into #613 / #614.

Running the rebased branch against gemma-4-26b-a4b (and 31B dense) initially failed every text-route request with `There is no Stream(gpu, 2) in current thread`. That is **not** this PR: pure `main` (`a48c86c`) fails identically, and `git bisect` lands on #595 — which merely enabled the TextModel route for Gemma 4, exposing a latent thread-affinity fault (lazy scaled-RoPE `_freqs`, excluded from `parameters()`, tagged to the load thread's stream; full diagnosis in #613). The targeted fix is #614.

With #614 applied on top of this branch, the full live gate passes on gemma-4-26b-a4b-it 4-bit (SimpleEngine, M3 Ultra), exercising exactly this PR's claims:

- startup: `Text route stop tokens: [1, 50, 106]` — the pinned Gemma 4 config EOS set
- tool-call leg: `stop_reason: "tool_use"`, well-formed `input`, `usage` populated
- canonical 3-turn tool-result follow-up (1.6k-token tool result): `stop_reason: "end_turn"`, `input_tokens: 1594` (was 0), 60 completion tokens, no `<turn|>` / `<|channel>` leakage, no free-running
- streaming with thinking disabled (server default `enable_thinking: false`): clean `text_delta`s, no raw channel markers, no thinking blocks

So #614 is a soft prerequisite for observing this PR's text-route fix live on current main, though the two are independent changes. Review order is your call; they don't conflict.

ursk:
Friendly ping on this.

janhilgard (approved):
Approving. Author's live-repro on gemma-4-26b-a4b-it 4-bit is convincing, and I can independently verify every claim in the PR body against upstream `main` and against a real Gemma 4 model on disk.

**Verified locally against `main` (post-#576):**

- **Current `SimpleEngine.start()`** stashes the raw HF processor tokenizer directly onto `self._text_tokenizer` (`vllm_mlx/engine/simple.py`, around the Qwen3.5 eos_token fix block). That tokenizer is what `_stream_generate_text` hands to `mlx_lm.stream_generate`, which auto-wraps it with `eos_token_ids={tokenizer.eos_token_id}` — so any EOS id declared only in the config JSON is silently dropped. Confirmed with a diff of `main` vs the PR: `main` has no `TokenizerWrapper` wrap, no `collect_eos_token_ids` call, no `Text route stop tokens` log line.

- **Current `MLLMScheduler._get_stop_tokens`** reads only `generation_config.json` (the code path in the diff is literally this: unions tokenizer attributes with `Path(model_path) / "generation_config.json"` and stops). It does not read `config.json`. That's the behavioral gap the PR points at, and the direction of the fix — reading both files and unioning — is the same widening the shared helper does for the text route. Good instinct to route both paths through one helper.

- **Live `config.json` on my disk for Gemma 4 12B and (historically) 31B** literally declares `"eos_token_id": [1, 106, 50]`, and `text_config.eos_token_id` is only `1`. So today, a chat model whose actual turn terminator lives in `config.json` gets that terminator lost on both paths — the batched one because it never opens `config.json`, the SimpleEngine text route because it hands the raw tokenizer to mlx_lm and defaults to the tokenizer's single scalar `eos_token_id`. The PR body's description of the symptom (`<turn|>` detokenizes into visible content, generation free-runs to `max_tokens`, `finish_reason: length`) is exactly what would happen.

- **`collect_eos_token_ids` (`utils/tokenizer.py`)** looks well-scoped: it unions `tokenizer.eos_token_id`, `tokenizer.eos_token_ids`, and both `config.json` / `generation_config.json` `eos_token_id` fields. Scalar and list shapes both accepted. `int(v)` coercion inside the update handles non-int stragglers. Missing directory / corrupt JSON both handled with silent fallback — right shape for a helper that runs at model-load time and mustn't fault the process.

- **The three commits after waybarrios's initial review are visible in the diff and address exactly what he flagged.** `2b931cf` adds the batched-path pin (`test_mllm_scheduler_batched_path_pins_gemma4_stop_set` covers both processor shapes: `processor.tokenizer` and bare `tokenizer`-as-processor). `a2d9a41` adds the streaming counterpart of the marker guard on all three streaming paths (chat completions, Anthropic messages, Responses). `b406ea1` fixes the `name_or_path` fallback for cases where the caller passes no explicit path.

**Behavioral change worth calling out (which the PR body does):**

The PR is honest that the batched-path change alters behavior for models that declare pad/bos ids in `config.json`'s EOS list — they'll now stop on those tokens where they previously ignored them. For the models I run in production (Qwen 3.6-27B/35B, historically Gemma 4 12B/31B), `config.json` EOS is either the tokenizer scalar or the specific chat-terminator set that *should* be stopping — the union is what fixes the bug. If any user has a model where `config.json` EOS lists spurious ids they were relying on being ignored, that model was rendering correctly by accident.

**Code-review notes:**

1. The streaming marker latch (`disabled_reasoning_latched`) is a **one-way** transition — once the parser turns on for a request, it stays on for the rest of the stream. That's the right choice: raw markers only appear when the model has decided to open a thinking block, and it's not going to un-decide mid-stream. If the same client somehow issued a follow-up request without the parser being re-armed, `reset_state()` handles it in the streaming reset path. Worth a one-line comment saying the latch is monotonic within a request, so a future reader doesn't try to make it two-way for cleverness.

2. The `_stream_anthropic_messages` change carries a subtle interaction with the already-open text block: when the latch fires mid-stream, parsed content flows into the *already-open* text block, not a new one. That's semantically correct (the parser produces content, not thinking, because thinking is disabled), but it means the Anthropic block sequence still reads as `text_delta` throughout and the client never sees a `thinking_delta`. That's the right choreography — the client asked for thinking off — but worth a comment in the streaming path explaining why no `thinking` block is ever started, because a reader might wonder if it's a missing branch.

3. The chat-completion streaming path (`stream_chat_completion`) introduces a `raw_stream_text` accumulator distinct from `accumulated_text`. Necessary — the reasoning branch below only updates `accumulated_text` when it runs, so under the latched-off condition it would fall behind — but the two-accumulator dance is a code smell. Not a blocker; the comment in the diff explains why. Might be worth a follow-up to consolidate on `raw_stream_text` as the primary and have the parser-flavored `accumulated_text` become a view.

4. `_explicit_reasoning_markers_present` reads `start_token` / `end_token` off the parser. Every reasoning parser I've seen in `vllm_mlx/reasoning/` exposes those (Gemma 4, Qwen 3, Harmony, GPT-OSS, DeepSeek R1, Glm4, Step3p5). Implicit-thinking parsers that don't expose them get `False` from this helper and stay skipped when thinking is disabled — which is exactly the #537 hazard-preserving behavior, as advertised.

5. The batched-path test (`test_mllm_scheduler_batched_path_pins_gemma4_stop_set`) uses `SimpleNamespace` to fake the scheduler shape (processor with a `.tokenizer` attribute vs. bare tokenizer). Both processor shapes are covered. Good.

6. The 12 hermetic unit tests in `test_collect_eos_token_ids.py` are the good kind of unit tests — they pin the exact `{1, 106, 50}` Gemma 4 set at multiple levels (tokenizer alone, config alone, both, generation_config alone, both configs, missing/corrupt JSON, missing path, `name_or_path` fallback). If a future refactor breaks any one of those seams, one of the 12 tests will fail with a specific reason.

7. `#613 / #614` context (from the author's third comment): the live E2E on rebased branch initially failed with `There is no Stream(gpu, 2) in current thread`. That's not this PR — pure `main` fails identically, and `git bisect` lands on `#595`. `#614` is the targeted fix and is already merged (I see `e2d3d95 fix(text-model-from-vlm): realize private lazy arrays before the model leaves the build thread (#614)` in my local `origin/main`). So the soft prereq is already satisfied on current `main`; this PR should now merge cleanly against it without needing operators to know about the #613/#614 dance.

8. CI: lint ✓, type-check ✓, test-matrix × 4 ✓, test-apple-silicon × 2 ✓, tests ✓. All 9 green on `b406ea1`. Full suite in the description: 2195 passed, 11 skipped.

I ran Gemma 4 12B on the SimpleEngine + MTP path (`--mllm-draft-model` requires SimpleEngine per the cli.py guard), and I ran Gemma 4 31B before that. Both models declare `config.eos_token_id: [1, 106, 50]`, both were rendering with the exact symptom the PR describes: `<turn|>` markers in output, `finish_reason: length` on continuation turns, `prompt_tokens: 0` in usage. I retired the 12B for unrelated single-user-cap reasons but the parsing tests I ran against it (a12b_parsing.py in a previous session) always showed the model going long — I chalked it up to model size limits. In hindsight the free-running was the config-EOS bug, not the model. This PR would have made those live tests noticeably cleaner.

The `--reasoning-parser gemma4` I use has `<|channel>thought`/`<channel|>` as explicit start/end markers, so `_explicit_reasoning_markers_present` will latch correctly on my deployments. If I ever re-enable Gemma 4 12B I'll pull this PR in before I do.

LGTM — ready to merge on top of current `main` (post-#614).

janhilgard:
Reinforcing evidence from a different model family — this bug isn't Gemma-specific, it hits any model that declares turn/tool terminators only in `config.json`.

**Case: Poolside Laguna-S-2.1 on `poolside/Laguna-S-2.1-NVFP4-mlx`** (deployed briefly last week for evaluation, since retired for unrelated quality reasons).

- `config.json` declares `eos_token_id: [2, 24]` — `<|EOS|>` = 2 and `</assistant>` = 24 (turn terminator, injected by the chat template).
- `tokenizer_config.json` only carries `eos_token: "<|EOS|>"` → tokenizer exposes `eos_token_id = 2` singleton.
- Repo ships **no `generation_config.json`** at all.

Result on the MLLM batched path (`MLLMScheduler._get_stop_tokens` — the pre-fix version this PR consolidates via `collect_eos_token_ids`): stop set = `{2}` only. Model correctly emitted `</assistant>` (id 24) as turn terminator, sampler didn't stop, `</assistant>` detokenized into visible content, and generation ran to `max_tokens` producing:

```
12 + 15 = **27**</assistant>
12 + 15 = **27**</assistant></assistant>
12 + 15 = **27**</assistant></assistant>...
```

Every simple request produced 500+ tokens with `finish_reason: "length"` and the leaked terminator repeated dozens of times. Passing `stop_token_ids: [2, 24]` in the request body didn't help — the scheduler's stop set is built at engine start, not per-request.

**Workaround I used before finding this PR**: manually created `generation_config.json` in the model directory with `{"eos_token_id": [2, 24]}`. Instant fix — `finish_reason` went to `stop`, 76 tokens for "what is 12+15?" instead of runaway.

The `collect_eos_token_ids` helper in this PR would have made that workaround unnecessary: it unions `tokenizer.eos_token_id` + `config.json` EOS + `generation_config.json` EOS, catching all three sources. The batched-path refactor (`MLLMScheduler._get_stop_tokens` onto the shared helper) is exactly the right consolidation — same code path, same behavior across text and MLLM routes, no more per-model workarounds.

Not blocking — I already approved on 2026-07-03 — but this is a second production model family (beyond Gemma 4) where the fix would have saved hours of debugging. Worth prioritizing the merge.

waybarrios:
Thanks for the PR. I prepared and verified the conflict resolution against current main (`61fab4b`).

GitHub shows “maintainers can edit”, but the head branch belongs to an organization-owned fork, so upstream maintainers do not receive push access. My attempted push was rejected with HTTP 403.

I published the resolved branch at `waybarrios/vllm-mlx:pr-610`. Please fast-forward your PR branch with:

git fetch https://github.com/waybarrios/vllm-mlx.git pr-610
git switch fix/mllm-text-route-eos
git merge --ff-only FETCH_HEAD
git push origin fix/mllm-text-route-eos

Everything else in the PR looks good. Once the branch is updated, it should be ready to merge.

waybarrios (approved):

Thump604 (commented):
I independently rechecked the exact PR head `b406ea16ca749117faca53813f85186997728cb2` against current `main` `22efb47968f96e6e7a79704f4eb0184cda4e0068`. The intended EOS fix is still relevant: current `vllm_mlx/engine/simple.py` still passes the raw text-route tokenizer to `mlx_lm.stream_generate`, and current `vllm_mlx/mllm_scheduler.py:_get_stop_tokens()` still reads only tokenizer attributes and `generation_config.json`. The PR head’s `simple.py:477-514` wraps the post-Qwen-fix tokenizer with the union from `utils/tokenizer.py:31-82`, and `mllm_scheduler.py:245-258` routes the batched path through the same helper. The added helper tests cover scalar/list tokenizer values, both config files, `name_or_path`, missing files, and corrupt JSON.

The branch is not merge-ready: GitHub reports `DIRTY` against base `a48c86c1a41900f7d26658471b5f67e5fdd35445`; the current-main comparison is divergent (`main` is 77 commits ahead and the PR head has 6 commits not on main), and the nine successful check-runs are from 2026-06-28 with no current status checks. Rebase the exact head onto current `main`, resolve the overlaps in the text-route startup and server streaming paths, then rerun the current CI matrix and focused EOS/reasoning tests. Please retain the existing two processor-shape batched test and the SimpleEngine wrapper/stop-token assertions after conflict resolution, and report the exact mlx/mlx-lm tree used for the live Gemma gate.

One small hardening gap is worth closing while rebasing: `collect_eos_token_ids()` catches unreadable/invalid JSON but `int(v)` can still raise on a malformed scalar/list value. Add a malformed-EOS-value regression and either ignore invalid optional values or fail with an explicit model-config error; do not let an optional config typo silently turn into an unexplained startup failure. This is independent of the standard numeric Gemma config and is not a claim that the reported live measurements are reproduced here.

This is a COMMENT review, not an approval or merge decision. After the rebase and current checks, maintainers can make the merge decision; no upstream source has been adopted locally.

waybarrios:
@ursk are you planning to still work on this pr?

mmcco:
Independent confirmation that this is still needed on current `main` (`ec8e493`), on a different Gemma 4 checkpoint: `mlx-community/gemma-4-12B-it-qat-mxfp8` @ `d8989284`, M5 Max, MLX 0.32.2, mlx-lm 0.31.3, SimpleEngine, `--mllm`.

The text route on that checkpoint is only reachable with #791 applied (without it, MXFP8 text requests fail earlier with a missing-kernel 500; see #783). With #791 on top of `ec8e493`, and the model served from its local snapshot directory:

- Loaded exactly as `MLXMultimodalLM.load()` loads it, `processor.tokenizer.eos_token_id` is `1`, and `TokenizerWrapper(processor.tokenizer).eos_token_ids` is `[1]`. The model config declares `[1, 106, 50]` (`106` = `<turn|>`, `50` = `<|tool_response>`).
- `"What is the capital of France? Answer in one word."` at `max_tokens=256` returns `finish_reason: "length"`, 256 completion tokens: `'Paris<turn|><turn|><turn|>thought\n<channel|>Paris<turn|><turn|><turn|>thought\n…'`. Served by Hub id, the same model takes the mlx-vlm route instead and answers in 3 tokens.
- A strict `json_schema` request on the same route stops normally (`finish_reason: "stop"`, 63 tokens), because the grammar closes it. So schema-only smoke tests will not catch this.

The symptom and cause match this PR's description exactly. Since #791 makes the route reachable for MXFP8 checkpoints, merging it alone will surface this runaway to more users.

ursk:
Yes — sorry for the long silence. GitHub notifications weren't reaching my inbox, so I missed both your Aug 20 resolution and the follow-up.

Branch is updated. Fast-forwarding onto `pr-610` alone wasn't enough: `main` had moved 40 commits past `61fab4b` and re-touched the same regions, so I merged current main on top of your resolution. How I resolved it:

- **`engine/simple.py`** — took the refactor as-is (the inline text-route build is now `_initialize_text_model()` with lazy deferral for draft routes, #650/#669) and re-applied the EOS wrap inside that helper, so both the eager and deferred paths get it.
- **`server.py`** — kept `_extract_streaming_reasoning_delta()` and widened its gates to `use_reasoning or disabled_reasoning_latched`, so the explicit-marker latch still fires when thinking is disabled. The no-delta early-`continue` needed the latch too, or a latched stream stops flushing on its final chunk.

Two existing tests needed updating. Flagging them rather than burying them, because this PR changes what they cover:

- `test_stream_chat_does_not_add_nemotron_prefix_when_thinking_disabled` — its stub parsers define `reset_state(self)`. `main` never calls `reset_state` on that path (it returns early when thinking is disabled); this PR builds the parser anyway so the latch can observe the stream, so the stubs now accept `implicit_mode`.
- `test_stream_generate_text_skips_system_cache_when_text_model_not_safe` — asserted `encode()` was never called, as a proxy for "manual system-cache path skipped". That proxy no longer holds, since the full prompt is now tokenized on every request so `usage.prompt_tokens` stops reporting 0. The path itself is still asserted by `fail_if_manual_cache_path_runs`; the assertion now pins exactly one encode, of the full prompt, and no system-prefix encode.

I also added one test for something your refactor surfaced: text-route construction now has an eager entry and a drafter-deferred one, and only the eager path asserted the EOS wrap. A model whose turn terminator is declared only in `config.json` would free-run on the deferred path alone while the eager path stayed green. Verified it fails with the wrap removed.

Re-verified end to end on `gemma-4-26b-a4b-it-4bit` against this branch — same model, same prompt, temperature 0, the only variable being the EOS wrap:

```
WITHOUT the wrap:  finish_reason "length", 400 completion tokens (hit the cap),
                   <turn|> leaking into visible content, 1766 chars
WITH the wrap:     finish_reason "stop", 180 completion tokens, clean answer
                   log: Text route stop tokens: [1, 50, 106]
```

The broken tail is the same shape as the original report — correct answer, then the marker as visible text, then free-running repetition:

```
...red, yellow, and blue.<turn|><|channel>thought
<channel|>The three primary colors are red, yellow, and blue.<turn|><turn|>
```

One caveat on that A/B: `prompt_tokens` was 27 in both runs, because removing only the EOS wrap leaves the always-tokenize-the-full-prompt change in place. So it isolates the EOS fix, not the `prompt_tokens` one.

CI is green across the board, both Apple Silicon jobs included.

PR #611, opened 2026-06-12
Keep assistant tool_calls and tool messages on the MLLM chat path

### LINKED ISSUE #608: [Bug] MLLM chat path drops assistant tool_calls + tool-role messages — multi-turn tool use loops forever
## Summary

On the MLLM path (vision models used for text-only chat), `MLLM.chat()` and `MLLM.stream_chat()` rebuild the message list via `_build_mllm_chat_messages` (`vllm_mlx/models/mllm.py:276` on `main` today), which appends only `{"role": role, "content": content}`. An OpenAI-style assistant message with `content: null` and `tool_calls` carries no text, so `has_content` is false and the message is **dropped entirely** (the `tool_calls` key is never read); the following `role:"tool"` result then loses its anchor and renders to nothing in templates that emit tool responses by forward-scan from the assistant `tool_calls` (e.g. Gemma 4's `chat_template.jinja`).

Net effect: the rendered continuation prompt is **byte-identical to the fresh prompt**, so the model deterministically re-issues the same tool call every round — an unconditional infinite tool loop for any agent client on an MLLM-served model. The non-MLLM (text LLM) path does not have this problem, which makes the failure look model-specific when it is not.

Affected: 0.3.0 (tested live), v0.4.0rc1, and **main** (the `_build_mllm_chat_messages` helper copies only role/content; `has_content` gates the append).

## Reproduction (gemma-4-12B-it-4bit; `--enable-auto-tool-choice --tool-call-parser gemma4 --reasoning-parser gemma4`)

- Probe A — fresh turn (system + user + 1 tool): correct `tool_calls` response, `usage.prompt_tokens = 132`.
- Probe B — continuation (system + user + assistant{content:null, tool_calls} + tool{result}): the **identical tool call again**, and `usage.prompt_tokens = 132` — equal to Probe A. The two history messages contributed zero prompt tokens: dropped before templating.
- Counter-proof: rendering the same 4-message conversation with the model's own `chat_template.jinja` (tool exchange renders, 176 tokens) and feeding it raw via `/v1/completions` makes the model close the turn correctly in 8 tokens (`DONE 3`). The model and template are fine; the rebuild loop is the bug.

## Suggested fix

In `_build_mllm_chat_messages` (and the two call sites in `chat()` / `stream_chat()`):
1. Keep assistant messages that carry `tool_calls` even when text is empty, forwarding `tool_calls` (arguments parsed str→mapping, as `engine/batched.py`'s `_normalize_tool_call_arguments_for_template` already does) and `reasoning_content` when present.
2. Forward `role:"tool"` messages with their `tool_call_id` and text content so template forward-scans can resolve them.
3. Consider `SUPPORTS_NATIVE_TOOL_FORMAT = True` on `Gemma4ToolParser` (the Gemma 4 template handles native tool messages).

I have a working local patch (applied to 0.3.0; verified: continuation renders +45 tokens and the model answers instead of looping) and can open a PR.

## Detection suggestion

A regression test that replays a 4-message tool exchange and asserts `prompt_tokens(continuation) > prompt_tokens(fresh)` — single-turn tool tests cannot catch this class.

waybarrios:
## Summary

Fixes #608. On the MLLM path, `_build_mllm_chat_messages` rebuilt every message as `{role, content}` and dropped any message without renderable text. An OpenAI assistant message with `content: null` + `tool_calls` was removed entirely, and `role:"tool"` messages lost their `tool_call_id`. The rendered continuation prompt came out byte-identical to the fresh prompt, so agent clients looped on the same tool call forever.

## What changed

All in `_build_mllm_chat_messages` (shared by `chat()` and `stream_chat()`, so both routes are covered with no call-site changes):

- Assistant messages carrying `tool_calls` survive even with empty text. Arguments are normalized str -> mapping via the same `normalize_messages_for_chat_template` helper the batched engine path wraps, so both paths render identically. `reasoning_content` is forwarded when present.
- `role:"tool"` messages keep their `tool_call_id`. Empty tool results (legit for action tools) survive with `content: ""` instead of being dropped, so template forward-scans can still pair calls to responses.
- Everything else is byte-identical to before: image/audio/video part ordering, plain-text messages, and the drop rule for genuinely empty messages (verified pre/post on the same inputs).

`SUPPORTS_NATIVE_TOOL_FORMAT` on `Gemma4ToolParser` (issue suggestion 3) is intentionally left out of scope.

## Tests

5 new regression tests in `tests/test_mllm_message_ordering.py` (written first, 3 failed on main for the exact bug reasons):

- assistant `{content: null, tool_calls}` survives with normalized arguments
- tool messages keep `tool_call_id` + content
- empty tool results survive with their anchor
- a 4-message tool exchange rebuilds strictly more messages than the fresh 3-message turn
- genuinely empty messages without `tool_calls` are still dropped

```
pytest tests/test_mllm_message_ordering.py -q   -> 8 passed
pytest tests/test_mllm*.py tests/test_server.py -> 256 passed
pytest tests/ -k "not slow"                     -> 2185 passed, 0 failed
ruff check                                       -> clean
```

## Live verification (issue repro)

Served `mlx-community/Qwen3-VL-8B-Instruct-4bit` with `--mllm --enable-auto-tool-choice --tool-call-parser hermes`, M4 Max:

- Probe A (fresh: system + user + 1 tool): tool_call emitted, `prompt_tokens = 166`
- Probe B (continuation: + assistant{content:null, tool_calls} + tool{result}): `prompt_tokens = 213 > 166` (history now contributes tokens), no repeated tool_call, final answer uses the tool result: "The weather in Paris is currently 19°C with sunny conditions."

On main, probe B renders the same prompt as probe A and re-issues the identical tool call.

### DIFF
--- a/tests/test_mllm_message_ordering.py
+++ b/tests/test_mllm_message_ordering.py
@@ -1,3 +1,5 @@
+import json
+
 from vllm_mlx.models.mllm import _build_mllm_chat_messages
 
 
@@ -108,3 +110,152 @@ def test_mllm_chat_messages_preserve_audio_text_order():
         },
     ]
     assert all_image_urls == []
+
+
+def test_mllm_chat_messages_keep_assistant_tool_calls_without_text():
+    # Issue #608: the standard OpenAI shape for an assistant tool-call turn is
+    # {content: None, tool_calls: [...]} and it must not be dropped.
+    messages = [
+        {
+            "role": "assistant",
+            "content": None,
+            "tool_calls": [
+                {
+                    "id": "call_1",
+                    "type": "function",
+                    "function": {
+                        "name": "get_weather",
+                        "arguments": '{"city": "Lima"}',
+                    },
+                }
+            ],
+            "reasoning_content": "User wants the weather.",
+        }
+    ]
+
+    chat_messages = _build_mllm_chat_messages(
+        messages,
+        all_image_urls=[],
+        video_frame_counts={},
+    )
+
+    assert chat_messages == [
+        {
+            "role": "assistant",
+            "content": "",
+            "tool_calls": [
+                {
+                    "id": "call_1",
+                    "type": "function",
+                    "function": {
+                        "name": "get_weather",
+                        # JSON argument strings are parsed to mappings for
+                        # templates that iterate argument keys.
+                        "arguments": {"city": "Lima"},
+                    },
+                }
+            ],
+            "reasoning_content": "User wants the weather.",
+        }
+    ]
+
+
+def test_mllm_chat_messages_keep_tool_call_id_on_tool_messages():
+    messages = [
+        {"role": "tool", "tool_call_id": "call_1", "content": '{"temp_c": 19}'}
+    ]
+
+    chat_messages = _build_mllm_chat_messages(
+        messages,
+        all_image_urls=[],
+        video_frame_counts={},
+    )
+
+    assert chat_messages == [
+        {
+            "role": "tool",
+            "content": [
+                {
+                    "type": "text",
+                    "text": '{"temp_c": 19}',
+                    "content": '{"temp_c": 19}',
+                }
+            ],
+            "tool_call_id": "call_1",
+        }
+    ]
+
+
+def test_mllm_chat_messages_keep_empty_tool_results():
+    """Tools may return empty output; the anchored message must survive."""
+    for empty_content in ("", None):
+        messages = [
+            {"role": "tool", "tool_call_id": "call_1", "content": empty_content}
+        ]
+
+        chat_messages = _build_mllm_chat_messages(
+            messages,
+            all_image_urls=[],
+            video_frame_counts={},
+        )
+
+        assert chat_messages == [
+            {"role": "tool", "content": "", "tool_call_id": "call_1"}
+        ]
+
+
+def test_mllm_chat_messages_keep_tool_exchange():
+    fresh = [
+        {"role": "system", "content": "You can call tools."},
+        {"role": "user", "content": "What is the weather in Lima?"},
+    ]
+    tool_call = {
+        "id": "call_1",
+        "type": "function",
+        "function": {"name": "get_weather", "arguments": '{"city": "Lima"}'},
+    }
+    continuation = fresh + [
+        {"role": "assistant", "content": None, "tool_calls": [tool_call]},
+        {"role": "tool", "tool_call_id": "call_1", "content": '{"temp_c": 19}'},
+    ]
+
+    fresh_out = _build_mllm_chat_messages(
+        fresh, all_image_urls=[], video_frame_counts={}
+    )
+    cont_out = _build_mllm_chat_messages(
+        continuation, all_image_urls=[], video_frame_counts={}
+    )
+
+    # The 4-message tool exchange must render strictly more prompt content
+    # than the fresh exchange (issue #608: it rendered byte-identical).
+    assert len(cont_out) == len(fresh_out) + 2
+    assert json.dumps(cont_out, sort_keys=True) != json.dumps(fresh_out, sort_keys=True)
+    assert len(json.dumps(cont_out)) > len(json.dumps(fresh_out))
+
+    # Assistant tool_calls survive even with empty text content.
+    assistant_msg = cont_out[2]
+    assert assistant_msg["role"] == "assistant"
+    assert assistant_msg["tool_calls"][0]["function"]["name"] == "get_weather"
+
+    # Tool result keeps its anchor for template forward-scans.
+    tool_msg = cont_out[3]
+    assert tool_msg["role"] == "tool"
+    assert tool_msg["tool_call_id"] == "call_1"
+
+
+def test_mllm_chat_messages_still_drop_empty_messages_without_tool_calls():
+    messages = [
+        {"role": "user", "content": "hi"},
+        {"role": "assistant", "content": ""},
+        {"role": "assistant", "content": None},
+    ]
+
+    chat_messages = _build_mllm_chat_messages(
+        messages,
+        all_image_urls=[],
+        video_frame_counts={},
+    )
+
+    assert [msg["role"] for msg in chat_messages] == ["user"]
+    # Kept messages carry only the rebuilt role/content keys.
+    assert set(chat_messages[0]) == {"role", "content"}
--- a/tests/test_mllm_message_ordering.py
+++ b/tests/test_mllm_message_ordering.py
@@ -161,9 +161,7 @@ def test_mllm_chat_messages_keep_assistant_tool_calls_without_text():
 
 
 def test_mllm_chat_messages_keep_tool_call_id_on_tool_messages():
-    messages = [
-        {"role": "tool", "tool_call_id": "call_1", "content": '{"temp_c": 19}'}
-    ]
+    messages = [{"role": "tool", "tool_call_id": "call_1", "content": '{"temp_c": 19}'}]
 
     chat_messages = _build_mllm_chat_messages(
         messages,
--- a/vllm_mlx/models/mllm.py
+++ b/vllm_mlx/models/mllm.py
@@ -32,6 +32,7 @@ from urllib.parse import urljoin, urlparse
 import numpy as np
 import requests
 
+from vllm_mlx.engine.chat_template_safety import normalize_messages_for_chat_template
 from vllm_mlx.mllm_cache import MLLMPrefixCacheManager
 
 logger = logging.getLogger(__name__)
@@ -253,6 +254,20 @@ def _build_ordered_mllm_message_content(
     return built_parts, bool(built_parts)
 
 
+def _normalize_mllm_tool_calls(tool_calls: list) -> list:
+    """Normalize replayed assistant tool calls for chat templates.
+
+    Mirrors ``_normalize_tool_call_arguments_for_template`` in
+    ``vllm_mlx/engine/batched.py``: JSON argument strings become mappings so
+    templates that iterate argument keys render correctly.
+    """
+    plain_calls = [_normalize_content_part(call) for call in tool_calls]
+    normalized = normalize_messages_for_chat_template(
+        [{"role": "assistant", "tool_calls": plain_calls}]
+    )
+    return normalized[0].get("tool_calls", plain_calls)
+
+
 def _build_mllm_chat_messages(
     messages: list[dict],
     *,
@@ -272,8 +287,31 @@ def _build_mllm_chat_messages(
             all_image_urls=all_image_urls,
             video_frame_count=video_frame_counts.get(msg_idx, 0),
         )
+        chat_message = {"role": role, "content": content}
+
+        if role == "assistant":
+            tool_calls = msg.get("tool_calls")
+            if isinstance(tool_calls, list) and tool_calls:
+                # Keep tool-call turns even when text content is empty so
+                # templates render the assistant -> tool exchange (issue #608).
+                chat_message["tool_calls"] = _normalize_mllm_tool_calls(tool_calls)
+                reasoning_content = msg.get("reasoning_content")
+                if reasoning_content:
+                    chat_message["reasoning_content"] = reasoning_content
+                has_content = True
+        elif role == "tool":
+            tool_call_id = msg.get("tool_call_id")
+            if tool_call_id:
+                chat_message["tool_call_id"] = tool_call_id
+                # Tools may legitimately return empty output; keep the message
+                # anyway so the assistant tool_call still has its anchor and
+                # template forward-scans pair calls to responses (issue #608).
+                if not has_content:
+                    chat_message["content"] = ""
+                    has_content = True
+
         if has_content:
-            chat_messages.append({"role": role, "content": content})
+            chat_messages.append(chat_message)
     return chat_messages
 
 


waybarrios:
Full test logs from the branch:

<details>
<summary>Regression tests (verbose) — 8 passed</summary>

```
$ pytest tests/test_mllm_message_ordering.py -v
tests/test_mllm_message_ordering.py::test_mllm_chat_messages_preserve_text_only_shape PASSED [ 12%]
tests/test_mllm_message_ordering.py::test_mllm_chat_messages_preserve_image_text_order PASSED [ 25%]
tests/test_mllm_message_ordering.py::test_mllm_chat_messages_preserve_audio_text_order PASSED [ 37%]
tests/test_mllm_message_ordering.py::test_mllm_chat_messages_keep_assistant_tool_calls_without_text PASSED [ 50%]
tests/test_mllm_message_ordering.py::test_mllm_chat_messages_keep_tool_call_id_on_tool_messages PASSED [ 62%]
tests/test_mllm_message_ordering.py::test_mllm_chat_messages_keep_empty_tool_results PASSED [ 75%]
tests/test_mllm_message_ordering.py::test_mllm_chat_messages_keep_tool_exchange PASSED [ 87%]
tests/test_mllm_message_ordering.py::test_mllm_chat_messages_still_drop_empty_messages_without_tool_calls PASSED [100%]
============================== 8 passed in 0.25s ===============================
```

</details>

<details>
<summary>MLLM + server suites — 250 passed</summary>

```
$ pytest tests/test_mllm_cache.py tests/test_mllm_continuous_batching.py \
    tests/test_mllm_message_ordering.py tests/test_mllm_mtp_routing.py \
    tests/test_mllm.py tests/test_server.py -q
====================== 250 passed, 13 deselected in 5.18s ======================
```

</details>

<details>
<summary>Full fast suite — 2185 passed, 0 failed</summary>

```
$ pytest tests/ -q -k "not slow"
=============== 2185 passed, 11 skipped, 23 deselected in 30.95s ===============

$ ruff check vllm_mlx/models/mllm.py tests/test_mllm_message_ordering.py
All checks passed!
```

</details>

<details>
<summary>Live repro verification (Qwen3-VL-8B-Instruct-4bit, M4 Max)</summary>

```
$ vllm-mlx serve mlx-community/Qwen3-VL-8B-Instruct-4bit --mllm \
    --enable-auto-tool-choice --tool-call-parser hermes --port 8077

# Probe A — fresh turn (system + user + 1 tool), temperature 0
prompt_tokens: 166
tool_calls: [{"id": "call_0a25c249", "type": "function",
              "function": {"name": "get_weather", "arguments": "{\"city\": \"Paris\"}"}}]
content: ''

# Probe B — continuation (+ assistant{content:null, tool_calls} + tool{result})
prompt_tokens: 213        # > 166: history now contributes tokens
tool_calls: null          # no repeated call — loop broken
content: 'The weather in Paris is currently 19°C with sunny conditions.'
```

On main, probe B renders byte-identical to probe A (166 tokens) and re-issues the same tool call.

</details>

eejd:
## Independent corroboration from a production fork

We hit the same bug documented in #608 and arrived at the same fix independently in our fork (`eejd/vllm-mlx`). We've been running it in production against Gemma-4 and Qwen3-VL MLLM models since our `v0.4.0rc1.post1` tag. The root cause, repro, and approach in this PR are exactly right — **+1 to merging**.

A few notes from our implementation experience that may be useful:

---

### One thing your fix has that ours is missing

Your use of `normalize_messages_for_chat_template` to parse JSON argument strings into mappings before passing them through is something we omitted on the native-template path. Our `_build_mllm_chat_messages` passes `tool_calls` through verbatim, which means templates that *iterate* argument keys (rather than just forwarding the raw string) will break on replayed turns. We plan to adopt your approach here. Concrete validation that the normalization step matters.

---

### Two additions worth considering as follow-ups (not blocking)

**1. Non-tool-aware templates (Gemma family)**

After this PR, models whose chat templates don't natively render `tool_calls` / `role:"tool"` (e.g. Gemma-4, which has only `user`/`model`/`system`) will silently drop the now-preserved metadata — the template engine just can't render it. `SUPPORTS_NATIVE_TOOL_FORMAT` is one signal (your suggestion 3); we went a different direction:

- A runtime probe (`_template_supports_tool_role`) checks the template string for `"tool_calls"` / `"'tool'"` / `'"tool"'` keywords, falling back to a `model_type` heuristic.
- When the probe returns False, a serializer (`_serialize_tool_messages_gemma_native`) rewrites the exchange *before* message collection: assistant `tool_calls` turns become ` ```tool_code\nfn(args)\n``` ` blocks in an `assistant` turn, and `role:"tool"` results become ` ```tool_output\n...\n``` ` blocks in a `user` turn — the inverse of what the Gemma4 output parser reads back out.
- This is applied in both `chat()` and `stream_chat()` before the rest of the MLLM message-build runs.

Happy to contribute this as a stacked follow-up PR on top of #611 if it's in scope.

**2. Broader replay metadata**

Your fix forwards `tool_calls`, `tool_call_id`, and `reasoning_content`. We also found it useful to carry `tool_responses`, `name`, and plain `reasoning` for templates/flows that use them. Minor addition but worth capturing somewhere.

---

### Separate adjacent gap (unrelated to #611, worth a separate PR)

`chat_template_kwargs` (including `enable_thinking`) is silently dropped on the MLLM path vs the text path in `llm.py`. This means `--default-chat-template-kwargs '{"enable_thinking": false}'` has no effect on MLLM models, so reasoning content leaks into responses when users expect it suppressed. Not a tool-call issue but closely adjacent if you're doing a housekeeping sweep of the MLLM path.

---

Great fix, clean test coverage. Thanks for digging into this.

eejd:
We have a related fix branch (`fix/mllm-tool-roundtrip`) in our fork that addresses the same gap in `_build_mllm_chat_messages`. Comparing our diff against this PR:

**This PR (#611):** ~40-line surgical fix — adds `_normalize_mllm_tool_calls()` using `normalize_messages_for_chat_template`, preserves `tool_calls` on assistant turns and `tool_call_id` + empty-string content on `role: tool` turns.

**Our branch:** broader — copies all known metadata keys (`tool_calls`, `tool_call_id`, `tool_responses`, `name`, `reasoning_content`) in a generic loop, plus adds Gemma-specific argument-rendering helpers (`_render_gemma_tool_call_args`, `_coerce_tool_call`) for the native-template path with mlx-vlm ≥ 0.6.3.

The two approaches conflict on the same lines in `_build_mllm_chat_messages`, so we're not opening a competing PR. We'll drop our branch once this merges.

One edge case worth noting from our testing: when the native Gemma 4 chat template is active (mlx-vlm ≥ 0.6.3, `_template_supports_tool_role()` → True), the template iterates tool-call argument keys directly — so `tool_calls` entries need their `arguments` field as a mapping (not a JSON string) at template-render time. Your `_normalize_mllm_tool_calls()` via `normalize_messages_for_chat_template` should handle this; just flagging it as a test case if you want to add coverage.

janhilgard (approved):
Approving. This is the right fix for a high-impact silent bug — the symptom (agent clients looping on the same tool call forever) is exactly the failure mode that doesn't surface in unit-test grids but burns expensive token budgets in production.

**Verified against upstream main on a fresh local checkout:**

- Current `_build_mllm_chat_messages` on `main` is the literal three lines the issue describes:

  ```python
  if has_content:
      chat_messages.append({"role": role, "content": content})
  ```

  No tool-calls forwarding, no `tool_call_id` preservation, no anchor for the tool response. An OpenAI `{role: "assistant", content: null, tool_calls: [...]}` message produces `has_content=False` (because `content=null` reduces to no built parts) and is **dropped entirely** before the chat template ever sees it. The following `{role: "tool", tool_call_id: "...", content: "..."}` then loses its anchor — Gemma 4's template forward-scans from the assistant `tool_calls` to pair the tool response, so without the anchor it renders nothing. The rendered continuation is byte-identical to the fresh prompt, the model deterministically re-issues the same tool call. Exactly the reproduction the issue describes.

- `normalize_messages_for_chat_template` in `vllm_mlx/engine/chat_template_safety.py` already does the right two transformations: closes dangling `<think>` before raw `<tool_call>` XML (Qwen 3.6 history repair, mirrors the Cheuk-Yiu Chan Jinja-template fix), and converts OpenAI tool-call `arguments` JSON-strings to mappings for templates that iterate argument keys. The BatchedEngine path in `engine/batched.py` already wraps history through this via `_normalize_tool_call_arguments_for_template`. PR #611 plugs the MLLM path into the same helper via `_normalize_mllm_tool_calls`. **Both routes now render identically** — that consistency is more valuable than the fix in isolation, because it eliminates the "works on the text path, fails on the MLLM path" debugging trap.

- `_normalize_content_part` (the Pydantic → dict helper used inside `_normalize_mllm_tool_calls`) already exists in `models/mllm.py` upstream. No new helper machinery introduced — the PR is reusing the existing patterns end-to-end. Good.

- The drop rule for genuinely-empty messages without `tool_calls` is preserved (`test_mllm_chat_messages_still_drop_empty_messages_without_tool_calls`). That's the right scope — if a user sends an empty assistant turn with no tool metadata, dropping it is the prior behavior and there's no reason to change it.

- `reasoning_content` forwarding on tool-call turns is a quiet but real win: thinking-aware models (Qwen 3.6, Gemma 4) carry the `<think>` block as `reasoning_content` on the assistant message, and templates that consume it (e.g. for verifier-style follow-up reasoning) would otherwise see a stripped turn.

**Code-review notes:**

1. The `_normalize_mllm_tool_calls` helper wraps the calls in a synthetic `{"role": "assistant", "tool_calls": [...]}` message just to run them through `normalize_messages_for_chat_template`, then unwraps. That works because `normalize_messages_for_chat_template` operates on a list of messages, but the wrap/unwrap dance is a hint that the underlying API would be cleaner as a `normalize_tool_calls(calls)` primitive. Not blocking — the chat-template-safety helper has only one consumer for tool calls today (BatchedEngine) and one new consumer (this PR), so factoring the inner loop out is premature. Worth a docstring line on `_normalize_mllm_tool_calls` explaining why the wrap exists, so the next reader doesn't try to "simplify" by calling some lower-level helper that doesn't yet exist.

2. The `chat_message["content"] = ""` assignment for empty tool results (the `if not has_content:` branch) flips the value from whatever `_build_ordered_mllm_message_content` returned (could be `[]`, could be `None`) to the empty string. That's the right thing for templates that do `{% if message.content %}` checks — they'll see truthy presence but no payload — but worth noting it changes the *shape* of the content from "list of parts" to "string" for that specific case. Templates that strictly type-check content parts would break, though I can't think of any in the supported model set that do. Test coverage in `test_mllm_chat_messages_keep_empty_tool_results` pins the shape explicitly (`{"role": "tool", "content": "", "tool_call_id": "call_1"}`), so any regression here would be caught.

3. The empty-string content fallback is *only* applied for `role:"tool"` messages. Assistant `tool_calls` messages with empty text content get `chat_message["content"] = content` (whatever the build helper returned, often `[]` or `""`). That's intentional — assistant turns are not the template-anchor, the `tool_calls` field is what templates iterate — but worth a comment in the code so the next reader understands why the two branches treat empty content differently.

4. eejd's two comments on the PR are useful corroboration: an independent fork hit the same bug, arrived at the same shape (with some additions for the Gemma native-template path that handles tool serialization differently). The note about the Gemma 4 native chat template (mlx-vlm ≥ 0.6.3) iterating tool-call argument keys directly — confirming that `arguments` MUST be a mapping rather than a JSON string at render time — is the practical validation that `_normalize_mllm_tool_calls` is doing real work, not just defensive prettification. Worth keeping the eejd comment in mind for the next PR that touches the Gemma path.

5. Test design is honest: 5 new tests, **3 of which fail on `main` for the exact bug reasons** per the PR body. That's the gold-standard test-first approach — the test suite explicitly proves the absence of the fix produces the observed bug, and the presence of the fix removes it. The fourth test (`test_mllm_chat_messages_keep_tool_exchange`) is particularly nice because it asserts at the **emergent behavior** level: a 4-message continuation must render strictly more content than the 3-message fresh exchange. That's the exact assertion that would have caught the original bug in code review without needing to read the rebuild logic line-by-line.

6. Scope discipline: `SUPPORTS_NATIVE_TOOL_FORMAT` on `Gemma4ToolParser` is explicitly out of scope (issue suggestion 3). Good call — that's a different surface (parser-side identification of native vs. plaintext tool format) and bundling it would make this PR harder to review and revert if needed.

7. CI: lint ✓, type-check ✓, test-matrix × 4 ✓, test-apple-silicon × 2 ✓, tests ✓. All 9 green on `875404fd`. Author's own logs: `pytest tests/ -q -k "not slow" → 2185 passed, 0 failed`.

8. Live verification with `mlx-community/Qwen3-VL-8B-Instruct-4bit` + Hermes parser is the right shape: probe A (fresh) → 166 prompt_tokens + tool_call; probe B (continuation with assistant tool_calls + tool result) → 213 prompt_tokens + actual integrated answer ("The weather in Paris is currently 19°C with sunny conditions."). The `213 > 166` assertion is the key one — it proves the history is now being *rendered* into the prompt, not silently elided. On `main`, probe B renders identically to probe A and re-issues the same tool call ad infinitum.

- I'm running three MLLM servers (Qwen 3.6-27B Heretic-v2 + MTP on Hermes-style tool parser, Qwen 3.6-35B-A3B, and a recently-retired Gemma 4 12B). All of them route assistant messages through `_build_mllm_chat_messages`. Tool-using agent flows on those servers would have exhibited exactly this bug — every continuation turn after a tool_call would have been a re-issuance — and I've been chalking up "agent went in circles" failure modes to model size limits when one of the active causes was the chat-template rebuild dropping the OpenAI-canonical assistant tool-call shape. This fix is going to materially improve agent throughput for my deploys.

- The eejd comment specifically flags that mlx-vlm ≥ 0.6.3 (which I'm on at 0.6.2 — one minor behind for unrelated reasons) iterates tool-call argument keys natively in the Gemma chat template. When I move to 0.6.3, `_normalize_mllm_tool_calls` will be doing real conversion work on every tool-call render. Without this PR, the symptom would have been "Gemma tool calls work for one round and then the template breaks because args are still JSON strings." Glad this is being closed before that hits production.

LGTM — please merge. This is the kind of small surgical fix that pays for itself in the first day after release.

PR #612, opened 2026-06-12
fix(ssd-cache): preserve original bfloat16 dtype across quantized spill

CBribiescas:
## Summary

Follow-up to #605 per @waybarrios's review comment: the dequant-on-spill path casts bf16 → fp16 (to dodge numpy's PEP 3118 buffer-protocol mismatch) but didn't record the original dtype. Reload then hands the model fp16 KV where it had computed bf16 — silent precision regression on every SSD reload of a quantized-cache model.

## Root cause

`_mx_to_numpy_safe` only stamps a dtype hint when *it* did the upcast (bf16 → fp32 fallback). The quantized-spill path pre-casts bf16 → fp16 before `snapshot_layer` is called, and since fp16 IS numpy-supported, `_mx_to_numpy_safe` returns `dtype=None`. The snapshot then carries no signal that the data originated as bf16, and `scheduler._reconstruct_ssd_layers` has nothing to cast back from.

## Fix

Mirror the existing plain-cache pattern: stash an explicit dtype sentinel on the layer before the cast, and have `KVCacheSerializer.snapshot_layer` honor it with priority over the autodetect path:

```python
# enqueue_spill (dequant branch):
if str(getattr(k, "dtype", "")).endswith("bfloat16"):
    layer._ssd_keys_original_dtype = "bfloat16"
    layer._ssd_values_original_dtype = "bfloat16"
    layer.keys = k.astype(mx.float16)
    layer.values = v.astype(mx.float16)

# KVCacheSerializer.snapshot_layer:
keys_orig_dtype = (
    getattr(layer, "_ssd_keys_original_dtype", None) or keys_orig_dtype
)
```

Sentinel name is `_ssd_`-prefixed to avoid colliding with any mlx-lm-side cache attribute.

Also add `"QuantizedKVCache": "supported_via_dequant_on_spill"` to `SERIALIZER_SUPPORT_MATRIX` so the diagnostics table matches what `enqueue_spill` actually handles.

## Reload side

No change. `scheduler.py:3019-3028` already reads `keys_original_dtype` from the deserialized dict and casts back via `mx.array(...).astype(mx.bfloat16)`. Only the spill-side signal was missing.

## Tests

Four regression tests in `TestLayerSerializer`:

- `test_support_matrix_includes_quantized_kv_cache` — guards the matrix entry against drift.
- `test_snapshot_layer_honors_original_dtype_sentinel` — sentinel set, no autodetect signal: must record `"bfloat16"`.
- `test_snapshot_layer_sentinel_overrides_autodetect` — sentinel wins even when `_mx_to_numpy_safe` also produces a hint.
- `test_snapshot_layer_no_sentinel_falls_back_to_autodetect` — plain fp16 layers still record no dtype hint (no regression on the existing path).

## Verification

```
pytest tests/test_ssd_cache.py
63 passed in 3.09s

black --check vllm_mlx/ssd_cache.py tests/test_ssd_cache.py
2 files would be left unchanged
```

Refs #605 #563.

### DIFF
--- a/tests/test_ssd_cache.py
+++ b/tests/test_ssd_cache.py
@@ -381,6 +381,88 @@ class TestLayerSerializer:
         with pytest.raises(ValueError, match="Unsupported"):
             get_serializer_for_layer("not a cache layer")
 
+    def test_support_matrix_includes_quantized_kv_cache(self):
+        """SUPPORT_MATRIX must surface both quantized cache types so the
+        diagnostics table reflects what enqueue_spill's dequant path handles."""
+        assert (
+            SERIALIZER_SUPPORT_MATRIX["QuantizedKVCache"]
+            == "supported_via_dequant_on_spill"
+        )
+        assert (
+            SERIALIZER_SUPPORT_MATRIX["_QuantizedCacheWrapper"]
+            == "supported_via_dequant_on_spill"
+        )
+
+    def test_snapshot_layer_honors_original_dtype_sentinel(self):
+        """When enqueue_spill's dequant path stashed a pre-cast dtype on the
+        layer, snapshot_layer must surface it in the snapshot — without this,
+        the reload path leaves the model holding fp16 KV where it computed
+        bf16. Regression for the dtype-round-trip gap @waybarrios flagged on
+        PR #605 (the fp16-cast layers carried no dtype hint because
+        _mx_to_numpy_safe only sets a hint on its own upcast path)."""
+        keys = MockMLXArray(np.random.randn(1, 8, 4, 16).astype(np.float16))
+        values = MockMLXArray(np.random.randn(1, 8, 4, 16).astype(np.float16))
+        layer = MockKVCacheLayer(keys=keys, values=values, offset=4)
+        layer._ssd_keys_original_dtype = "bfloat16"
+        layer._ssd_values_original_dtype = "bfloat16"
+
+        snapshot = KVCacheSerializer().snapshot_layer(layer)
+
+        assert snapshot["keys_original_dtype"] == "bfloat16"
+        assert snapshot["values_original_dtype"] == "bfloat16"
+
+    def test_snapshot_layer_sentinel_overrides_autodetect(self):
+        """When both the sentinel and the _mx_to_numpy_safe autodetect produce
+        a dtype, the explicit sentinel wins. This pins down the precedence so
+        a future cast-chain (e.g. bf16 → fp16 → bf16 round-trip pre-snapshot)
+        can't silently lose the original dtype."""
+
+        class _BF16RaisingArray(MockMLXArray):
+            def __init__(self, data):
+                super().__init__(data)
+                # Simulate the bf16 buffer-protocol RuntimeError so
+                # _mx_to_numpy_safe reports "bfloat16" via autodetect; the
+                # sentinel below should still win.
+                self.dtype = type(
+                    "dtype",
+                    (),
+                    {"size": 2, "__str__": lambda self: "mlx.core.bfloat16"},
+                )()
+
+            def __array__(self):
+                raise RuntimeError("buffer format string mismatch")
+
+            def astype(self, _dt):
+                return MockMLXArray(self._data)
+
+        keys = _BF16RaisingArray(np.random.randn(1, 4, 2, 8).astype(np.float32))
+        values = _BF16RaisingArray(np.random.randn(1, 4, 2, 8).astype(np.float32))
+        layer = MockKVCacheLayer(keys=keys, values=values, offset=2)
+        layer._ssd_keys_original_dtype = "bfloat16"
+        layer._ssd_values_original_dtype = "bfloat16"
+
+        snapshot = KVCacheSerializer().snapshot_layer(layer)
+
+        # Sentinel must win over the upcast-recovered hint either way; both
+        # paths point to bfloat16 here, but the assertion guards that the
+        # sentinel branch is the one actually consulted.
+        assert snapshot["keys_original_dtype"] == "bfloat16"
+        assert snapshot["values_original_dtype"] == "bfloat16"
+
+    def test_snapshot_layer_no_sentinel_falls_back_to_autodetect(self):
+        """Without a sentinel, snapshot_layer must keep the prior behavior:
+        plain fp16/fp32 layers record no dtype hint (None), bf16 layers get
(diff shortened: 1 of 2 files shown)


janhilgard (approved):
Approving. Tight, well-scoped follow-up to #605 that closes a real precision-loss path without touching the reload-side, which already does the right thing.

**Verified by reading upstream main directly:**

- `scheduler._reconstruct_ssd_layers` (the place the PR body cites) does already read `keys_original_dtype`/`values_original_dtype` from the deserialized dict and cast back via `mx.array(np_payload).astype(dt)`. That branch landed with #605 for the plain-cache bf16-via-fp32-upcast path:

  ```python
  keys_orig = ld.get("keys_original_dtype")
  if keys_orig is not None:
      dt = _mx_dtype_from_name(keys_orig)
      if dt is not None:
          kv.keys = kv.keys.astype(dt)
  ```

  So the reload side is already wired — this PR is purely about ensuring the spill side actually puts that signal in the dict for the quantized-spill code path. No reload-side change required, which is exactly what the PR body claims.

- `_mx_to_numpy_safe` returns `(np.ndarray, original_dtype_name_or_None)`, where the `_name_or_None` is set **only when an upcast happened inside `_mx_to_numpy_safe`** (bf16 → fp32 fallback after the RuntimeError catch). The quantized-spill path in `enqueue_spill` pre-casts bf16 → fp16 *before* calling `snapshot_layer`, and since `np.array(fp16_mlx_array)` succeeds without raising, `_mx_to_numpy_safe` returns `dtype=None`. Bug shape exactly as described — the autodetect path can't see what happened upstream of it.

- Sentinel placement on the **layer** (not the cache or snapshot dict) is the right seam:
  - `enqueue_spill` already mutates the layer (`layer.keys = k.astype(mx.float16)`), so adding two more attribute assignments on the same object is consistent.
  - `snapshot_layer` receives the layer first thing, so reading the sentinel adds zero indirection.
  - The `_ssd_`-prefix on the attribute name avoids colliding with mlx-lm-side KVCache attributes (mlx-lm uses unprefixed names like `keys`, `values`, `offset`, `max_size`, `keep`, `step`, `_idx`). Safe namespace.

- `getattr(layer, "_ssd_keys_original_dtype", None) or keys_orig_dtype` precedence — sentinel wins over autodetect when both exist. That's the right call: in the quantized-spill path the autodetect can only ever produce `None` (fp16 is numpy-readable), but if a future refactor changes the cast pre-snapshot to land in some other dtype, the sentinel should still record what the model originally computed in. The `test_snapshot_layer_sentinel_overrides_autodetect` test pins exactly this precedence.

- `SERIALIZER_SUPPORT_MATRIX["QuantizedKVCache"] = "supported_via_dequant_on_spill"` matches the existing `_QuantizedCacheWrapper` entry the matrix already had. The PR is adding the **native** mlx-lm name alongside the wrapper name — both code paths flow through the same dequant-on-spill handler in `enqueue_spill`, so both should advertise the same support state in diagnostics. Catches the case where someone has a native `QuantizedKVCache` (no wrapper) and the diagnostics table previously implied "unknown / not_supported". Good catch.

**Code-review notes:**

1. Test `test_snapshot_layer_sentinel_overrides_autodetect` builds a `_BF16RaisingArray` that raises `RuntimeError("buffer format string mismatch")` on `__array__()`, forcing `_mx_to_numpy_safe` down its fallback branch where it stamps `dtype="bfloat16"` itself. Then it asserts the sentinel `"bfloat16"` is what shows up in the snapshot. As written, both branches converge on the same string, so the assertion only proves "snapshot dtype is bfloat16", not "sentinel won over autodetect". A stricter pin would have the sentinel set to a different value (`"bf16_via_sentinel"` or similar) and the autodetect to `"bfloat16"`, then assert the sentinel value is what came through. Cosmetic — both paths produce the same answer in practice — but the test as-is would still pass if `getattr(..., None) or keys_orig_dtype` were silently reordered to `keys_orig_dtype or getattr(...)`.

2. The `bool(string) → True` part of `getattr(layer, "_ssd_keys_original_dtype", None) or keys_orig_dtype` works for any truthy string but would fall through to `keys_orig_dtype` on `""`. Sentinel should never be set to empty, but `if getattr(layer, "_ssd_keys_original_dtype", None) is not None else keys_orig_dtype` is the strictly-safer form and avoids the empty-string footgun. Stylistic.

3. The dequant branch sets two sentinels (keys + values) but in practice they will always be the same string (`"bfloat16"`). Worth a one-line comment explaining why both fields exist anyway — future enhancements may legitimately split (e.g. a half-quantized cache where keys and values stored in different precisions), and having two separate fields is forward-compatible. Saves a future contributor from "simplifying" them down to one shared sentinel.

4. The `test_snapshot_layer_no_sentinel_falls_back_to_autodetect` test asserts `"keys_original_dtype" not in snapshot` rather than `snapshot.get("keys_original_dtype") is None`. The current `snapshot_layer` only stamps the key when the value is non-`None` (`if keys_orig_dtype is not None: snapshot["keys_original_dtype"] = ...`), so the `not in` form is correct, but it's tightly coupled to that exact emit pattern. If anyone ever changes `snapshot_layer` to unconditionally write `keys_original_dtype` (even as `None`), this test breaks. Worth either using `.get(...)` for resilience or adding a comment explaining the contract.

5. CI: lint ✓, type-check ✓, test-matrix × 4 ✓, test-apple-silicon × 2 ✓, tests ✓. All 9 green on `12370054`.

6. The PR fits the rule of thumb that **every silent-correctness fix should ship with a unit test that fails without the fix and passes with it**. The four new tests collectively pin: matrix entry presence (so the diagnostics-table fix doesn't regress), sentinel-set-no-autodetect (the happy path), sentinel-wins-over-autodetect (the precedence contract), and no-sentinel-keeps-old-behavior (no regression on the plain-cache path). That's complete coverage for the four code paths the PR touches.

- I'm currently running KV-cache quantization (`--kv-cache-quantization` + `--kv-cache-quantization-bits 8`) on all three production servers (Qwen 3.6-27B, 3.6-35B-A3B). My SSD-spill setup hasn't been exercised hard enough to surface this silently, but the precision regression #612 closes is exactly the kind that doesn't fail loudly — outputs would just be slightly off in the second decimal of logit magnitudes after a reload, with no error in any log. Glad this is being chased down before someone has to bisect a model-quality regression back to "we restarted last Tuesday".

LGTM — please merge.

PR #615, opened 2026-06-13
fix(engine): preserve SimpleEngine wait admission

Thump604:
## Summary

- Preserve `VLLM_MLX_SIMPLE_ENGINE_LOCK_ADMISSION=wait` after `SimpleEngine` validates the env value.
- Add a regression test that checks both the internal admission mode and `get_stats()` reporting.

## Root cause

`SimpleEngine.__init__` read and validated `VLLM_MLX_SIMPLE_ENGINE_LOCK_ADMISSION`, but then unconditionally reassigned `_generation_lock_admission = "fail_fast"` immediately after the invalid-value warning block. That made the documented `wait` opt-in unreachable.

## Repro

Before the fix, this focused regression failed because the engine reported `fail_fast` even with the env set to `wait`:

```sh
VLLM_MLX_SIMPLE_ENGINE_LOCK_ADMISSION=wait pytest tests/test_simple_engine.py::TestSimpleEngineConcurrency::test_lock_admission_env_wait_is_preserved -q
```

## Expected behavior

`VLLM_MLX_SIMPLE_ENGINE_LOCK_ADMISSION=wait` should preserve queued admission behavior, while invalid values should still fall back to `fail_fast`.

## Validation

```sh
uv run --python 3.12 --extra dev --extra vision python -m pytest tests/test_simple_engine.py -q
uv run --python 3.12 --with ruff ruff check vllm_mlx/engine/simple.py tests/test_simple_engine.py --select E,F,W --ignore E402,E501,E731,F811,F841
uv run --python 3.12 --with black black --check vllm_mlx/engine/simple.py tests/test_simple_engine.py
git diff --check upstream/main..HEAD
```

## Not claimed

This does not address the separate streaming pre-admission probe question from the #540 follow-up comment. This PR only restores the existing `wait` env contract.

### DIFF
--- a/tests/test_simple_engine.py
+++ b/tests/test_simple_engine.py
@@ -186,6 +186,18 @@ class TestSimpleEngineConcurrency:
             release.set()
             await first
 
+    def test_lock_admission_env_wait_is_preserved(self, monkeypatch):
+        """VLLM_MLX_SIMPLE_ENGINE_LOCK_ADMISSION=wait keeps queued behavior."""
+        from vllm_mlx.engine.simple import SimpleEngine
+
+        monkeypatch.setenv("VLLM_MLX_SIMPLE_ENGINE_LOCK_ADMISSION", "wait")
+
+        with patch("vllm_mlx.engine.simple.is_mllm_model", return_value=False):
+            engine = SimpleEngine("test-model")
+
+        assert engine._generation_lock_admission == "wait"
+        assert engine.get_stats()["generation_lock"]["admission"] == "wait"
+
     @pytest.mark.anyio
     async def test_blocking_serialized_tracks_active_request(self):
         """Busy errors include the current blocking serialized holder."""
--- a/vllm_mlx/engine/simple.py
+++ b/vllm_mlx/engine/simple.py
@@ -225,7 +225,7 @@ class SimpleEngine(BaseEngine):
                 "Invalid VLLM_MLX_SIMPLE_ENGINE_LOCK_ADMISSION=%r; using fail_fast",
                 self._generation_lock_admission,
             )
-        self._generation_lock_admission = "fail_fast"
+            self._generation_lock_admission = "fail_fast"
         self._generation_waiters = 0
         self._generation_busy_rejections = 0
 


janhilgard (approved):
Approving. Surgical one-line dedent fix for a clear logic-bug regression introduced in #540, backed by a focused test that pins both the internal state and the externally observable `/v1/status` shape.

**Verified by reading the diff:**

- The bug is exactly the kind that survives a clean code review because the offending statement is syntactically valid in either position. Before #615, the validation block read

  ```python
  if value not in {"fail_fast", "wait"}:
      logger.warning("Invalid VLLM_MLX_SIMPLE_ENGINE_LOCK_ADMISSION=%r; using fail_fast", value)
  self._generation_lock_admission = "fail_fast"   # always executed
  ```

  so the documented `wait` opt-in path was unreachable for everyone, not just users with a typo. The fix moves the assignment one column to the right so it lives inside the `if invalid:` branch, which is the only place that should be deciding the fallback. Reading the diff that's literally the entire change to `simple.py`.

- The regression test is well-scoped:
  - `monkeypatch.setenv("VLLM_MLX_SIMPLE_ENGINE_LOCK_ADMISSION", "wait")` exercises the exact env contract.
  - `patch("vllm_mlx.engine.simple.is_mllm_model", return_value=False)` keeps it a pure unit test (no model load), so it'll run cheaply on every CI job.
  - Asserting both `engine._generation_lock_admission == "wait"` and `engine.get_stats()["generation_lock"]["admission"] == "wait"` pins the internal state *and* the externally observable contract — the latter is the one operators will key dashboards on, so it's the right thing to assert.

- Behavior for invalid env values is unchanged (still warns + falls back to `fail_fast`). The existing `wait` callers — anyone who set the env to opt into queued admission — were silently getting `fail_fast` since #540 landed; this restores the documented behavior.

- All 9 CI checks green on `8a9c351d` (lint, type-check, test-matrix × 4, test-apple-silicon × 2, tests).

**Non-blocking notes:**

1. The original `if invalid:` branch in #540 should probably have asserted against the resulting `_generation_lock_admission` value at PR time; the absence of that assertion is what let the dedent typo slip through. The new test in this PR closes that hole for `wait`; a symmetric one-liner for the invalid-value path (e.g. `"banana"` → `_generation_lock_admission == "fail_fast"` and a `caplog` assertion on the warning) would make future regressions on either branch obvious in CI. Small follow-up, not blocking.

2. As a side observation while reading this: PR #540 introduced this surface only a few days ago (merged 2026-06-09) and PR #615 is correcting it now. Worth a one-line note in the next release notes so anyone who landed in those few days knowing they wanted `wait` realises they were silently getting `fail_fast`.

LGTM — ship it.

PR #622, opened 2026-06-21
fix(gemma4): parse paren and tool_code fallback tool-call forms

eejd:
## Problem

Under long system prompts, multi-turn conversations, or when several tools are
registered, Gemma 4 frequently abandons its canonical brace form:

```
<|tool_call>call:fn{key:<|"|>val<|"|>}<tool_call|>
```

and instead emits the call as plain content using Python-style syntax. Two forms
are seen in practice:

**e4b — parenthesized with `call:` prefix:**
```
<|tool_call>call:get_weather(city="Tokyo", unit="celsius")
```

**e2b — fenced `tool_code` block:**
````
```tool_code
get_weather(city="Tokyo", unit="celsius")
```
````

When the canonical parser finds neither `<|tool_call>` nor a parseable brace
block it returns `tools_called=False`, causing the host to treat the output as
plain text and never dispatch the tool.

## Fix

Adds a fallback layer in `Gemma4ToolParser.extract_tool_calls` that fires when
the canonical parse yields no calls. The fallback handles:

- **`call:fn(...)`** — parenthesized call with required `call:` prefix (prevents
  ordinary prose like `"see foo()"` from being mistaken for a tool call).
- **``` ```tool_code``` blocks** — one or more bare `fn(...)` call statements
  inside a fenced block; uses `ast.parse` in exec mode so multiple calls in one
  block are all dispatched.

Each form is detected with a narrow regex, then the call expression is parsed via
`ast.literal_eval` on every argument value — no `eval`, no exec.

Matching spans are stripped from the returned `content` so the host doesn't echo
the raw call text back to the client.

The streaming path gains a cheap `_FALLBACK_MARKER_RE` guard so it doesn't stall
waiting for `<tool_call|>` on the fallback forms.

## Tests

`tests/test_gemma4_tool_parser.py` — covers both fallback forms plus the
canonical brace form, multi-call blocks, and streaming path.

Live-tested against `mlx-community/gemma-4-e2b-it-4bit` and
`mlx-community/gemma-4-e4b-it-4bit` via an OpenAI-compatible tool-call round-trip
(`tool_choice: required` → tool result → natural-language summary).

Ref: #80

### DIFF
--- a/tests/test_gemma4_tool_parser.py
+++ b/tests/test_gemma4_tool_parser.py
@@ -205,6 +205,99 @@ class TestGemma4ToolParserExtract:
         assert args == {"tags": ["alpha", "beta", "gamma"]}
 
 
+class TestGemma4ToolParserFallbackForms:
+    """Fallback Python-style forms Gemma 4 emits under load (issue #80).
+
+    Under a long system prompt + multi-turn + several tools, Gemma 4 stops using
+    the canonical <|"|>-brace form and leaks its call as plain content using
+    Python call syntax. These must still parse into tool_calls.
+    """
+
+    def setup_method(self):
+        self.parser = Gemma4ToolParser()
+
+    def test_e4b_paren_form_with_tool_call_marker(self):
+        """e4b: <|tool_call>call:fn(kwargs) — parens instead of braces, no end."""
+        output = '<|tool_call>call:radarr_get_movies(search="Dune")'
+        result = self.parser.extract_tool_calls(output)
+        assert result.tools_called is True
+        assert len(result.tool_calls) == 1
+        tc = result.tool_calls[0]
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Thump604** [comment]
Quick CI triage: this appears blocked only on Black formatting. Ruff, type-check, all Python matrix tests, Apple Silicon tests, and the aggregate tests job are green.

The lint log says Black would reformat:

```sh
tests/test_gemma4_tool_parser.py
vllm_mlx/tool_parsers/gemma4_tool_parser.py
```

Please run:

```sh
black tests/test_gemma4_tool_parser.py vllm_mlx/tool_parsers/gemma4_tool_parser.py
```

After that push, this should be ready for review.

Issue #83, opened 2026-02-13
TypeError in `extract_multimodal_content` when processing messages with tool_calls


wep4you:
# [Bug] TypeError in `extract_multimodal_content` when processing messages with tool_calls

## Description

When sending chat completion requests that include `tool` role messages in the conversation history (e.g. from agentic coding tools like OpenClaw/Claude Code), vllm-mlx crashes with a `TypeError` in `extract_multimodal_content()`. Every request containing tool call history fails with a 500 Internal Server Error.

## Environment

- **vllm-mlx**: installed via `uv tool install`
- **Model**: `mlx-community/Qwen2.5-Coder-14B-Instruct-4bit`
- **Hardware**: Mac Mini M4 Pro, 24GB
- **macOS**: Sequoia
- **Client**: OpenClaw (sends OpenAI-compatible chat completions with tool calling)
- **Version**: vllm-mlx v0.2.6

## Server command

```bash
vllm-mlx serve mlx-community/Qwen2.5-Coder-14B-Instruct-4bit \
  --port 1234 \
  --continuous-batching \
  --use-paged-cache \
  --stream-interval 1 \
  --api-key "token" \
  --max-tokens 32768
```

## Error

```
File "/…/vllm_mlx/api/utils.py", line 227, in extract_multimodal_content
    text = (text + "\n" if text else "") + "\n".join(tool_calls_text)
            ~~~~~^~~~~~
TypeError: can only concatenate list (not "str") to list
```

## Request details

The request includes a conversation with `tool` roles — typical for agentic tool-calling workflows:

```
roles=['system', 'user', 'user', 'assistant', 'user', 'assistant', 'tool',
       'assistant', 'user', 'assistant', 'tool', 'assistant', 'tool',
       'assistant', 'tool', 'assistant', 'user']
tools=23
total_chars=70448
```

## Root cause

In `vllm_mlx/api/utils.py` line 227, the code assumes `text` is a `str`, but when processing messages that contain `tool_calls`, `text` is actually a `list` of `ContentPart` objects. The string concatenation then fails because Python cannot concatenate a list with a string.

## Fix

Add a type check to convert `text` from list to string before concatenation:

```python
# Before (line 227):
text = (text + "\n" if text else "") + "\n".join(tool_calls_text)

# After:
if isinstance(text, list):
    text = "\n".join(getattr(t, 'text', str(t)) for t in text)
text = (text + "\n" if text else "") + "\n".join(tool_calls_text)
```

[...]

PR #623, opened 2026-06-21
fix(gemma4): parse unfenced tool_code = fn(...) assignment form

eejd:
## Problem

The e2b variant of Gemma 4 occasionally emits a tool call in an unfenced
assignment form rather than inside a ` ```tool_code``` ` fence (which is handled
by #622):

```python
tool_code = get_weather(city="London", unit="celsius")
print(tool_code)
```

The ` ```tool_code``` ` fence fallback added in #622 doesn't match this because
there is no fence delimiter. The canonical brace parser also misses it. The
result is a silent failure: `tools_called=False`, tool never dispatched.

## Fix

Adds a third fallback pattern (`_TOOL_CODE_ASSIGN_RE`) that matches lines of the
form `tool_code = fn(...)` anchored at the start of a line. The literal
`tool_code =` anchor prevents ordinary assignments like `x = foo()` from being
mistaken for a tool call.

The call expression after `= ` is parsed via the same `ast.literal_eval` path as
the other fallback forms. Matched spans (including any trailing `print(tool_code)`
line emitted as bookkeeping noise) are stripped from `content`.

The cheap `_FALLBACK_MARKER_RE` guard in the streaming path is extended to also
match `tool_code\s*=\s*\w+\s*\(` so streaming doesn't stall.

## Stacking

This PR is stacked on #622 — the `fix/gemma4-fallback-tool-parse` branch must
merge first (or this can be rebased onto main once #622 lands).

## Tests

`tests/test_gemma4_tool_parser.py` — covers the unfenced assignment form,
`print(tool_code)` noise stripping, and the streaming marker.

Ref: issue #83

### DIFF
--- a/tests/test_gemma4_tool_parser.py
+++ b/tests/test_gemma4_tool_parser.py
@@ -297,6 +297,44 @@ class TestGemma4ToolParserFallbackForms:
         tc = self.parser.extract_tool_calls(output).tool_calls[0]
         assert tc["id"].startswith("call_")
 
+    def test_e2b_unfenced_tool_code_assignment(self):
+        """e2b: `tool_code = fn(...)\nprint(tool_code)` — no fence, assignment form.
+
+        Exact output shape observed in live trajectories (2026-06-13).  The parser
+        must dispatch the call and leave content=None (the print line is noise).
+        """
+        output = 'tool_code = radarr_get_movies(search="Dune")\nprint(tool_code)'
+        result = self.parser.extract_tool_calls(output)
+        assert result.tools_called is True
+        assert len(result.tool_calls) == 1
+        tc = result.tool_calls[0]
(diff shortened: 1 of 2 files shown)


Thump604:
Quick CI triage: this appears blocked only on Black formatting. Ruff, type-check, all Python matrix tests, Apple Silicon tests, and the aggregate tests job are green.

The lint log says Black would reformat:

```sh
tests/test_gemma4_tool_parser.py
vllm_mlx/tool_parsers/gemma4_tool_parser.py
```

Please run:

```sh
black tests/test_gemma4_tool_parser.py vllm_mlx/tool_parsers/gemma4_tool_parser.py
```

After that push, this should be ready for review. Since this is stacked with the nearby Gemma parser work, it may still need a rebase if #622 lands first.

PR #626, opened 2026-07-02
fix(embeddings,rerank): resolve truncation length from model config instead of hardcoded 512

brandy975:
## Summary

`EmbeddingEngine.embed()` and the reranker's `SigmoidAdapter.tokenize_pair()`
both hardcode `max_length=512` when tokenizing input — regardless of what the
loaded model actually supports. Inputs longer than 512 tokens are silently
truncated (`truncation=True`), with no warning to the caller.

This is invisible for classic BERT-class models (512 is their real window),
but modern embedding/reranker checkpoints support much more:
`mlx-community/nomicai-modernbert-embed-base-8bit` and ModernBERT-base
support **8192** tokens, `embeddinggemma-300m` supports 2048. Any request
using one of these models currently loses everything past token 512 without
the caller knowing.

This PR resolves the truncation length from each model's own
`config.max_position_embeddings`, trusted as-is and uncapped since it's an
explicit architecture value. Only when the config doesn't supply one does it
fall back to `tokenizer.model_max_length` — subject to sentinel detection,
since HuggingFace defaults that field to a huge placeholder (`~1e30`) when
never actually configured (any tokenizer-derived value ≥ 1,000,000 is
treated as unset). Otherwise it defaults to 512. Classic 512-token models are
unaffected — the resolved value stays 512, byte-identical to today's
behavior.

## Type of change

- Bug fix

## Surface touched

- OpenAI API endpoints (`/v1/embeddings`, `/v1/rerank`)

## Test plan

New unit tests cover the resolver in isolation plus both call sites, positive
(model reports a large window → truncation length follows it) and negative
(no usable config/tokenizer value → falls back to 512):

    pytest tests/test_truncation.py tests/test_embeddings.py tests/test_rerank.py -q

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,6 +85,7 @@ jobs:
             tests/test_dependency_floors.py \
             tests/test_harmony_parsers.py \
             tests/test_endpoint_model_policies.py \
+            tests/test_truncation.py \
             tests/test_gemma4_openai_format.py \
             tests/test_gemma4_streaming_edge.py \
             tests/test_gemma4_tool_parser.py \
@@ -157,6 +158,8 @@ jobs:
             tests/test_tool_choice_forced.py \
             tests/test_tool_choice_none.py \
             tests/test_harmony_render.py \
+            tests/test_embeddings.py \
+            tests/test_rerank.py \
             -v --tb=short \
             -m "not slow" \
             -k "not Integration"
--- a/tests/test_embeddings.py
+++ b/tests/test_embeddings.py
@@ -131,6 +131,63 @@ class TestEmbeddingEngine:
 
         assert result[0] == [0.1, 0.2]
 
+    def test_embed_uses_config_max_length(self):
+        """Positive: truncation length follows model.config.max_position_embeddings."""
+        import numpy as np
+
+        from vllm_mlx.embedding import EmbeddingEngine
+
+        engine = EmbeddingEngine("test-model")
+
+        mock_output = MagicMock()
+        mock_output.text_embeds.tolist.return_value = [[0.1, 0.2]]
+        mock_model = MagicMock(return_value=mock_output)
+        mock_model.config.max_position_embeddings = 8192
+
+        mock_inner_tokenizer = MagicMock()
+        mock_inner_tokenizer.return_value = {
+            "input_ids": np.array([[1, 2]]),
+            "attention_mask": np.array([[1, 1]]),
+        }
+        mock_tokenizer = MagicMock()
+        mock_tokenizer._tokenizer = mock_inner_tokenizer
+
+        with patch.object(engine, "_ensure_loaded"):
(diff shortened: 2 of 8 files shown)


Thump604 (changes requested):
The focused tests pass, and replacing the hardcoded 512 limit is worthwhile. The global 8192 clamp, however, preserves the same class of silent truncation for every model whose explicit `config.max_position_embeddings` exceeds 8192.

The Hugging Face sentinel concern applies to an unset or unreliable `tokenizer.model_max_length`; it does not justify discarding a positive architecture value supplied by the model config. Please separate those cases: trust a valid explicit config limit, and apply sentinel detection or a documented fallback policy only to the tokenizer-derived value. If an operational cap is still required, expose it explicitly rather than presenting 8192 as the model-derived context.

Please add a regression test with `config.max_position_embeddings > 8192` that proves the resolver does not silently reduce an explicit model limit. The current `test_resolve_caps_at_8192` instead locks in the behavior the PR is intended to remove.

brandy975:
Thanks @Thump604 for the detailed review — you're right, and I've split the two cases apart.

`resolve_max_length()` now trusts `config.max_position_embeddings` as-is, uncapped, when it's present. Sentinel detection only applies to the `tokenizer.model_max_length` fallback: HuggingFace defaults that field to a huge placeholder (~1e30) when it was never actually configured, so any tokenizer-derived value ≥ 1,000,000 is treated as unset rather than a real length. An explicit config value never hits that check.

`test_resolve_caps_at_8192` is replaced by `test_resolve_trusts_large_explicit_config_value`, which asserts a `max_position_embeddings` of 100,000 passes through unreduced — the regression case you asked for. I also added a boundary test (`test_resolve_config_value_bypasses_sentinel_threshold`) showing the config path is never subject to the sentinel check, even at values that would trip it on the tokenizer side.

I didn't add an operational cap — nothing in this codebase needs one today, and I'd rather not ship an unused knob. If a deployment-level ceiling turns out to be needed, `resolve_max_length()` is a natural place to add an explicit, clearly-named parameter for it later.

Thump604 (approved):
The follow-up now keeps the two sources distinct: an explicit model-config context value is used uncapped, while only implausible tokenizer fallback values are treated as sentinels. The added regressions cover both the large explicit-config value and the tokenizer sentinel path.

Focused truncation, embedding, and rerank tests pass (67 passed); Ruff and whitespace checks are clean. Approved.

brandy975:
Rebased onto current `main` to fix a conflict in `tests/test_embeddings.py` (both this PR and #667 added a new test at the same spot). No functional changes — just conflict resolution.

waybarrios:
While reviewing this, I found a couple of edge-case issues around XLM-R token limits and CI coverage. The latest commit fixes them and adds the missing regression and CI tests.

waybarrios (approved):
created a small patch on this PR. It seems ready to go

Issue #632, opened 2026-07-07
mlx-vlm 0.6.4 corrupts MLX-format qwen3_5 weights at load — pin !=0.6.4


ethanelasky:
The `mlx-vlm>=0.6.2` constraint currently resolves to 0.6.4, which has a weight-loading regression (Blaizzy/mlx-vlm#1521, confirmed): it re-runs `sanitize_weights()` on already-converted MLX-format checkpoints, which scrambles qwen3_5 / qwen3_5_moe weights at load. Any MLX-community qwen3.5 model served through the MLLM path generates deterministic garbage — temp 0 gives multilingual token soup on every request.

Repro:

```shell
vllm-mlx serve mlx-community/Qwen3.5-4B-4bit --continuous-batching
# any chat completion at temperature 0
```

What makes it nasty to diagnose is that `mlx_lm.generate` on the same checkpoint in the same venv is perfectly coherent (mlx-lm's loader is unaffected), so it looks like a vllm-mlx serving bug. It isn't — I bisected my way to the dep: downgrading only mlx-vlm to 0.6.3 with everything else unchanged fixes serving completely.

Suggested fix until the upstream guard lands (their issue already has an agreed `_is_mlx_safetensors_format()` check): pin `mlx-vlm>=0.6.2,!=0.6.4`.

Thump604:
Confirmed the dependency boundary against current main. PR #633 excludes exactly mlx-vlm 0.6.4, retains the existing >=0.6.2 floor, and adds a dependency-contract regression test. A fresh resolution selects mlx-vlm 0.6.3. This prevents vllm-mlx from resolving the affected release; it does not claim to fix the underlying mlx-vlm loader bug.

PR #633, opened 2026-07-10
fix(deps): exclude corrupt mlx-vlm 0.6.4 release

Thump604:
Refs #632.

## Problem

`mlx-vlm 0.6.4` re-sanitizes already-converted Qwen3.5 MLX weights. The affected dependency release can therefore produce corrupted Qwen3.5 output when `vllm-mlx` resolves its current unbounded `mlx-vlm>=0.6.2` requirement.

The same Qwen3.5 checkpoint remains coherent with `mlx-vlm 0.6.3`; the reported regression is tracked upstream as `Blaizzy/mlx-vlm#1521`.

## Change

- Exclude exactly `mlx-vlm==0.6.4` while retaining the existing `>=0.6.2` floor.
- Add a dependency-contract test for the exclusion.

## Verification

- `uv run --python 3.12 --extra dev pytest tests/test_dependency_floors.py -q` -> `3 passed`
- `uv run --python 3.12 --extra dev ruff check tests/test_dependency_floors.py` -> passed
- Fresh dependency resolution selected `mlx-vlm 0.6.3`, not 0.6.4.
- `git diff --check` -> passed

## Not claimed

This does not fix the loader regression inside `mlx-vlm`. It prevents `vllm-mlx` installations from resolving the known-bad release until the upstream package ships a corrected version.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ classifiers = [
 dependencies = [
     "mlx>=0.29.0",
     "mlx-lm>=0.31.3",  # 0.31.3+ required by current mlx-vlm runtime support
-    "mlx-vlm>=0.6.2",  # 0.6.2+ includes Step 3.7 Flash and follow-up fixes
+    "mlx-vlm>=0.6.2,!=0.6.4",  # 0.6.4 corrupts converted Qwen3.5 MLX weights (#632)
     "transformers>=5.0.0",  # mlx-lm 0.30.5+ requires transformers 5.0 (rc3 bug fixed in stable)
     "tokenizers>=0.19.0",
     "huggingface-hub>=0.23.0",
--- a/tests/test_dependency_floors.py
+++ b/tests/test_dependency_floors.py
@@ -32,10 +32,20 @@ def _has_minimum(requirement: str, minimum: str) -> bool:
     return _version_tuple(version) >= _version_tuple(minimum)
 
 
+def _excludes_version(requirement: str, version: str) -> bool:
+    return f"!={version}" in {part.strip() for part in requirement.split(",")}
+
+
 def test_mlx_vlm_floor_includes_step37_flash_support_and_followups():
     dependencies = _project_dependencies()
 
     assert _has_minimum(dependencies["mlx-vlm"], "0.6.2")
+
+
+def test_mlx_vlm_excludes_corrupt_qwen35_release():
+    dependencies = _project_dependencies()
+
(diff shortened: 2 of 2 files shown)


janhilgard (approved):
LGTM — clean, targeted fix for a real production regression.

## What I verified

- **Upstream issue confirmed & fixed**: `Blaizzy/mlx-vlm#1521` reproduces the `sanitize_weights` double-run on already-converted MLX safetensors. Closed 2026-07-15, and `mlx-vlm 0.6.5` (2026-07-16) shipped the day after — the ecosystem now has a coherent upgrade path.
- **Available versions on PyPI right now**: 0.6.0 → 0.6.8. Only `0.6.4` is affected; everything from 0.6.5 onward includes the loader guard.
- **Production evidence**: I've been running `mlx-vlm 0.6.6` against MLX-community qwen3_5 checkpoints (Qwen 3.6-27B, 35B-A3B) on Apple Silicon since 2026-07-23. No sanitize corruption observed. Confirms 0.6.5+ is safe.

## Design choice — surgical exclude vs. floor bump

The PR keeps the floor at `>=0.6.2` and adds `!=0.6.4` instead of just bumping to `>=0.6.5`. This is the right call:

- 0.6.2 and 0.6.3 are known-good; raising the floor would needlessly block users pinned there.
- Only 0.6.4 is toxic — exclude exactly the toxic version, no more, no less.
- The composite constraint is explicit about *why* the exclusion exists (comment cites #632).

Bumping to `>=0.6.5` would also be acceptable, but surgical exclude minimizes user disruption.

## Code review

- `pyproject.toml`: one-line change, comment references issue #632 — traceable.
- `tests/test_dependency_floors.py`: new `_excludes_version` helper is small, correct, and reusable if future bad releases need exclusion. Reasonable pattern.
- Existing `test_mlx_vlm_floor_includes_step37_flash_support_and_followups` still passes → no regression in floor semantics.

## CI

All 9 checks green (lint, type-check, apple-silicon 3.11/3.13, test-matrix 3.10-3.13).

## Follow-up (non-blocking)

Once you're confident 0.6.5+ has been stable in the wild for a few weeks, consider bumping the floor to `>=0.6.5` (or the latest known-good) and dropping the `!=0.6.4` clause — keeps the constraint expression clean long-term. Not for this PR.

Ship it.

waybarrios (approved):
Approving after full review (6-agent + empirical mutation test): exclusion is valid PEP 440, test fails when the clause is removed, CI all green, no findings.

PR #634, opened 2026-07-11
fix(engine): close SSD cache tier on stop paths

djacobsmeyer:
## Summary

**Defect:** SSD writer threads never stopped on shutdown. `close_ssd_tier()` was only reachable from `Scheduler.reset()`, never from the actual engine stop paths.

**Fix:** Wired `close_ssd_tier()` into both engines' stop paths:
- **Text engine (`vllm_mlx/engine_core.py`):** Added call to `self.scheduler.close_ssd_tier()` in `_engine_loop()`'s `finally` block (before `worker.shutdown(wait=True)`) and as an idempotent safety net in `EngineCore.stop()`.
- **MLLM engine (`vllm_mlx/mllm_scheduler.py`):** `MLLMScheduler.stop()` now closes `self._ssd_tier` (guarded for `None`) and clears it. Also initialized `_ssd_tier: Optional[Any] = None` in `__init__` to prevent `AttributeError` when stop is called before any request is processed.

## Test Results

**New test file:** `tests/test_ssd_shutdown_wiring.py` (8 tests) — verifies writer thread is joined, tier is cleared, queued spills are flushed, and stop is a safe no-op when no tier is configured.

**Existing tests fixed:** Added `close_ssd_tier()` no-op to `FakeScheduler` in `tests/test_engine_core_thread_streams.py` to maintain test double compatibility.

**Test suite results:**
- New SSD shutdown tests: 8 passed
- Targeted regression sweep (SSD + scheduler + MLLM + engine-core): 155 passed
- Full suite: 2257 passed, 24 skipped, 23 deselected

No failures. Code formatting (black, ruff) verified clean on all changed files.

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,6 +83,7 @@ jobs:
             tests/test_anthropic_models.py \
             tests/test_anthropic_adapter.py \
             tests/test_dependency_floors.py \
+            tests/test_ssd_cache_shutdown.py \
             tests/test_simple_engine_thread_pinning.py \
             tests/test_harmony_parsers.py \
             tests/test_endpoint_model_policies.py \
@@ -156,6 +157,9 @@ jobs:
             tests/test_batching.py \
             tests/test_continuous_batching.py \
             tests/test_memory_cache_mlx.py \
+            tests/test_mllm_ssd_spill.py \
+            tests/test_ssd_cache.py \
+            tests/test_ssd_shutdown_wiring.py \
             tests/test_lifecycle_cli.py \
             tests/test_lifecycle_server.py \
             tests/test_normalize_messages.py \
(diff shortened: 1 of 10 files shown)


Thump604 (changes requested):
The stop-path wiring is needed, but the current close contract does not guarantee the queued-spill behavior claimed by the PR. `SSDCacheTier.close()` sets `_writer_stop` before the writer drains `_spill_queue`; `_writer_loop()` checks that event at the top of its loop, so it can exit with queued items unwritten. The new flush test is race-dependent because the writer is started before the spill is enqueued. There is a second race after `join(timeout=5)`: the code clears the writer reference and closes SQLite even if the writer is still alive and inside `_write_entry`.

Please make shutdown explicitly drain via queue/poison-pill semantics (or another deterministic contract), keep the index open until writer termination is proven, and add a test that holds the writer before dequeue so a queued item is known to still be pending when close begins. Because both callers are async stop paths, also avoid introducing an event-loop-blocking join.

waybarrios:
I found a remaining gap in the SSD shutdown path where queued spills could be dropped and cleanup could block the event loop. I fixed it by draining the queue safely, handling timeouts without closing live resources, restoring the tier on restart, and adding regression coverage.

waybarrios:
CI done and ready to go!. I already fixed the gap

Issue #635, opened 2026-07-13
response_format can return successful output that violates the requested JSON shape


Thump604:
# `response_format` can return successful output that violates the requested JSON shape

## Reproduction

Run `vllm-mlx` with a Qwen 3.6 model and send streaming chat-completion requests
with either:

1. `response_format={"type":"json_object"}` and a prompt requesting
   `[ {"ok": true} ]`.
2. A strict JSON schema whose root is an object and whose `sc` property uses
   nine homogeneous `prefixItems`, `minItems: 9`, and `maxItems: 9`, with each
   item constrained to an integer from 1 through 5.

The local reproduction used `qwen3.6-35b-a3b`, the upstream v0.4.0 code paths,
and the request bodies preserved in:

- `run/qwen-constrained-output-investigation/20260713T192436Z/pre-fix-json-object-request.json`
- `run/qwen-constrained-output-investigation/20260713T192436Z/pre-fix-json-schema-request.json`

Send either body to `POST /v1/chat/completions`:

```json
{"model":"qwen3.6-35b-a3b","messages":[{"role":"user","content":"Return exactly this JSON array and nothing else: [{\"ok\":true}]"}],"stream":true,"max_tokens":256,"temperature":1.0,"top_p":0.95,"top_k":20,"min_p":0.0,"presence_penalty":0.0,"repetition_penalty":1.0,"response_format":{"type":"json_object"}}
```

```json
[...]

PR #636, opened 2026-07-13
Fix fail-open structured-output enforcement

Thump604:
# Fix fail-open structured-output enforcement

## Summary

- require an object root for `response_format.type=json_object`
- route strict `json_schema` through token-level JSON Schema enforcement
- mask numeric-range violations and premature array closure before emission
- require schema-valid completion before EOS or HTTP success
- fail closed when constrained-decoder setup or execution fails

## Exact reproduction

On upstream v0.4.0, a streaming `json_object` request could return an array root with `finish_reason=stop`. A strict compact schema requiring exactly nine integers from 1 through 5 could return strings and report `finish_reason=stop`.

A follow-up full Jobs-shaped request reproduced a second failure: the existing processor emitted nine `0` values despite `minimum: 1` and `maximum: 5`, then reached an empty allowed-token set and returned HTTP 500.

The compared paths are:

- `vllm_mlx/constrained/json_schema_processor.py`
- `vllm_mlx/constrained/llguidance_schema_processor.py`
- `vllm_mlx/api/tool_calling.py`
- `vllm_mlx/server.py`

## Expected behavior

Declared structured output remains constrained for the entire request. Invalid numeric values and premature structural closure are excluded before token selection. If the schema cannot be compiled or the processor fails, the request fails instead of continuing unconstrained.

## Minimal patch shape

`json_object` retains the object-root LMFE path. Strict `json_schema` uses a request-local `llguidance` matcher with MLX token masking, padded-vocabulary handling, schema-aware EOS, and a bounded non-progress whitespace guard. For streaming requests with response_format, the endpoint buffers the complete SSE sequence and validates it before returning a success response. This intentionally changes time-to-first-byte and memory behavior for structured streams; ordinary text streams remain incremental.

## Validation

- full upstream suite: 2,261 passed, 24 skipped, 23 deselected
- focused structured/server suite: 222 passed, 2 skipped, 3 deselected
- Ruff, formatting, and `git diff --check`: passed
- exact 51,014-character live request repeated 3 times: HTTP 200, `finish_reason=stop`, independent schema validation passed, all scores in range
- conflict run where the prompt requested `0`: token masking emitted legal value `1`

Live summary:

`/opt/ai-runtime/run/qwen-constrained-output-investigation/20260713T-live-numeric-range-llguidance-certification-pass/summary.json`

## Explicit non-claims

This does not claim a model-quality fix or universal JSON Schema coverage. Unsupported or invalid constrained-decoder setup remains fail-closed. The patch does not change model templates, sampling, Runtime contracts, Jobs routing, or feature defaults.

Refs #635.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,6 +69,7 @@ dependencies = [
     "jsonschema>=4.0.0",
     # Constrained decoding (grammar-guided generation for response_format)
     "lm-format-enforcer>=0.10.9",
+    "llguidance>=1.7.6",
     # pytz is required by gradio but not declared as a dependency
     # See: https://github.com/waybarrios/vllm-mlx/issues/23
     "pytz>=2024.1",
--- a/tests/test_constrained_decoding.py
+++ b/tests/test_constrained_decoding.py
@@ -20,6 +20,9 @@ running with a minimal dependency set still passes.
 
 from __future__ import annotations
 
+import math
+
+import mlx.core as mx
 import pytest
 
 from vllm_mlx.api.anthropic_adapter import anthropic_to_openai
@@ -398,8 +401,44 @@ class TestSimplifySchema:
 
         result = _simplify_schema(schema)
 
-        assert result["items"] == integer_score
+        assert result["items"] == {
+            **integer_score,
+            "enum": [1, 2, 3, 4, 5],
+        }
         assert "prefixItems" not in result
+
+    @pytestmark_lmfe
+    def test_integer_range_masks_preferred_out_of_range_token(self):
+        """A preferred zero must lose to a legal 1-5 score token."""
+        from vllm_mlx.constrained import JSONSchemaLogitsProcessor
+
+        tokenizer = _FakeTokenizer()
+        processor = JSONSchemaLogitsProcessor(
+            {"type": "integer", "minimum": 1, "maximum": 5},
+            tokenizer,
+        )
+        logits = mx.full((1, tokenizer.vocab_size), -100.0)
+        logits[0, tokenizer._tok_to_id["0"]] = 100.0
+        for score in range(1, 6):
+            logits[0, tokenizer._tok_to_id[str(score)]] = float(6 - score)
+
+        masked = processor(mx.array([999]), logits)
+        values = masked.tolist()[0]
+
+        zero_id = tokenizer._tok_to_id["0"]
+        assert math.isinf(values[zero_id]) and values[zero_id] < 0
+        assert int(mx.argmax(masked, axis=-1).item()) == tokenizer._tok_to_id["1"]
+
+    def test_integer_range_rejects_unenforceable_shapes(self):
+        from vllm_mlx.constrained.json_schema_processor import (
+            UnsupportedJSONSchemaError,
+            _simplify_schema,
+        )
+
+        with pytest.raises(UnsupportedJSONSchemaError, match="lower and upper"):
+            _simplify_schema({"type": "integer", "minimum": 1})
+        with pytest.raises(UnsupportedJSONSchemaError, match="exceeds"):
+            _simplify_schema({"type": "integer", "minimum": 0, "maximum": 5000})
 
     @pytestmark_lmfe
     def test_complete_json_requires_schema_valid_value(self):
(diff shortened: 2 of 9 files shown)


janhilgard (approved):
LGTM — this is the right defense-in-depth rewrite. Ships fail-closed semantics that structured output was supposed to have from day one.

## Why this matters

The old path had three silent-failure modes that all reported `finish_reason=stop` to the client:

1. `json_object` accepting array roots — schema says "object", spec allows any valid JSON, output could be `[...]`.
2. Strict `json_schema` returning out-of-range integers — `minimum: 1` respected only by validation at buffer flush, not at token selection, so mid-stream out-of-range values slipped through.
3. Constrained decoder failures degrading to unconstrained generation — schema compile error → `except: return None` → prompt-only "please emit valid JSON" mode, indistinguishable from an unconstrained request.

All three now fail closed. Client gets an error or gets valid output — no third state.

## Code review

**`LLGuidanceJSONSchemaLogitsProcessor` (new, 206 lines):**
- Request-local matcher (`_matcher: LLMatcher` per-instance) → no cross-request state contamination.
- Token-level masking via `llguidance.mlx.apply_token_bitmask` before sampling → constraint is *arithmetic*, not post-hoc validation. This is the correct architecture.
- Padded vocabulary handling — `_clear_mask_token` correctly indexes into the bitmask array (word_index, bit_index) and the tests explicitly verify padded-vocab masking (vocab+64 padding, all padding bits stay `-inf`). Important — MLX tokenizers commonly ship padded vocabs and a naive mask would allow padding token IDs.
- Whitespace guard (`_MAX_NONPROGRESS_WHITESPACE_CHARS = 256`) prevents the "runaway whitespace" degeneracy where a matcher accepts trailing whitespace forever. `_trailing_json_whitespace` correctly ignores whitespace inside JSON strings.
- Terminal state → EOS-only mask is the right way to guarantee schema-complete finalization. Rejecting EOS before `_matcher.is_accepting()` closes the "premature stop" hole.
- Error surface: `ConstrainedDecodingError` with decoded suffix tail → actionable debug info without leaking full state. Nice touch.

**`json_schema_processor.py` refactor:**
- Existing LMFE path preserved for `json_object` (only requires object root).
- Split points clean: `json_object` → LMFE, strict `json_schema` → llguidance. No mixing.

**`api/tool_calling.py:parse_json_output` change:**
- `if not isinstance(parsed, dict): return ... "JSON object root must be an object"` — closes reproduction case #1 at the HTTP boundary as a belt-and-suspenders behind the token mask.
- Return type widening from `Dict[str, Any]` to `Any` is correct — an object-root failure returns the parsed value for debug logging, not just `None`.

**`build_json_logits_processor` rework:**
- Old: silent `return None` on missing dep or setup error → client got unconstrained output with no signal.
- New: `RuntimeError` on setup failure → client gets HTTP 500 with a diagnosable message.
- This is a **behavior change** but a correct one. If someone was relying on "constrained fails silently and I hope the model does the right thing" — that's not a contract, it's a bug.

## Tests

Well-targeted:
- `test_strict_schema_masks_preferred_out_of_range_integer` — the actual repro. Sets `logits[zero_id] = 100.0`, verifies `masked[zero_id] == -inf`, verifies `masked[one_id]` is finite. Direct assertion on the fix.
- `test_strict_schema_masks_model_vocab_padding` — vocab_size 64 tokens smaller than logits shape, checks all padding indices masked to `-inf`. Prevents the "sampled padding token → cascade failure" class of bug.
- `test_strict_schema_accepts_eos_after_complete_value` — EOS gate.

Full suite pass (2,261) plus focused 222-test constrained suite. Live replication (51,014-char request × 3, conflict test with prompt requesting `0`) is documented with artifact path. Solid evidence.

## Dependency

Adds `llguidance>=1.7.6` as required (not optional). Given the module now hard-imports on the strict schema path, this is correct — you can't have optional-behind-import-error and fail-closed at the same time. Production installs will need this pinned; worth calling out in release notes.

## CI

All 9 checks green.

## Non-blocking notes

- `_MAX_NONPROGRESS_WHITESPACE_CHARS = 256` is a magic number. Fine as a threshold, but a brief comment about the choice (e.g., "long enough for legitimate indentation, short enough to catch runaway whitespace") would help future readers.
- `_tokenizer_cache: dict[int, tuple[...]]` uses `id(candidate)` — safe since tokenizers are long-lived and rarely GC'd during a request, but worth a docstring note if this ever becomes a lifetime concern.
- The `except Exception` in `_decoded_token_is_whitespace` swallows all errors. Reasonable for a whitespace check, but if a tokenizer.decode crashes systemically, this hides it silently. Non-critical.

Ship it. This closes a real correctness hole in structured output.

Thump604:
Rebased the five-commit fail-closed structured-output series onto current `main` (`7afa61a`). The sole conflict retained main’s `_streaming_json_fence_stripper(request)` helper while preserving structured-stream accumulation and validation. I also changed the linkage to `Refs #635` and made the structured-stream buffering/TTFB tradeoff explicit in the body. Verification: `2585 passed, 25 skipped, 27 deselected`; focused constrained/structured/server `221 passed, 2 skipped, 3 deselected`; Black, compile, and diff checks clean. New head: `e5d5f54`.

Issue #627, opened 2026-07-03
Multi-model manager's memory_budget_gb is never reconciled with the Metal allocation ceiling


danmackinlay:
**Environment:** vllm-mlx v0.4.0, macOS (Apple Silicon), MLX backend, `--models-config` registry mode.

## Summary

In registry mode the model manager makes eviction decisions purely against `manager.memory_budget_gb`, but it never consults the actual Metal allocation ceiling (`--gpu-memory-utilization`, `--cache-memory-mb`, or the OS `iogpu.wired_limit_mb`). The two numbers are independent and unreconciled. If the budget is set higher than what is actually allocatable once KV cache and activations are counted, the manager keeps N models resident because *its* arithmetic says they fit — and then MLX hits the ceiling and the process OOM-crashes, instead of the manager gracefully evicting an older model. So the failure mode the multi-model budget exists to prevent (running out of memory) is the one you hit, just as a hard crash rather than an eviction.

## Mechanism (line numbers approximate, from the installed 0.4.0 wheel)

- `manager.memory_budget_gb` is **required from the YAML and has no CLI/env override**: `model_registry._parse_memory_budget_bytes` raises `"models-config manager.memory_budget_gb is required"` when absent, and it is read only from the `manager` block (`memory_budget_gb` / `memory_budget`).
- The eviction fit-check is **weights-only vs the budget**: `projected_bytes + required_bytes <= self._config.memory_budget_bytes` (three call sites in `model_registry.py`). `required_bytes` is the summed on-disk `.safetensors`/`.gguf` size for local models (`path.stat().st_size`) or the operator's `estimated_memory_gb` for HF ids — in both cases weights only, no KV or activation accounting.
- The manager **never references** `gpu_memory_utilization`, `cache_memory_mb`, or the wired limit in that decision. `cli.py` has no budget argument (it only validates that `--models-config` is not combined with a positional model / `--served-model-name`).

## Reproduction

1. Registry with two large models (e.g. two ~32 GB MoE builds), `manager.memory_budget_gb: 68`.
2. Launch with a ceiling that cannot actually hold both plus cache, e.g. `--gpu-memory-utilization 0.62` (~79 GB on a 128 GB machine) and `--cache-memory-mb 20480`.
3. Send a request to model A, then to model B.
[...]

PR #638, opened 2026-07-15
fix(registry): honor auto-unload-idle-seconds and release MLX memory on unload

brandy975:
# fix(registry): honor auto-unload-idle-seconds and release MLX memory on unload

## Summary

`--auto-unload-idle-seconds` already exists and works for the default single-model server, but it is a silent no-op in `--models-config` (multi-model registry) mode, and even where it does work today it doesn't actually release memory back to the OS. This PR fixes both:

- `ModelManager` (registry mode) now proactively unloads idle models via a background timer, matching the behavior already present for the single-model `ResidencyManager` path.
- `SimpleEngine.stop()` / `BatchedEngine.stop()` now call `mx.clear_cache()`, so an unload actually releases MLX's Metal buffer cache instead of only dropping Python references.

This is a bug fix to an existing, incompletely-implemented feature, not a new capability.

## Problem

1. **`--auto-unload-idle-seconds` is silently ignored in registry mode.** `cli.py` only threads `args.auto_unload_idle_seconds` into `load_model()` (the single-model path); the registry branch (`load_model_registry()`) never received it. In registry mode, `ModelManager` only evicts models *reactively* — when a different model is requested and the configured memory budget would be exceeded (`_collect_idle_unloads_locked`, invoked from inside `acquire()`). If a deployment only ever serves one model through `--models-config`, or requests happen infrequently enough that a second model is never concurrently requested, that model stays resident indefinitely, regardless of the configured timeout — with no error or warning that the flag has no effect.
2. **Unload doesn't actually release memory.** Neither `SimpleEngine.stop()` nor `BatchedEngine.stop()` called `mx.clear_cache()` — they only dropped Python references (`self._model = None`, etc.). MLX's Metal buffer-cache pool was never explicitly released, so resident memory did not shrink the way "unload" implies, even in the already-working single-model path.

## Tests

14 new tests across 5 files:

- `tests/test_model_registry.py` — `unload_idle()` unloads stale models and skips busy ones; no-op when disabled; leaves recently-used models alone; `run_idle_reaper()` unloads after the timeout elapses and returns immediately when disabled; YAML parsing of `manager.idle_unload_seconds` (explicit value wins, falls back to the CLI default, defaults to disabled).
- `tests/test_lifecycle_cli.py` — regression test that `--auto-unload-idle-seconds` combined with `--models-config` is now actually threaded through; fixture updated to track/clean up `_registry_idle_reaper_task`.
- `tests/test_lifecycle_server.py` — the reaper task starts and stops correctly via `lifespan()`, both when enabled and when disabled.
- `tests/test_simple_engine.py`, `tests/test_batched_engine.py` — spy-based tests asserting `stop()` calls `mx.clear_cache()` exactly once.
- `tests/test_registry_idle_unload_real_model.py` — new, manual-only real-server/real-model script (see "Manual verification" below). Not collected by pytest.

Full suite: `pytest tests/` — 2260 passed, 23 skipped. Three pre-existing failures are unrelated to this change (verified identical on unpatched `main` via `git stash` before making any changes):

- `tests/test_lifecycle_manager.py::TestResidencyManagerContracts::test_shutdown_canceled_prepare_error_unwinds_to_unloaded`
- `tests/test_simple_engine.py::TestSimpleEngineConcurrency::test_cancelling_specprefill_request_stops_during_scoring`
- `tests/test_simple_engine.py::TestSimpleEngineConcurrency::test_cancelling_specprefill_request_stops_during_sparse_prefill`

These are pre-existing timing-sensitive cancellation tests (SpecPrefill scoring/sparse-prefill cancellation, residency shutdown-during-prepare) unrelated to the registry/idle-unload/engine-stop code paths touched here.

## Manual verification

End-to-end tested against a real model (`mlx-community/Qwen3-0.6B-8bit`) in registry mode: correct idle-unload timing, transparent reload on the next request, and a measurable RSS drop after unload.

This is captured as a repo-committed, manually-runnable script: `tests/test_registry_idle_unload_real_model.py`. It starts a real `vllm-mlx serve --models-config ...` subprocess, drives it over HTTP, and prints RSS at each stage. It intentionally has no `test_`-prefixed function, so `pytest` collects zero items from it — it never runs as part of the normal suite or CI, only via:

```
python tests/test_registry_idle_unload_real_model.py
python tests/test_registry_idle_unload_real_model.py --model mlx-community/Qwen3-0.6B-8bit  # override, same default
```

Sample output:

```
RSS before load:  369.4 MB
RSS after load:   1230.6 MB  (+861.2 MB)
RSS after unload: 606.7 MB  (-623.9 MB reclaimed)
RSS after reload: 1315.2 MB  (+708.5 MB)
```

Also confirmed the script fails as expected against unpatched `main` (model never unloads) and passes once this fix is applied — a real regression check, not just a happy-path run.

### DIFF
--- a/docs/guides/model-registry.md
+++ b/docs/guides/model-registry.md
@@ -53,6 +53,7 @@ Example:
 ```yaml
 manager:
   memory_budget_gb: 100
+  idle_unload_seconds: 300
   contention_policy:
     strategy: wait_then_preempt
     wait_timeout_s: 45
@@ -94,6 +95,16 @@ It does not include, and does not reserve room for:
 - other colocated services
 
 On a 128 GB machine, a practical starting point is often `80-100 GB`.
+
+### `idle_unload_seconds`
+
+Automatically unload a model after it has had no active requests for this many
+seconds. Values less than or equal to `0` disable idle unloading.
+
+If this setting is omitted, registry mode inherits
+`--auto-unload-idle-seconds`; that flag defaults to `0`. A value in the YAML
+file takes precedence over the CLI fallback. Preloaded models are also eligible
+for idle unloading after their preload lease is released.
 
 ### Budget vs. the Metal allocation ceiling
 
@@ -263,6 +274,13 @@ Registry-backed responses include the configured model ids and current state suc
 - `loading`
 - `unloaded`
 - `preempting`
+
+Each model entry also includes `last_used_at` when it is loaded. For the
+effective idle timeout together with all model states, inspect `/v1/status`:
+
+```bash
+curl http://localhost:8000/v1/status
+```
 
 ### Verify a cold-load path
 
--- a/tests/test_batched_engine.py
+++ b/tests/test_batched_engine.py
@@ -320,3 +320,36 @@ class TestToolCallReplayNormalization:
         normalized = _normalize_tool_call_arguments_for_template([MessageLike()])
 
         assert normalized == [{"role": "assistant", "content": "plain response"}]
+
+
+class TestBatchedEngineStop:
+    """stop() must actually release MLX's Metal buffer cache, not just drop
+    Python references — otherwise idle-unload frees objects but not memory.
+    """
+
+    def _make_engine(self):
+        from vllm_mlx.engine.batched import BatchedEngine
+
+        with patch("vllm_mlx.engine.batched.is_mllm_model", return_value=False):
+            engine = BatchedEngine("test-model")
+        engine._loaded = True
+        return engine
+
+    @pytest.mark.anyio
+    async def test_stop_calls_mx_clear_cache(self, monkeypatch):
+        from vllm_mlx.engine import batched as batched_mod
+
+        calls = {"count": 0}
+        monkeypatch.setattr(
+            batched_mod.mx,
+            "clear_cache",
+            lambda: calls.__setitem__("count", calls["count"] + 1),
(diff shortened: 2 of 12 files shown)


brandy975:
Related: #627

Thump604 (changes requested):
The substantive registry/lifecycle behavior looks sound in review, and focused local validation passed (`168 passed`). The current-head CI run also passed every test matrix, Apple Silicon job, and type check.

The remaining blocker is mechanical but required: Black reports four files that need formatting:

- `tests/test_model_registry.py`
- `tests/test_registry_idle_unload_real_model.py`
- `tests/test_simple_engine.py`
- `vllm_mlx/model_registry.py`

Please run `black vllm_mlx/ tests/`, push the formatting result, and let CI rerun. I found no separate behavioral blocker in the current patch.

brandy975:
@Thump604 Fixed in 89b8fd1 — black vllm_mlx/ tests/ applied and black --check now passes clean across the repo. Pushed, ready for re-review.

waybarrios:
Updated this PR to make registry idle unloads cancellation-safe during shutdown, keep the lifecycle CLI test compatible with current main, and document the idle-timeout behavior. The branch is now synced with main; focused registry and CLI tests plus Ruff, Black, and compilation checks pass. Fresh CI is running.

Issue #641, opened 2026-07-22
[Bug]: MemoryAwarePrefixCache stores caches by reference — leaks Metal buffer handles on hybrid models until `Resource limit (499000) exceeded` abort


rohnach29:
### Environment

python 3.12.13 (main, Jun 11 2026, 03:55:21) [Clang 22.1.3]
mlx device Device(gpu, 0)

vllm-mlx: 0.4.0rc1
mlx: 0.31.2
mlx-lm: 0.31.3
mlx-vlm: 0.6.3
mlx-audio: 0.4.4
mlx-embeddings: 0.1.0

macOS 26.2 (25C56)

Chip (M1 / M2 / M3 / M4 / M5, including Pro / Max / Ultra): M3 Ultra 
RAM: 512 GB
macOS version: 26.2 (25C56)
Python version: 3.12.13
vllm-mlx version: 0.4.0rc1 
mlx / mlx-lm / mlx-vlm / mlx-audio versions: 0.31.2 / 0.31.3 / 0.6.3 / 0.4.4

### Model and quantization

mlx-community/Qwen3.6-35B-A3B-8bit

### Reproduction

```shell
# Server (continuous batching + prefix cache on a hybrid model):
vllm-mlx serve mlx-community/Qwen3.6-35B-A3B-8bit \
  --continuous-batching --max-num-seqs 10 \
  --enable-prefix-cache --cache-memory-mb 16384 \
  --reasoning-parser qwen3 --enable-auto-tool-choice --tool-call-parser qwen3_coder \
  --port 8000

# Then drive sustained mixed chat-completions traffic (we serve ~350 req/h
# production traffic, prompts 0.1K-20K tokens). Watch
# mx.get_active_memory() / vmmap: memory ratchets ~0.9 GB per completed
# request and never comes back down. Process aborts in ~47 minutes.

# Minimal proof that the fatal "Resource limit (499000)" is a live-buffer
# COUNT, not bytes (crashes at ~0.5 GB allocated):
python -c "
import mlx.core as mx
arrs = []
for b in range(60):
    batch = [mx.zeros((256,)) for _ in range(10000)]  # 1 KB each
    mx.eval(*batch)
    arrs.extend(batch)
# -> RuntimeError: [metal::malloc] Resource limit (499000) exceeded at ~490K live buffers
"
```

### Expected behavior

[...]

PR #642, opened 2026-07-22
fix: detach stored prefix-cache entries from live batch state

rohnach29:
## Summary

Refs #641

`store()` keeps per-request cache layers **by reference**. On the continuous-batching path those layers are lazy slices of live batch-wide arrays, and hybrid layers (`ArraysCache`) expose their mutable state container via `.state`. A stored entry therefore (a) **aliases** containers the batch generator keeps mutating — same bug class as #575, fixed for SimpleEngine in #576 but not here — and (b) **retains each entry's upstream lazy graph**, pinning ~1,700 Metal buffer handles and ~0.9 GB per entry until eviction. On a hybrid model (Qwen3.6-35B-A3B, 30/40 `ArraysCache` layers) at ~350 req/h this exhausts the Metal handle limit (`[metal::malloc] Resource limit (499000) exceeded` — a count, not bytes; 10-line repro in #641) and aborted the server every ~47 minutes. The `shape × dtype` accounting cannot see the retention (it prices a lazy array's promised output, not what its graph pins), so `--cache-memory-mb` never contained it.

**The fix:** `store()` becomes the single owner of copying and evaluation, with the pipeline **trim (lazy) → quantize (lazy) → preflight → detach → account**. The pipeline runs outside `_memory_lock`; the entry-sized copy is serialized by a dedicated copy lock; `store()` never raises — any pipeline failure rejects the entry with a warning (counted in a new `store_rejections` stat).

- **Detach:** every real `mx.array` is replaced with a guaranteed freshly-allocated, dtype-faithful copy (`arr + 0`, per review — `mx.contiguous` may share an already-contiguous input's buffer), with all containers cloned: KVCache-style layers (incl. quantized tuple keys and batch variants' `left_padding`/`lengths`/`offset` metadata arrays), `.state` layers, `CacheList` children (recursively — its `state` setter writes through to live children), extracted state dicts, and nested mappings. One batched `mx.eval` per store; `copy.copy()` so `__slots__` classes work.
- **Preflight before materialization:** `_trim_to_offset` now slices lazily, so over-limit entries are rejected from `shape × dtype` pricing with nothing evaluated — an oversized lazy entry costs ~0 instead of an entry-sized allocation spike.
- **Honest accounting:** keys-carrying layers are priced on their **raw buffers**, byte-exact with what detachment copies (ring/chunked semantics forbid slicing to offset); `CacheList` is priced by recursing into children; state-layer metadata arrays are priced too. The old sliced-view pricing under-counted padded ring/chunked layers by the padding ratio — up to 25.6× on short entries — silently breaching the byte cap. **Consequence:** a given `--cache-memory-mb` now admits fewer heavily-padded entries; the cap tracks real residency. Trimming ring buffers could recover much of that padding — left as a follow-up.
- **Quantize before detach:** with `kv_quantize`, the stored arrays are the final quantized representation; the old order retained the full-precision snapshot as graph input.
- **Fail-closed, as a verified postcondition:** after snapshotting, any `mx.array` in the snapshot that is still the same object as one in the live input rejects the store. (Branch dispatch alone can't guarantee this: every mlx_lm `_BaseCache` subclass inherits a default `state` property returning `[]`, so a layer with arrays outside `.state` would previously "detach" an empty state and store everything else aliased, priced at 0 bytes.) Unknown array-bearing layers, dict layers with arrays outside `state`, and arrays in unsupported containers are rejected; array-free unknown layers pass through.
- **#683 integration:** `Scheduler._copy_cache_state` is removed; the snapshot path hands `store()` a structure mirror whose state deliberately aliases the live caches, and `store()` detaches or rejects. #683's scheduler/key semantics (`_make_snapshot_destination`, `_cache_key_for_snapshot`, usefulness gating, refresh threshold, evict_prefixes) are unchanged.

Follows the approach of #576. With detachment plus recursive accounting, `shape × dtype` pricing matches resident memory, so `--cache-memory-mb` governs real bytes.

## Type of change

- Bug fix

## Surface touched

- KV cache / prefix cache / scheduler snapshot path

## Test plan

```
python -m pytest tests/test_memory_cache.py tests/test_memory_cache_mlx.py \
  tests/test_prefix_cache.py tests/test_prefix_cache_untrimmable.py \
  tests/test_mllm_cache.py tests/test_kv_cache_quantization.py \
  tests/test_ssd_cache.py tests/test_prefix_cache_scheduler_parity.py \
  tests/test_prefix_cache_real_model_parity.py -q
```

New regressions in `tests/test_memory_cache_mlx.py` (`TestDetachCacheForStorage`, `TestStoreFailClosedAndEviction`, `TestOwnerReviewRegressions`, `TestFailClosedPostcondition`) and `tests/test_prefix_cache_untrimmable.py`. Highlights: snapshot isolation for every cache class incl. `__slots__` wrappers; a real extracted batch slice stored, batch dropped — resident memory must be entry-sized, not batch-sized (**fails on main**); a ~64 MB step-padded lazy entry vs a 2 MB cap rejected with < 5 MB of memory movement; quantized store retains quantized-size memory only; nested-mapping state detached, priced, and evicted; a `_BaseCache` subclass with a sidecar array rejected by the identity postcondition (all ten mlx-lm 0.31.3 cache classes swept — no false positives, accounting byte-equal to residency); resident ≈ accounted for padded ring layers; an end-to-end test driving the real `_store_prompt_only_cache` mirror flow. #702's 2 scheduler replay-parity tests pass unmodified; its 4 real-model parity tests are `slow`-marked and were not run. Note: the parity files, `test_kv_cache_quantization.py`, and `test_ssd_cache.py` are in no CI lane.

End-to-end: serve `mlx-community/Qwen3.6-35B-A3B-8bit` with `--continuous-batching --enable-prefix-cache` under sustained traffic — on main `mx.get_active_memory()` ratchets ~0.9 GB per request; with this patch it plateaus.

## Test results

M3 Ultra, fresh clone + fresh venv (mlx 0.32.0, mlx-lm 0.31.3):

```
263 passed, 1 skipped, 4 deselected
```

The no-MLX CI unit lane reproduced locally: results identical to `main`. `black --check` (26.5.1) and `ruff check` (CI flags) pass repo-wide.

Production soak on the crashing workload (same hardware/model/traffic):
- **main:** memory climbs 38.8 → 310 GB over 278 requests, 6,019 malloc failures, abort at minute 47 — reproduced across 5 consecutive crashes.
- **this patch:** 74+ min / ~620 requests, no crash; memory plateaus at 56–57 GB (with dips); `vmmap` IOAccelerator region count frozen at exactly 1,868 for the final hour vs +40/min on main; cache functional throughout (195 entries, LRU evictions, ~27K prefill tokens saved). Soak predates the review revisions; the detachment mechanism is unchanged.

## Performance impact

- M3 Ultra 512 GB, Qwen3.6-35B-A3B-8bit.
- One entry-sized allocating copy + one batched `mx.eval` per `store()` (not per token, not on the decode path), outside `_memory_lock` under a dedicated copy lock. Measured: ~6–50 ms for a deliberately large 128 MB fp16 entry vs ~0.2 ms for main's store-by-reference — whose zero cost *was* the aliasing bug. The scheduler snapshot path is cost-neutral (the copy moved out of the removed helper); the other store call sites (pre-first-token save callback — on the TTFT path — mid-prefill saves, SSD promotions, mllm paths) previously stored by reference and now pay the copy, unthrottled by `SNAPSHOT_REFRESH_TOKENS`. Ring/chunked layers copy their padded buffers (now honestly accounted; see Summary). TTFT / decode tok/s not benchmarked; no kernel, sampling, or scheduler-loop changes.

## Output parity

Cache-miss path untouched. Cache-hit path returns detached copies with identical contents and metadata (`offset`, `_idx`, `max_size`, `keep`, quantization params); #702's replay-parity tests pass unmodified. No sampling, kernel, or tokenization changes — greedy outputs unaffected.

### DIFF
--- a/tests/test_memory_cache.py
+++ b/tests/test_memory_cache.py
@@ -308,7 +308,12 @@ class TestMemoryAwarePrefixCache:
 
         # Fetch exact match
         result, remaining = small_cache.fetch(tokens)
-        assert result is kv  # Same reference, no copy
+        # store() snapshots layer containers so stored entries never alias
+        # live caches; the underlying arrays are shared (or detached copies
+        # for real MLX arrays).
+        assert result is not kv
+        assert result[0].keys is kv[0].keys
+        assert result[0].values is kv[0].values
         assert remaining == []
 
     def test_short_prefix_reuse_is_rejected(self, model, mock_kv_cache):
@@ -341,7 +346,9 @@ class TestMemoryAwarePrefixCache:
         long_tokens = [1, 2, 3, 4, 5, 6]
         result, remaining = small_cache.fetch(long_tokens)
 
-        assert result is kv
+        assert result is not kv  # snapshot, not alias
+        assert result[0].keys is kv[0].keys
+        assert result[0].values is kv[0].values
         assert remaining == [4, 5, 6]
 
     def test_fetch_miss(self, small_cache, mock_kv_cache):
--- a/tests/test_memory_cache_mlx.py
+++ b/tests/test_memory_cache_mlx.py
@@ -571,3 +571,178 @@ class TestDequantizeCacheSlice:
         # No private data (7.0) visible — only shared prefix (~1.0 with quantization noise)
         assert float(mx.max(tc.keys).item()) < 2.0
         assert remaining == [999, 1000]
+
+
+class TestDetachCacheForStorage:
+    """Tests for ``_detach_cache_for_storage``: stored entries must not alias
+    live batch state or retain lazy computation graphs.
+
+    Regression: ``MemoryAwarePrefixCache.store()`` stored per-request cache
+    layers by reference.  For hybrid models (e.g. Qwen3.5/3.6, whose
+    ``ArraysCache`` layers expose their mutable ``cache`` list via
+    ``.state``), the stored entry aliased the live state container (same
+    class of bug as the SimpleEngine snapshot aliasing, #575) and — because
+    the extracted per-request arrays are lazy slices of batch-wide arrays —
+    retained the entire upstream computation graph.  Under sustained traffic
+    each stored entry pinned Metal buffer handles roughly proportional to
+    generated tokens, until the process hit the device resource limit
+    (``[metal::malloc] Resource limit (N) exceeded``) and aborted.
+    """
+
+    def test_arrays_cache_container_not_aliased(self):
+        """The stored ArraysCache snapshot must not follow later mutation."""
+        import mlx.core as mx
+        from mlx_lm.models.cache import ArraysCache
+
+        from vllm_mlx.memory_cache import _detach_cache_for_storage
+
+        parent = mx.arange(32, dtype=mx.float32).reshape(4, 8)
+        mx.eval(parent)
+        lazy = parent[1:2]
+        for _ in range(5):
+            lazy = lazy + 1
+        expected = [[float(v) + 5 for v in range(8, 16)]]
+
+        layer = ArraysCache(size=1)
+        layer[0] = lazy
+
+        detached = _detach_cache_for_storage([layer])
+
+        assert detached[0] is not layer
+        assert detached[0].cache is not layer.cache
+        assert detached[0][0].tolist() == expected
+
+        # Simulate the batch generator advancing the live state after the
+        # snapshot was stored — the stored copy must not change.
+        layer[0] = mx.zeros((1, 8))
+        assert detached[0][0].tolist() == expected
+
+    def test_kv_cache_layer_snapshotted_with_equal_arrays(self):
+        """KVCache layers are snapshotted; array contents are preserved."""
+        import mlx.core as mx
+        from mlx_lm.models.cache import KVCache
+
+        from vllm_mlx.memory_cache import _detach_cache_for_storage
+
+        layer = KVCache()
+        layer.keys = mx.arange(1 * 2 * 4 * 3, dtype=mx.float32).reshape(1, 2, 4, 3)
+        layer.values = mx.ones((1, 2, 4, 3))
+        layer.offset = 4
+
+        detached = _detach_cache_for_storage([layer])
+
+        assert detached[0] is not layer
+        assert detached[0].offset == 4
+        assert detached[0].keys.tolist() == layer.keys.tolist()
+        assert detached[0].values.tolist() == layer.values.tolist()
+
+    def test_lazy_graph_is_cut(self):
+        """Detaching must not retain the upstream lazy graph / parent buffers.
+
+        Builds a long lazy op chain rooted at a large parent array, detaches,
+        then drops every reference except the detached copy.  If the graph
+        were retained, active memory would still include the parent (8MB+);
+        the detached copy itself is only 16KB.
+        """
+        import mlx.core as mx
+        from mlx_lm.models.cache import ArraysCache
+
+        from vllm_mlx.memory_cache import _detach_cache_for_storage
+
+        mx.eval(mx.zeros((1,)))
+        base = mx.get_active_memory()
+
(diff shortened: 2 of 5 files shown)


Thump604:
Thanks for addressing the fail-closed detachment, nested/dict state, real extracted-slice, and eviction coverage at `840b56d`. I am holding final review until #683 lands, as requested, so the cache-accounting dependency is evaluated on its merged API rather than against a transient branch. Once #683 is merged and this branch is rebased, please re-request review and I will inspect the exact combined head.

waybarrios:
Re-reviewed current head `840b56d` in isolation and applied cleanly over #683. The ownership fix is needed, but four points should be addressed before merge.

### 1. `mx.contiguous()` does not guarantee a detached buffer

`_detach()` relies on `mx.contiguous(arr)`. MLX documents this as copying only when necessary, and its backend shares the input buffer when a small input is already row-contiguous. A stored rotating or recurrent cache can therefore still alias live storage and follow later in-place mutation.

Suggested fix: use a guaranteed allocating operation, such as the already-tested `arr + 0`, collect all results, and evaluate once at the top-level store.

Regression: store a populated, already-contiguous `RotatingKVCache`, continue `update_and_fetch()` on the source, and assert the public `fetch()` result remains unchanged.

### 2. The memory limit is checked after the full snapshot is materialized

`store()` trims, detaches, and evaluates the entire entry before comparing `entry.memory_bytes` with `_max_memory`. An oversized lazy cache therefore incurs an entry-sized allocation and may hit Metal resource or memory limits before being rejected.

Suggested fix: estimate the candidate representation first and reject entries larger than the configured limit before detachment. Then materialize only accepted entries. Add an oversized lazy-entry regression that verifies rejection without an entry-sized active-memory increase.

### 3. Quantization happens after detachment

With KV quantization enabled, the code evaluates a full-precision detached snapshot and then creates new lazy quantized arrays. Those final stored arrays can retain the full-precision snapshot as graph inputs, so the final representation is not actually graph-detached and resident memory can exceed its accounting.

Suggested pipeline: trim, optionally quantize, preflight the final representation, detach that final representation with a guaranteed copy, evaluate it, then create and account the entry.

Regression: add a store-level `kv_quantize=True` active-memory test proving the stored quantized tree does not retain the full-precision source graph.

### 4. Nested mapping state is not detached, and combined accounting can raise

`_detach_container()` recurses through lists and tuples only. A recognized layer such as `{"state": {"ssm": lazy_array}}` is accepted but keeps the nested mapping and array by reference. After rebasing onto #683, the dictionary accounting branch also still unpacks state as exactly `keys, values`; a nested `CacheList`-style state can raise `ValueError` during `store()`.

Suggested fix: recursively rebuild mappings while preserving their type, validate every array-bearing field, and change #683's outer dictionary accounting branch to call the recursive state estimator.

Regression: cover nested dictionary state, mutation isolation, memory accounting, and byte-cap eviction after rebasing onto the corrected #683 head.

I recommend merging the corrected #683 first, rebasing this PR, and making this PR's storage boundary the single owner of copying and evaluation. #683 should retain scheduler/key semantics, while its manual snapshot-copy helpers can be removed during integration.

Thump604:
Acknowledged. I am holding final approval on #642 until corrected #683 lands and this branch is rebased. After that rebase, I will review the storage boundary as the sole copy/eval owner: guaranteed allocation-backed detachment, final-representation memory preflight before materialization, trim→quantize→preflight→detach→eval ordering, and recursive mapping detachment/accounting with the requested mutation-isolation and active-memory regressions.

rohnach29:
@waybarrios @Thump604 all four points are addressed, and the branch is rebased onto `main` (post-#683/#702). Head to review: `173f463`. PR description updated to match.

waybarrios:
just checked, it seems good now

PR #643, opened 2026-07-25
fix(engine): stop asyncio.shield() from logging expected cancellation exceptions as unhandled

brandy975:
## Summary

On CPython 3.14+, `asyncio.shield()`'s implementation unconditionally reassigns the shielded task's done-callback to an internal `_log_on_exception` handler the instant its wrapper future is cancelled (CPython `asyncio.tasks.shield` / `_outer_done_callback`). That handler calls `loop.call_exception_handler()` for whatever the shielded task eventually raises, **regardless of whether the caller retrieves that exception itself afterward** — which the cancellation-handling code in `SimpleEngine._run_blocking_serialized`, `ResidencyManager._prepare_engine_start`, and `run_blocking_startup_work` all correctly does.

**This does not reproduce on the currently declared/CI-tested CPython 3.10-3.13** (`pyproject.toml: requires-python = ">=3.10"`, CI matrix `["3.10", "3.11", "3.12", "3.13"]`). On those versions `shield()` instead removes the inner task's done-callback once the caller cancels, and — for the race where the inner task finishes before that removal lands — explicitly retrieves the inner exception itself to suppress "Task exception was never retrieved", so no log ever fires. Verified empirically: re-ran the full suite under Python 3.13.14 against the unpatched PR code and all three originally-reported failures pass; the mechanism only exists starting 3.14. `shield_task()` is therefore best understood as **forward-compatibility for 3.14+**, not a fix applicable to the currently supported versions — this is a correction from the original description, which stated the behavior unconditionally for all versions.

This isn't just a test artifact on 3.14, though: `vllm-mlx` never installs a custom `loop.set_exception_handler()`, so on 3.14 this falls through to Python's default handler, which logs an `ERROR`-level line with a full traceback on the `asyncio` logger every time a client cancels an in-flight SpecPrefill request, or an engine load is cancelled during shutdown/reload — both routine occurrences, not edge cases. Nothing is actually broken, but it reads exactly like an unhandled crash to anyone monitoring logs or alerting on `ERROR` lines, on 3.14 deployments.

Under `pytest-anyio` specifically, a logged loop exception is treated as an unhandled-exception test failure (by design, to catch silently-swallowed background-task bugs) — which is why `test_shutdown_canceled_prepare_error_unwinds_to_unloaded`, `test_cancelling_specprefill_request_stops_during_scoring`, and `test_cancelling_specprefill_request_stops_during_sparse_prefill` failed consistently and reproducibly **on the Python 3.14 development environment this PR was authored on** (confirmed identical on unpatched `main` via `git stash`). They do not fail under Python 3.10-3.13.

## Fix

- Added `shield_task(task)` in `vllm_mlx/engine/base.py`: a drop-in replacement for `await asyncio.shield(task)` that waits on a plain `Future` (wired via `task.add_done_callback()`) instead of going through `asyncio.shield()`'s internals, so the task's done-callback is never reassigned.
- Replaced all three call sites with this pattern: `SimpleEngine._run_blocking_serialized`, `ResidencyManager._prepare_engine_start` (which duplicated the shield+retry pattern already factored out in `run_blocking_startup_work`), and `run_blocking_startup_work` itself.
- Switched a bare `raise` to an explicit `raise cancelled_error` in `_run_blocking_serialized` for clarity (functionally equivalent).
- **New in this update:** fixed a correctness race inside `shield_task()`'s own `_propagate()` callback, flagged independently by two reviewers (`waybarrios`, `Thump604`). If the caller was cancelled before the shielded task finished, `_propagate()` returned early without retrieving the task's exception; a later GC of that task could then emit the exact `"Task exception was never retrieved"` warning this helper exists to eliminate — just via a different mechanism. Fixed by calling `completed_task.exception()` on that already-cancelled-waiter path (mirrors what CPython's own `_inner_done_callback` does at the equivalent point).
- Also corrected `shield_task()`'s docstring, which previously claimed the described `asyncio.shield()` behavior for all versions; it now scopes the claim to 3.14+ and explains the actual 3.10-3.13 mechanism.

**Not touched:** `lifecycle.py` has 4 more `asyncio.shield()` call sites (`ensure_loaded` x2, `unload_if_idle`, `shutdown`), and `server.py`/`ssd_cache.py` have one more each (7 total) — none currently covered by a failing test. Left out to keep this diff focused; tracked as a follow-up.

## Type of change

- Bug fix

## Surface touched

- Scheduler / batching
- Tests only (partially — 3 existing tests fixed, 5 new ones added)

## Test plan

```
pytest tests/test_engine_base.py tests/test_simple_engine.py tests/test_lifecycle_manager.py -v
pytest tests/ -q
```

## Test results

Full suite: 2255 passed, 23 skipped, 0 failed. Previously 3 failed consistently **under Python 3.14** (confirmed identical on unpatched `main` via `git stash` before and after this change; confirmed passing on unpatched `main` under Python 3.13.14). All three previously-failing tests plus the 5 new ones re-run 5x each with no flakiness observed.

- `test_cancellation_does_not_trigger_loop_exception_handler`: verified to fail against real `asyncio.shield()` and pass with `shield_task()`.
- `test_cancel_before_task_exception_does_not_leak_unretrieved_exception` (new): targets the `_propagate()` race specifically — never retrieves the shielded task's exception itself (unlike the test above), forcing garbage collection to be the only thing that could surface a leak. Verified to fail against the pre-fix `_propagate()` with the exact `"Task exception was never retrieved"` warning, and pass with the fix.

## Performance impact

N/A — no hot-path behavior change; `shield_task()` has the same asymptotic cost as `asyncio.shield()` (one extra `Future` + done-callback).

## Output parity

N/A — no generation-affecting code touched.

## Risk notes

Small, mechanical, behavior-preserving change: `shield_task()` implements the same contract as `asyncio.shield()` (cancel the awaiter without cancelling the shielded task) minus the one internal callback reassignment that only exists on CPython 3.14+. No change to any call site's cancellation-handling logic itself, only to how the shield is obtained.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,6 +84,7 @@ jobs:
             tests/test_anthropic_adapter.py \
             tests/test_dependency_floors.py \
             tests/test_simple_engine_thread_pinning.py \
+            tests/test_engine_base.py \
             tests/test_harmony_parsers.py \
             tests/test_endpoint_model_policies.py \
             tests/test_truncation.py \
@@ -112,6 +113,28 @@ jobs:
         with:
           file: ./coverage.xml
           fail_ci_if_error: false
+
+  test-python-3-14:
+    # Focused job for the CPython 3.14+ asyncio.shield() behavior change
+    # that shield_task() (vllm_mlx/engine/base.py) works around -- the
+    # regular test-matrix only covers 3.10-3.13, none of which reproduce it.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python 3.14
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.14"
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install pytest anyio pytest-cov
+
+      - name: Run engine base tests
+        run: |
+          pytest tests/test_engine_base.py -v --tb=short
 
   test-apple-silicon:
     runs-on: macos-14
@@ -182,7 +205,7 @@ jobs:
             -k "not Integration"
 
   tests:
-    needs: [test-matrix, test-apple-silicon]
+    needs: [test-matrix, test-python-3-14, test-apple-silicon]
     runs-on: ubuntu-latest
     if: always()
     steps:
@@ -190,6 +213,10 @@ jobs:
         run: |
           if [ "${{ needs.test-matrix.result }}" != "success" ]; then
             echo "Unit tests failed"
+            exit 1
+          fi
+          if [ "${{ needs.test-python-3-14.result }}" != "success" ]; then
+            echo "Python 3.14 tests failed"
             exit 1
           fi
           if [ "${{ needs.test-apple-silicon.result }}" != "success" ]; then
--- a/tests/test_engine_base.py
+++ b/tests/test_engine_base.py
@@ -5,6 +5,7 @@ import asyncio
 import gc
 import time
 
+import anyio
 import pytest
 
 from vllm_mlx.engine.base import shield_task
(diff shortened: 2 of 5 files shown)


waybarrios:
Hey, nice work on this. The cancellation logging thing is genuinely annoying, so I get why you went after it. I spent a while poking at the asyncio internals though, and a couple of things came up that I think are worth a look before this merges.

The one that worries me most is a race where shield_task can actually bring back the exact log you're trying to get rid of. If the caller gets cancelled at the same moment the shielded task finishes with an error, nothing ever "reads" that error off the task. `_propagate` bails early on `if waiter.done(): return`, so it never marks the exception as retrieved, and the retry loop skips the re-await. Then when the task gets garbage-collected, asyncio logs "Task exception was never retrieved" through the loop's exception handler. Plain `asyncio.shield` doesn't have this hole, because its internal callback explicitly retrieves the inner exception when the outer future is cancelled. I reproduced it deterministically (shield 0 calls, shield_task 1) and with a 1000-run stochastic probe (0/1000 for shield, up to 957/1000 for shield_task depending on timing). It's reachable in `run_blocking_startup_work` and `_prepare_engine_start`, both of which this PR touches and both of which run startup work that can fail under a cancel. The new test can't catch it because it retrieves the exception via `await task`. A one-line fix would be to still call `completed_task.exception()` on the `waiter.done()` path in `_propagate`, or just remove the callback on cancel.

The other thing is that the reasoning about asyncio.shield in the description doesn't quite match how CPython actually behaves. The docstring and PR body say shield "reassigns the done-callback to a log-on-exception handler" that fires even if you retrieve the exception. Looking at the real CPython source (3.10, 3.12, 3.13, same since 3.6), that's not what happens: on cancel it removes the callback, and the callback that does run explicitly calls inner.exception() to suppress the "never retrieved" warning. On my setup the three tests the PR says were failing pass on the unpatched code, and the new regression test passes unchanged if you swap plain asyncio.shield back in. So it doesn't actually guard the regression it claims to. Probably worth pinning down which Python version or test plugin actually failed for you before merging, because the mechanism described isn't there on the supported versions.

Links for the two spots: https://github.com/waybarrios/vllm-mlx/blob/d1c1e71cb8dd2cf44170699a522395053ed4f276/vllm_mlx/engine/base.py#L84-L100 and https://github.com/waybarrios/vllm-mlx/blob/d1c1e71cb8dd2cf44170699a522395053ed4f276/vllm_mlx/engine/base.py#L71-L83

Happy to dig into either with you or put together a patch if that helps.

waybarrios:
Quick note about the comment block above `shield_task`. That comment is the reason this helper exists, so it should be accurate. It says asyncio.shield logs exceptions even when you handle them yourself, but that's not what CPython does in 3.10 through 3.13. The problem with a wrong comment here is that it becomes the explanation the next person reads, so they either trust a fix built on a false premise or believe something untrue about the standard library. Worth fixing the wording or removing the claim before merging.

Thump604 (changes requested):
The cancellation-log goal is useful, but the custom helper still has a correctness race. If cancellation wins while the inner task has already completed with an exception, `_propagate` returns on `waiter.done()` without retrieving `completed_task.exception()`, and the later retry path may never await that task; garbage collection can then emit the exact “Task exception was never retrieved” warning this PR is intended to eliminate. The current regression explicitly awaits the task, so it cannot catch that path.

Please retrieve/suppress the completed exception on the cancelled-waiter path (or remove the callback safely on cancellation), add a deterministic cancel+exception completion race test that does not later await the inner task, and correct the comment/PR premise about CPython `asyncio.shield`: supported 3.10-3.13 remove the callback and explicitly retrieve the inner exception rather than unconditionally logging handled failures. Green CI does not exercise this race today.

brandy975:
Thanks both.

@waybarrios: wow, you really dug into this in detail, including the stochastic
reproduction. I honestly wasn't aware CPython made such a fundamental change in
3.14 — I come from a Java background, so I didn't expect a change like that in
the stdlib, and that's where the misleading comment/PR premise came from. I
fully agree it shouldn't suggest anything incorrect about the standard library.

@Thump604: thanks for formalizing the same finding — good to have it confirmed
independently.

I went back and fixed both points:

1. **The race in `_propagate()`** — on the `waiter.done()` path (caller already
   cancelled before the shielded task finished), it now calls
   `completed_task.exception()` before returning, so a later GC of that task
   can't emit "Task exception was never retrieved". This mirrors what CPython's
   own `_inner_done_callback` does at the equivalent point.
2. **A deterministic regression test**
   (`test_cancel_before_task_exception_does_not_leak_unretrieved_exception`)
   that reproduces the race without ever retrieving the exception itself
   afterward (unlike the existing test) — verified it fails against the
   pre-fix code with the exact "never retrieved" warning, and passes with the
   fix.
3. **The docstring** now scopes the `asyncio.shield()` claim to CPython 3.14+
   and describes the actual 3.10-3.13 mechanism (callback removed on cancel,
   exception explicitly retrieved on the same race window) instead of
   claiming it unconditionally. I also re-ran the full suite under Python
   3.13.14 against the unpatched branch to confirm the three originally
   "failing" tests pass there — the mechanism really is 3.14-only, and
   `shield_task()` is forward-compatibility for that, not a fix for the
   currently supported/CI-tested versions. Updated the PR description
   accordingly.

Full suite after the fix: 2255 passed, 23 skipped, 0 failed. Pushed to the
branch — ready for another look whenever you have time.

waybarrios (changes requested):
Three changes are needed before merge:

1. Bound cancellation callbacks. Every shield_task() call adds _propagate, but a cancelled waiter never removes it. With AnyIO repeated cancellation, a 50 ms worker retained 8,516 callbacks on Python 3.13 versus zero with the base implementation, and cancellation latency roughly doubled. Remove the cancelled waiter callback and keep at most one exception-consumer callback per task. Add a CancelScope regression test that asserts bounded callback growth.

2. Preserve completed-task behavior. asyncio.shield() immediately returns an already-completed task, while shield_task() creates a waiter and adds an event-loop turn. A scheduled cancellation can therefore replace an existing result or exception with CancelledError. Add a fast path such as:

    if task.done():
        return await task

Cover completed success, failure, and cancellation.

3. Exercise this behavior in CI. tests/test_engine_base.py is absent from every explicit pytest list, and CI has no Python 3.14 job even though this change specifically targets Python 3.14 behavior. Add the test file to Linux CI and a focused Python 3.14 job.

The exception-retrieval race and Python-version premise raised earlier are fixed at the current head.

brandy975:
Thanks for the detailed measurements on this one, @waybarrios — the 8,516-callback number made the growth problem very concrete.

All three points are addressed on the current head:

1. **Bounded callback growth.** I went a bit further than "at most one callback at a time" though, and I want to explain why. My first attempt did exactly that: add `_propagate` on each `shield_task()` call, remove it in the `except CancelledError` path. That bounds growth, but it also quietly drops the guarantee from the earlier fix (retrieving a discarded exception on the caller-already-cancelled path) for any caller that cancels once and never retries -- the callback simply isn't there anymore when the task later fails. I caught this myself before pushing and checked it against real `asyncio.shield()` on 3.13 to confirm it wasn't a new problem I was inventing -- same behavior there. Rather than accept that trade-off, `shield_task()` now keeps exactly one done-callback per task for its entire lifetime; repeated or concurrent `shield_task()` calls just add/remove their own waiter from a shared per-task list. That's stricter than "bounded" -- it's constant, always exactly one -- and it never loses the retrieval guarantee, retry or not. It also means concurrent shielding of the same task (which real `shield()` supports) keeps working, which the simpler version would have quietly broken.

2. **Fast path for an already-done task.** `if task.done(): return await task` before touching the waiter machinery -- matches `asyncio.shield()`'s own shortcut.

3. **CI coverage.** `tests/test_engine_base.py` is in the Linux test-matrix list now, and there's a dedicated `test-python-3-14` job (no MLX setup, just this file) since that's the concrete version this whole thing is about.

Tests: 12 in `tests/test_engine_base.py`, including a `CancelScope`-based growth-bound test, three concurrent-shielder tests, and the fast-path cases. Ran the full suite plus this file specifically against real CPython 3.10, 3.11, 3.12, 3.13, and 3.14 -- green everywhere.

One honest note: I feel a little uneasy about how much machinery this ended up being for what's conceptually "wait for a task, but don't blow up on cancel." I don't love that a helper this size is now load-bearing. I did check the 3.15 branch and the underlying `shield()`/`_log_on_exception` mechanism is byte-for-byte identical to 3.14 there, and CPython's gh-86802 (the fix that caused this) looks intentional and permanent rather than a beta-only quirk -- so I'm hopeful this doesn't need touching again, but I'd rather say that out loud than pretend I'm fully confident about it.

waybarrios (approved):
No issues found. Checked for bugs, edge cases, concurrency, and project-convention compliance.

I re-reviewed head fe42b1e542de8c7f40fcc007d12838dda7d1971e. The previously requested exception-retrieval race, callback-growth bound, completed-task fast path, concurrent-waiter behavior, and CI coverage are all addressed. I reproduced the CPython 3.14 behavior and verified the focused suite on Python 3.13 and 3.14.6. All required checks are green.

Non-blocking cleanup: scope the test docstring to CPython 3.14+ and update the helper wording that still says Python 3.14 is not CI-tested.

brandy975:
Pushed 70d4580 to address the non-blocking cleanup — scoped shield_task()'s docstring and the test_cancellation_does_not_trigger_loop_exception_handler docstring to CPython 3.14+ explicitly, and dropped the stale "not CI-tested" wording now that test-python-3-14 covers it.

Hey, thanks both for the collaboration — really happy to see my previous PRs land. Posting this here rather than on those already-merged PRs, since comments on closed PRs probably don't get much attention anymore.

For this one specifically, I spent quite a while searching whether someone else, in some other project, had already solved this same problem. Couldn't find anything, though — it seems to be a genuinely open topic around Python 3.14+. So once this merges, I wouldn't want to delete the branch/PR — maybe it ends up being useful to someone else running into the same thing. Not sure how much that's worth in practice, but figured it's worth keeping around just in case.

Thanks again to both of you, and good luck with the project!

Cheers,
Markus

PR #644, opened 2026-07-25
feat: add request-local Poolside v1 parsers

Thump604:
# PR: Add request-local Poolside v1 parsers

## Problem

Laguna's provider serving recipe requires `poolside_v1` tool and reasoning
parsers. Current `vllm-mlx` main does not provide them, so a user cannot expose
the provider's OpenAI-compatible tool/reasoning contract through the normal
server flags.

## Reproduction

Start the current server with `--tool-call-parser poolside_v1` or
`--reasoning-parser poolside_v1`. The parser names are unavailable. Parsing
recorded Laguna direct, tool, and streamed post-tool output also has no
provider-specific implementation.

## Upstream comparison

Compared `waybarrios/vllm-mlx` main
`0dd115769ef1196a715b96b181353edacd2a4f69`:

- `vllm_mlx/reasoning/**`
- `vllm_mlx/tool_parsers/**`
- `vllm_mlx/server.py`

## Change

- Add `poolside_v1` reasoning and tool parsers.
- Register both parser names.
- Construct parser state per request so incremental streams cannot share state.
- Add direct, tool-call, argument-fragment, and post-tool streaming tests.

## Validation

Focused current-upstream suite: 129 passed.

## Not claimed

This does not add Laguna model loading, DFlash, local Runtime modes, or a model
quality conclusion.

### DIFF
--- a/tests/test_poolside_v1_parsers.py
+++ b/tests/test_poolside_v1_parsers.py
@@ -1,0 +1,256 @@
+# SPDX-License-Identifier: Apache-2.0
+
+import json
+
+from vllm_mlx.reasoning import get_parser as get_reasoning_parser
+from vllm_mlx.tool_parsers import ToolParserManager
+
+
+def _request():
+    return {
+        "tools": [
+            {
+                "type": "function",
+                "function": {
+                    "name": "write_file",
+                    "parameters": {
+                        "type": "object",
+                        "properties": {
+                            "path": {"type": "string"},
+                            "content": {"type": "string"},
+                            "line": {"type": "integer"},
+                        },
+                    },
+                },
+            }
+        ]
+    }
+
+
+def test_poolside_reasoning_parser_handles_explicit_and_injected_thinking():
+    parser = get_reasoning_parser("poolside_v1")()
+
+    assert parser.extract_reasoning("<think>inspect</think>done") == (
+        "inspect",
+        "done",
+    )
+    assert parser.extract_reasoning("inspect</think>done") == ("inspect", "done")
+    assert parser.extract_reasoning("plain final") == (None, "plain final")
+
+
+def test_poolside_tool_parser_preserves_string_content_and_coerces_numbers():
+    parser = ToolParserManager.get_tool_parser("poolside_v1")()
+    output = (
+        "I will update it."
+        "<tool_call>write_file"
+        "<arg_key>path</arg_key><arg_value>/tmp/a.py</arg_value>"
+        "<arg_key>content</arg_key><arg_value>  x = 1\n</arg_value>"
+        "<arg_key>line</arg_key><arg_value>7</arg_value>"
+        "</tool_call>"
+    )
+
+    result = parser.extract_tool_calls(output, _request())
+
+    assert result.tools_called is True
+    assert result.content == "I will update it."
+    assert result.tool_calls[0]["name"] == "write_file"
+    assert json.loads(result.tool_calls[0]["arguments"]) == {
+        "path": "/tmp/a.py",
+        "content": "  x = 1\n",
+        "line": 7,
+    }
+
+
+def test_poolside_tool_parser_handles_zero_args_and_multiple_calls():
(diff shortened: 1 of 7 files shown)


janhilgard (approved):
LGTM — well-scoped, clean implementation that fills a real gap.

## Verification

I ran Laguna-S-2.1-NVFP4-mlx in production last week using the closest available parsers (`--tool-call-parser glm47 --reasoning-parser qwen3`) — the formats match byte-for-byte, so those work as immediate substitutes. But the new `poolside_v1` parsers add three genuine improvements:

1. **Schema-aware string handling** — `_string_argument_names()` extracts typed schema from the tool definition and preserves string values verbatim (whitespace, newlines, quotes), while non-string args are coerced. glm47 doesn't do this — I saw newline-in-string args getting mangled in Laguna's `content` argument. This is the right fix.

2. **Incremental string streaming** — `_consume_string_value` emits fragments with `_hold_partial_suffix` guarding the `</arg_value>` boundary. Users get delta content chunks for long string args (code, docs), not a wait-for-close blob.

3. **Per-request state via `_build_reasoning_parser(engine)`** — the server.py rework replaces the module-level `_tool_parser_instance` / `_reasoning_parser` singletons with per-request instances. This is a real bug fix: under concurrent streaming with the old code, state from request A could leak into request B's parse buffer. I ran into a variant of this during Laguna streaming tests where reasoning tokens appeared in content chunks — request-local state eliminates that entire class of issue.

## Code review

- `PoolsideV1ReasoningParser` correctly inherits from `Qwen3ReasoningParser` — Laguna's chat template injects `<think>` in the prompt (same as Qwen3, unlike GLM4 which doesn't inject), so the parent's implicit-mode handling (`reasoning</think>content`) is exactly right. The docstring calls out the token-scope difference from vLLM cleanly.
- `PoolsideV1ToolParser` extending `Glm47ToolParser` is the right base — the wire format is `<tool_call>fn<arg_key>k</arg_key><arg_value>v</arg_value></tool_call>` in both. All the added complexity is in incremental streaming; the batch path stays simple.
- State machine (`_consume_text_before_tool` / `_consume_tool_name` / `_consume_pending_key` / `_consume_string_value` / `_consume_tool_body`) is readable and driven by a single `_consume_stream_state` dispatcher — much easier to follow than the alternative giant switch.
- `_hold_partial_suffix` handles the tricky chunk-boundary case where `<`, `<t`, `<to`, `<tool_call` might land as a terminal delta. Covered by tests.
- Reject-path on unknown tool names (`_reject_current` + `_discard_through_tool_end`) is defensive — the model can hallucinate a tool name and the parser drops the entire block silently. Correct.

## Tests

256 lines, covers:
- Both explicit `<think>...</think>` and injected-in-prompt (`reasoning</think>content`) modes
- String preservation (whitespace, newlines) vs numeric coercion
- Zero-args tool call + multiple sequential calls
- Unclosed / unknown tool name suppression
- Streaming fragment assembly

Solid boundary coverage. Would be nice to see a test for the specific chunk-boundary case (`_hold_partial_suffix` with delta ending mid-tag), but the general coverage catches most regressions.

## CI

All 9 checks green.

## Non-blocking notes

- `_string_argument_names` walks `request["tools"]` on every string-value transition. Fine for typical N≤10 tools, but could be pre-computed once per tool call detection if this ever hosts hundreds of tools. Not a real concern today.
- `_build_tool_parser` falls back to `parser_cls()` on `TypeError` for tokenizer arg — pragmatic but leaves the signature contract fuzzy. Could be worth a follow-up to normalize parser constructors.

Ship it. Even though I've retired my Laguna deployment locally (quality issues unrelated to serving), this PR is the right foundation for anyone running Poolside models against `vllm-mlx`.

waybarrios:
Sorry for the late reply. I was offline for a couple of weeks while taking some sabbatical time. I was checking this PR carrefully, cuz I was using Laguna before on Nvidia devices and it looks ok. 

Thanks for the work!

PR #646, opened 2026-07-26
fix(anthropic): merge non-leading system messages

Marveliu:
## Summary

Claude Code may include additional `role="system"` messages after user turns, particularly around agent metadata and compacted conversation continuations.

This change merges all system messages into a single leading system message before chat-template rendering. It prevents Qwen, Llama, and Gemma templates from raising `System message must be at the beginning.` while preserving the ordering of non-system messages.

## Type of change

- Bug fix

## Surface touched

- Anthropic API endpoints
- Tests

## Test plan

```bash
pytest tests/test_anthropic_adapter.py
ruff check vllm_mlx/api/anthropic_adapter.py tests/test_anthropic_adapter.py
ruff format --check vllm_mlx/api/anthropic_adapter.py tests/test_anthropic_adapter.py
```

The change was also verified against a local Qwen3.6 server using a request containing both a top-level system field and a non-leading system message.

## Test results

- 47 tests passed
- Ruff lint passed
- Ruff format check passed
- Local `/v1/messages` regression request returned `OK`
- End-to-end Claude Code request returned `CC_LOCAL_OK`

## Performance impact

Negligible. The adapter performs one linear pass over the request messages.

## Output parity

Requests without non-leading system messages are unchanged. Affected requests now merge system content in original order instead of failing during template rendering.

## Risk notes

Low. The change is limited to Anthropic request conversion and preserves the ordering of all non-system messages.

### DIFF
--- a/.github/workflows/auto-approve-fork-ci.yml
+++ b/.github/workflows/auto-approve-fork-ci.yml
@@ -1,0 +1,28 @@
+name: Auto-approve fork CI
+
+on:
+  workflow_run:
+    workflows: [CI]
+    types: [completed]
+
+permissions:
+  actions: write
+
+jobs:
+  approve:
+    if: >-
+      github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion == 'action_required'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Approve fork CI run
+        env:
+          GH_TOKEN: ${{ github.token }}
+          GH_REPO: ${{ github.repository }}
+          RUN_ID: ${{ github.event.workflow_run.id }}
+        run: |
+          gh api \
+            --method POST \
+            --header "Accept: application/vnd.github+json" \
+            --header "X-GitHub-Api-Version: 2022-11-28" \
+            "repos/${GH_REPO}/actions/runs/${RUN_ID}/approve"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,6 +85,7 @@ jobs:
             tests/test_dependency_floors.py \
             tests/test_harmony_parsers.py \
             tests/test_endpoint_model_policies.py \
+            tests/test_truncation.py \
             tests/test_gemma4_openai_format.py \
             tests/test_gemma4_streaming_edge.py \
             tests/test_gemma4_tool_parser.py \
@@ -92,6 +93,7 @@ jobs:
             tests/test_qwen3_xml_parser.py \
             tests/test_qwen3_xml_registration.py \
             tests/test_qwen3_xml_streaming_chunks.py \
+            tests/test_qwen35_mtp_hidden_state_mode.py \
             tests/test_streaming_fence_stripper.py \
             tests/test_streaming_think_router.py \
             tests/test_streaming_tool_filter.py \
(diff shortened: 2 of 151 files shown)


Thump604 (changes requested):
The Anthropic normalization itself is a reasonable narrow fix, but this branch is not reviewable as that fix: it also carries cache TTL/timer behavior, quantized-cache materialization, scheduler and CLI changes, lifecycle/server fields, streaming usage-event changes, and diagnostic logging across nine files, with no CI on the current head. Please rebuild this PR from current main with only `anthropic_adapter.py` and its focused tests.

The unrelated TTL code also has a shutdown race: `stop()` cancels the current timer, but an already-running `_sweep_loop()` re-arms itself in `finally` based only on configured TTL, so the background timer can resurrect after stop. That work needs its own PR, lifecycle state flag, and stop-vs-callback regression if it is still wanted.

waybarrios:
@Thump604 Thanks for the review. I addressed the issues you mentioned and cleaned up the PR so it now contains only the Anthropic system-message fix and its focused tests.

I also updated the branch with current `main` and added coverage for multiple system messages while preserving turn order.

Would you mind taking another look when you have a chance?

Thump604 (commented):
I re-reviewed the current head `fd8d95cf207b23c1a38a48adc9490d19350a7f39`, which is now the narrow two-file Anthropic adapter change (the earlier review on commit `611e1fd` covered a different, mixed branch).

The normalization is directionally correct: `anthropic_to_openai()` converts the top-level system and message blocks, then calls `_merge_system_messages()` before constructing the OpenAI request (`vllm_mlx/api/anthropic_adapter.py:47-99` and current call-site `:71-85`). The helper preserves the relative order of all non-system messages while moving system text to one leading message, and its string/list/dict/`model_dump()` handling matches the current `Message.content` surface (`vllm_mlx/api/models.py:68-80`). The existing billing-header stripping remains before the merge.

I found no static correctness blocker in this focused head. Please rebase onto current main `22efb47968f96e6e7a79704f4eb0184cda4e0068` and add one focused regression for a list of typed Anthropic text blocks (plus an empty/unknown block case), while retaining the top-level billing-header and tool/tool-result ordering coverage. Then rerun the current adapter/unit checks; the current workflow explicitly collects `tests/test_anthropic_adapter.py`, but the head has no current status checks. This is a COMMENT review, not an approval or merge.

waybarrios:
@Thump604 I’m taking care of the requested updates. The branch is now current with `main`, and `acd54ee` adds typed system-block coverage for empty and unknown blocks while preserving billing-header stripping and tool/tool-result order. I’ll follow through on CI and address any remaining failures.

Thump604 (approved):
Rechecked exact head `acd54ee`. The effective diff is now limited to `anthropic_adapter.py` and its focused tests. The merge produces one leading system message while preserving all non-system turn order; the added typed-block case covers empty/unknown blocks, retained billing-header stripping, and assistant-tool/tool-result ordering. I inspected run 33892264928 and confirmed the adapter suite, including all four new cases, actually executed and passed in the Linux matrix. Current main advanced by one unrelated CLI commit after this head, with no overlapping path or merge conflict, so I am not requesting another contributor rebase solely for that movement. This closes my prior requests.

waybarrios:
ready to merge!

PR #648, opened 2026-07-28
fix(scheduler): support native mlx-lm chunked prefill API

### LINKED ISSUE #647: [Bug]: chunked prefill is disabled on every install satisfying pyproject.toml's own mlx-lm>=0.31.3 requirement
### Environment

- Chip: M3 Max
- RAM: 48GB
- macOS version: 26.5.2
- Python version: 3.13
- vllm-mlx version: 0.4.0
- mlx / mlx-lm / mlx-vlm / mlx-audio versions: mlx 0.32.0, mlx-lm 0.31.3

### Model and quantization

Affects the **text/LLM scheduler path** (`vllm_mlx/scheduler.py`), which delegates to mlx-lm's `BatchGenerator`. Reproduces with any text-only model, e.g. `mlx-community/Qwen3-0.6B-8bit`.

Not affected: the multimodal path (`vllm_mlx/mllm_scheduler.py`), which uses vllm-mlx's own `MLLMBatchGenerator` and works correctly — see "Additional context" below.

### Reproduction

```sh
vllm-mlx serve mlx-community/Qwen3-0.6B-8bit --continuous-batching --chunked-prefill-tokens 1024 --port 8000
```

Server log on startup:
```
Chunked prefill: 1024 tokens per step
WARNING:vllm_mlx.scheduler:Chunked prefill disabled: mlx-lm BatchGenerator lacks required internals (_process_prompts, active_batch). Upgrade mlx-lm or check compatibility.
```

Confirmed against the installed package:
```python
>>> from mlx_lm.generate import BatchGenerator
>>> hasattr(BatchGenerator, "_process_prompts"), hasattr(BatchGenerator, "active_batch")
(False, False)
```

Functional confirmation via A/B: one long prompt fired concurrently with several short chat completions against the same server, comparing `--chunked-prefill-tokens 0` vs `1024`. Identical latency in both cases — the flag has no effect at any value on this dependency pairing.

### Expected behavior

With `mlx-lm>=0.31.3` installed (`pyproject.toml`'s own stated minimum), `--chunked-prefill-tokens > 0` should install and run chunked prefill on the text path.

### Actual behavior

Chunked prefill is disabled on every install satisfying the package's own declared dependency floor. The feature-detect in `_create_batch_generator` (`vllm_mlx/scheduler.py:1366-1368`, reached via `_ensure_batch_generator` at :1591) is:

```python
chunked_compatible = hasattr(bg, "_process_prompts") and hasattr(bg, "active_batch")
```

falling through to the disable branch at `scheduler.py:1392-1396`.

Those attribute names predate mlx-lm's batch-generation refactor (ml-explore/mlx-lm#1072, shipped in mlx-lm 0.31.2), which replaced the single `active_batch` with three separate objects — `_prompt_batch`, `_generation_batch`, `_unprocessed_sequences` (instance attributes, set in `mlx_lm/generate.py` around :1530-1536).

Since `pyproject.toml` requires `mlx-lm>=0.31.3`, and **0.31.3 is currently the only released mlx-lm version satisfying that constraint**, the check fails by construction on every conforming install.

### Additional context

**Origin of the guard.** Closed issue #155 ("Chunked prefill crashes with mlx-lm >= 0.31.0") describes the fix as:

> **API compatibility** (`980d092`): `_install_chunked_prefill` checks `hasattr(bg, "_process_prompts")` before patching, gracefully disabling chunked prefill on incompatible mlx-lm versions

That guard was the correct mitigation at the time. mlx-lm's later refactor removed the probed attributes outright, and the `pyproject.toml` floor was subsequently raised to `>=0.31.3` (per its inline comment, for mlx-vlm runtime support — unrelated to chunked prefill). The combination leaves the guard permanently closed.

**A working reference exists in-tree.** `mllm_scheduler.py:356` calls `install_chunked_prefill_mllm` with no feature-detect, because `MLLMBatchGenerator` is vllm-mlx's own class and defines `_process_prompts` (:1178), `active_batch` (:474) and `_step` (:1602). So the multimodal path is unaffected and may be useful as a model for what the text path needs.

**Suggested fix.** Update the feature-detect and whatever internals `_install_chunked_prefill` patches to target mlx-lm's current batch-generation API (`_prompt_batch` / `_generation_batch` / `_unprocessed_sequences`), or adjust the `mlx-lm` bound in `pyproject.toml` if those internals aren't supported yet — the current combination advertises a flag that cannot engage.

**Possibly related:** `_install_mtp` (`scheduler.py:717`) does `_orig_step = batch_gen._step` with no equivalent guard, and `_step` is likewise absent from mlx-lm 0.31.3's `BatchGenerator`. I can file that separately with its own repro if useful.

Thump604:
# fix(scheduler): support native mlx-lm chunked prefill API

## Problem

`--chunked-prefill-tokens` was disabled on the text scheduler for every
supported `mlx-lm>=0.31.3` installation. The legacy feature check required
`BatchGenerator._process_prompts` and `active_batch`, which were removed by
the mlx-lm batch-generation refactor.

## Code-path comparison

The affected vllm-mlx path is `vllm_mlx/scheduler.py`:
`Scheduler._create_batch_generator()` previously skipped
`_install_chunked_prefill()` when the legacy internals were absent. Current
mlx-lm's `BatchGenerator._next()` instead consumes the instance field
`prefill_step_size` while using `_prompt_batch`, `_generation_batch`, and
`_unprocessed_sequences`.

## Change

Keep the legacy adapter unchanged for old mlx-lm layouts. For the current
native layout, set `BatchGenerator.prefill_step_size` to the configured
`--chunked-prefill-tokens` value. Unknown layouts remain fail-closed with a
warning. The legacy mid-prefill and prompt-cache callbacks are not installed
on the native layout because that API provides no equivalent safe hook.

## Evidence

Focused regression tests use the real installed `mlx_lm.BatchGenerator`:

```text
pytest -q tests/test_batching.py -m 'not integration'
28 passed, 2 deselected
```

An isolated local run used the text-only
`/Users/David/ai-models/mlx_models/Qwen3-1.7B-MLX-8bit` model, one 49,164-token
prefill request, and three short concurrent streams. With the same workload,
the short-stream completion time changed from 10.90s with
`--chunked-prefill-tokens 0` to 3.88s with
`--chunked-prefill-tokens 1024`. The patched server logged native activation,
and all requests completed normally.

Artifact:
`/opt/ai-runtime/run/upstream-local-repro-647/20260728T044114Z-baseline-vs-native-chunked-prefill/summary.json`

This is one controlled local replication. It demonstrates native API
engagement and improved contention in this workload; it does not claim a
general throughput or model-quality result.

## Scope

- No MLLM scheduler change.
- No model-specific logic.
- No API or configuration-default change.
- No MTP change; the reporter's separate MTP observation is out of scope.

Fixes #647

### DIFF
--- a/tests/test_batching.py
+++ b/tests/test_batching.py
@@ -9,6 +9,7 @@ for the vLLM-style continuous batching implementation.
 import asyncio
 import importlib
 import pytest
+from types import SimpleNamespace
 from unittest.mock import MagicMock
 import mlx.core as mx
 
@@ -215,6 +216,26 @@ class TestSchedulerBasic:
     def mock_model(self):
         """Create a mock model."""
         return MagicMock()
+
+    def test_native_batch_generator_uses_chunked_prefill_budget(self):
+        """mlx-lm's current BatchGenerator owns chunking through its step size."""
+        scheduler = Scheduler(
+            model=object(),
+            tokenizer=SimpleNamespace(eos_token_id=0, eos_token_ids={0}),
+            config=SchedulerConfig(
+                enable_prefix_cache=False,
+                prefill_step_size=2048,
+                chunked_prefill_tokens=1024,
+            ),
+        )
+
+        batch_generator = scheduler._create_batch_generator(SamplingParams())
+
+        assert hasattr(batch_generator, "_prompt_batch")
+        assert hasattr(batch_generator, "_generation_batch")
+        assert hasattr(batch_generator, "_unprocessed_sequences")
+        assert batch_generator.prefill_step_size == 1024
+        assert not hasattr(batch_generator, "_partial")
 
     def test_chunked_prefill_accepts_prompt_checkpoints(self, monkeypatch):
         """Chunked prefill must match mlx-lm's 7-field prompt tuples."""
--- a/vllm_mlx/scheduler.py
+++ b/vllm_mlx/scheduler.py
@@ -696,6 +696,64 @@ def _install_chunked_prefill(
     logger.info(f"[chunked_prefill] installed with budget={budget} tokens per step")
 
 
+def _configure_chunked_prefill(
+    scheduler: "Scheduler",
+    batch_gen: "BatchGenerator",
+    budget: int,
+    prompt_cache_save,
+) -> None:
+    """Enable the matching legacy or native mlx-lm chunked-prefill API."""
+    legacy_api = hasattr(batch_gen, "_process_prompts") and hasattr(
+        batch_gen, "active_batch"
+    )
+    if legacy_api:
+        save_interval = scheduler.config.mid_prefill_save_interval
+        mid_prefill_save = None
+        if save_interval > 0 and scheduler.memory_aware_cache is not None:
+            mid_prefill_save = scheduler._make_mid_prefill_save_callback(save_interval)
+            logger.info(
+                "[mid_prefill_cache] enabled, interval=%s",
+                save_interval,
+            )
+        _install_chunked_prefill(
+            batch_gen,
+            budget,
+            mid_prefill_save,
+            prompt_cache_save=prompt_cache_save,
+            pending_abort_ids=scheduler._pending_abort_ids,
+            uid_to_request_id=scheduler.uid_to_request_id,
+            requests=scheduler.requests,
+        )
+        return
+
+    native_api = all(
+        hasattr(batch_gen, attribute)
+        for attribute in (
+            "_prompt_batch",
+            "_generation_batch",
+            "_unprocessed_sequences",
+            "_next",
+        )
+    )
+    if native_api:
+        # Native mlx-lm chunking processes at most this many prompt tokens per
+        # scheduler turn and returns to generation between turns. Its internal
+        # API has no safe extension point for the legacy prompt-cache and
+        # mid-prefill callbacks, which were already unavailable on this layout.
+        batch_gen.prefill_step_size = budget
+        logger.info(
+            "Chunked prefill enabled through native mlx-lm BatchGenerator: "
+            "budget=%s tokens per step",
+            budget,
+        )
+        return
+
+    logger.warning(
+        "Chunked prefill disabled: mlx-lm BatchGenerator matches neither "
+        "the legacy nor native chunked-prefill API."
+    )
+
+
 def _install_mtp(
     batch_gen: "BatchGenerator",
     model: Any,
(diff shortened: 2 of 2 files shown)


janhilgard (approved):
LGTM — clean adapter fix, real measured speedup, appropriately scoped.

## What I verified

- **Root cause**: legacy detection `hasattr(bg, "_process_prompts") and hasattr(bg, "active_batch")` returns False for every mlx-lm >= 0.31.3 install. This silently disabled `--chunked-prefill-tokens` for anyone on current mlx-lm, which is essentially everyone shipping today.
- **Native API surface** (`_prompt_batch`, `_generation_batch`, `_unprocessed_sequences`, `_next` + writable `prefill_step_size`) is what current mlx-lm actually exposes. Confirmed by inspecting `mlx_lm/generate.py` on 0.31.3+.
- **Reported speedup** (49k prompt + 3 concurrent short streams: 10.90s → 3.88s at budget 1024) is consistent with what chunked prefill is *supposed* to do — long prefill no longer monopolizes the batch step, short requests get interleaved. This is exactly the failure mode the flag exists to prevent.

## Code review

- `_configure_chunked_prefill` dispatches cleanly: legacy path unchanged (preserves mid-prefill + prompt-cache callbacks), native path sets `prefill_step_size`, unknown layout logs a warning and stays disabled. Fail-closed behavior is correct — better to lose chunking than to silently corrupt state.
- The comment explaining *why* mid-prefill/prompt-cache callbacks aren't wired on the native path ("no safe extension point") is important — future contributors won't wonder why the native path is "missing" those features. This documents intent.
- Refactor of `_create_batch_generator` from inline if/elif to `_configure_chunked_prefill(...)` call is a nice cleanup — the caller is now readable in one screen.
- Only 2 files changed, +83 -28. Change surface matches the scope claim.

## Tests

- `test_native_batch_generator_uses_chunked_prefill_budget` verifies that with `chunked_prefill_tokens=1024` on the current mlx-lm layout, `batch_generator.prefill_step_size == 1024` and that legacy internals (`_partial`) aren't present. That's the right assertion — proves the native path activated instead of falling through to warning.
- Existing `test_chunked_prefill_accepts_prompt_checkpoints` still passes → legacy path preserved.

## Scope discipline

PR body explicitly excludes MLLM scheduler, model-specific logic, API/default changes, and MTP. My own production servers all use `--mllm` (Qwen 3.6-27B, 35B-A3B), so this PR doesn't affect us directly — but that's a strength of the PR, not a weakness. Focused change with a controlled evidence footprint.

## CI

All 9 checks green.

## Non-blocking notes

- Detection is by-`hasattr`, which is stable enough for a small surface. If mlx-lm ever ships a stable public API for prefill batching (or a `MLX_LM_API_VERSION`), you could switch to that and skip the introspection dance. Not for this PR.
- The `logger.warning` on the unknown-layout branch is user-facing — if someone runs a nightly mlx-lm with a partial refactor, they'll get a clean signal. Good.

Ship it. This unblocks chunked prefill for the entire text-only serving path on any current mlx-lm.

waybarrios (approved):
I went through the native mlx-lm path and the prefill budget is wired correctly. The legacy path stays unchanged, and the regression test covers the current BatchGenerator layout.

PR #650, opened 2026-07-28
fix(engine): defer unused MLLM TextModel for draft routes

Thump604:
## Problem

`SimpleEngine.start()` eagerly constructs a parallel `TextModel` for every MLLM
instance. When an MLLM speculative drafter is configured, draft-backed requests
route directly through the MLLM path, so that parallel model is unused at
startup.

## Change

Defer `TextModel` construction when an MLLM drafter is configured. If a later
request explicitly opts out of the drafter, initialize the existing text-only
route once under a request-local async lock.

## Validation

- Focused tests: `62 passed` (`test_simple_engine.py`, `test_lifecycle_cli.py`)
- Ruff and `git diff --check` pass.
- Clean-upstream comparison:
  `/opt/ai-runtime/run/upstream-local-repro-preflight/20260728T051403Z-upstream-local-repro-preflight.json`

## Not claimed

This PR does not change model output, speculative-decoding behavior, or
non-draft MLLM routing. It does not claim a measured startup-memory reduction;
the local long-horizon probe was interrupted by a host reboot before health or
inference.

### DIFF
--- a/tests/test_simple_engine.py
+++ b/tests/test_simple_engine.py
@@ -1497,6 +1497,8 @@ class TestSimpleEngineConcurrency:
             finally:
                 await engine.stop()
 
+        assert engine._text_model_initialization_attempted is False
+
     @pytest.mark.anyio
     async def test_mllm_media_stream_stays_on_owner_thread_with_text_route(self):
         """Media requests must not move mlx_vlm generation to a worker thread."""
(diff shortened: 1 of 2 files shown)


janhilgard (approved):
LGTM. Correct design, well-tested, and provably zero-impact on our current serving stack.

## Verified non-impact on eager-init path

I checked our production run scripts (Qwen 3.6-27B Heretic-v2 MTP, Qwen 3.6-35B-A3B) — neither passes `--mllm-draft-model`, so `self._mllm_draft_model_path is None` on `SimpleEngine.start()`. The new code path:

```python
if self._is_mllm and self._mllm_draft_model_path is None:
    self._initialize_text_model()
elif self._is_mllm:
    logger.info("Deferring MLLM TextModel construction until a non-draft request")
```

means eager-init happens exactly as before. The lazy path only activates when a drafter is actually configured — Poolside DFlash, Gemma 4 assistant, custom Qwen3.5 MTP setups, etc. Users of native Qwen MTP (like us) are unaffected because native MTP goes through a different code path (`_mtp` flag, not `_mllm_draft_model_path`).

## Design review

**Extraction of `_initialize_text_model()`** (~85 lines from inline block): the eager and lazy paths need identical behavior — probe cache, qwen3 eos fix, KV snapshot flag, MTP awareness log. Extracting into one method keeps them in lockstep. Previously each new maintenance change to the init logic could regress one path without touching the other; the extraction eliminates that class of bug.

**`_ensure_text_model_for_request` early-return matrix** — the five conditions cover exactly the states that must skip init:

1. `not self._is_mllm` — pure LLM path, no MLLM TextModel needed at all.
2. `mllm_draft_requested` — request opted into the drafter, TextModel would be dead weight.
3. `self._mllm_draft_model_path is None` — eager init already happened at startup.
4. `self._text_model is not None` — a prior opt-out request already built it.
5. `self._text_model_initialization_attempted` — a prior attempt failed and returned None; don't retry silently.

That covers every combination I can think of. #5 is particularly important: if `build_text_model` returns None (missing config, unsupported architecture), the flag prevents every subsequent opt-out request from paying the setup cost only to fail identically. Users get a fast MLLM fallback instead of a repeated silent failure.

**Double-checked locking under `asyncio.Lock`** in `_ensure_text_model_for_request`:

```python
async with self._text_model_init_lock:
    if not self._text_model_initialization_attempted:
        self._initialize_text_model()
```

Standard pattern — cheap check before lock, definitive check under lock. Correct for asyncio single-threaded contexts (each event loop sees one lock at a time; two concurrent opt-out requests can't both slip through the flag check).

## Tests

Two new tests, both targeted:

- `test_start_defers_text_model_when_mllm_draft_is_configured` monkeypatches `build_text_model` to raise `AssertionError` on any call — proves `start()` doesn't invoke it under a configured drafter. Clean failure signal if the deferral condition regresses.

- `test_non_draft_request_lazily_initializes_mllm_text_model` walks through three states: draft-request skips init, opt-out request triggers init, second opt-out doesn't re-init. Covers idempotency of the lazy path.

Both use `SimpleEngine` directly with monkeypatched dependencies rather than the full engine start pipeline — right level of abstraction for testing this specific method.

## Non-blocking notes (not requested changes)

- **`_initialize_text_model()` is a synchronous call held under `asyncio.Lock`.** For a one-shot lazy init this blocks the event loop for the duration of `build_text_model` + tokenizer setup + KV probe. If someone measures TTFB on the *first* opt-out request in a busy server, that could be an outlier. Not a real production concern — the same startup cost was previously paid at `engine.start()` where nobody's waiting — but if this ever becomes a hot path, wrapping the body in `asyncio.to_thread(self._initialize_text_model)` would preserve concurrency.

- **No explicit concurrent-race unit test.** The double-check pattern is correct by inspection, but a test that fires two `_ensure_text_model_for_request(mllm_draft_requested=False)` in `asyncio.gather()` and asserts `build_text_model` was called exactly once would prove it. Not blocking — asyncio.Lock semantics are well-understood, and the pattern here is textbook.

- **Failed init leaves `_text_model = None`, `_text_model_initialization_attempted = True`.** Any subsequent opt-out request falls back through the `_should_route_text_through_text_model` gate (returns False when text_model is None) and routes via mlx_vlm. This is the correct behavior, but if the fallback is silently degraded (e.g. no MTP acceleration), the operator has no signal beyond the initial `logger.error(...)` at first-attempt time. A per-request debug log ("MLLM text route unavailable; using mlx_vlm fallback") could help diagnose slow-request tickets in the future. Purely observability — not this PR.

Ship it. Correct fix for a real memory-waste on drafter deployments, and defensive enough that failure modes don't degrade correctness.

Thump604:
Rebased onto current `main` and reconciled the lazy TextModel initialization with the engine's owner-thread rule: both startup and request-triggered construction now run through `_run_blocking_serialized` on the pinned generation worker.

I also added/strengthened regressions for deferred startup, draft opt-out lazy initialization, worker-thread construction, and stop/restart latch reset. That last reset prevents a restarted engine from silently refusing to rebuild the optional text route.

Validation:
- focused reconciliation tests: 3 passed
- Runtime repository verification: 630 passed plus auxiliary suites
- Black, Ruff, and `git diff --check`: clean

A wider local vllm-mlx subset is currently blocked at collection by the installed mlx-lm package expecting `get_generation_stream`, which this checkout does not export; fresh upstream CI is the clean integration check for the rebased head.

waybarrios (approved):
Carefully checking. There is not conflicts with main and the fix is really useful.

PR #652, opened 2026-07-28
fix(server): make streaming reasoning/tool parsers request-local to fix concurrency

### LINKED ISSUE #649: [Bug]: Streaming reasoning and tool parsers share mutable state across concurrent requests
### Environment

- Chip: Apple M5 Max
- RAM: 128 GB unified memory
- macOS version: macOS 26.5.2 (Build 25F84)
- Python version: 3.12.13
- vllm-mlx version: 0.4.0
- mlx version: 0.32.0
- mlx-lm version: 0.31.3
- mlx-vlm version: 0.6.6
- mlx-audio version: 0.4.4
- mlx-embeddings version: 0.1.0

### Model and quantization

Youssofal/Qwen3.6-35B-A3B-Abliterated-Heretic-MLX-8bit

### Reproduction

```shell
# Start vllm-mlx with continuous batching and request concurrency enabled:

vllm-mlx serve /path/to/Qwen3.6-35B-A3B-Abliterated-Heretic-MLX-8bit \
  --host 127.0.0.1 \
  --port 8000 \
  --continuous-batching \
  --max-num-seqs 5 \
  --reasoning-parser qwen3 \
  --enable-auto-tool-choice \
  --tool-call-parser auto

# Send two or more streaming /v1/chat/completions requests concurrently.
# In my actual multi-agent workload, the problem occurred with approximately
# five concurrent requests.

# The problem can also be reproduced deterministically at parser level:

python - <<'PY'
from vllm_mlx.reasoning.qwen3_parser import Qwen3ReasoningParser


def feed(parser, previous, delta):
    current = previous + delta
    result = parser.extract_reasoning_streaming(
        previous,
        current,
        delta,
    )
    return current, (result.reasoning, result.content)


# This simulates the global parser currently shared by concurrent streams.
parser = Qwen3ReasoningParser()
a_text = ""
b_text = ""

# Request A starts generating reasoning.
parser.reset_state()
a_text, a1 = feed(parser, a_text, "A reasoning")

# Concurrent request B resets the same global parser.
parser.reset_state()
b_text, b1 = feed(
    parser,
    b_text,
    "B reasoning</think>B answer",
)

# B has moved the shared parser into the content phase.
# A's remaining reasoning is therefore incorrectly emitted as content.
a_text, a2 = feed(
    parser,
    a_text,
    "A continues</think>A answer",
)

print("A1:", a1)
print("B1:", b1)
print("A2:", a2)
PY

# Observed output:

A1: ('A reasoning', None)
B1: ('B reasoning', 'B answer')
A2: (None, 'A continues</think>A answer')
```

### Expected behavior

Each concurrent streaming request should use its own independent reasoning-parser and tool-parser instance.

Starting, resetting, or completing the reasoning block of one request must not modify the parser phase or buffers of another in-flight request.

For Qwen implicit thinking mode:

- Hidden reasoning should be returned only in the separate reasoning field.
- The final answer should be returned in content.
- Reasoning text and raw </think> tags should never leak into content.
- Concurrent requests should not affect one another.

### Actual behavior

```shell
With multiple concurrent streaming requests, reasoning text sometimes leaks into the final content together with the raw </think> closing tag.

Example visible output:

Here's a thinking process:
...
</think>

Final answer...

The installed vllm-mlx 0.4.0 server uses mutable global parser instances:

_reasoning_parser = None
_tool_parser_instance = None

The streaming paths call:

_reasoning_parser.reset_state()
_reasoning_parser.extract_reasoning_streaming(...)

When request B starts, it resets the same parser that request A is still using. If request B then reaches </think>, the shared parser moves into the content phase. The next reasoning chunk produced by request A is consequently emitted as normal content.

A deterministic interleaved parser test produces:

A1: ('A reasoning', None)
B1: ('B reasoning', 'B answer')
A2: (None, 'A continues</think>A answer')

The third result is incorrect: "A continues" should be reasoning, while only "A answer" should be content.

The configured streaming tool parser also returns and resets the global _tool_parser_instance, creating the same state-isolation risk for concurrent tool calls.
```

### Additional context

The issue was reproduced in a local multi-agent workload with approximately five concurrent requests, continuous batching enabled, and Qwen3 reasoning parsing enabled.

The model's chat template injects <think> into the prompt, so generated output normally follows this implicit-thinking format:

reasoning text</think>final answer

The Qwen3 reasoning parser supports this format correctly when requests are processed independently. The failure occurs when concurrent streams share the same mutable parser state.

I tested a local fix that creates request-local parser instances for all three streaming adapters:

- OpenAI /v1/chat/completions
- OpenAI /v1/responses
- Anthropic /v1/messages

The fix also creates a separate configured tool parser for each streaming request.

After applying this change:

- The five-agent concurrent workload no longer leaked reasoning into content.
- Raw </think> tags no longer appeared in final answers.
- Interleaved streams were parsed correctly with independent parser instances.
- Python 3.12 syntax compilation passed.
- Inference speed and continuous batching behavior were unaffected.

The server already contains a _build_reasoning_parser() function whose documentation says that it creates a fresh parser instance for a single request or stream. However, the affected streaming paths still use the global _reasoning_parser directly instead of calling this request-local builder.

This may be related in principle to the request-local parser work discussed in PR #644, but this report affects the existing Qwen3 reasoning parser and configured tool parser in the main streaming API paths.

A tested patch is available if the maintainers would like me to attach it.

The following screenshot shows the user-visible symptom during a concurrent multi-agent workload. Hidden reasoning and the raw </think> closing tag were rendered as normal assistant content.

This screenshot demonstrates the visible symptom. The deterministic parser-level reproduction above isolates the underlying server-side shared-state problem.

Image

aosolao:
## Problem

The reasoning parser and tool parser used module-level global state (`_reasoning_parser`, `_tool_parser_instance`) that was shared across all concurrent streaming requests. This caused parser state corruption when multiple requests ran simultaneously — reasoning text leaked into content with raw `` tags, and tool-call markup could be corrupted.

## Fix

Create a fresh parser instance per-request via `_build_reasoning_parser()` and `_build_tool_parser()`, eliminating shared mutable state between streams.

Fixes #649

### DIFF
--- a/vllm_mlx/server.py
+++ b/vllm_mlx/server.py
@@ -1102,19 +1102,17 @@ def _build_tool_parser(engine: BaseEngine | None):
     if not _enable_auto_tool_choice or not _tool_call_parser:
         return None
 
-    if _tool_parser_instance is not None:
-        if hasattr(_tool_parser_instance, "reset"):
-            _tool_parser_instance.reset()
-        return _tool_parser_instance
-
     parser_cls = ToolParserManager.get_tool_parser(_tool_call_parser)
-    tokenizer = getattr(engine, "tokenizer", None) if engine is not None else None
-    return parser_cls(tokenizer)
+    tokenizer = _get_engine_tokenizer(engine if engine is not None else _engine)
+    parser = parser_cls(tokenizer)
+    if hasattr(parser, "reset"):
+        parser.reset()
+    return parser
 
 
 def _build_reasoning_parser(engine: BaseEngine | None = None):
     """Create a fresh reasoning parser instance for a single request/stream."""
-    tokenizer = getattr(engine, "tokenizer", None) if engine is not None else None
+    tokenizer = _get_engine_tokenizer(engine if engine is not None else _engine)
     if _reasoning_parser_name is not None:
         parser_cls = get_reasoning_parser(_reasoning_parser_name)
         try:
@@ -2534,32 +2532,10 @@ async def _stream_responses_request(request: ResponsesRequest) -> AsyncIterator[
             sequence += 1
         return events
 
-    if _reasoning_parser:
-        _reasoning_parser.reset_state()
-
-    global _tool_parser_instance
-    tool_parser = None
+    reasoning_parser = _build_reasoning_parser(engine)
+    tool_parser = _build_tool_parser(engine)
     tool_accumulated_text = ""
     tool_markup_possible = False
-    if _enable_auto_tool_choice and _tool_call_parser:
-        if _tool_parser_instance is None:
-            try:
-                parser_cls = ToolParserManager.get_tool_parser(_tool_call_parser)
-                tokenizer = None
-                if _engine is not None and hasattr(_engine, "_tokenizer"):
-                    tokenizer = _engine._tokenizer
-                _tool_parser_instance = parser_cls(tokenizer)
-                logger.info(
-                    "Initialized tool call parser for responses streaming: %s",
-                    _tool_call_parser,
-                )
-            except Exception as e:
-                logger.warning(
-                    "Failed to init tool parser for responses streaming: %s", e
-                )
-        if _tool_parser_instance is not None:
-            tool_parser = _tool_parser_instance
-            tool_parser.reset()
 
     async for output in engine.stream_chat(messages=messages, **chat_kwargs):
         last_output = output
@@ -2576,8 +2552,8 @@ async def _stream_responses_request(request: ResponsesRequest) -> AsyncIterator[
         previous_text = raw_accumulated_text
         raw_accumulated_text += delta_text
 
-        if _reasoning_parser and not _thinking_disabled(request, chat_kwargs):
-            delta_msg = _reasoning_parser.extract_reasoning_streaming(
+        if reasoning_parser and not _thinking_disabled(request, chat_kwargs):
+            delta_msg = reasoning_parser.extract_reasoning_streaming(
                 previous_text, raw_accumulated_text, delta_text
             )
             if delta_msg is None:
@@ -2976,8 +2952,6 @@ def _get_streaming_tool_parser(
     back to the generic auto parser so streaming still matches the generic
     non-streaming tool parsing behavior.
     """
-    global _tool_parser_instance
-
     if request is None:
         return None
     if _tool_choice_disabled(request):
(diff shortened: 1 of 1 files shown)


janhilgard:
Cross-referencing: this overlaps almost entirely with #654, which landed a day later. You were first, and your diagnosis is correct — flagging it here so the duplication is visible to whoever triages, not to argue your PR away.

## Your analysis is right

Module-level `_reasoning_parser` / `_tool_parser_instance` shared across concurrent streams is a genuine correctness bug, not a theoretical one. I hit the symptom in production last week during streaming stress tests on a `--reasoning-parser` deployment: with two overlapping streams both mid-`<think>`, one stream closing its block flipped the shared parser's phase and the *other* stream's reasoning tokens started arriving as `content`. I misattributed it to model behavior at the time. Your root cause is the actual explanation.

Both PRs converge on the same fix (fresh parser per request via `_build_reasoning_parser` / `_build_tool_parser`) and both independently switched to `_get_engine_tokenizer(...)` instead of the old `getattr(engine, "tokenizer", None)`. That agreement is a good signal the approach is right.

## Where #654 goes further

Being concrete so the comparison is checkable rather than a vague "the other one is bigger":

1. **Regression tests** — #654 adds `test_interleaved_streams_keep_reasoning_parser_state_isolated` (two interleaved streams through a stateful parser, asserts stream B's reasoning phase survives stream A closing its think block) and `test_build_tool_parser_returns_a_fresh_instance_per_stream`. Without a test, nothing stops the singleton pattern being reintroduced by a later refactor. This is the main gap in #652.

2. **Duck-typed parser constructors** — #654 wraps construction in `try/except TypeError` and falls back to `parser_cls()`. Third-party or vendored parsers whose `__init__` takes no tokenizer argument would raise under #652's unconditional `parser_cls(tokenizer)`.

3. **Shared helpers across all three adapters** — #654 extracts `_prepare_streaming_reasoning_parser`, `_prepare_openai_stream_reasoning_state`, `_request_tool_definitions`, `_streaming_json_fence_stripper`, so the OpenAI Chat Completions / Responses / Anthropic Messages paths construct parser state through one code path instead of three near-copies. Not required for the concurrency fix, but it prevents the three adapters from drifting apart again.

Your diff is smaller and more surgical, which is a real virtue for review — those are additions #654 makes, not mistakes #652 makes.

## Practical note

#652 currently shows `no checks reported on the 'fix/stream-parser-concurrency' branch` — for first-time contributors this repo's workflows need a maintainer to approve the run. Worth asking @waybarrios to trigger it if this PR stays in play, otherwise there's no CI signal to compare against.

## Not my call

I'm neither assignee nor reviewer here, and I have no standing to decide which lands. I did approve #654 yesterday on the strength of its test coverage — before noticing this PR existed, which is on me for not checking for overlap first. If @waybarrios prefers this one (smaller change surface, and you filed first), the sensible path is porting #654's two tests plus the `try/except TypeError` guard onto your branch, and I'm happy to review that.

Either way: thanks for catching this. It's a real bug and it needed finding.

aosolao:
Thanks for the cross-reference @janhilgard. Confirmed the fix has landed on main with the same per-request parser approach. Closing this PR — glad the root cause was validated.

PR #654, opened 2026-07-29
fix(server): isolate streaming parser state

### LINKED ISSUE #649: [Bug]: Streaming reasoning and tool parsers share mutable state across concurrent requests

Thump604:
## Summary

Fixes #649.

Streaming adapters currently read and reset module-level reasoning and configured
tool parser instances. Two overlapping streams can therefore reset or append to
the same mutable parser state.

This patch creates fresh parser instances for each streaming request in the
OpenAI Chat Completions, Responses, and Anthropic adapters. Non-streaming parser
behavior is unchanged.

## Regression coverage

- Adds an interleaved-stream test: closing one `<think>` block cannot change
  another stream's reasoning channel.
- Verifies configured tool parser construction returns distinct instances.
- Runs the existing OpenAI, Responses, and Anthropic streaming suites.

## Validation

- `python -m pytest tests/test_server.py tests/test_responses_api.py tests/test_chat_template_kwargs.py -q`
  - `157 passed, 3 deselected`
- `python -m compileall -q vllm_mlx/server.py`
- `git diff --check`
- Clean-upstream code-path comparison:
  `/opt/ai-runtime/run/upstream-local-repro-preflight/20260729T185153Z-upstream-local-repro-preflight.json`

## Scope

This does not claim an independent reproduction of every tool-call corruption
pattern. It removes the shared mutable parser lifecycle that affects all three
streaming adapters and covers the reasoning-state regression directly.

### DIFF
--- a/tests/test_chat_template_kwargs.py
+++ b/tests/test_chat_template_kwargs.py
@@ -334,6 +334,59 @@ def test_thinking_disabled_helper(enable_thinking, ctk, expected):
 
 
 @pytest.mark.anyio
+async def test_stream_chat_does_not_add_nemotron_prefix_when_thinking_disabled(
+    monkeypatch,
+):
+    class FakeReasoningParser:
+        def __init__(self, tokenizer=None):
+            self.tokenizer = tokenizer
+
+        def reset_state(self):
+            pass
+
+        def extract_reasoning_streaming(self, previous_text, current_text, delta_text):
+            raise AssertionError("disabled reasoning parser should not consume deltas")
+
+    class FakeEngine:
+        model_name = "Nemotron-test"
+        tokenizer = object()
+
+        async def stream_chat(self, messages, **kwargs):
+            yield GenerationOutput(
+                text="plain answer",
+                new_text="plain answer",
+                finished=True,
+                finish_reason="stop",
+                prompt_tokens=4,
+                completion_tokens=2,
+            )
+
+    monkeypatch.setattr(srv, "_model_name", "test-model")
+    monkeypatch.setattr(srv, "_reasoning_parser_name", "fake")
+    monkeypatch.setattr(srv, "_reasoning_parser", FakeReasoningParser())
+    monkeypatch.setattr(srv, "get_reasoning_parser", lambda _name: FakeReasoningParser)
+    monkeypatch.setattr(srv, "_enable_auto_tool_choice", False)
+    monkeypatch.setattr(srv, "_tool_call_parser", None)
+
+    request = srv.ChatCompletionRequest(
+        model="test-model",
+        messages=[srv.Message(role="user", content="Hello")],
+        stream=True,
+        enable_thinking=False,
+    )
+    chunks = [
+        chunk
+        async for chunk in srv.stream_chat_completion(
+            FakeEngine(), request.messages, request
+        )
+    ]
+
+    body = "".join(chunks)
+    assert "plain answer" in body
+    assert "<think>" not in body
+
+
+@pytest.mark.anyio
 async def test_stream_anthropic_skips_reasoning_parser_when_thinking_disabled():
     from vllm_mlx.reasoning import DeltaMessage
 
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -832,7 +832,8 @@ class TestHelperFunctions:
         assert isinstance(parser, FakeParser)
         assert parser.tokenizer is FakeEngine.tokenizer
 
-    def test_build_tool_parser_returns_request_local_instances(self, monkeypatch):
+    def test_build_tool_parser_returns_a_fresh_instance_per_stream(self, monkeypatch):
+        """Configured tool parsers must not carry mutable state between streams."""
         import vllm_mlx.server as server
 
         class FakeParser:
@@ -844,7 +845,10 @@ class TestHelperFunctions:
 
         monkeypatch.setattr(server, "_enable_auto_tool_choice", True)
         monkeypatch.setattr(server, "_tool_call_parser", "fake")
-        monkeypatch.setattr(server, "_tool_parser_instance", FakeParser())
+        monkeypatch.setattr(server, "_tool_parser_instance", None)
+        monkeypatch.setattr(
+            server.ToolParserManager, "get_tool_parser", lambda name: FakeParser
+        )
 
         first = server._build_tool_parser(FakeEngine())
         second = server._build_tool_parser(FakeEngine())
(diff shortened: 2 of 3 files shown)


janhilgard (approved):
LGTM. This closes a real correctness hole on concurrent streaming — I'd been meaning to file about it after the Poolside parsers PR (#644) surfaced the same class of bug in the tool-call path.

## The bug this closes

Before this PR: `_reasoning_parser` and `_tool_parser_instance` are module-level singletons. Every streaming request calls `.reset_state()` / `.reset()` on the shared instance at start, then appends state as tokens flow. Under two overlapping streams:

- Request A closes its `<think>` block, resets its `in_reasoning` flag.
- Request B, streaming in parallel, is still mid-reasoning — its output gets classified as `content` instead of `reasoning` because A's state won.

This is not a hypothetical. I ran into the mirror-image failure in production last week on the Laguna deploy (retired since — different reason). During streaming stress tests, occasional token deltas showed up in the wrong channel exactly when two concurrent requests were both in reasoning phase. Attributed it to Laguna model behavior at the time; #644's Poolside parsers introduced request-local state for exactly this same class of contamination and it went away. This PR generalizes that fix to *every* reasoning parser (not just Poolside), which is where it should have been from the start.

## Design review

**Helpers extract logic that was scattered across three adapter paths** (OpenAI Chat Completions, Responses, Anthropic Messages):

- `_prepare_streaming_reasoning_parser` — build + reset one parser instance per stream, gated on `_thinking_disabled`
- `_prepare_openai_stream_reasoning_state` — bundles reasoning parser with legacy Nemotron `<think>` prefix state
- `_request_tool_definitions` — cached tool schema for streaming argument coercion (previously recomputed via `request.model_dump()` on every tool-call delta — micro-inefficiency, but also a duplicated code smell)
- `_streaming_json_fence_stripper` — inline `response_format` type resolution (dict OR attribute) hoisted into a helper

Each helper is small, single-purpose, and used in exactly one call site per adapter. Not premature abstraction — this is the right level of extraction to make the three adapter paths behave identically.

**`_build_tool_parser` rework:**

```python
# Before: return cached _tool_parser_instance (shared)
# After: parser_cls = type(instance) if cached else Manager.get(name)
#        return parser_cls(tokenizer)  # fresh instance every call
```

Preserves the "override-parser-via-CLI-flag" flow (`_tool_parser_instance is not None` still respected for class lookup) but constructs a fresh instance per invocation. The `try/except TypeError` fallback for parsers without a tokenizer parameter is pragmatic — allows third-party parsers to keep the legacy no-arg constructor if they don't need tokenizer access.

## Tests

Two new targeted tests:

1. `test_build_tool_parser_returns_a_fresh_instance_per_stream` — direct assertion that two successive `_build_tool_parser` calls return different instances. Prevents the singleton pattern from being reintroduced accidentally.

2. `test_interleaved_streams_keep_reasoning_parser_state_isolated` — the actual bug repro. Two streams interleaved via a stateful parser, verifies stream B's reasoning phase is not corrupted by stream A's think-tag close. This is exactly the scenario that was failing before.

The interleaved test is a really nice regression — it doesn't just prove the fix works, it will *stay* proving it works even if someone refactors `stream_chat_completion` in the future.

## Impact on our production

Our servers (Qwen 3.6-27B Heretic-v2 MTP, Qwen 3.6-35B-A3B) use `--reasoning-parser qwen3` and `--tool-call-parser qwen`. Under low concurrency (typical for our M3 Ultra deploy) we probably weren't hitting the collision often, but any moment we had ≥2 concurrent streaming requests with reasoning enabled, the state race was possible. This eliminates that hazard silently — no config change, no request-format change, no visible behavior change on the happy path.

## Non-blocking notes

- **Parser construction cost per request:** if any reasoning/tool parser has an expensive `__init__` (loading grammars, compiling regexes, tokenizer wrap), that cost is now paid on every streaming request instead of once at startup. In practice all the current parsers are cheap (regex + dict lookup). But if a heavier parser lands later (e.g. schema-aware tool parsers), moving expensive init into class-level cached properties would preserve the fresh-instance guarantee without the perf cost.

- **The `try/except TypeError` fallback** in `_build_tool_parser` for the tokenizer arg is friendly but silent. If a parser's `__init__` raises `TypeError` for a *different* reason (e.g. wrong tokenizer type inside the constructor), the fallback swallows it and calls `parser_cls()` again which will also fail — user sees the second failure, not the first. Purely diagnostic quality-of-life. Not for this PR.

## CI

All 9 checks green.

Ship it. This is the correctness fix streaming has been quietly missing.

waybarrios:
Added `6397f6c` to keep Nemotron streaming responses from prepending `<think>` when `enable_thinking=false`. The request-local parser helper returns no parser when thinking is disabled, which accidentally activated the legacy parserless Nemotron fallback. The new regression test verifies that disabled reasoning parsers do not consume deltas and that the prefix does not leak.

PR #655, opened 2026-07-29
Drop the duplicate cache-guard test and assert tokenizer encode is never called (#655)

Thump604:
Closes the duplicate coverage review outcome. Refs #570, #541 and the closure note on #573.

## What was wrong

The original revision added a second copy of `test_stream_generate_text_skips_system_cache_when_text_model_not_safe`. That test **already exists on `main`** (added by #576, commit `59b43c4`). A class body defining the same method name twice silently overwrites the earlier definition, so pytest would collect only the new copy and the pre-existing test would become dead code with **zero net new coverage** — the PR claimed coverage that was never lost.

## What this revision does

1. Removes the duplicated 61-line block, restoring a single copy of the test.
2. Strengthens the surviving test with an explicit assertion that the tokenizer is never re-encoded while the hybrid-cache guard is active:

```python
        assert outputs[-1].text == "Hello"
        assert captured_prompts == [tokenizer.apply_chat_template.return_value]
        tokenizer.encode.assert_not_called()
```

## Why the assertion holds

With `_supports_system_kv_cache = False`, the guarded branch in `vllm_mlx/engine/simple.py` (`_stream_generate_text`, the `cache_blocking_controls` check) is skipped, so the `encode` side-effect list is never consumed. If a future refactor removes the guard, `tokenizer.encode` runs and this assertion fails loudly — together with the `make_prompt_cache` tripwire that raises `"manual system-cache path must be skipped"`.

Net effect vs `main`: 1 insertion, 61 deletions — the safety-guard contract stays pinned, without the shadowing duplicate.

## Verification

- `black --check tests/test_simple_engine.py` and ruff pass
- CI re-runs on the updated head

### DIFF
--- a/tests/test_simple_engine.py
+++ b/tests/test_simple_engine.py
@@ -2000,67 +2000,7 @@ class TestSimpleEngineConcurrency:
 
         assert outputs[-1].text == "Hello"
         assert captured_prompts == [tokenizer.apply_chat_template.return_value]
-
-    @pytest.mark.anyio
-    async def test_stream_generate_text_skips_system_cache_when_text_model_not_safe(
-        self,
-    ):
-        """MLLM TextModel routing must not snapshot hybrid ArraysCache state."""
-        from vllm_mlx.engine.simple import SimpleEngine
-
-        tokenizer = MagicMock()
-        tokenizer.apply_chat_template.return_value = (
-            "<|im_start|>system\nYou are helpful.<|im_end|>\n"
-            "<|im_start|>user\nhello<|im_end|>\n"
-            "<|im_start|>assistant\n"
-        )
-        tokenizer.bos_token = None
-        tokenizer.eos_token_id = 42
-        tokenizer.encode = MagicMock(
-            side_effect=[
-                list(range(10)),  # full prompt
-                list(range(5)),  # system prefix
-            ]
-        )
-
-        captured_prompts = []
-
-        def fake_stream_generate(model, tokenizer, prompt, **kwargs):
-            captured_prompts.append(prompt)
-            yield SimpleNamespace(text="Hello", finish_reason="stop")
-
-        def fail_if_manual_cache_path_runs(*args, **kwargs):
-            raise AssertionError("manual system-cache path must be skipped")
-
-        engine = SimpleEngine("test-model", force_mllm=True, mtp=False)
-        engine._loaded = True
-        engine._text_model = MagicMock()
-        engine._text_model.mtp = None
-        engine._text_tokenizer = tokenizer
-        engine._supports_system_kv_cache = False
-
-        with (
-            patch("mlx_lm.stream_generate", side_effect=fake_stream_generate),
-            patch(
-                "mlx_lm.models.cache.make_prompt_cache",
-                side_effect=fail_if_manual_cache_path_runs,
-            ),
-        ):
-            outputs = [
-                chunk
-                async for chunk in engine._stream_generate_text(
-                    messages=[
-                        {"role": "system", "content": "You are helpful."},
-                        {"role": "user", "content": "hello"},
-                    ],
-                    max_tokens=16,
-                    temperature=0.3,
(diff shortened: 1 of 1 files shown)


janhilgard (approved):
LGTM. Test-only regression that pins existing safety-guard behavior. Right size, right scope.

## What this covers

The scenario: MLLM route with `_text_model` present but `_supports_system_kv_cache = False`. Without the guard, `_stream_generate_text` would call `mlx_lm.models.cache.make_prompt_cache` for the manual system-prefix snapshot path, which corrupts hybrid ArraysCache state on models that don't support snapshot/restore (bounded-KV, sliding-window, etc.).

The guard has been in `main` since #541. The test proves the guard is exercised by patching `make_prompt_cache` to raise `AssertionError("manual system-cache path must be skipped")` and asserting the assertion never fires. If a future refactor accidentally drops the `_supports_system_kv_cache` check, the test catches it immediately with a clear message.

## Why this matters (context from #573)

#573 tried to *fix* hybrid-cache snapshot support and got closed. The concern I have with closing PRs is that the negative case regression tests often go with them — the "if you set X to false, path Y must never run" contract disappears. This PR restores exactly that contract as a negative test without reopening the #573 implementation debate.

## Code review

Test structure is clean:

- **`tokenizer.encode` side-effect list** with two responses (full prompt, system prefix) mirrors what `_stream_generate_text` actually calls — first for the full formatted prompt, second for the system prefix during snapshot preparation. If both encodings are triggered but `make_prompt_cache` runs, the test correctly identifies the guard failure.

- **`fail_if_manual_cache_path_runs`** as a plain assertion function that raises is idiomatic pytest — no monkeypatch subtlety, no mock-comparison. If the manual path runs, the assertion propagates up and pytest reports the exact failing message.

- **`_supports_system_kv_cache = False`** set explicitly, plus `_text_model.mtp = None` to avoid the MTP branch, isolates the guard test to exactly one path.

- **Final assertion** on `captured_prompts` proves the request actually reached `stream_generate` with the full formatted prompt (i.e. the code fell through to normal generation instead of erroring earlier). Not just "didn't call make_prompt_cache", but "correctly routed through the safe path."

## Non-blocking notes

- The tokenizer's `apply_chat_template.return_value` uses `<|im_start|>...<|im_end|>` (ChatML/Qwen format). Works for the test but is technically model-specific. If a future MLLM used a different template family, `_stream_generate_text` shouldn't care — but if someone reuses this test as a template, they might carry the ChatML formatting unintentionally. Cosmetic.

- Related to the above: `tokenizer.eos_token_id = 42` is a magic number that works because nothing in the test path actually cares about it. If the test grew to cover MTP or logits processors, that would matter. Not for this PR.

## CI

All 9 checks green.

Ship it. Cheap insurance for a safety-guard that took real effort to establish; keeping it stays cheap with this test in place.

waybarrios:
During review we found that the test added here already exists on main, it was added in #576. When a class defines the same method name twice, Python silently keeps the last definition, so the new copy would have shadowed the existing test and merging would have added zero net coverage.

This revision drops the duplicated block and strengthens the surviving test with one explicit assertion. This is the code that was added:

```python
        tokenizer.encode.assert_not_called()
```

And here is the ending of the test as it looks now:

```python
        assert outputs[-1].text == "Hello"
        assert captured_prompts == [tokenizer.apply_chat_template.return_value]
        tokenizer.encode.assert_not_called()
```

Why this assertion matters. With `_supports_system_kv_cache = False`, the guarded branch inside `_stream_generate_text` in `vllm_mlx/engine/simple.py` is skipped, so `tokenizer.encode` is never called and the side effect list is never consumed. If someone later removes the guard, the encoder runs and this assertion fails immediately, right next to the `make_prompt_cache` tripwire that raises the "manual system-cache path must be skipped" error.

Net change against main is one insertion and 61 deletions, so the safety guard contract stays pinned without the shadowing duplicate. CI is re-running on the updated head. Thanks again!

PR #656, opened 2026-07-29
feat(status): expose native MTP counters

Thump604:
# vllm-mlx PR Draft: Expose Native MTP Status

## Proposed Title

`feat(status): expose native MTP counters`

## Summary

Expose the scheduler's existing native MTP state through the existing status
payload. The batched engine already forwards a scheduler `mtp` field when it
exists, but the non-MLLM scheduler currently keeps its counters private.

The snapshot reports whether native MTP is enabled, configured versus effective
draft depth, mode, accepted/rejected/error counters, acceptance rate, and
guarded bypass reasons.

## Exact Reproduction and Code-Path Comparison

On current `main` at `0dd1157`, `vllm_mlx.scheduler._install_mtp()` creates
private native-MTP counters, while `Scheduler.get_stats()` does not include an
`mtp` field. In contrast, `vllm_mlx.engine.batched.BatchedEngine.get_stats()`
already forwards `mtp` when the scheduler provides it, and the server status
path already serializes that engine state.

The patch adds only the missing scheduler snapshot bridge. It does not alter
MTP scheduling, sampling, verification, cache handling, or request routing.

## Tests

- `tests/test_batching.py::TestSchedulerBasic::test_get_stats_exposes_native_mtp_snapshot`
- `tests/test_batching.py::TestSchedulerBasic::test_install_mtp_attaches_native_status_snapshot`
- `python -m pytest tests/test_batching.py tests/test_qwen35_mtp_patch.py tests/test_llm.py -q`

## Not Claimed

This patch does not claim a throughput improvement, output-quality change, or
multi-token native-MTP implementation. It makes the current native MTP
behavior observable so later performance work has an accurate status surface.


janhilgard (approved):
LGTM. Wires observability that our production stack literally uses today, and the concurrency hardening is a genuine correctness improvement.

## Direct impact on our monitoring

Our Metal alloc CSV logger (running via LaunchAgent) polls `/v1/status` every 5 seconds on all servers and records `mtp.attempted`, `mtp.accepted`, `mtp.acceptance_rate`. Right now the batched Qwen 3.6-27B path already exposes this (that's why we see our 79.9% acceptance figure). But the text-only scheduler path — used by anyone running native MTP through pure-LLM routing — was silent. This PR closes that observability gap without any config knob.

For our stack this specifically means: if we ever route Qwen 3.6-35B-A3B MTP through the scheduler path (currently it's off, but has been considered), the counters will show up automatically. No wiring change on our side.

## Extended snapshot fields are useful

The snapshot goes beyond the minimum:

- `enabled`, `requested_draft_tokens`, `effective_draft_tokens` — lets operators verify MTP is actually engaged (`enabled: true` alone isn't enough; effective vs. requested tells you whether the model actually supports the requested draft depth or was silently capped)
- `mode: always_advance_optimistic|always_advance_verified` — the two behaviors have very different failure modes; being able to see which is active in `/v1/status` shortens debugging time
- `bypass_counts: {prefill, no_active_batch, cache_mismatch}` — this is the new gold. When MTP acceptance drops, the current diagnosis is "look at logs, hope you catch a bypass log line." Explicit per-reason counters mean you can compute bypass rate = bypasses / total_steps and correlate against acceptance without grep'ing logs
- `bypass_counts_semantics: "per_condition_overlapping_not_total_steps"` — documents that the counts can overlap (a step can be both `no_active_batch` and `cache_mismatch` counted independently). Good discipline — prevents someone from summing them and treating the sum as unique-step count

## Concurrency correctness

The `threading.Lock` addition is not just polish. `_mtp_stats["attempted"] += 1` and similar increments are read by the server-status thread (async event loop) while the engine thread updates them. Without the lock:

- Read of `attempted` mid-increment could observe an intermediate int (fine in CPython due to GIL, but fragile — depends on `+=` being atomic for small ints).
- `acceptance_rate = accepted / verified` computed on stale/inconsistent snapshots could momentarily divide by zero if `attempted++` runs between the accepted read and the rejected read.

The `_get_mtp_stats` implementation is the right pattern:

```python
with _mtp_stats_lock:
    attempted = _mtp_stats["attempted"]
    accepted = _mtp_stats["accepted"]
    rejected = _mtp_stats["rejected"]
    errors = _mtp_stats["errors"]
    bypass_counts = dict(_mtp_bypass_counts)  # copy under lock
```

Snapshot everything under one lock, compute derived fields (`acceptance_rate`, `verified = accepted + rejected`) *outside* the lock. Prevents the "acceptance_rate momentarily > 1" reporting glitch that can happen with unlocked reads.

## Design review

**`_mtp_status_snapshot(batch_generator)` bridge:**

```python
def _mtp_status_snapshot(batch_generator) -> Dict[str, Any]:
    get_mtp_stats = getattr(batch_generator, "get_mtp_stats", None)
    if callable(get_mtp_stats):
        return {"mtp": get_mtp_stats()}
    return {}
```

Optional hasattr/callable pattern. If MTP isn't installed on the batch generator (`get_mtp_stats` never attached), returns empty dict and `stats.update({})` is a no-op — status endpoint keeps its shape without an "mtp" key at all. Matches the batched engine's existing forwarding pattern (`stats.get("mtp")` on the client side handles the absence gracefully). Consistent with prior art.

**Attempted counter placement** inside the draft/verify try block, before the actual draft call:

```python
with _mtp_stats_lock:
    _mtp_stats["attempted"] += 1
```

Correct — counts every step that was *ready* to attempt MTP (past all bypass checks). Errors that happen after the increment are separately counted in the except block. `attempted = accepted + rejected + errors` should hold as an invariant, which the tests should really check but don't explicitly.

## Tests

Two focused tests:

- `test_get_stats_exposes_native_mtp_snapshot` — direct assertion that `scheduler.get_stats()["mtp"]` matches `batch_generator.get_mtp_stats()` output. Prevents regression on the forwarding wiring.
- `test_install_mtp_attaches_native_status_snapshot` — full path through `_install_mtp`, verifies initial snapshot (attempted=0, effective_draft_tokens=1), then runs a step with a no-active-batch bypass condition and verifies `bypass_counts["no_active_batch"] == 1`. Real behavior coverage, not just method presence.

## Non-blocking notes

- Would be nice to have an invariant test that `attempted == accepted + rejected + errors` after N steps. Not for this PR — the current tests are proportional to the change scope.

- `effective_draft_tokens: 1` is hardcoded because the implementation only drafts one token per step (documented in the source comments elsewhere). If someone later extends to multi-token drafting, this field will need to become dynamic. Fine placeholder for now.

- `bypass_counts_semantics` string field is unusual — normally you'd document semantics in a schema or docs rather than in the payload. But given operators are reading this via `curl /v1/status | jq`, embedding the semantics inline is actually pragmatic. Just noting the choice.

## CI

All 9 checks green.

Ship it. This gives operators the counters they've been reconstructing from logs, and the lock work moves reporting correctness from "coincidentally works" to "provably works."

waybarrios (changes requested):
The MTP counters reset whenever sampling changes recreate the BatchGenerator, so /v1/status can move backward during normal traffic. Please keep the counter state on Scheduler and cover sampler-driven generator replacement in tests.

waybarrios (approved):
I pushed the fix and resolved the conflict with main. MTP counters now survive sampler-driven BatchGenerator replacement, and the regression is covered in tests.

waybarrios:
I addressed the counter reset in `2c7c8e2` by keeping MTP stats on `Scheduler` and covering sampler-driven `BatchGenerator` replacement. `ebdfe30` updates the branch from `main` while preserving the fix and the newer chunked-prefill path. The earlier feedback is fully addressed.

Issue #657, opened 2026-07-29
Qwen MTP receives post-final-norm hidden states


Thump604:
## Observed behavior

`vllm_mlx/patches/qwen3_5_mtp.py` requests hidden states from the Qwen
TextModel, but the patched `__call__` returned the state after the backbone's
final normalization. `mtp_forward()` then applies
`pre_fc_norm_hidden`, so the MTP head receives a differently normalized input
than its checkpoint contract expects.

## Exact local repro

With `Qwen3.6-35B-A3B-VLM-MTP-8bit`, run:

```sh
AI_RUNTIME_BYPASS_SAFETY_GATE=1 \
VLLM_MLX_QWEN36_VLM_MTP_MODEL=/opt/ai-runtime/run/qwen-mtp-evidence/20260729T202502Z-qwen36-35b-clean-base-native-mtp-view-manifest \
python -m pytest tests/test_text_model_from_vlm.py::test_qwen36_text_model_return_hidden_is_pre_norm_for_mtp -q
```

The regression proves that `return_hidden=True` must return the pre-final-norm
backbone state when MTP will apply `pre_fc_norm_hidden` itself.

## Upstream code path compared

Compared local and `main` at `0dd115769ef1196a715b96b181353edacd2a4f69`:

- `vllm_mlx/patches/qwen3_5_mtp.py`

## Expected behavior

The MTP head receives the same pre-final-norm hidden state assumed by its
checkpoint, and applies its own MTP normalization once.

## Minimal fix

Return `hidden_states` before the final backbone norm only for the
`return_hidden` MTP path. Add an opt-in real-artifact regression.

## Not claimed

This report does not claim a throughput improvement, multi-token draft depth,
or qualification for tool, constrained-output, sampled, or concurrent
workloads.

Thump604:
Resolved by #660, which is merged. Closing as completed.

PR #660, opened 2026-07-29
fix(mtp): pass Qwen pre-norm hidden states

Thump604:
## Summary

Return Qwen backbone hidden states before the final norm when native MTP asks
for `return_hidden=True`.

Refs #657.

`mtp_forward()` applies `pre_fc_norm_hidden` itself. Returning a final-normalized
state causes that MTP-specific norm to see the wrong representation.

## Scope

- `vllm_mlx/patches/qwen3_5_mtp.py`
- `tests/test_text_model_from_vlm.py`

## Validation

```sh
AI_RUNTIME_BYPASS_SAFETY_GATE=1 \
VLLM_MLX_QWEN36_VLM_MTP_MODEL=/opt/ai-runtime/run/qwen-mtp-evidence/20260729T202502Z-qwen36-35b-clean-base-native-mtp-view-manifest \
python -m pytest tests/test_text_model_from_vlm.py::test_qwen36_text_model_return_hidden_is_pre_norm_for_mtp -q
```

The opt-in test passed against a Qwen 3.6 35B VLM artifact. The matching
batched MTP integration run is tracked separately from this narrow fix.

## Upstream comparison

Compared against `main` at `0dd115769ef1196a715b96b181353edacd2a4f69` in
`vllm_mlx/patches/qwen3_5_mtp.py`.

## Not claimed

This PR does not change MTP scheduling, sampling, cache behavior, or model
selection. It does not claim a performance result.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,6 +92,7 @@ jobs:
             tests/test_qwen3_xml_parser.py \
             tests/test_qwen3_xml_registration.py \
             tests/test_qwen3_xml_streaming_chunks.py \
+            tests/test_qwen35_mtp_hidden_state_mode.py \
             tests/test_streaming_fence_stripper.py \
             tests/test_streaming_think_router.py \
             tests/test_streaming_tool_filter.py \
--- a/docs/guides/continuous-batching.md
+++ b/docs/guides/continuous-batching.md
@@ -38,6 +38,12 @@ that envelope fall back to the normal scheduler path instead of using MTP. This
 keeps sampling correctness ahead of throughput until the MLLM verifier is
 sampler-aware.
 
+Injected Qwen3.5 and Qwen3.6 MTP checkpoints use the established post-norm
+backbone state by default. A checkpoint qualified for pre-norm input can opt in
+by setting `mtp_hidden_state_mode` to `pre_norm` in its `text_config`. Leave the
+setting unset for existing checkpoints. Unknown values fall back to post-norm
+and emit a warning.
+
 Thinking/logits processors stay active by default for the whole request. The
 experimental retirement-to-MTP handoff is opt-in via
 `VLLM_MLX_ENABLE_THINKING_RETIREMENT_RESUME=1`; leave it unset unless you have
--- a/tests/test_qwen35_mtp_hidden_state_mode.py
+++ b/tests/test_qwen35_mtp_hidden_state_mode.py
@@ -1,0 +1,56 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for the Qwen3.5 MTP hidden-state checkpoint contract."""
+
+import logging
+
+import pytest
+
+from vllm_mlx.patches.qwen3_5_mtp import (
+    _resolve_qwen_mtp_hidden_state_mode,
+    _select_qwen_mtp_hidden_state,
+)
+
+
+def test_qwen_mtp_hidden_state_mode_defaults_to_post_norm():
+    assert _resolve_qwen_mtp_hidden_state_mode({}) == "post_norm"
+    assert _resolve_qwen_mtp_hidden_state_mode({"text_config": {}}) == "post_norm"
+
+
+@pytest.mark.parametrize(
+    "config",
+    [
+        {"mtp_hidden_state_mode": "pre_norm"},
+        {"text_config": {"mtp_hidden_state_mode": "pre_norm"}},
+    ],
+)
+def test_qwen_mtp_hidden_state_mode_accepts_explicit_pre_norm(config):
+    assert _resolve_qwen_mtp_hidden_state_mode(config) == "pre_norm"
+
+
+def test_qwen_mtp_hidden_state_mode_uses_nested_setting_first():
+    config = {
+        "mtp_hidden_state_mode": "pre_norm",
+        "text_config": {"mtp_hidden_state_mode": "post_norm"},
+    }
+
+    assert _resolve_qwen_mtp_hidden_state_mode(config) == "post_norm"
+
+
+@pytest.mark.parametrize("value", ["unknown", None, 1])
+def test_qwen_mtp_hidden_state_mode_rejects_unknown_value(value, caplog):
+    caplog.set_level(logging.WARNING)
+
+    mode = _resolve_qwen_mtp_hidden_state_mode(
+        {"text_config": {"mtp_hidden_state_mode": value}}
+    )
+
+    assert mode == "post_norm"
+    assert "Unsupported mtp_hidden_state_mode=" in caplog.text
+
+
+def test_qwen_mtp_hidden_state_selection_preserves_checkpoint_contract():
+    pre_norm = object()
+    post_norm = object()
+
+    assert _select_qwen_mtp_hidden_state("post_norm", pre_norm, post_norm) is post_norm
+    assert _select_qwen_mtp_hidden_state("pre_norm", pre_norm, post_norm) is pre_norm
--- a/tests/test_text_model_from_vlm.py
+++ b/tests/test_text_model_from_vlm.py
@@ -202,6 +202,12 @@ def test_qwen36_text_model_return_hidden_is_pre_norm_for_mtp():
     from mlx_vlm import load as vlm_load
 
     assert QWEN36_VLM_MTP_MODEL is not None
+    config = json.loads((QWEN36_VLM_MTP_MODEL / "config.json").read_text())
+    text_config = config.get("text_config", config)
+    assert text_config.get("mtp_hidden_state_mode") == "pre_norm", (
+        "qualification artifact must explicitly declare "
+        "mtp_hidden_state_mode=pre_norm"
+    )
     vlm_model, _ = vlm_load(str(QWEN36_VLM_MTP_MODEL))
     text_model = build_text_model(vlm_model, QWEN36_VLM_MTP_MODEL)
     tokens = mx.array([[1, 2, 3]])
--- a/tests/test_text_model_from_vlm.py
+++ b/tests/test_text_model_from_vlm.py
@@ -2,6 +2,7 @@
 """Tests for building mlx_lm TextModel from mlx_vlm-loaded weights."""
 
 import json
+import os
 import sys
 import types
 from pathlib import Path
@@ -16,6 +17,15 @@ VLM_MTP_MODEL = Path.home() / "ai-models/mlx_models/Qwen3.5-35B-A3B-VLM-MTP-8bit
 
 # Text-only MTP model (no vision tower — can't test VLM loading)
 TEXT_MTP_MODEL = Path.home() / "ai-models/mlx_models/Qwen3.5-35B-A3B-8bit"
+
+# Deliberately opt in to the large Qwen3.6 MTP artifact. This regression is
+# exercised in an isolated model-qualification window, never by the ordinary
+# unit suite on developer machines or CI.
+QWEN36_VLM_MTP_MODEL = (
+    Path(os.environ["VLLM_MLX_QWEN36_VLM_MTP_MODEL"])
+    if os.environ.get("VLLM_MLX_QWEN36_VLM_MTP_MODEL")
+    else None
+)
 
 
 def test_build_text_model_no_config():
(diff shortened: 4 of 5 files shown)


janhilgard (commented):
Comment only (not a verdict) — I'd love to see this land, but there's an empirical conflict against a different Qwen variant that I want to surface before someone merges.

## Historical context I want to reconcile

In 2026-04 we deployed native MTP on three Qwen models (27B Dense Heretic, 35B-A3B MoE, 122B-A10B MoE) through this same `qwen3_5_mtp.py` patch. During that bring-up we tested **both** pre-norm and post-norm on `return_hidden` — the working answer at the time was post-norm, and that shipped as the current `return normed  # post-norm hidden states (MTP expects post-norm)` line. My working notes from that period read literally:

> Post-norm hidden states: MTP expects OUTPUT of model's final RMSNorm. Return `normed` (post-norm) NOT `hidden_states` (pre-norm) from `return_hidden`.

I can't reconstruct the exact acceptance numbers we saw with pre-norm at the time (only that post-norm won), but that decision has been in production for three months and my current live baseline is:

- **Qwen 3.6-27B Heretic-v2 dense** (`qwen3_5`, ordinary text MTP checkpoint, not the VLM variant): **79.9% acceptance** (1093/1368) on a warm production server right now, with the post-norm path.

That's not "kind of works" — 78-80% is exactly the acceptance range these MTP heads were reported to hit in the Qwen 3.5 paper's own results.

## What the double-norm question looks like from `mtp_forward`

The head does:

```python
h = self.mtp.pre_fc_norm_hidden(hidden_states)
```

So if `hidden_states` is post-norm, RMSNorm runs twice. RMSNorm is *not* idempotent in general — it re-normalizes to unit-variance, so a second pass rescales without dividing by zero — but it changes both magnitude and the effect of the learned `weight` offset. The PR's argument that "the checkpoint's `pre_fc_norm_hidden` is trained assuming a pre-norm input" is theoretically clean.

But then either (a) our 79.9% is actually degraded from a hidden ceiling of ~85%+ that a single-norm path would unlock, or (b) something else in the current stack compensates for the double-norm — most likely `_apply_qwen_mtp_rmsnorm_offset_fixups()`, which adds +1.0 to the raw HF offset on `pre_fc_norm_hidden.weight` when the weight mean is <0.5. That fixup was empirically necessary to make MTP work at all (0% → 78% acceptance in one of our earlier passes), and it may be masking the norm-count question.

## What I'd want to see before merge

1. **A dense-model regression alongside the VLM one.** The test as written targets `Qwen3.6-35B-A3B-VLM-MTP-8bit` (VLM variant), env-gated so it doesn't run in CI. Native MTP is deployed against both dense text models (27B) and non-VLM MoE (35B-A3B, 122B-A10B) in the field. A dense regression showing "pre-norm improves acceptance from X% to Y% on 27B Dense" would be dispositive. If it *drops* acceptance on dense, this is a VLM-specific change and needs to be gated on VLM path.

2. **Interaction with `_apply_qwen_mtp_rmsnorm_offset_fixups`.** Does the pre-norm return path assume the offset fixup is already applied to the checkpoint's `pre_fc_norm_hidden.weight`? Or would the fixup double-shift under the new logic? Worth a sentence in the PR body.

3. **Behavior on a checkpoint that was extracted with the post-norm assumption.** If someone converted their MTP `mtp/weights.safetensors` under the old post-norm contract (e.g. because that was the working path at the time), does switching `return_hidden` to pre-norm mid-flight break them? The one-line change would silently regress every existing deployment on the same day of the merge — no config knob, no fallback.

## What I can offer

I'm running Heretic-v2 27B in production with the current `qwen3_5_mtp.py` and can reproduce against pre-norm on the same artifact if you want a dense-variant data point. Point me at a branch (or I can apply the 3-line diff locally) and I'll send you before/after acceptance + a smoke-test transcript within an hour. Bear in mind we're on Apple Silicon MLX 0.32.0 / mlx-vlm 0.6.6, so it's a genuine second implementation stack, not just a rerun of the VLM 8bit test.

The upstream comparison sha (`0dd1157...`) matches what I have installed, so I'd be testing against the same base.

Not blocking — the fix is theoretically sound and the test methodology is clean. Just want dense-model evidence in the record before this ships, given how easy it is to silently kick production MTP acceptance down.

Thump604:
Thanks, Jan. I completed the requested three-artifact check locally, and it surfaced a separate issue rather than resolving the pre-norm question.

For our VLM composite 35B-A3B and 122B-A10B artifacts, the mlx-vlm loader selected the base RMSNorm convention before filtering draft `mtp.*` keys. The draft shard therefore changed base-model loading even with MTP later disabled. Filtering draft keys first restores normal final content; the narrow fixes are tracked separately in mlx-vlm #1754 and vllm-mlx #664/#665.

This does **not** establish that #660 should replace the current post-norm path universally. Our 27B control was a final-content diagnostic, not an acceptance-rate comparison against a native dense text checkpoint, so it cannot adjudicate your 79.9% Heretic-v2 result or the interaction with `_apply_qwen_mtp_rmsnorm_offset_fixups()`.

I am keeping #660 unmerged and unactivated. The next evidence needed here is the paired, same-artifact dense comparison you offered: pre-norm versus post-norm with the current offset fixup, recording acceptance/rejection counts and a smoke transcript. That will tell us whether #660 is a general improvement, a checkpoint-specific path, or should remain separate from the composite-shard correction.

janhilgard:
That composite-shard finding is the more important result here, and it changes how I read both of our datasets.

## Confirming the scope boundary from my side

I checked my production artifact against the failure mode you describe. `Qwen3.6-27B-Heretic-v2-MTP-mlx-mixed-4-8` is **not** affected, and the reason is layout:

```
Qwen3.6-27B-Heretic-v2-MTP-mlx-mixed-4-8/
├── model-0000{1..4}-of-00004.safetensors   ← glob("*.safetensors") sees only these
├── model.safetensors.index.json
└── mtp/
    └── weights.safetensors                  ← 15 tensors, all prefixed `mtp.*`
```

The draft tensors live in a `mtp/` subdirectory, so `mlx_vlm.utils`' root-level `glob("*.safetensors")` never sees them, and sanitization computes `shift_norm_weights` on base tensors only. Prefix is plain `mtp.*`, not `language_model.mtp.*`.

That subdirectory placement was originally a workaround for a different problem — `mlx_vlm` raising `ValueError: Received N parameters not in model` when MTP tensors sat in the root — but it incidentally sidesteps the sanitization bug entirely.

Two consequences:

1. **My 79.9% figure is not contaminated** by the convention-selection bug. Whatever it measures, it measures on a base model that loaded with the correct RMSNorm convention.
2. **The 27B is therefore a clean control** for the paired comparison — a dense checkpoint where the only variable is the `return_hidden` path, with no composite-shard interference. That is precisely the artifact class your 35B run could not isolate.

Worth stating explicitly in #1754's reproduction boundary: the bug requires draft tensors visible to the root glob. Artifacts using the `mtp/` subdirectory convention are out of scope, which may matter for anyone triaging whether their deployment is affected.

## The paired comparison — methodology I intend to run

Same artifact, same server config, same request set, only `qwen3_5_mtp.py:365` differing:

| Arm | `return_hidden` returns | `_apply_qwen_mtp_rmsnorm_offset_fixups` |
|-----|------------------------|------------------------------------------|
| A (current) | `normed` (post-norm) | active — my checkpoint's norms have mean < 0.5, so +1.0 is applied |
| B (#660) | `hidden_states` (pre-norm) | unchanged, still active |

Recorded per arm:
- `/v1/status` `mtp.attempted / accepted / rejected / acceptance_rate` over a fixed request set, warm server, greedy sampling (`temperature=0, top_p=1, top_k=0, min_p=0`, single-request batch — otherwise the bypass guards fire and the counters measure nothing)
- Smoke transcript on identical prompts, checking for degradation the acceptance number would not reveal
- Norm-weight means for `pre_fc_norm_hidden` / `pre_fc_norm_embedding` so the offset-fixup interaction is on record rather than inferred

The offset fixup staying active in both arms is deliberate. If B improves acceptance, that tells us the double-norm was costing us. If B degrades it, the likely reading is that the fixup was compensating for post-norm input all along, and the two changes are coupled rather than independent.

## Timing

This needs a service window — patching `qwen3_5_mtp.py` in an editable install means restarting the 27B server, and I run it as a production endpoint. I need to clear that with the machine's owner before touching it, so I can't commit to a specific time here. It will not be tonight.

Holding #660 unmerged until then is the right call, and I'd extend that to not merging it as part of the #664/#665/#1754 group either — the composite-shard correction stands on its own evidence, this one does not yet.

I'll report the numbers here either way, including if they contradict my prior.

janhilgard:
Ran the paired comparison. **Pre-norm measures worse on my dense artifact — I'd hold #660.**

## Setup

`Qwen3.6-27B-Heretic-v2-MTP-mlx-mixed-4-8` (dense `qwen3_5`, MTP in `mtp/` subdirectory, unaffected by the composite-shard bug per my earlier comment). Served via `--mllm --continuous-batching --enable-mtp`, mlx 0.32.0 / mlx-vlm 0.6.6, M3 Ultra.

Only variable: `qwen3_5_mtp.py:366`. `_apply_qwen_mtp_rmsnorm_offset_fixups()` left active in both arms.

10 fixed prompts (arithmetic, code, explanation, Czech, logic, JSON, SQL, summarization), 180 tokens each, fully greedy (`temperature=0, top_p=1, top_k=0, min_p=0`), one request at a time. Server restarted before each arm so `/v1/status` counters start at zero.

## Results

| Arm | `return_hidden` | Acceptance | attempted | accepted | rejected | errors | mean tok/s |
|-----|----------------|-----------:|----------:|---------:|---------:|-------:|-----------:|
| **A** | `normed` (current) | **80.30 %** | 1005 | 807 | 198 | 0 | 36.2 |
| **B** | `hidden_states` (#660) | **75.24 %** | 1034 | 778 | 256 | 0 | 34.3 |
| **A′** | `normed` (repeat) | **80.30 %** | 1005 | 807 | 198 | 0 | 36.2 |

**A′ reproduces A exactly** — identical counters, identical throughput. Greedy decoding is deterministic here, so the A↔B gap is not run-to-run noise.

Delta: **−5.06 pp acceptance, +58 rejections, −1.9 tok/s.**

For reference, this artifact's lifetime counter before the experiment was 1093/1368 = 79.90 % over ~7 days of mixed traffic, so 80.30 % on the fixed set is consistent with its normal operating point.

## Norm weights, since the fixup interaction matters

```
mtp.layers.0.input_layernorm.weight           mean=+0.0408   → +1.0 applied
mtp.layers.0.post_attention_layernorm.weight  mean=+0.2109   → +1.0 applied
mtp.pre_fc_norm_embedding.weight              mean=-0.4414   → +1.0 applied
mtp.pre_fc_norm_hidden.weight                 mean=-0.1709   → +1.0 applied
mtp.layers.0.self_attn.k_norm.weight          mean=+0.7422   → skipped
mtp.layers.0.self_attn.q_norm.weight          mean=+0.7578   → skipped
mtp.norm.weight                               mean=+1.2734   → skipped
```

`pre_fc_norm_hidden` — the norm #660 is reasoning about — is one of the four that **does** receive the +1.0 shift on this checkpoint. So on this artifact the two changes are coupled, not independent: the head's `pre_fc_norm_hidden` was fitted (or at least is being used) in a way that expects the post-norm input it currently gets.

That is consistent with your framing that this might be checkpoint-specific rather than a general correction. I can't distinguish "this checkpoint's MTP was extracted under a post-norm assumption" from "post-norm is correct generally" — my single artifact can't separate those.

## Separate observation: verified mode is not output-stable across this change

9 of 10 prompts produced byte-identical output between arms. One diverged, and A′ reproduced A's version exactly:

```
pos 317, prompt "Write a Python function that reverses a string. Code only."

A: ...several ways to reverse a string.\n    *   *Slicing:* `s[::-1]` (Most Pythonic, efficient, concise).\n    *   *Bui
B: ...several ways to reverse a string.\n    *   Method 1: Slicing `s[::-1]`. This is the most Pythonic and efficient w
```

Under `always_advance_verified` I'd expect the emitted token sequence to be independent of acceptance rate — a draft is only taken when it equals the verify argmax, so changing which drafts get accepted should change *speed*, not *content*. Divergence means something in the accept/reject path is not output-neutral. The rejection replay is the obvious suspect: it manipulates cache state, and #662's own comment notes the batch cache cannot roll back a single row.

I'd treat that as its own issue rather than evidence about pre- vs post-norm. It is reachable today by anything that shifts the accept pattern, not just this PR. Happy to open it separately with this repro if you agree it's distinct — or if you'd rather fold it into the #662 discussion, that works too.

## Where this leaves #660

On the only artifact I can measure, pre-norm costs 5 points of acceptance and produces different text. That is evidence against merging it as a universal change, but it is **not** evidence that the theory is wrong — your reading of the checkpoint contract may still be correct for artifacts whose MTP tensors were extracted differently from mine.

What would settle it: the same paired run on an artifact where `pre_fc_norm_hidden` has mean ≥ 0.5, i.e. where the offset fixup does *not* fire. If pre-norm wins there, the two changes are coupled and #660 needs to land together with a fixup-aware guard rather than alone. If it loses there too, post-norm is simply correct and the docstring in `mtp_forward` should say why.

Raw per-prompt data (counters, token counts, throughput, transcripts) is on my side if you want it in any particular format.

Production is back on `normed`, file verified bit-identical to the pre-experiment state.

waybarrios:
I pushed a small follow-up that keeps post-norm as the default while allowing pre-norm through an explicit `mtp_hidden_state_mode` setting. I also added CI coverage for both paths, keeping existing checkpoints stable without blocking qualified pre-norm artifacts.

Issue #658, opened 2026-07-29
Validate Qwen MoE MTP conversion artifacts before decode


Thump604:
## Observed behavior

The Qwen MTP converter used `key.replace(".weight", suffix)` for quantization
companions. Qwen 3.6 publishes fused MoE MTP keys such as
`experts.gate_up_proj` and `experts.down_proj` without a `.weight` suffix.
For those keys, scales and biases reused the weight key and overwrote the
tensor during conversion.

The resulting artifact can load far enough to reach decode, then fails because
the supposed expert weight has a bias-shaped last dimension.

## Exact local repro

Compare the source Qwen MTP tensors with a converted artifact:

```text
source gate_up_proj: [256, 1024, 2048]
source down_proj:    [256, 2048, 512]
corrupt converted gate_up_proj: [256, 1024, 32]
corrupt converted down_proj:    [256, 2048, 8]
```

The repaired converter rebuilt all 19 published MTP tensors and the exact
Qwen 3.6 35B MTP text route completed `MTP_MANIFEST_OK` with
`finish_reason=stop`.

## Upstream code paths compared

Compared local and `main` at `0dd115769ef1196a715b96b181353edacd2a4f69`:

- `scripts/add_mtp_weights_qwen35.py`
- `vllm_mlx/patches/qwen3_5_mtp.py`

## Expected behavior

The converter gives fused tensors distinct `.scales` and `.biases` companion
keys. Runtime injection rejects incompatible MoE tensor shapes before decode.

## Minimal fix

- Generate companion names that preserve fused tensor keys.
- Verify all source-indexed MTP tensors are extracted.
- Reject malformed fused or split expert schemas before attaching MTP.
- Write a converter sidecar declaring whether RMSNorm tensors were translated
  to MLX gamma format, instead of inferring it from tensor means at load time.

## Not claimed

This report does not claim that every legacy MTP artifact can be repaired
automatically or that artifact conversion changes model quality.

PR #661, opened 2026-07-29
fix(mtp): validate Qwen MTP artifact conversion

Thump604:
## Summary

Validate Qwen MoE MTP artifacts before they reach decode and preserve fused
tensor names while writing quantization companions.

Refs #658.

The converter previously overwrote fused expert weights because Qwen 3.6
fused keys do not end in `.weight`. The runtime now rejects incompatible
expert shapes before attaching MTP. Converter output carries a small manifest
so already translated RMSNorm gamma tensors are not guessed from their mean.

## Scope

- `scripts/add_mtp_weights_qwen35.py`
- `vllm_mlx/patches/qwen3_5_mtp.py`
- `tests/test_qwen35_mtp_patch.py`

## Validation

```sh
python -m pytest tests/test_qwen35_mtp_patch.py tests/test_text_model_from_vlm.py -q
ruff check scripts/add_mtp_weights_qwen35.py vllm_mlx/patches/qwen3_5_mtp.py tests/test_qwen35_mtp_patch.py
```

Focused result: 12 passed, 4 artifact-dependent tests skipped.

A rebuilt Qwen 3.6 35B artifact contained all 19 upstream MTP tensors and
completed a text request with `finish_reason=stop`.

## Upstream comparison

Compared against `main` at `0dd115769ef1196a715b96b181353edacd2a4f69` in the
converter and runtime injection paths listed above.

## Not claimed

This PR does not change the MTP scheduler, sampling algorithm, or serving
defaults.

### DIFF
--- a/scripts/add_mtp_weights_qwen35.py
+++ b/scripts/add_mtp_weights_qwen35.py
@@ -98,6 +98,18 @@ def identify_mtp_shards(index: dict) -> tuple[dict[str, str], set[str]]:
     return mtp_keys, shards_needed
 
 
+def quantization_companion_key(key: str, suffix: str) -> str:
+    """Return a scale/bias key without overwriting fused Qwen tensors.
+
+    Qwen 3.6 names fused expert tensors ``gate_up_proj`` and ``down_proj``
+    rather than appending ``.weight``.  Quantized artifacts must keep their
+    companion tensors under distinct names for the runtime to dequantize them.
+    """
+    if key.endswith(".weight"):
+        return f"{key[:-len('.weight')]}.{suffix}"
+    return f"{key}.{suffix}"
+
+
 def download_shards(
     shards: set[str], source_model: str, download_dir: Path
 ) -> dict[str, Path]:
@@ -176,6 +188,9 @@ def extract_and_quantize_mtp_weights(
         del shard_data
 
     print(f"Loaded {len(all_mtp_weights)} MTP weight tensors")
+    missing = sorted(set(mtp_keys) - set(all_mtp_weights))
+    if missing:
+        raise RuntimeError(f"MTP source shards omitted tensors: {missing}")
 
     # Norm keys that need +1.0 shift (HF centered ~0 → MLX centered ~1)
     norm_suffixes = (
@@ -221,8 +236,8 @@ def extract_and_quantize_mtp_weights(
             print(f"  Quantize {bits}-bit: {key} {q_w.shape}")
             return {
                 key: q_w,
-                key.replace(".weight", ".scales"): q_s,
-                key.replace(".weight", ".biases"): q_b,
+                quantization_companion_key(key, "scales"): q_s,
+                quantization_companion_key(key, "biases"): q_b,
             }
         else:
             print(f"  Keep FP (small): {key} {weight.shape}")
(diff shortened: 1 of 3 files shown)


janhilgard (approved):
LGTM. Correct fix for the fused-expert overwrite, and legacy artifacts stay on the exact code path they had before.

## Verified against our production MTP artifact

Our production 27B MTP artifact is `Qwen3.6-27B-Heretic-v2-MTP-mlx-mixed-4-8/mtp/weights.safetensors` (built with the older `scripts/add_mtp_weights_qwen35.py`, before this PR's manifest and fused-key changes existed):

- **Dense model, no fused experts** → `_validate_qwen_moe_mtp_tensor_shapes` short-circuits on the "no fused, no split MoE keys" branch. Not exercised at all.
- **No `mtp/manifest.json`** → `_mtp_artifact_uses_mlx_norm_weights(mtp_file)` returns False → legacy path runs.
- **Norm weights have mean < 0.5** (verified: `pre_fc_norm_hidden.weight` mean=-0.17, `pre_fc_norm_embedding.weight` mean=-0.44, `layers.0.post_attention_layernorm.weight` mean=0.21) → `_apply_qwen_mtp_rmsnorm_offset_fixups` applies +1.0 exactly as before.

Zero behavioral change for our stack. `/v1/status.mtp` currently reports 79.9% acceptance (1093/1368) on the 27B server; that stays put after this merges.

## The fused-key overwrite is a real bug

I hadn't hit this because our production model is dense, but the analysis is correct. Old converter:

```python
key.replace(".weight", ".scales")
```

For Qwen 3.6 MoE fused keys like `layers.0.mlp.experts.gate_up_proj` (no `.weight` suffix), `str.replace` finds nothing and returns the *same string* — so `q_s` and `q_b` were both written under the base weight name, silently overwriting the fused weight tensor. Any MoE MTP artifact converted with the old script would be broken in a way that wouldn't fail at load time — it'd just decode garbage. The `_mtp_quantization_companion_key` helper is the right fix: check for `.weight` suffix explicitly and fall back to appending the companion suffix when absent.

## Runtime shape validation is defensive gold

`_validate_qwen_moe_mtp_tensor_shapes` catching `ValueError("Qwen MoE MTP expert shape mismatch")` before injection means an already-corrupted artifact (from the old converter, sitting on someone's disk) fails at engine startup instead of silently degrading MTP acceptance to noise. `return False` from `inject_mtp_support` cleanly disables MTP for that model — server keeps serving without MTP acceleration. Much better than the "0% acceptance forever" failure mode.

## Manifest as authoritative RMSNorm state

The `mtp/manifest.json` sidecar is the right design. The old logic (`mean < 0.5` heuristic) is a proxy that:

- Works for our artifact (both norm formats sit clearly on one side of 0.5)
- Could misfire on a low-mean gamma tensor that legitimately sits below 0.5 (rare but possible in downstream fine-tunes)
- Is silent on new-converter output (which already writes MLX gamma but stores back means < 0.5 → old runtime would double-shift them)

Manifest → boolean → correct path. No inference, no ambiguity. Legacy artifacts (like ours) still get the heuristic fallback so existing deployments don't need reconversion.

## Non-blocking notes

- `_mtp_artifact_uses_mlx_norm_weights` checks `artifact_schema_version == 1` implicitly by parsing the file successfully. If future schema versions land, the check probably wants to be explicit about version compatibility (a v2 schema with different keys shouldn't silently be trusted as v1). Not for this PR — worth remembering when v2 lands.

- Manifest `weight_keys` field is sorted, informative, unused by runtime. Fine — it's a debugging aid for artifact inspection. Just noting the loose contract in case someone wonders whether omitting it breaks something (it doesn't).

## CI

All 9 checks green.

Ship it. Fixes a silent-corruption class of bug on MoE MTP conversions and leaves the dense-model legacy path (that we actually run) untouched.

Thump604:
Rebased the single artifact-schema commit onto current `main` (`7afa61a`). The only conflict retained both current-main shard-prefix coverage and this PR’s schema/manifest tests. The adversarial pass also found and fixed a documented fail-safe gap: syntactically valid non-object `manifest.json` values now fall back to legacy detection instead of raising `AttributeError`. Verification: full suite `2578 passed, 25 skipped, 27 deselected`; focused Qwen MTP tests `9 passed`; converter smoke, Black, Ruff, and diff checks clean. New head: `debc069`.

Issue #659, opened 2026-07-29
Support sampled concurrent verification in MLLM MTP scheduling


Thump604:
## Capability gap

The MLLM MTP scheduler uses greedy argmax equality for verification and
previously bypassed sampled or concurrent requests. That prevents native MTP
from serving normal request sampling tuples through continuous batching.

## Exact local repro

Run a Qwen 3.6 35B MLLM server with continuous batching, MTP, and normal
sampling requests:

```sh
python -m runtime_patches serve <qwen-model> --mllm --continuous-batching \
  --enable-mtp --mtp-num-draft-tokens 1
```

Two concurrent requests using `temperature`, `top_p`, `top_k`, and `min_p`
completed with visible content and `finish_reason=stop`. Scheduler metadata
reported 6 draft attempts, 5 accepted drafts, 1 exact fallback rejection, and
0 errors.

## Upstream code path compared

Compared local and `main` at `0dd115769ef1196a715b96b181353edacd2a4f69`:

- `vllm_mlx/mllm_batch_generator.py::install_mtp_mllm`

## Expected behavior

MTP verification uses each request's post-filter sampling distribution. State
that crosses scheduler turns is keyed by request UID, not the batch position.
When one row rejects in a concurrent batch, the scheduler replays the primary
token for every row rather than using an invalid per-row cache rollback.

## Minimal patch

- Add exact sampled speculative acceptance and residual sampling.
- Preserve request-local verified state across UID reordering.
- Hold newly queued work for the one verified-state transition.
- Keep MTP disabled for logits-processor requests because their transformed
  distributions are not available to the draft verifier.

## Not claimed

This request does not add multi-token MTP recurrence, constrained-output
support under MTP, or tool-call MTP support.

Thump604:
Resolved by #662, which is merged. Closing as completed.

PR #662, opened 2026-07-29
feat(mtp): support sampled concurrent MLLM decoding

Thump604:
## Summary

Enable sampler-aware, request-local MTP verification for MLLM continuous
batches.

Refs #659.

The prior greedy equality check is correct only for argmax decoding. This PR
computes acceptance and residual sampling from each request's post-filter
distribution, and keys deferred/verified state by request UID so batch
reordering cannot cross-contaminate requests.

## Scope

- `vllm_mlx/mllm_batch_generator.py`
- `tests/test_mllm_continuous_batching.py`

## Correctness boundaries

- Prefill and requests with logits processors remain non-speculative.
- Concurrent rejection replays primary tokens for the batch because the cache
  cannot roll back a single row safely.
- The implementation still drafts one token per verification step.

## Validation

```sh
python -m pytest tests/test_mllm_continuous_batching.py -q
ruff check vllm_mlx/mllm_batch_generator.py tests/test_mllm_continuous_batching.py
```

Focused result: 46 passed, 4 deselected.

In a Qwen 3.6 35B continuous-batching run, two concurrent normal-sampling
requests completed with `finish_reason=stop`; scheduler metadata recorded 6
attempts, 5 accepted drafts, 1 exact fallback rejection, and 0 errors.

## Upstream comparison

Compared against `main` at `0dd115769ef1196a715b96b181353edacd2a4f69` in
`install_mtp_mllm`.

## Not claimed

This PR does not add multi-token MTP recurrence, tool-call MTP, or
constrained-output MTP.

### DIFF
--- a/tests/test_mllm_continuous_batching.py
+++ b/tests/test_mllm_continuous_batching.py
@@ -19,6 +19,7 @@ import os
 import threading
 import tempfile
 from contextlib import nullcontext
+from types import SimpleNamespace
 from unittest.mock import MagicMock
 
 import pytest
@@ -1018,7 +1019,6 @@ class TestMLLMBatchGeneratorMTPGuards:
             "prefill": 0,
             "no_active_batch": 0,
             "concurrent_batch": 0,
-            "non_greedy_sampling": 0,
             "logits_processors": 0,
         }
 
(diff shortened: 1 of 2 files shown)


janhilgard (approved):
LGTM. Significant correctness improvement — this moves MLLM MTP from "greedy-only proof-of-concept" to genuinely usable speculative decoding, and the concurrent-batch handling is defensible.

## What this unblocks for our production

Our current MLLM MTP config on Qwen 3.6-27B Heretic-v2 hits the greedy-only bypass on almost every real client request. Chat template default parameters bypass MTP (`top_p=0.9` triggers `non_greedy_sampling`), only manually-crafted greedy requests (temp=0, top_p=1, top_k=0, min_p=0, single-request batch) actually engage MTP. That's how we got 79.9% acceptance — cherry-picked greedy calls. Under real client load with typical `temperature=0.7, top_p=0.9`, MTP is effectively off.

After this PR:
- **Sampled requests engage MTP** through per-request `_sampling_logprobs` matching the client's sampler
- **Concurrent batches engage MTP** via UID-keyed `_skip_state_by_uid` instead of positional `_skip_state[0]`
- Only prefill and request-local logits processors still bypass — both are correct exclusions (multi-token input has no draft target; custom processors can't be replayed under speculation)

The `bypass_counts` dict removes `concurrent_batch` and `non_greedy_sampling` reasons because they're no longer bypasses. That's the right shape change.

## Mode string change is a wire-format break

`mode: "always_advance_verified"` → `mode: "request_local_sampler_aware_verified"`.

Any consumer parsing exact strings will need to update. I checked our monitoring dashboard (`~/monitoring/src/`) — we grep for `mtp.mode` and just display it verbatim, so we're unaffected. But third-party clients or homerolled dashboards that string-compare against `"always_advance_verified"` will silently stop matching.

Suggest either:
1. Note in release notes that the mode string changed
2. Or bump a schema version somewhere so consumers can detect the change

Non-blocking — the more descriptive name is worth the churn.

## Correctness review

**`_sampling_logprobs`** — reproduces mlx-lm's sampler transformations in log-prob space:

```python
if 0.0 < top_p < 1.0: logprobs = apply_top_p(logprobs, top_p)
if min_p != 0.0:      logprobs = apply_min_p(logprobs, min_p)
if top_k > 0:         logprobs = apply_top_k(logprobs, top_k)
logprobs = logprobs * (1.0 / temperature)
logprobs = logprobs - logsumexp(...)
```

Filter order matches mlx-lm's canonical order (top_p, min_p, top_k, then temperature scaling). Renormalization at end is correct — after filtering the un-normalized log-probs need to be re-normalized before being interpreted as a distribution.

Temperature=0 special-case correctly collapses to a one-hot at argmax (log-prob = -inf everywhere else, 0 at chosen token). This preserves the greedy path exactly.

**`_residual_logprobs`** — the standard rejection-sampling correction:

```python
residual = max(exp(target) - exp(draft), 0)
normalized = log(residual) - log(mass)
```

`mx.where(mass > 1e-12, normalized, fallback)` handles the degenerate case (target ≤ draft everywhere) by falling back to target distribution. This is the standard behavior — when the residual has no probability mass, the next token comes from the target directly. Correct.

**`_accept_sampled_draft`** — implements `min(1, p/q)` in log space:

```python
log_acceptance = target_logprob - draft_logprob
return log_acceptance >= 0 or log(u) < log_acceptance
```

When `target > draft`, always accept (log ratio ≥ 0). Otherwise flip a coin with `p = target/draft`. This is textbook stochastic speculative sampling. `max(u, 1e-35)` guards against `log(0)`.

**UID-keyed skip state:**

```python
_skip_state_by_uid: Dict[int, dict] = {}
```

Old positional `_skip_state[0]` had a subtle bug: if the batch composition changed between the accept step (which wrote skip state) and the next verify step (which read skip state), the state was applied to the wrong request. This particularly affected text-only work joining/leaving a continuous batch mid-decode. Keying by UID eliminates that class of bug. The invariant check:

```python
if any(skip_entries) and not all(skip_entries):
    raise RuntimeError("verified-state invariant failed: a batch mixed requests with and without request-local skip state")
```

Is defensive, but correct — a batch reordering that leaves *some* requests with skip state and *others* without means someone made a scheduling mistake. Fail loud, don't silently corrupt.

**Concurrent rejection handling:**

```python
sampled_single_reject = (
    uses_stochastic_sampling and batch_size == 1 and bool(residual_tokens_by_uid)
)
```

Concurrent batch with any rejection → replay only primary tokens for the whole batch, no residual sampling. Loses that step's speedup but preserves cache consistency. Single-request batch with rejection → can sample residual and advance two positions. Trade-off is honestly documented in the code comment ("A batch cache cannot roll back an individual row"). Correct.

## Test coverage

`test_sampled_mtp_uses_post_filter_distributions_and_residuals` — direct math verification of the three helpers. Nice property tests (target/residual sums == 1.0, acceptance rule at boundaries).

`test_sampled_mtp_attempts_request_with_normal_sampling` — proves that a request with `temperature=0.6, top_p=0.95, top_k=3` no longer takes the bypass path (the fake `_original_step` raises AssertionError if invoked, which would fire on a bypass). Clean negative test.

Old parametrized bypass tests correctly removed since those bypasses no longer exist.

## Non-blocking notes

- **`math.log(max(uniform_draw, 1e-35))`** — this is `-80.6` for the clamped case, which is well below any realistic `log_acceptance`. Fine as a floor. Just noting the value in case anyone wonders about the magic constant.

- **`_residual_logprobs`** returns `target_logprobs` as the fallback when residual mass is negligible. Consider: is that mathematically the right choice, or should it be the draft distribution? Both are defensible — target is what the "correct" model wants next, but you've just paid the draft cost. In practice `mass > 1e-12` catches essentially all real cases, so this only matters in pathological setups.

- **Batch=1 sampled-rejection residual sampling adds latency**: `residual_token = mx.random.categorical(residual); mx.eval(residual_token)` is a Metal round-trip. If real workloads have high sampled rejection rate, that's a real cost. Worth benchmarking rejected-vs-accepted throughput under our actual sampling defaults post-merge.

- **Assertion invariant on skip state mix** is `RuntimeError` — kills the request. Non-blocking, but a log + graceful fallback (`_skip_state_by_uid.clear(); return _orig_step(...)`) might be more operator-friendly than crashing. Only fires on scheduler bugs, so exceedingly rare in practice.

## What I'd want to see post-merge

For our production stack specifically, I'd want to measure:
1. **Real-world MTP acceptance rate under default client sampling** (`temperature=0.7, top_p=0.9`) — hypothesis is 40-60% based on the sampled acceptance formula
2. **Concurrent-batch acceptance** with 2-4 simultaneous requests
3. **Rejection replay overhead** — how much time is lost when a rejection triggers primary-only replay for the whole batch

Happy to run these locally on the 27B Heretic-v2 server after this merges and share the numbers — the current 79.9% figure comes from cherry-picked greedy requests, so we've never seen an honest under-load number.

## CI

All 9 checks green.

Ship it. This unlocks MTP for real client traffic instead of just synthetic greedy tests. The mode string change is the only wire-visible concern, and it's a defensible break for the added correctness.

waybarrios:
Ran an automated multi-angle review of this PR (10 finder passes + independent verification + a gap sweep) against the `install_mtp_mllm` reject/residual code paths, since those are the parts the new tests don't exercise. Found and fixed several real bugs on top of the existing branch:

**Correctness fixes**
- `sampled_single_reject` (single-request, non-greedy draft rejected) stored the residual replay's skip-state logits without reducing the model's sequence dimension, unlike every other skip-state write in this function. This corrupted the next decode step's sampler input shape (rank-2 instead of rank-1 primary tokens). Fixed by reducing to the same 2-D convention used everywhere else, and added a regression test that reproduces the shape crash pre-fix.
- The residual token's reported `logprobs` was taken from the raw unfiltered target distribution instead of the residual distribution it was actually sampled from. Now stores and reports the correct residual logprobs.
- The new UID-mismatch invariant check raised a hard `RuntimeError` (outside the function's own try/except) whenever batch membership changed since skip state was last populated -- reachable in practice when chunked prefill (installed underneath MTP) merges a new request into the active batch mid-generation. Replaced with a graceful fallback: discard stale skip state and force one full forward for the batch, trading that step's acceleration for correctness, consistent with how the concurrent-rejection path already handles this tradeoff.
- The "hold queued requests back for one decode step" mechanism in `_mtp_next` actually held them back indefinitely (it re-evaluated the same non-empty-skip-state condition on every call, and skip state is repopulated on almost every step), silently starving continuous batching's mid-batch-extend admission path for any request sharing a generator with an actively-MTP-decoding one. Removed it -- it's no longer needed for correctness now that `_mtp_step` degrades gracefully on membership mismatches.
- `_request_uses_stochastic_sampling` misclassified greedy (temperature=0) requests left at a non-default `top_p`/`top_k`/`min_p` (e.g. the common `top_p=0.9` default) as needing the more expensive sampled draft/verify path, even though `_sampling_logprobs` ignores those filters once temperature is 0. Fixed to short-circuit on temperature=0.
- Fixed a small resource-lifetime gap where a request's skip-state entry could outlive it when it finished on a step with no pending deferred draft (the cleanup sweep was unreachable behind an early return).

**Test coverage**
- Added a regression test for the sampled single-request reject path (the one with the shape bug above).
- Added a regression test for a concurrent (batch_size>1) rejection, which previously had zero test coverage -- the only existing concurrent test only exercises the accept path.

**Cleanup**
- Removed a dead `draft_distribution = None` assignment in the greedy branch.

Validated with `ruff check` and `python -m py_compile` locally (this sandbox can't run the mlx-dependent suite -- no Apple Silicon here), so letting CI run the actual test suite to confirm. Also flagged but did not touch (out of scope for a minimal fix): the dead `concurrent_batch` bypass-counter key, duplicated greedy-vs-sampled verifier math, and a per-row Python loop that could be vectorized for the common all-greedy-batch case -- happy to follow up on any of those if useful.

waybarrios:
Now it seems all set to me!  Please @Thump604 check it out before I accept it.

Thump604:
Reviewed current PR head 54fcd2f.

Local validation:
- Exact upstream test path: `tests/test_mllm_continuous_batching.py`
- Result from a clean checkout at 54fcd2f: 48 passed, 4 deselected
- Runtime carry is promoted from the same head into the durable local patch cache and venv-live
- Full Runtime verification: 622 Runtime tests passed, 5 mlx-qsdpa tests passed, patch parity 19/19
- Live Qwen 3.6 27B sampled concurrent smoke on the same `install_mtp_mllm` path: 323 MTP attempts, 192 accepted, 131 rejected, 0 errors, and no concurrent-batch bypasses

Review result: I found no additional blocking issue in the current PR head. The follow-up fixes cover reject-replay shape normalization, residual-distribution logprobs, stale mixed-batch fallback, queued-request admission, greedy-temperature classification, and request-state cleanup.

Expected behavior confirmed: normal model-card sampling and concurrent MLLM requests enter request-local sampler-aware MTP without the prior hard state-mismatch failure.

Not claimed: this is not a throughput certification; it does not add multi-token recurrence, tool-context MTP, or constrained-output MTP. One bounded live request reached `finish_reason=length` because Qwen consumed the 256-token budget in reasoning; that was not an MTP error.

waybarrios:
good!

PR #663, opened 2026-07-29
fix(server): flush reasoning parser at stream end (supersedes #653)

janhilgard:
Supersedes #653 with the CI failure resolved.

## Background

@Thump604 opened #653 on 2026-07-28 to flush parser-buffered terminal text (e.g. Gemma 4's truncated `<|channel>` markers) before emitting the terminal chunk of a stream. The fix is correct and needed — Gemma 4's `finalize_stream()` already implements the fallback, `stream_chat_completion` just wasn't calling it.

CI on that PR broke four existing streaming tests:

```
tests/test_server.py::TestStreamChatCompletion::test_reasoning_stream_emits_structured_tool_calls
tests/test_server.py::TestStreamChatCompletion::test_reasoning_stream_redirects_gemma4_tool_marker
tests/test_server.py::TestStreamChatCompletion::test_reasoning_stream_skips_tool_parser_until_markup_appears
tests/test_server.py::TestStreamChatCompletion::test_response_format_stream_promotes_reasoning_json_to_content
```

All four failed with `AttributeError: 'FakeReasoningParser'/'ReasoningOnlyParser' object has no attribute 'finalize_stream'`.

## Root cause

Those tests declare inline test fixtures (`FakeReasoningParser`, `ReasoningOnlyParser`) that duck-type `extract_reasoning_streaming` and `reset_state` but don't inherit from `ReasoningParser`. The base class provides `finalize_stream()` as a documented no-op, so any parser that *does* inherit gets the default automatically. #653 called `reasoning_parser.finalize_stream()` unconditionally, which broke the duck-typed fixtures.

The alternatives are:
1. Add `hasattr` / `getattr` guard in `_resolve_reasoning_stream_delta`.
2. Add `finalize_stream = lambda self: None` stubs to every fake fixture.

I picked (1). The base class's own docstring establishes `finalize_stream` as an *optional* extension point ("Default implementation is a no-op"), so the calling code should tolerate parsers that don't implement it — that keeps the interface contract loose for third-party parsers as well, not only tests.

## Change

One-line change in `_resolve_reasoning_stream_delta` (`vllm_mlx/server.py`):

```python
# Before (from #653):
final_delta = reasoning_parser.finalize_stream()

# After:
finalize = getattr(reasoning_parser, "finalize_stream", None)
final_delta = finalize() if callable(finalize) else None
```

Everything else is @Thump604's original diff, unchanged. Authorship on this commit is preserved as `Thump604 <...>`; I'm listed as `Co-Authored-By` for the guard change.

## Validation

Ran the 5 relevant tests locally (Python 3.14, macOS ARM64, uv-installed dev extras against fresh clone):

```
tests/test_server.py::TestStreamChatCompletion::test_reasoning_stream_emits_structured_tool_calls PASSED
tests/test_server.py::TestStreamChatCompletion::test_reasoning_stream_redirects_gemma4_tool_marker PASSED
tests/test_server.py::TestStreamChatCompletion::test_reasoning_stream_skips_tool_parser_until_markup_appears PASSED
tests/test_server.py::TestStreamChatCompletion::test_response_format_stream_promotes_reasoning_json_to_content PASSED
tests/test_server.py::TestStreamChatCompletion::test_gemma_stream_flushes_pending_content_before_length_finish PASSED
5 passed in 1.72s
```

The four previously-failing tests now pass without modification — the guard tolerates duck-typed fixtures. The new test (@Thump604's Gemma flush regression) still passes as designed.

## Not claimed

Same non-claims as #653:
- Preserves template selection, sampling, request limits, scheduler behavior.
- Does not change Jobs integration, model configuration, constrained decoding.
- Does not claim every length-bound Gemma generation produces final content — only preserves parser-buffered terminal content when it exists.

## Followup

Please close #653 after merging this. The intent is identical; this is just a rebased single-commit variant with the CI regression fixed and the base-class no-op contract respected.

Closes #653.

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -1950,6 +1950,94 @@ class TestStreamChatCompletion:
         assert payloads[2]["choices"][0]["finish_reason"] == "stop"
 
     @pytest.mark.anyio
+    async def test_gemma_stream_flushes_pending_content_before_length_finish(
+        self, monkeypatch
+    ):
+        """A terminal length stop must not discard Gemma's buffered tail."""
+        from vllm_mlx.engine.base import GenerationOutput
+        from vllm_mlx.reasoning.gemma4_parser import Gemma4ReasoningParser
+        from vllm_mlx.server import (
+            ChatCompletionRequest,
+            Message,
+            stream_chat_completion,
+        )
+        import vllm_mlx.server as server
+
+        class FakeEngine:
+            model_name = "gemma4-31b-it-8bit"
+
+            async def stream_chat(self, messages, **kwargs):
+                chunks = [
+                    GenerationOutput(
+                        text="<|channel>thought\n",
+                        new_text="<|channel>thought\n",
+                        finished=False,
+                        prompt_tokens=4,
+                        completion_tokens=3,
+                    ),
+                    GenerationOutput(
+                        text="<|channel>thought\nplan",
+                        new_text="plan",
+                        finished=False,
+                        prompt_tokens=4,
+                        completion_tokens=5,
+                    ),
+                    GenerationOutput(
+                        text="<|channel>thought\nplan<channel|>FINAL<|chan",
+                        new_text="<channel|>FINAL<|chan",
+                        finished=True,
+                        finish_reason="length",
+                        prompt_tokens=4,
+                        completion_tokens=9,
+                    ),
+                ]
+                for chunk in chunks:
+                    yield chunk
+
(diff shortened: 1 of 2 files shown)


janhilgard:
Closing in favor of #653 — I pushed the same fix directly to the PR branch (amended into the original commit, preserving @Thump604 as author with co-author credit for the guard change). See https://github.com/waybarrios/vllm-mlx/pull/653 for the updated PR.

PR #664, opened 2026-07-29
fix: accept standard Qwen MTP shard prefixes

Thump604:
## Summary

Accept both supported standalone Qwen MTP shard key layouts when injecting MTP into a text model:

- `mtp.*`
- `language_model.mtp.*`

The latter is used by composite VLM artifacts that store their dedicated draft tensors in `model-mtp.safetensors`. Previously the injector found that file but discarded every tensor, leaving MTP unavailable.

## Tests

Adds focused coverage for both accepted prefixes and the unrelated-key rejection case.

## Reproduction boundary

Validated against a local Qwen 3.5 122B-A10B VLM MTP composite artifact. The model loaded and returned correct final text after the prefix was normalized. This PR does not change model sampling, MTP scheduling, or performance claims.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,6 +140,7 @@ jobs:
             tests/test_mllm.py \
             tests/test_server.py \
             tests/test_paged_cache.py \
+            tests/test_qwen35_mtp_patch.py \
             tests/test_mllm_continuous_batching.py \
             tests/test_mllm_cache.py \
             tests/test_optimizations.py \
--- a/tests/test_qwen35_mtp_patch.py
+++ b/tests/test_qwen35_mtp_patch.py
@@ -5,6 +5,7 @@ import mlx.core as mx
 
 from vllm_mlx.patches.qwen3_5_mtp import (
     _apply_qwen_mtp_rmsnorm_offset_fixups,
+    _strip_mtp_key_prefix,
 )
 
 
@@ -45,3 +46,12 @@ def test_qwen_mtp_non_norm_one_dimensional_weights_are_not_shifted():
 
     assert shifted == 0
     assert mx.allclose(weights["layers.0.mlp.shared_expert_gate.weight"], original)
+
+
+def test_qwen_mtp_standalone_shards_accept_both_supported_prefixes():
+    assert _strip_mtp_key_prefix("mtp.layers.0.fc.weight") == "layers.0.fc.weight"
+    assert (
+        _strip_mtp_key_prefix("language_model.mtp.layers.0.fc.weight")
+        == "layers.0.fc.weight"
+    )
+    assert _strip_mtp_key_prefix("language_model.model.norm.weight") is None
--- a/vllm_mlx/patches/qwen3_5_mtp.py
+++ b/vllm_mlx/patches/qwen3_5_mtp.py
@@ -25,6 +25,16 @@ from typing import Any
 
 logger = logging.getLogger(__name__)
 
+_MTP_KEY_PREFIXES = ("mtp.", "language_model.mtp.")
+
+
+def _strip_mtp_key_prefix(key: str) -> str | None:
+    """Return an MTP-relative key for supported standalone shard layouts."""
+    for prefix in _MTP_KEY_PREFIXES:
+        if key.startswith(prefix):
+            return key.removeprefix(prefix)
+    return None
+
 _QWEN_MTP_RMSNORM_WEIGHT_SUFFIXES = (
     "input_layernorm.weight",
     "post_attention_layernorm.weight",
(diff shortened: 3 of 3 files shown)


janhilgard (approved):
LGTM. Verified against my production artifact — behavior is byte-identical for the existing layout.

## Regression check on the `mtp.*` layout

`Qwen3.6-27B-Heretic-v2-MTP-mlx-mixed-4-8` stores 15 draft tensors in `mtp/weights.safetensors`, all with the plain `mtp.` prefix. I ran both the old and new stripping logic over the actual key set:

```
15 MTP tensors
old vs new logic — differences: 0
sample: 'mtp.fc.weight' → 'fc.weight'
all accepted by new logic: True
```

`_MTP_KEY_PREFIXES` puts `"mtp."` first, so the existing layout hits the same branch it always did. No behavior change for artifacts that were already working, which is the property that matters most here — this file is in the load path of every Qwen MTP deployment.

## The bug is real and the failure mode was bad

Old code:

```python
raw_mtp = {k.removeprefix("mtp."): v for k, v in raw.items() if k.startswith("mtp.")}
```

For a composite VLM artifact where every key is `language_model.mtp.*`, the `startswith` filter rejects all of them and `raw_mtp` comes out empty. The injector then proceeds with zero weights — no exception, no error log, just MTP silently unavailable on a model whose `model-mtp.safetensors` is sitting right there on disk. That is exactly the class of failure that costs someone an afternoon.

## Implementation

- `_strip_mtp_key_prefix` returning `str | None` rather than a bare bool is the right shape — one pass does both the accept decision and the transform, so the caller can't accidentally strip a prefix it didn't match against.
- The walrus in the dict comprehension keeps that single-pass property visible instead of hiding it behind a helper called twice.
- Module-level `_MTP_KEY_PREFIXES` tuple makes the supported set greppable, which matters when the next composite layout shows up.

## Tests

Covers both accepted prefixes and — importantly — the rejection case (`language_model.model.norm.weight` → `None`). That last assertion is what stops a future "just be permissive" refactor from letting base-model tensors leak into the MTP module.

Given #1754 is about draft keys contaminating base sanitization, having a test that pins the *opposite* direction (base keys must not reach the MTP path) is well placed.

## CI

All 9 checks green.

Ship it.

waybarrios:
Hi, during review we found that the new regression test never ran in CI. The test lives in tests/test_qwen35_mtp_patch.py, and that file was missing from both pytest file lists in .github/workflows/ci.yml. Since the file imports mlx.core at module level, the ubuntu no-MLX job cannot collect it either. So the coverage added by this PR was only executed when someone ran it manually on Apple Silicon.

This fix registers the file in the test-apple-silicon job. The change is one line in .github/workflows/ci.yml:

```yaml
            tests/test_paged_cache.py \
            tests/test_qwen35_mtp_patch.py \
            tests/test_mllm_continuous_batching.py \
```

That is the pytest list of the test-apple-silicon job with the new entry inserted after test_paged_cache.py. With this in place, test_qwen_mtp_standalone_shards_accept_both_supported_prefixes runs on every CI push and the prefix regression is enforced automatically. Thanks again!

waybarrios:
The CI re-run on the updated head (b84c0c7) picked up the file, and the regression test now executes and passes. Log excerpt from the test-apple-silicon (3.11) job, run 30785822055:

```
tests/test_qwen35_mtp_patch.py::test_qwen_mtp_actual_gamma_norm_weights_are_not_shifted_again PASSED [ 35%]
tests/test_qwen35_mtp_patch.py::test_qwen_mtp_non_norm_one_dimensional_weights_are_not_shifted PASSED [ 35%]
tests/test_qwen35_mtp_patch.py::test_qwen_mtp_standalone_shards_accept_both_supported_prefixes PASSED [ 35%]
```

The third line is the new prefix regression test, and it runs as part of the regular Apple Silicon suite. Full job summary:

```
===================== 577 passed, 25 deselected in 22.91s ======================
```

The same suite also passed on the 3.13 leg, and all nine CI checks are green on the updated head. The prefix coverage is now enforced on every push. Thanks again!

waybarrios (approved):
Approving after a full review pass (six review agents plus my own verification of the CI wiring).

What was verified:

- The fix is correct and complete. Both layouts (mtp. and language_model.mtp.) strip to identical module-relative keys, every downstream consumer behaves identically, and no other code path still assumes the single-prefix layout for Qwen3.5 composites.
- No regression for existing artifacts. The new filter is a strict superset of the old one, byte-identical for every key the previous code accepted. The rejection of base-model keys (language_model.model.norm.weight) is deliberate and protective.
- The regression test is mutation-robust: removing either prefix, reverting to the old filter, or returning None each kills the test, and the three assertions cover all helper branches.
- One gap found and fixed during review: the test file was not collected by any CI job. Commit b84c0c7 registers tests/test_qwen35_mtp_patch.py in the test-apple-silicon job, and the re-run confirms the new test executes and passes (577 passed, 25 deselected) on both 3.11 and 3.13.
- All nine CI checks are green on the updated head and the merge state is CLEAN.

The squash merge should carry the natural-language message "Accept both standard Qwen MTP shard key prefixes when loading MTP weights (#664)" per the current commit convention. Ship it.

PR #665, opened 2026-07-29
fix: honor disabled MTP in VLM text routing

Thump604:
## Summary

Make the VLM-derived text route honor the serving engine MTP setting. When `--disable-mtp` is selected, Runtime now builds an ordinary TextModel without loading or injecting draft MTP tensors.

The helper remains backward compatible for direct callers; SimpleEngine passes its explicit MTP state.

## Tests

Adds coverage that the TextModel builder accepts the engine MTP decision. Existing TextModel tests continue to cover the default MTP-capable build path.

## Reproduction boundary

Validated locally with a Qwen 3.6 35B-A3B VLM MTP composite artifact. This PR does not alter model sampling, chat templates, MTP verification, or scheduling.

### DIFF
--- a/tests/test_simple_engine.py
+++ b/tests/test_simple_engine.py
@@ -1469,8 +1469,9 @@ class TestSimpleEngineConcurrency:
         mock_mllm.model = MagicMock()
         mock_mllm.get_tokenizer.return_value = tokenizer
 
-        def build_text_model(*_args, **_kwargs):
+        def build_text_model(*_args, **kwargs):
             captured["build_thread"] = threading.get_ident()
+            captured["enable_mtp"] = kwargs["enable_mtp"]
             return text_model
 
         with (
@@ -1494,6 +1495,7 @@ class TestSimpleEngineConcurrency:
                 assert engine._text_tokenizer is tokenizer
                 assert captured["build_thread"] == worker_thread
                 assert captured["build_thread"] != event_loop_thread
+                assert captured["enable_mtp"] is False
             finally:
                 await engine.stop()
 
--- a/tests/test_text_model_from_vlm.py
+++ b/tests/test_text_model_from_vlm.py
@@ -3,6 +3,7 @@
 
 import json
 import os
+import inspect
 import sys
 import types
 from pathlib import Path
@@ -38,6 +39,12 @@ def test_build_text_model_none_vlm():
     """Returns None when vlm_model is None."""
     result = build_text_model(None, TEXT_MTP_MODEL)
     assert result is None
+
+
+def test_build_text_model_accepts_engine_mtp_decision():
+    parameter = inspect.signature(build_text_model).parameters["enable_mtp"]
+
+    assert parameter.default is True
 
 
 def test_build_text_model_dispatches_gemma4_text_model(tmp_path, monkeypatch):
--- a/tests/test_text_model_from_vlm.py
+++ b/tests/test_text_model_from_vlm.py
@@ -47,6 +47,69 @@ def test_build_text_model_accepts_engine_mtp_decision():
     assert parameter.default is True
 
 
+def test_build_text_model_disabled_mtp_skips_draft_loading_and_injection(
+    tmp_path, monkeypatch
+):
+    """The non-speculative SimpleEngine path must not touch MTP state."""
+
+    model_path = tmp_path / "qwen"
+    model_path.mkdir()
+    (model_path / "config.json").write_text(
+        json.dumps(
+            {
+                "model_type": "qwen3_5",
+                "mtp_num_hidden_layers": 1,
+            }
+        )
+    )
+
+    class FakeLanguageModel:
+        def parameters(self):
+            return {}
+
+    class FakeVlmModel:
+        language_model = FakeLanguageModel()
+
+    class FakeTextModelArgs:
+        @classmethod
+        def from_dict(cls, config):
+            return config
+
+    class FakeTextModel:
+        def __init__(self, args):
+            self.args = args
+            self.mtp = None
+            self.loaded_weights = []
+
+        def load_weights(self, weights, strict=False):
+            self.loaded_weights.append((weights, strict))
+
+        def train(self, mode=True):
+            self.training = mode
+            return self
+
+    qwen_module = types.ModuleType("mlx_lm.models.qwen3_5")
+    qwen_module.TextModel = FakeTextModel
+    qwen_module.TextModelArgs = FakeTextModelArgs
+
+    mtp_loads = []
+
+    def record_mtp_load(path):
+        mtp_loads.append(path)
+        return [("mtp.layers.0.weight", object())]
+
+    monkeypatch.setitem(sys.modules, "mlx_lm.models.qwen3_5", qwen_module)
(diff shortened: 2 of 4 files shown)


janhilgard (approved):
LGTM on the fix. One reservation about the test, non-blocking.

## The fix is correct

`build_text_model` unconditionally called `_load_mtp_weights(model_path)` and then injected an MTP module regardless of what the engine asked for. So `--disable-mtp` on a VLM text route still paid for draft tensor loading and still got an MTP-capable model. Combined with the sanitization ordering bug in `Blaizzy/mlx-vlm#1754`, that meant the presence of a draft shard changed base-model loading *even with MTP explicitly disabled* — the operator's off switch didn't reach the code that mattered.

Threading `enable_mtp` through from `SimpleEngine._mtp` closes that. Three call sites all gated consistently:

- `_load_mtp_weights(...) if enable_mtp else []`
- `if enable_mtp and mtp_weights:` for the load, with the "no MTP weights found" warning correctly moved under `elif enable_mtp` so a deliberate opt-out doesn't emit a spurious warning
- `if enable_mtp and (not hasattr(...) or ... is None)` for the injection

Keyword-only with `default=True` keeps direct callers working. Good.

## My reservation: the test asserts the signature, not the behavior

```python
def test_build_text_model_accepts_engine_mtp_decision():
    parameter = inspect.signature(build_text_model).parameters["enable_mtp"]
    assert parameter.default is True
```

This pins the default value and nothing else. It passes if someone adds `enable_mtp=True` to the signature and never reads it — which is precisely the bug class this PR exists to fix. The three gates that carry the actual behavior are untested.

What would cover it, roughly:

```python
def test_build_text_model_skips_draft_tensors_when_mtp_disabled(monkeypatch):
    calls = []
    monkeypatch.setattr(
        "vllm_mlx.text_model_from_vlm._load_mtp_weights",
        lambda p: calls.append(p) or [],
    )
    build_text_model(fake_vlm, tmp_path, enable_mtp=False)
    assert calls == []                       # loader never invoked
    # and: no mtp attribute injected on the returned model
```

That fails if any of the three gates regress; the current test doesn't.

Not blocking the merge — the fix itself reads correctly and the reproduction boundary is documented. But if this ever gets refactored, the signature test won't catch it, and the composite-artifact path is exactly where a silent regression would hurt.

Happy to push that test onto the branch if you'd rather not context-switch — say the word.

## Impact on my deployments: none

Both my servers (Qwen3.6-27B Heretic-v2 with `--enable-mtp`, Qwen3.6-35B-A3B without) run `--continuous-batching`, so they're on BatchedEngine and never reach `SimpleEngine.start()`'s `build_text_model` call. Nothing for me to verify empirically here — flagging that as a limit on my review rather than a clean bill of health.

## CI

All 9 checks green.

Thump604:
@janhilgard Addressed the behavioral-test reservation in f985137. The new regression calls the actual builder with MTP declared in config and verifies `enable_mtp=False` does not load draft weights, does not inject MTP, and transfers only backbone weights.

Focused verification: `python -m pytest -q tests/test_text_model_from_vlm.py` (`6 passed, 4 skipped`). The new CI run is in progress. Please re-review when convenient.

janhilgard (approved):
Re-reviewed `f985137`. The new regression does what I was asking for — I verified that by trying to break it.

## Empirical check: does the test actually catch the regression?

A behavioral test is only worth the failure it produces, so I ran it against three variants of the production code in a clean clone.

**1. PR as submitted** — `6 passed, 4 skipped`, matching your local run.

**2. Production code reverted to `upstream/main`, test kept:**

```
FAILED test_build_text_model_accepts_engine_mtp_decision
  KeyError: 'enable_mtp'
FAILED test_build_text_model_disabled_mtp_skips_draft_loading_and_injection
  TypeError: build_text_model() got an unexpected keyword argument 'enable_mtp'
```

Both fail, but only because the parameter is gone. That is the easy case.

**3. The case my original review was actually about** — parameter kept in the signature, every gate stripped so nothing reads it:

```python
mtp_weights = _load_mtp_weights(model_path)            # was: ... if enable_mtp else []
if mtp_weights:                                        # was: if enable_mtp and mtp_weights:
else:                                                  # was: elif enable_mtp:
if not hasattr(text_model, "mtp") or ... is None:      # was: if enable_mtp and (...)
```

Result:

```
tests/test_text_model_from_vlm.py ...F.ssss.

FAILED test_build_text_model_disabled_mtp_skips_draft_loading_and_injection
  assert mtp_loads == []
E   Left contains one more item: PosixPath('.../qwen')

1 failed, 5 passed, 4 skipped
```

The signature test **passed** — the parameter was still there, so it had nothing to say. The behavioral test **failed on the loader assertion**. That is exactly the "added the parameter, never read it" regression I was worried about, and it is now covered.

## What makes the test work

Three things I'd have gotten wrong writing it myself:

- **`record_mtp_load` returns a non-empty list.** If the gate regresses, the loader is not only called but returns weights, so `loaded_weights` catches it too even if someone later weakens the `mtp_loads` assertion. Two independent tripwires on one gate.
- **`config.json` declares `mtp_num_hidden_layers: 1`.** Without that the injection path would be skipped for an unrelated reason and the test would pass vacuously. Easy to miss.
- **`assert text_model.loaded_weights == [([], False)]`** pins that backbone transfer still happens exactly once. So the test distinguishes "MTP correctly skipped" from "the whole load path broke", which a bare `assert mtp_loads == []` would not.

Keeping the signature test alongside is fine — it costs nothing and covers the parameter-removal case that variant 2 showed.

## Standing

My earlier approval was on the fix with a reservation about coverage. The reservation is resolved; this is now an unqualified approve.

Unchanged from before: my own deployments run `--continuous-batching` on BatchedEngine, so they never reach this code path. My review is code-and-test reading plus the sabotage runs above, not production evidence.

CI on `f985137` was still in flight when I looked — worth a glance before merge, though the focused suite passed locally for both of us.

Thump604:
@waybarrios Jan has re-reviewed f985137 and resolved the behavioral-test reservation. All CI checks are green. This is now ready for your merge review.

Thump604:
Rebased the two-commit disabled-MTP routing fix onto current `main` (`7afa61a`). Conflict resolution preserved main’s owner-thread `build_text_route()` and forwards `enable_mtp=self._mtp`. The adversarial pass also strengthened the startup regression to directly assert that `SimpleEngine(mtp=False)` passes `enable_mtp=False` to the loader. Verification: full suite `2575 passed, 25 skipped, 27 deselected`; focused TextModel/SimpleEngine `64 passed, 5 skipped`; Black and diff checks clean. New head: `74cd292`.

waybarrios:
@Thump604 could fix the conflicts in this PR?

waybarrios:
I checked the diff in your implementation. I liked, just green lighted you. Update code and can merge right away 

cc @Thump604

Thump604:
Rebased the two-commit fix onto current `main` (`7c7232a`). The conflict came from merged #650; its lazy, owner-thread-safe `_initialize_text_model()` path is preserved, and that refactored builder now receives `enable_mtp=self._mtp`.

Verification: focused TextModel/SimpleEngine `67 passed, 5 skipped`; full local suite `2657 passed, 25 skipped, 27 deselected`; Black, Ruff, and `git diff --check` passed. All 9 GitHub checks are green. New head: `4ad2fe0`.

PR #666, opened 2026-07-30
fix(server): preserve active streaming responses

Thump604:
## Problem
Streaming requests were cancelled after the total request timeout even while the generator continued emitting SSE chunks.

## Change
Treat the streaming timeout as generator-output inactivity. Real chunks reset the timeout; SSE heartbeats continue to probe client disconnects but do not count as generator progress.

## Coverage
- continuous chunks that run longer than the configured timeout complete
- a generator that produces no chunk still times out

## Scope
No changes to sampling, thinking, model routing, response-format enforcement, or clients.

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 """Tests for the OpenAI-compatible API server."""
 
+import asyncio
 import json
 import platform
 import sys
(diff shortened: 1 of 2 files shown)


janhilgard (approved):
LGTM. Verified it does not collide with #644, and it fixes a timeout that actually bites my deployment.

## First: no silent-merge risk here

I checked this specifically because #653 turned out to merge cleanly onto current `main` while being semantically broken — #644 (`94c008b`, today 09:56) moved streaming parser handling to request-local instances, and #653 was still handing the module-level global to its new finalizer.

`_disconnect_guard` is untouched by #644 — zero occurrences in that diff. I test-merged this PR onto `94c008b`: clean merge, and the merged region is internally consistent (`last_chunk_at` at 4351/4413, `_chunk_timeout` at 4328/4360/4381, no global/local split). `-k "disconnect or timeout or stream"` → **27 passed** on the merged tree.

So unlike #653, this one is safe to land on current main as-is.

## The bug is real on my stack, with numbers

`_default_timeout` is set from `--timeout`, and both my servers run `--timeout 1800` with `--max-tokens 32768`. Against the old absolute-deadline behaviour:

| scenario | tok/s | time to max_tokens | old behaviour |
|---|---:|---:|---|
| 27B isolated, MTP engaged | 50 | 10.9 min | completes |
| **27B under 35B GPU contention** | **13** | **42.0 min** | **killed at 30 min** |
| 35B single-user | 70 | 7.8 min | completes |
| 35B at C=8, per-user | 25 | 21.8 min | completes |

The contention row is not hypothetical — two servers share one Apple Silicon GPU, and 27B decode drops to 12–15 tok/s whenever 35B is under load. A long reasoning response that legitimately runs to `max_tokens` gets cut mid-stream with no error the client can distinguish from a model stop. Treating the timeout as generator inactivity fixes exactly that.

## Implementation

- `last_chunk_at` updated only on real chunks (line 4413), heartbeats deliberately excluded — correct, since a heartbeat proves the socket is alive, not that generation is progressing. Counting them would make the timeout unreachable.
- `timeout=min(heartbeat_interval, _chunk_timeout - idle_seconds)` on the `asyncio.wait` is a nice detail: without it the loop could sleep a full heartbeat interval past the inactivity deadline. The subtraction can't go negative because the `idle_seconds >= _chunk_timeout` break runs first.
- Log message changed to `OUTPUT INACTIVITY TIMEOUT ... without a generator chunk` — operators reading logs will not mistake this for the old semantics.
- Disconnect detection is unchanged, so the zombie-stream case that motivated the original absolute timeout is still covered by `is_disconnected()` plus heartbeat-forced writes.

## One consequence worth stating plainly

The absolute ceiling is gone. A generator that keeps emitting *anything* faster than `_chunk_timeout` now runs unbounded. In practice `max_tokens` bounds it, and a wedged generator produces nothing and still times out — so I don't think this needs a second knob today.

But it does mean `--timeout` changes meaning: it was "maximum stream lifetime", it is now "maximum silence". Anyone who set it as a cost guard rather than a liveness guard will get different behaviour without any config change. Worth a line in the release notes; not worth blocking on.

## Tests

Both cases covered — continuous chunks past the configured interval complete, silent generator still times out. That's the pair that matters.

## CI

All 9 checks green. Note they ran at 00:52, before #644 landed — but since `_disconnect_guard` is disjoint from that change and I re-ran the relevant tests on the merged tree myself, that's fine here.

Ship it.

Thump604:
@waybarrios Jan approved the current clean head after test-merging it onto post-#644 main and rerunning the relevant stream suite. All CI checks are green. This is ready for merge review.

waybarrios:
I was checking carefully and it seems good to me.

PR #667, opened 2026-07-31
fix(embeddings): release MLX buffers after each batch

SvaRgoS:
## Problem

`/v1/embeddings` never releases the Metal buffers it allocates, so the server's memory only grows. Every LLM path calls `mx.clear_cache()` (`worker.py`, `engine_core.py`, `specprefill.py`); `embedding.py` does not.

MLX keeps freed buffers in its allocator pool, keyed by size. `EmbeddingEngine.embed()` tokenizes with `padding=True`, so the sequence length is whatever the longest text in that batch happens to be — with varied inputs almost every request asks for buffer sizes the pool has never seen and cannot reuse. Nothing ever shrinks it.

`DELETE /v1/cache` does not help: it only reaches `mlx_vlm` caches and returns `{"status":"cleared","error":"Cache clear not available (mlx_vlm not loaded)"}` on an embeddings-only server.

## Reproduction

Server started for embeddings only, with the memory flags already recommended for that setup:

```
vllm-mlx serve <small-stub-llm> \
  --embedding-model mlx-community/multilingual-e5-large-mlx \
  --lazy-load-model --auto-unload-idle-seconds 180 --cache-memory-percent 0.05
```

Then batches of 32 texts of randomly varied length (10–320 words), measuring `footprint -p <pid>`:

| after | phys_footprint |
|---|---|
| model load | 2363 MB |
| 32 texts | 4673 MB |
| 160 texts | 13 GB |
| 320 texts | 24 GB |

Roughly 70 MB retained per input text, monotonic, no plateau. None of `--lazy-load-model`, `--auto-unload-idle-seconds` or `--cache-memory-percent` affects it — the growth is not the main model and not the KV/prefix cache.

In production this reached **50 GB** `phys_footprint` on a week-old process. `footprint` attributed 49 GB of it to `IOAccelerator (graphics)` across 3442 regions, with `Reclaimable` = 0, which exhausted swap on a 64 GB machine.

## Fix

Call `mx.clear_cache()` after the embeddings are converted to Python lists — matching what every other engine path already does.

Same measurement with the patch applied:

| after | before | after fix |
|---|---|---|
| model load | 2363 MB | 1923 MB |
| 32 texts | 4673 MB | 3636 MB |
| 160 texts | 13 GB | 3214 MB |
| 320 texts | 24 GB | 3571 MB |

Flat at ~3.5 GB, of which the model itself is 1.1 GB.

### Does clearing the pool cost throughput?

The pool exists so buffers get reused, so the fair objection is that clearing it unconditionally throws away reuse. Measured on both builds, same seeded inputs:

| workload | throughput (fix / no fix) | growth over 320 texts (fix / no fix) |
|---|---|---|
| varied lengths (10–320 words) | 19.2 / 19.1 texts/s | flat ~3.5 GB / **2.3 → 24 GB** |
| uniform length (150 words) | 40.4 / 40.1 texts/s | +1.6 GB / **+3.2 GB** |

Throughput is identical in both. And the uniform case — the one where the pool *should* be reused — still grows twice as fast without the fix, so the reuse being protected is largely theoretical here.

That is what makes the unconditional call reasonable rather than a threshold on `mx.get_cache_memory()`: it costs nothing measurable on either workload, and there is no cross-request state worth keeping for embeddings, unlike generation where a KV cache spans requests.

## Environment

- macOS 15 (Darwin 25.5.0), Apple M2 Max, 64 GB
- vllm-mlx 0.4.0, mlx 0.31.2, mlx-embeddings 0.1.0
- model: `mlx-community/multilingual-e5-large-mlx`

## Note

This is adjacent to #641 (`MemoryAwarePrefixCache` retaining Metal buffers) — a different code path, but the same failure mode of Metal buffers never being handed back.

### DIFF
--- a/tests/test_embeddings.py
+++ b/tests/test_embeddings.py
@@ -96,6 +96,41 @@ class TestEmbeddingEngine:
         assert len(result) == 2
         assert result[0] == [0.1, 0.2]
 
+    @patch("vllm_mlx.embedding.EmbeddingEngine.load")
+    @patch(
+        "vllm_mlx.embedding.EmbeddingEngine.is_loaded",
+        new_callable=lambda: property(lambda self: True),
+    )
+    def test_embed_clears_mlx_cache_after_batch(self, _mock_loaded, mock_load):
+        """Test embed releases MLX buffers after converting to Python lists."""
+        import numpy as np
+
+        from vllm_mlx.embedding import EmbeddingEngine
+
+        engine = EmbeddingEngine("test-model")
+
+        mock_output = MagicMock()
+        mock_output.text_embeds.tolist.return_value = [[0.1, 0.2], [0.3, 0.4]]
+
+        mock_model = MagicMock(return_value=mock_output)
+
+        mock_inner_tokenizer = MagicMock()
+        mock_inner_tokenizer.return_value = {
+            "input_ids": np.array([[1, 2], [3, 4]]),
+            "attention_mask": np.array([[1, 1], [1, 1]]),
+        }
+        mock_tokenizer = MagicMock()
+        mock_tokenizer._tokenizer = mock_inner_tokenizer
+
+        engine._model = mock_model
+        engine._tokenizer = mock_tokenizer
+
+        with patch("vllm_mlx.embedding.mx.clear_cache") as mock_clear_cache:
+            result = engine.embed(["hello", "world"])
+            mock_clear_cache.assert_called_once()
+
+        assert result[0] == [0.1, 0.2]
+
     def test_embed_normalises_single_string(self):
         """Test that a single string input is wrapped into a list."""
         import numpy as np
--- a/vllm_mlx/embedding.py
+++ b/vllm_mlx/embedding.py
@@ -82,7 +82,17 @@ class EmbeddingEngine:
         embeds: mx.array = output.text_embeds
 
         # Convert to Python lists for JSON serialization
-        return embeds.tolist()
+        result = embeds.tolist()
+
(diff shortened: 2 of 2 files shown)


waybarrios:
Nice fix, and a genuinely great writeup — the repro table, the before/after footprint measurements, and especially the throughput comparison (varied vs. uniform-length batches) preemptively answer the obvious objection ("doesn't clearing the pool throw away reuse?") with actual data instead of hand-waving. That's exactly the kind of evidence a one-line change touching a hot path deserves.

A few things I checked and liked:
- Root cause makes sense: `padding=True` means the sequence length varies per batch, so the allocator pool keyed by size almost never gets a hit here — unlike the KV cache in the generation path, there's no cross-request state worth protecting, so clearing unconditionally is the right call rather than a `mx.get_active_memory()` threshold like `engine_core.py` uses for the generation hot path.
- Consistent with the existing convention: every other MLX compute path in the codebase (`worker.py`, `engine_core.py`, `specprefill.py`) already calls `mx.clear_cache()`; this was just the one path that didn't.
- No correctness/concurrency concern: `clear_cache()` only releases already-unreferenced buffers, so it's safe regardless of what else might be running.

Only minor, non-blocking suggestion: `tests/test_embeddings.py` doesn't have a test asserting `mx.clear_cache()` gets called — a quick mock/spy would make this leak fix regression-proof. Not needed to merge.

waybarrios:
Added a regression test (`test_embed_clears_mlx_cache_after_batch`) that mocks `mx.clear_cache` and asserts it's called once per `embed()` batch — the leak fix is now regression-proof.

PR #669, opened 2026-08-01
fix: enable configured MLLM assistant drafters by default

Thump604:
## Problem

A server configured with `--mllm-draft-model` loads the assistant drafter, but text-only requests remain on the TextModel route unless every client sends the private `mllm_draft=true` field. The configured drafter therefore does not participate by default, and terminal MTP counts are not exposed in `generation_metadata` or SimpleEngine status.

## Change

- Add `--default-mllm-draft` for explicitly enabling a configured assistant drafter at server startup.
- Preserve per-request `mllm_draft=false` opt-out.
- Route the default-enabled path through the MLLM adapter and report terminal MTP draft/accept counters.
- Report assistant-drafter implementation and its SimpleEngine-only constraint in status.

## Scope

This changes only configured MLLM assistant-drafter behavior. It does not add continuous-batching support, alter native TextModel MTP, change model templates or sampling, or make a model-specific qualification claim.

## Validation

- `30 passed`: assistant-drafter, metadata, CLI, and server forwarding tests.
- `ruff check` on all touched source and test files.
- Manual BF16 Gemma 4 assistant-drafter smoke verified final content plus nonzero terminal draft/accept counters with no request opt-in.

## Not claimed

This PR does not claim batch-scheduler compatibility, workload quality, tool-flow qualification, or performance improvement.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,7 +84,9 @@ jobs:
             tests/test_anthropic_adapter.py \
             tests/test_dependency_floors.py \
             tests/test_ssd_cache_shutdown.py \
+            tests/test_chat_template_safety.py \
             tests/test_simple_engine_thread_pinning.py \
+            tests/test_engine_base.py \
             tests/test_harmony_parsers.py \
             tests/test_endpoint_model_policies.py \
             tests/test_truncation.py \
@@ -113,6 +115,28 @@ jobs:
         with:
           file: ./coverage.xml
           fail_ci_if_error: false
+
+  test-python-3-14:
+    # Focused job for the CPython 3.14+ asyncio.shield() behavior change
+    # that shield_task() (vllm_mlx/engine/base.py) works around -- the
+    # regular test-matrix only covers 3.10-3.13, none of which reproduce it.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python 3.14
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.14"
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install pytest anyio pytest-cov
+
+      - name: Run engine base tests
+        run: |
+          pytest tests/test_engine_base.py -v --tb=short
 
   test-apple-silicon:
     runs-on: macos-14
@@ -147,6 +171,7 @@ jobs:
             tests/test_paged_cache.py \
             tests/test_prefix_cache_untrimmable.py \
             tests/test_qwen35_mtp_patch.py \
+            tests/test_batched_engine_mllm_config.py \
             tests/test_mllm_continuous_batching.py \
             tests/test_mllm_cache.py \
             tests/test_optimizations.py \
@@ -191,7 +216,7 @@ jobs:
             -k "not Integration"
 
   tests:
-    needs: [test-matrix, test-apple-silicon]
+    needs: [test-matrix, test-python-3-14, test-apple-silicon]
     runs-on: ubuntu-latest
     if: always()
     steps:
@@ -199,6 +224,10 @@ jobs:
         run: |
           if [ "${{ needs.test-matrix.result }}" != "success" ]; then
             echo "Unit tests failed"
+            exit 1
+          fi
+          if [ "${{ needs.test-python-3-14.result }}" != "success" ]; then
+            echo "Python 3.14 tests failed"
             exit 1
           fi
           if [ "${{ needs.test-apple-silicon.result }}" != "success" ]; then
--- a/docs/guides/continuous-batching.md
+++ b/docs/guides/continuous-batching.md
@@ -38,6 +38,25 @@ that envelope fall back to the normal scheduler path instead of using MTP. This
 keeps sampling correctness ahead of throughput until the MLLM verifier is
 sampler-aware.
 
+An external mlx-vlm assistant drafter can be configured for either simple or
+continuous-batching mode:
+
+```bash
+vllm-mlx serve TARGET_MODEL \
+  --mllm \
+  --continuous-batching \
+  --mllm-draft-model ASSISTANT_MODEL \
+  --mllm-draft-kind mtp \
+  --default-mllm-draft
+```
+
+`--default-mllm-draft` enables the assistant when a chat or completion request
+does not specify an override. Send `"mllm_draft": false` to opt out for one
+request, or `true` to opt in when the server default is disabled. Batched
+assistant drafters currently require `--mllm-draft-kind mtp`. `/v1/status`
+shows the configured model and default after startup. In continuous-batching
+mode, it also adds runtime acceptance counters as requests run.
+
 Injected Qwen3.5 and Qwen3.6 MTP checkpoints use the established post-norm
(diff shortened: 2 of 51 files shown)


janhilgard (changes requested):
Both CI failures come from this PR, and I traced them to a specific cause rather than guessing. The fixes are small.

## `lint` — `black` not run

```
would reformat tests/test_mllm_assistant_drafter.py
would reformat vllm_mlx/engine/simple.py
```

## `tests` — the new test leaks module-level state into later tests

Three failures, all `assert 404 == 400`:

```
TestMaxTokensLimit::test_create_completion_rejects_over_limit_before_engine_lookup
TestMaxTokensLimit::test_create_chat_completion_rejects_over_limit_before_engine_lookup
TestMaxTokensLimit::test_create_anthropic_message_rejects_over_limit_before_engine_lookup

HTTPException(404, 'The model `test-model` does not exist. Available model: `gemma4`')
```

That `gemma4` in the error is the tell. The chain:

1. `server.py:3237` — `load_model()` writes the module global `_model_name = served_model_name or model_name`
2. The pre-existing tests in `TestLoadModelTrustRemoteCode` (`test_server.py:720+`) all call `load_model("test-model", …)`, so the global is left as `"test-model"`
3. `TestMaxTokensLimit` (`test_server.py:1273+`, later in the file) sends requests with `model="test-model"` — the name matches, the handler proceeds to the max-tokens check, and returns **400**
4. This PR adds `test_load_model_forwards_default_mllm_draft_to_simple_engine` at `test_server.py:759`, calling `load_model("gemma4", …)` — the global becomes `"gemma4"`
5. The later tests no longer match, model lookup returns **404** before max-tokens validation is reached

## Verified by running, not by reading

Four runs on a clean checkout of `2e0ca73`:

| Selection | Result |
|---|---|
| `-k "MaxTokensLimit"` alone | **6 passed** |
| `-k "load_model_forwards_default or MaxTokensLimit"` | **3 failed**, 4 passed |
| `-k "load_model_batched_forwards or MaxTokensLimit"` (existing `"test-model"` test) | **7 passed** |
| `black --check` on the two files | 2 files would be reformatted |

The third row is the control: it rules out `load_model()` as such and pins the cause to the model *name* the new test passes.

I also confirmed both candidate fixes make the combined selection green (**7 passed** each):

- **A — minimal:** change `"gemma4"` to `"test-model"` in the new test
- **B — robust:** take `monkeypatch` in the test and restore `_model_name` after it

## Which fix I'd pick

**B.** The existing suite only passes because every `load_model` test happens to use the same name — an unwritten convention nothing enforces. Your test is the first to deviate, and it will not be the last. Restoring the global turns a latent ordering dependency into a local concern.

A works today and is one word, so if you want the PR unblocked quickly that is legitimate — it just leaves the trap in place for whoever writes the next `load_model` test.

Worth noting for later: any test calling `load_model()` mutates `_default_max_tokens`, `_max_request_tokens`, `_model_path`, `_model_name`, `_force_mllm_model` and more. A shared autouse fixture snapshotting and restoring those would fix the whole class of problem, but that is a separate change, not yours to carry.

## The feature itself reads fine

The `--default-mllm-draft` plumbing is consistent — CLI → `load_model` → `SimpleEngine.__init__` → `prepare_for_start`, with the `default_mllm_draft and not mllm_draft_model` guard rejecting the meaningless combination early. Per-request `mllm_draft=false` opt-out is preserved, so the flag widens the default without removing control.

`_generation_metadata(thinking_processor, output)` gaining a second parameter with a default keeps existing callers working, and the `thinking_processor is None and not (mtp_drafts or mtp_accepted)` condition correctly still returns `None` when there is genuinely nothing to report.

## No production evidence from me

Both my servers run `--continuous-batching`, so they are on BatchedEngine and never reach the SimpleEngine assistant-drafter path this PR changes. I have not exercised this feature against a real drafter — my review is code reading plus the test runs above. The `SimpleEngine-only` constraint you surface in status is the right call to make visible.

Requesting changes only for the two CI failures. Happy to push either fix onto the branch if you'd rather not context-switch — say the word.

Thump604:
Addressed both requested changes in `623aaf8`.

- Ran Black on `vllm_mlx/engine/simple.py`, `tests/test_mllm_assistant_drafter.py`, and the touched server test.
- Registered `server._model_name` with `monkeypatch` in the new `load_model` test so teardown restores module state and later request-validation tests cannot inherit `gemma4`.

Verification:
- combined regression selection: `7 passed`
- assistant drafter + watchdog + server suites: `136 passed, 3 deselected`
- Ruff: passed
- `git diff --check`: passed

janhilgard (approved):
Verified `623aaf8`. You picked option B, and I confirmed the guard actually guards — by removing it.

## Runs on a fresh checkout

| Check | Result |
|---|---|
| `-k "load_model_forwards_default or MaxTokensLimit"` (the combination that failed) | **7 passed** |
| Full `tests/test_server.py` | **120 passed**, 3 deselected |
| `black --check` on all three touched files | clean |

The full-file run matters more than the targeted one here: a state leak can surface anywhere downstream of the mutation, not only in the tests I happened to name. 120/120 says nothing else was inheriting `gemma4` either.

## The guard is load-bearing, not decorative

A passing test proves nothing about a protection unless you can make it fail. So I deleted exactly one line from your test — `monkeypatch.setattr(server, "_model_name", server._model_name)` — and left everything else intact:

```
FAILED TestMaxTokensLimit::test_create_completion_rejects_over_limit_before_engine_lookup
FAILED TestMaxTokensLimit::test_create_chat_completion_rejects_over_limit_before_engine_lookup
FAILED TestMaxTokensLimit::test_create_anthropic_message_rejects_over_limit_before_engine_lookup
3 failed, 4 passed
```

Restored it: **7 passed** again. So the line is doing the work, and a future refactor that drops it will be caught rather than silently reintroducing the ordering dependency.

## On the idiom

`monkeypatch.setattr(server, "_model_name", server._model_name)` sets the attribute to its own current value purely to register it for teardown. That reads as a no-op at a glance, which is exactly why the comment above it earns its place:

```python
# load_model mutates module-level serving state. Register the current
# value with monkeypatch so this test cannot affect later request tests.
```

Without that, someone would delete the line as dead code within a release or two. With it, the intent survives.

## Standing

Earlier request-changes was purely about the two CI failures. Both are resolved, the fix is the more robust of the two options, and it is self-verifying. Approving.

Unchanged from my first review: my servers run `--continuous-batching` and never reach the SimpleEngine assistant-drafter path, so I have no production evidence for the feature itself — this is code reading plus the runs above.

The broader point from last time still stands as a separate item: `load_model()` also mutates `_default_max_tokens`, `_max_request_tokens`, `_model_path` and `_force_mllm_model`, and only `_model_name` is now protected. An autouse fixture snapshotting the set would close the class rather than the instance. Not this PR's job.

Thump604:
Rebased the two-commit configured-assistant-default and metadata/state-isolation patch onto current `main` (`7afa61a`). Conflicts were limited to preserving current prefix-trie additions in `simple.py`/`server.py` plus formatting. Verification: full suite `2578 passed, 25 skipped, 27 deselected`; focused assistant/server/lifecycle suites `209 passed, 9 deselected`; Black, compile, and diff checks clean. Scope remains limited to configured-drafter default behavior and the covered metadata paths; no broader streaming/zero-counter observability claim. New head: `f293ccf`.

waybarrios (approved):

PR #670, opened 2026-08-01
feat(mtp): support assistant drafters with MLLM batching

Thump604:
## Problem

External MLLM assistant drafters are currently rejected when continuous batching is enabled, even though the batched scheduler already has request-local speculative state, verification, rollback, and cancellation handling for native MTP. This prevents Gemma 4 assistant-MTP from using the same concurrent serving path.

## Change

- pass an external MLLM drafter and its kind/block size through `BatchedEngine` into `MLLMScheduler`
- install that drafter in the existing UID-scoped MLLM speculative verifier
- bind shared target KV/positions for each request before proposing one token
- preserve request-local sampler verification, rollback, deferred-step state, cancellation cleanup, and metadata counters
- stop rejecting `--mllm-draft-model` with `--continuous-batching`

The one-token proposal size intentionally matches the effective concurrent MTP path already used by Qwen. This does not change native MTP selection and remains opt-in through the existing drafter flags.

## Verification

- full suite: `2268 passed, 24 skipped, 23 deselected`
- focused MLLM/CB/drafter suite: `90 passed, 4 deselected`
- Black, Ruff, and `git diff --check`: pass
- live Gemma 4 31B target plus matching assistant drafter at 128K KV:
  - single request: final content preserved, `finish_reason=stop`
  - two concurrent requests: both final contents preserved, both `finish_reason=stop`
  - streamed request: reasoning/final separation preserved through `[DONE]`
  - final counters: 142 drafted, 95 accepted, 47 rejected, 0 speculative errors

## Scope

This PR does not add a default drafter, change sampling, alter model templates, or claim a throughput improvement. It adds the missing batched execution path and validates correctness/engagement; performance measurement is separate.

## Compatibility boundary

`_model_parts()` preserves the existing tuple behavior and bypasses objects without `hidden_states`. It intentionally widens speculation to model outputs exposing both `.logits` and `.hidden_states`; those objects previously bypassed speculation. External assistant verification also depends on `draft_block(..., greedy=True)`: the draft is a point mass, so sample-and-compare preserves the target distribution. Changing the external draft to stochastic sampling requires a rejection-sampling verifier that accounts for the draft distribution.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,6 +147,7 @@ jobs:
             tests/test_paged_cache.py \
             tests/test_prefix_cache_untrimmable.py \
             tests/test_qwen35_mtp_patch.py \
+            tests/test_batched_engine_mllm_config.py \
             tests/test_mllm_continuous_batching.py \
             tests/test_mllm_cache.py \
             tests/test_optimizations.py \
--- a/tests/test_batched_engine_mllm_config.py
+++ b/tests/test_batched_engine_mllm_config.py
@@ -82,6 +82,66 @@ def test_start_mllm_forwards_prefix_cache_disable_to_mllm_scheduler(monkeypatch)
     assert captured["config_kwargs"]["ssd_cache_max_gb"] == 10.0
 
 
+def test_start_mllm_forwards_external_assistant_drafter(monkeypatch):
+    from vllm_mlx.engine.batched import BatchedEngine
+
+    captured = {}
+    loaded_drafter = object()
+
+    class FakeMLXMultimodalLM:
+        def __init__(self, model_name, trust_remote_code=True, **kwargs):
+            captured["model_kwargs"] = kwargs
+            self.model = object()
+            self.processor = object()
+            self._draft_model = loaded_drafter
+
+        def load(self):
+            return None
+
+    class FakeMLLMSchedulerConfig:
+        def __init__(self, **kwargs):
+            self.__dict__.update(kwargs)
+
+    class FakeMLLMScheduler:
+        def __init__(self, model, processor, config, **kwargs):
+            captured["scheduler_kwargs"] = kwargs
+
+        async def start(self):
+            return None
+
+    import vllm_mlx.engine.batched as batched_mod
+
+    fake_scheduler = types.ModuleType("vllm_mlx.mllm_scheduler")
+    fake_scheduler.MLLMScheduler = FakeMLLMScheduler
+    fake_scheduler.MLLMSchedulerConfig = FakeMLLMSchedulerConfig
+    fake_model = types.ModuleType("vllm_mlx.models.mllm")
+    fake_model.MLXMultimodalLM = FakeMLXMultimodalLM
+    monkeypatch.setitem(sys.modules, "vllm_mlx.mllm_scheduler", fake_scheduler)
+    monkeypatch.setitem(sys.modules, "vllm_mlx.models.mllm", fake_model)
+    monkeypatch.setattr(
+        batched_mod.BatchedEngine, "_inject_mtp_mllm", lambda self: None
+    )
+
+    engine = BatchedEngine(
+        model_name="gemma4",
+        scheduler_config=_base_scheduler_config(enable_mtp=False),
+        force_mllm=True,
+        mllm_draft_model="assistant",
+        mllm_draft_kind="mtp",
+        mllm_draft_block_size=6,
+    )
+    asyncio.run(engine._start_mllm())
+
+    assert captured["model_kwargs"]["draft_model"] == "assistant"
+    assert captured["model_kwargs"]["draft_kind"] == "mtp"
+    assert captured["model_kwargs"]["draft_block_size"] == 6
+    assert captured["scheduler_kwargs"] == {
+        "draft_model": loaded_drafter,
+        "draft_kind": "mtp",
+        "draft_block_size": 6,
+    }
+
+
 def _run_start_mllm(monkeypatch, scheduler_config):
     """Run BatchedEngine._start_mllm with fakes, return captured kwargs."""
     from vllm_mlx.engine.batched import BatchedEngine
--- a/tests/test_batched_engine_mllm_config.py
+++ b/tests/test_batched_engine_mllm_config.py
@@ -8,6 +8,7 @@ import asyncio
 import sys
 import types
 from types import SimpleNamespace
+from unittest.mock import AsyncMock
 
 
 def test_start_mllm_forwards_prefix_cache_disable_to_mllm_scheduler(monkeypatch):
@@ -140,6 +141,32 @@ def test_start_mllm_forwards_external_assistant_drafter(monkeypatch):
         "draft_kind": "mtp",
         "draft_block_size": 6,
     }
+
+
+def test_generate_forwards_mllm_draft_opt_in():
+    from vllm_mlx.engine.batched import BatchedEngine
+
+    scheduler = SimpleNamespace(
+        generate=AsyncMock(
+            return_value=SimpleNamespace(
+                output_text="ok",
+                output_token_ids=[1],
+                prompt_tokens=2,
+                completion_tokens=1,
+                finish_reason="stop",
+                mtp_drafts=1,
+                mtp_accepted=1,
+            )
+        )
+    )
+    engine = BatchedEngine.__new__(BatchedEngine)
+    engine._loaded = True
+    engine._is_mllm = True
+    engine._mllm_scheduler = scheduler
+
+    asyncio.run(engine.generate("hello", mllm_draft=True))
+
+    assert scheduler.generate.await_args.kwargs["mllm_draft"] is True
 
 
 def _run_start_mllm(monkeypatch, scheduler_config):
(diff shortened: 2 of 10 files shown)


janhilgard (approved):
This is the one I've been waiting for. It unblocks a deployment I retired two months ago, and the acceptance criterion is correct — I checked it numerically rather than trusting the shape.

## Direct impact on my stack

I retired Gemma 4 12B-IT + assistant MTP on 2026-06-07 for exactly the constraint this PR removes. My note at the time:

> `--max-num-seqs 1` cannot be enforced on the SimpleEngine path (required for drafter MTP) — the flag is accepted but never read.

`--mllm-draft-model` forced SimpleEngine, SimpleEngine ignores `max_num_seqs` (`engine/simple.py` never reads it; only `scheduler.py` and `mllm_scheduler.py` do), so a drafter deployment had no concurrency ceiling and no queue rejection. Under N clients everyone got 60/N tok/s with nothing to bound it. Routing the drafter through BatchedEngine makes that flag real again.

Both artifacts are still on disk here — `gemma-4-12B-it-mlx-mixed-4-8` (7.5 GB) + `gemma4-12b-assistant` (837 MB), and the 31B pair (20 GB + 926 MB), plists parked at `RunAtLoad: false`. So this is a redeploy, not a rebuild.

## The acceptance criterion is correct — and correct for a non-obvious reason

The external path diverges from what #662 established for native MTP:

```python
# native (#662): stochastic draft, rejection sampling
accepted = _accept_sampled_draft(target_logprob, draft_logprob, uniform_draw)   # min(1, p/q)

# external (#670): greedy draft, sample-and-compare
draft_tokens = draft_model.draft_block(..., greedy=True)[:, 0]
sampled_target = sampler(verify_distribution)
all_accepted = sampled_target.tolist() == draft_list
```

That is *not* `min(1, p/q)`, and at first read it looks like it drops the rejection-sampling guarantee. It does not — but only because the draft is greedy.

Monte Carlo, 400k draws, target `p = [0.45, 0.25, 0.20, 0.10]`:

```
NATIVE   (stochastic draft q + min(1,p/q)):  0.8493 sim vs 0.8500 theory
EXTERNAL (greedy draft + sample/compare):    0.4502 sim vs 0.4500 = p(argmax)

emitted distribution: 0.4502  0.2493  0.2008  0.0997
target p:             0.4500  0.2500  0.2000  0.1000
max deviation: 0.00079
```

The output distribution is preserved. The reason: with a greedy draft, `q` is a point mass, so accepting exactly when a target sample equals the draft accepts with probability `p(x_draft)` — and on rejection the emitted token *is* that same target sample. Every emitted token is a valid draw from `p` either way.

**This correctness depends entirely on `greedy=True`.** Flip that flag and the scheme silently starts biasing output, with no test failure to catch it — acceptance rate would simply drift. Worth a comment at the `draft_block(..., greedy=True)` call saying the verifier assumes a point-mass draft, so a future change to that argument is understood as a correctness change rather than a tuning knob.

Also note the acceptance rate is structurally lower on this path (0.45 vs 0.85 in my toy case) because a greedy draft can only ever match the target's mode. Your live numbers — 142 drafted / 95 accepted = 66.9% — are consistent with a peaked real distribution. That is expected, not a defect, but it means "external acceptance %" and "native acceptance %" are not comparable quantities and shouldn't be charted on one axis.

## Native path is untouched — verified

My production 27B runs `--enable-mtp` (native head), never `--mllm-draft-model`, so it takes `external_drafter = False` throughout. The only thing that changes on that path is the `_model_parts()` extraction. I checked the equivalence directly:

| input | old behaviour | `_model_parts` |
|---|---|---|
| `tuple` (what Qwen `return_hidden=True` gives) | unpack | **identical unpack** |
| object without `hidden_states` | bypass to `_orig_step` | `hidden` is `None` → same bypass |
| object *with* `hidden_states` | bypass to `_orig_step` | **now enters the MTP path** |

Row three is a real behaviour change. It's a widening rather than a regression, and it's presumably deliberate to accommodate the assistant drafter's output shape — but any model returning a `.logits`/`.hidden_states` object was previously guaranteed to skip speculation and no longer is. Worth stating in the PR body so it isn't discovered by surprise.

Test runs on `e5a607b`: `test_mllm_continuous_batching.py` + `test_batched_engine_mllm_config.py` + `test_lifecycle_cli.py` → **63 passed, 4 deselected**.

## Guard removal

Dropping the `cli.py` check is correct now that the path exists. Worth remembering that with `KeepAlive=true` under launchd, that error used to produce a restart loop rather than a clean failure — so anyone who worked around it by removing `--continuous-batching` from a launch script will want to put it back.

## What I'm not vouching for

Throughput. The PR explicitly doesn't claim it and I haven't measured it. Your correctness evidence (single, concurrent, streamed, all `finish_reason=stop`, 0 speculative errors) is the right bar for this change; the interesting question — whether assistant MTP under batching beats no-MTP under batching — is separate work.

I intend to redeploy Gemma 4 12B on this path once it lands and will report acceptance and throughput under real concurrency. That's the data point neither of us has yet.

Approving. Clean design, honest scope, and the one subtlety in it is load-bearing rather than accidental.

Thump604:
Thank you for checking the acceptance rule numerically and for identifying the load-bearing invariant. I added an inline comment at `draft_block(..., greedy=True)` stating that sample-and-compare depends on a point-mass draft and that stochastic drafting requires a rejection-sampling verifier accounting for q(x). I also updated the PR body to make the `.logits`/`.hidden_states` object-result widening explicit. Focused verification remains 63 passed, 4 deselected. Commit: `1ae4fb9`.

Thump604:
Added commit 34af62f after a production-scale Gemma 4 assistant-MTP failure exposed two correctness gaps.

When a second request joined an active batch at a different KV position, the external drafter used one batched position and eventually reached `mlx_vlm/models/cache.py::_update_in_place` with negative dimensions. The scheduler then retried the partially mutated batch indefinitely while SSE emitted heartbeats.

The update routes mixed-position assistant drafting through mlx-vlm’s rowwise active-batch helper and fails/drains every request sharing a batch after an unrecoverable scheduler step. Regression coverage verifies positions 10,377/10,327 are drafted independently and that all affected request streams terminate exactly once.

Validation: `tests/test_mllm_continuous_batching.py -m "not slow"` — 50 passed, 4 deselected.

waybarrios:
Pushed d53af16 to preserve assistant-drafter correctness. It now validates batched MTP configuration early, honors per-request `mllm_draft` opt-in with safe batch isolation, and replays stochastic rejection samples without biasing the target distribution. Regression tests cover request propagation, queue isolation, validation, and rejection replay.

waybarrios:
Just fixed the ci conflict.

waybarrios (approved):
Just added additional tests in the final PR version. @janhilgard could you take a quick look?

janhilgard (approved):
Re-approving after the five commits since my August 1 review. The substantive one is `fix(mtp): isolate mixed-position assistant batches`, and the sampler-aware verification alongside it is a bigger deal than the PR title suggests.

**This fixes a limitation we have documented in production and worked around.** Our notes on native MTP say the acceptance figure only holds for deliberately greedy requests, and that ordinary client traffic at `temperature=0.7, top_p=0.9` never engages MTP at all — the argmax-equality check bypasses it silently. Your comment names exactly that:

> The old argmax equality check was valid only for greedy decoding and silently bypassed Qwen's normal temperature/top-p/top-k requests.

Verifying in each request's sampler space via `_sampling_logprobs` is the right fix, and drafting greedily under it is sound because the external draft is a point mass — worth that the comment at 2317 says so rather than leaving it to be re-derived.

**Checked the greedy invariant against our own configuration**, since `_request_uses_stochastic_sampling` keys on temperature alone and our servers pass `top_p` from the chat template even at temperature 0:

```
temp=0,   top_p=0.9   -> stochastic=False   (our greedy path, still greedy)
temp=0.7, top_p=0.9   -> stochastic=True    (ordinary client, now verified)
temp=0,   defaults    -> stochastic=False
temp=1.0              -> stochastic=True
```

That is correct and the docstring's reasoning holds: with temperature 0 the filters can only remove tokens below the maximum, so the argmax is invariant to them. Keying on temperature alone is not an oversimplification here.

`_draft_external_mtp_active_batch` threading explicit `positions` rather than letting mixed KV positions conflate is the kind of thing that produces plausible-but-wrong drafts rather than a crash, so I am glad it is isolated in a named function with the invariant in the docstring.

`tests/test_mllm_continuous_batching.py`, `test_batched_engine_mllm_config.py`, `test_qwen35_mtp_patch.py` at your head: **62 passed, 4 deselected**. CI green on all nine.

One thing I cannot verify and would not want assumed from this approval: whether sampler-aware verification actually **helps** on our workload. Practitioners on the oMLX side report ~80–85% acceptance in logs with negligible end-to-end gain, and specifically "nothing or slower" on MoE — which is what our 35B-A3B is. Correctness-wise this is clearly better than silently bypassing; whether it earns its cost on a MoE model at concurrency is a separate measurement, and I would rather flag that than let the acceptance-rate framing imply a speedup nobody has shown yet.

Thump604:
Rebased the five substantive assistant-MTP commits onto current `main` (`7afa61a`), dropping only the old merge commit. The sole conflict was CI-list composition; both current-main coverage and this PR’s batched-config test were retained. Verification: full suite `2584 passed, 25 skipped, 27 deselected`; focused MLLM/CB/MTP/lifecycle `343 passed, 1 skipped, 10 deselected`; Black, compile, and diff checks clean. Scope remains correctness/engagement, not throughput; mixed opt-in batches may still bypass assistant MTP and effective external drafting remains one token. New head: `a6b64b7`.

Issue #672, opened 2026-08-03
[Bug]: Streaming tool calls can end without finish_reason when the terminal engine output is swallowed by a parser


funkymonkeymonk:
## Summary

With `--enable-auto-tool-choice` (reproduced with `--tool-call-parser gemma4` and `--reasoning-parser gemma4` serving `mlx-community/gemma-4-31b-it-4bit`), a streamed chat completion can end with the `tool_calls` chunk (`finish_reason: null`) followed directly by `data: [DONE]` — no chunk ever carries `finish_reason`. Strict OpenAI-compatible clients (e.g. the pi coding agent) abort with `Stream ended without finish_reason` and retry the turn.

## Reproducer

Minimal engine-output sequence (drives `stream_chat_completion` directly; also observed live against the real server — captured SSE below):

```python
from vllm_mlx.engine.base import GenerationOutput

# 1. Model emits the complete canonical tool call, end marker included, in one delta:
yield GenerationOutput(
    text=''<|tool_call>call:get_weather{<|"|>city<|"|>: <|"|>Paris<|"|>}<tool_call|>'',
    new_text=<same>, prompt_tokens=10, completion_tokens=20,
    finish_reason=None, finished=False,
)
# 2. Terminal engine output is a bare end-of-turn token (stop-string hit):
yield GenerationOutput(
    text="<turn|>", new_text="<turn|>", prompt_tokens=10, completion_tokens=21,
    finish_reason="stop", finished=True,
)
```

**Observed stream** (real server, production traffic):

```
data: {...,"delta":{"role":"assistant"},"finish_reason":null}
data: {...,"delta":{"reasoning_content":"..."},"finish_reason":null}   # xN
data: {...,"delta":{"tool_calls":[{"id":"call_01ed9d3d","type":"function","function":{"name":"yak_list","arguments":"{\"filter\":\"not-done\"}"}}]},"finish_reason":null}
data: [DONE]
```

**Expected:** a terminal chunk with `"finish_reason":"tool_calls"` before `[DONE]`.

## Root cause

[...]

Issue #628, opened 2026-07-03
[Bug]: Streaming chat completions never emit finish_reason "stop" on natural stops (length works)


mvmories:
### Environment

- Chip: M4 Max
- RAM: 64 GB
- macOS version: 26.5.1
- Python version:  Python 3.12.13
- vllm-mlx version: 0.4.0 (git @ `0dd115769ef1196a715b96b181353edacd2a4f69`) — also reproduced on 0.4.0rc1
- mlx:                            0.31.2
- mlx-audio:                 0.4.4
- mlx-lm:                      0.31.3
- mlx-metal:                 0.31.2
- mlx-vlm:                    0.6.3
- vllm-mlx:                   0.4.0
- Model: `mlx-community/Qwen3.6-35B-A3B-4bit`, SimpleEngine (`MLLM=False`)

## Bug

When a streamed `/v1/chat/completions` generation ends naturally (EOS), the final chunk carries `"finish_reason":null` followed by `data: [DONE]` — no chunk ever carries `"finish_reason":"stop"`. Strict OpenAI-compatible clients (e.g. the pi coding agent) treat the stream as truncated: `Stream ended without finish_reason`.

When generation is cut by `max_tokens`, `"finish_reason":"length"` **is** emitted correctly — only the natural-stop path is affected.

Reproduced identically **with and without** `--reasoning-parser qwen3 --enable-auto-tool-choice --tool-call-parser qwen3_coder`, so it is not parser-related.

## Reproduction

Serve:

```bash
vllm-mlx serve mlx-community/Qwen3.6-35B-A3B-4bit \
  --port 8000 --served-model-name current --offline \
  --max-kv-size 32768 --timeout 1200
```

Natural stop — **bug**:

```bash
curl -sN http://localhost:8000/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{"model":"current","stream":true,"max_tokens":4096,
       "messages":[{"role":"user","content":"Write a short haiku about servers. Do not think long."}]}' \
  | grep -o '"finish_reason":"[a-z]*"'
```

Output: *(empty — final chunk has `"finish_reason":null`, then `[DONE]`)*

Same request with `"max_tokens":300` (length stop) — **works**: `"finish_reason":"length"`

Server logs show a clean exit, no exception:

```
Chat completion (stream): 1416 tokens in 20.39s (69.4 tok/s)
[disconnect_guard] generator exhausted normally, 1417 chunks
[...]

PR #673, opened 2026-08-03
fix(server): emit terminal finish_reason chunk when parsers swallow the finished output

funkymonkeymonk:
Refs #672

## Problem

A streamed tool call can end with the `tool_calls` chunk (`finish_reason: null`) followed directly by `data: [DONE]` — no chunk ever carries `finish_reason`, and strict OpenAI clients (e.g. the pi coding agent) abort with `Stream ended without finish_reason` and retry the turn.

Root cause: in `stream_chat_completion`, the only code attaching `finish_reason` to a chunk lives in the per-output loop. When the engine's `finished=True` output is consumed by a parser `continue` — e.g. gemma4 emits the complete canonical call (`<|tool_call>call:fn{...}<tool_call|>`) in one delta, then a bare `<turn|>` end-of-turn token as the terminal delta, which the tool parser suppresses with `continue` (nothing new to emit) — the terminal chunk is never emitted. There is no post-loop recovery.

This is a bug class, not gemma4-specific: **any** parser `continue` on a `finished=True` output (tool parser, reasoning parser buffering) loses the stream's `finish_reason`. Non-deterministic in production: bites when the model keeps its thought channel open through the tool call, so the terminal delta travels the reasoning path into the tool parser.

## Fix

Post-loop terminal-chunk guard: track whether any emitted chunk carried `finish_reason`; if the engine's finished output was suppressed, emit the terminal chunk (`finish_reason="tool_calls"` when tool calls were detected, else the engine's reason or `"stop"`).

- Single insertion point, parser-agnostic
- Never alters existing chunks (unlike a buffered-text flush, this cannot change parser-visible content — the property that broke CI on #653/#663)

## Tests

Two regression tests in `TestStreamChatCompletion`, covering both swallow branches:

- `test_stream_terminal_finish_reason_when_tool_parser_suppresses_eot` — plain tool-parser branch, real `Gemma4ToolParser`, exact production delta sequence
- `test_stream_terminal_finish_reason_when_reasoning_path_suppresses_eot` — reasoning-parser branch

Both fail on unpatched main (`assert None == 'tool_calls'`) and pass with the fix. Full `tests/test_server.py` suite: **121 passed**.

## Live verification

Reproduced against a real server (`gemma-4-31b-it-4bit`, production agent payload with 21 tools): unpatched stream ended `tool_calls(finish_reason=null)` → `[DONE]`; patched stream ends `tool_calls` → `finish_reason: "tool_calls"` → `[DONE]`.

## Related

- #628 / #629 — natural-stop engine epilogue (engine side; complementary — even a stamped reason is lost if the chunk is swallowed server-side)
- #653 / #663 — reasoning-parser flush at stream end (different mechanism; closed)

### DIFF
--- a/vllm_mlx/server.py
+++ b/vllm_mlx/server.py
@@ -168,7 +168,9 @@ from .model_registry import (
     ModelLease,
     ModelManager,
     RegistryServeDefaults,
+    build_memory_budget_report,
     load_registry_config,
+    log_memory_budget_report,
 )
 from .metrics import metrics as _metrics
 from .models.mllm import UnsafeRemoteURLError, _validate_url_safety, is_url
@@ -967,6 +969,11 @@ _STREAMING_TOOL_MARKERS = (
     "[TOOL_CALLS]",
     "<minimax:tool_call>",
     '<invoke name="',
+    # gpt-oss harmony: commentary channel addresses tools, and <|call|>
+    # terminates a tool call. Note the raw (unstripped) delta is what the
+    # streaming gates scan, so the harmony tokens remain present.
+    "<|channel|>commentary",
+    "<|call|>",
 )
 _STREAMING_BARE_BRACKET_MARKER = re.compile(r"\[\w+\(\{")
 _STREAMING_BARE_BRACKET_PARTIAL = re.compile(r"\[\w+\($")
@@ -1129,6 +1136,57 @@ def _build_reasoning_parser(engine: BaseEngine | None = None):
         return type(_reasoning_parser)(tokenizer)
     except TypeError:
         return type(_reasoning_parser)()
+
+
+def _prepare_streaming_reasoning_parser(
+    engine: BaseEngine,
+    request: ChatCompletionRequest | ResponsesRequest | None,
+    chat_kwargs: dict[str, object],
+    *,
+    allowed: bool = True,
+):
+    """Build and reset request-local reasoning state when thinking is enabled."""
+    if not allowed or _thinking_disabled(request, chat_kwargs):
+        return None
+    parser = _build_reasoning_parser(engine)
+    if parser is not None:
+        parser.reset_state()
+    return parser
+
+
+def _prepare_openai_stream_reasoning_state(
+    engine: BaseEngine,
+    request: ChatCompletionRequest,
+    chat_kwargs: dict[str, object],
+) -> tuple[object | None, bool]:
+    """Return request-local reasoning state and the legacy Nemotron marker state."""
+    parser = _prepare_streaming_reasoning_parser(engine, request, chat_kwargs)
+    is_thinking_model = (
+        "nemotron" in (engine.model_name or "").lower()
+        and not parser
+        and not _thinking_disabled(request, chat_kwargs)
+    )
+    return parser, is_thinking_model
+
+
+def _request_tool_definitions(request: ChatCompletionRequest) -> list | None:
+    """Return the request tool schema once for streaming argument coercion."""
+    if request and request.tools:
+        return request.model_dump(include={"tools"}).get("tools")
+    return None
+
+
+def _streaming_json_fence_stripper(
+    request: ChatCompletionRequest,
+) -> StreamingJsonFenceStripper | None:
+    """Create a fence stripper only for JSON-constrained streaming responses."""
+    response_format = getattr(request, "response_format", None)
+    response_format_type = getattr(response_format, "type", None)
+    if response_format_type is None and isinstance(response_format, dict):
+        response_format_type = response_format.get("type")
+    if response_format_type in ("json_object", "json_schema"):
+        return StreamingJsonFenceStripper()
+    return None
 
 
 # Lifecycle startup coordination — an Event lets the lifecycle loop block
@@ -1532,7 +1590,7 @@ async def lifespan(app: FastAPI):
 app = FastAPI(
     title="vllm-mlx API",
     description="OpenAI-compatible API for MLX LLM/MLLM inference on Apple Silicon",
-    version="0.4.0",
+    version="0.4.1",
     lifespan=lifespan,
 )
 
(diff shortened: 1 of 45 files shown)


funkymonkeymonk:
@Thump604 thank you for the thorough review and for catching the tracker bug. I have applied all the requested changes:

1. **Tracker now reflects actual non-null serialized finish_reason** — switched from  to  so a  chunk does not falsely claim emission.

2. **Added  regression test** () — verifies the guard still stamps  when the engine leaves the reason null.

3. **Cherry-picked the engine-side fix from #681** —  now stamps  on natural-stop streaming epilogue and  on token-limit hits, so the combined path is green.

4. **Changed  → ** in the commit message until the full combined path is merged.

Please let me know if anything else needs adjustment.

waybarrios:
just updated this branch from main. Ill review it soon, meanwhile ci is running.

vllm_mlx/server.py
```diff
@@ -6436,6 +6453,34 @@ async def stream_chat_completion(
...
+                    ChatCompletionChunkChoice(
+                        delta=ChatCompletionChunkDelta(),
+                        finish_reason=(
+                            "tool_calls"
+                            if tool_calls_detected
+                            else (getattr(last_output, "finish_reason", None) or "stop")
+                        ),
+                    )
+                ],
+                usage=get_usage(last_output),
```
waybarrios, line 6491: I found one edge case here: when the engine finishes with `finish_reason=None`, the normal path already sends usage at [the preceding final chunk](https://github.com/waybarrios/vllm-mlx/blob/ac2500096af807248915747d448972d6a62b99db/vllm_mlx/server.py#L6392-L6411). Since no finish reason was emitted, this fallback then sends another terminal chunk with the same usage.

That can make clients double-count prompt and completion tokens. Could we ensure usage is emitted only once, ideally on the fallback chunk when it supplies `"stop"`? A regression assertion that exactly one payload contains usage would cover this.

vllm_mlx/server.py
```diff
@@ -6436,6 +6453,34 @@ async def stream_chat_completion(
...
                 )
                 yield f"data: {tool_chunk.model_dump_json()}\n\n"
+                finish_reason_emitted = True
+
+        # Terminal-chunk guard: if the engine's finished output was swallowed
+        # by a parser `continue` above (e.g. a bare end-of-turn token arriving
+        # after a completed tool call), no chunk carried finish_reason and
+        # OpenAI clients abort with "stream ended without finish_reason".
+        # Emit the terminal chunk now.
+        if (
```
waybarrios, line 6473: Could we add coverage for the reasoning parser swallowing the final delta directly? The relevant early `continue` is at [vllm_mlx/server.py#L6172-L6174](https://github.com/waybarrios/vllm-mlx/blob/ac2500096af807248915747d448972d6a62b99db/vllm_mlx/server.py#L6172-L6174).

The current tests cover tool-parser suppression, but not the `delta_msg is None` path. A small regression test should verify that the stream still ends with `finish_reason: "stop"` and includes the final usage values.

vllm_mlx/server.py
Thump604: Fixed in Thump604/vllm-mlx@7e454a0. Finished outputs with no terminal reason no longer attach usage to the preceding nonterminal chunk; the fallback `stop` chunk is now the sole usage-bearing payload. The regression asserts exactly one usage payload with the expected prompt/completion totals. Focused terminal matrix: 4 passed; full `tests/test_server.py`: 126 passed, 3 deselected. The author can cherry-pick `7e454a0`.

vllm_mlx/server.py
Thump604: Added in Thump604/vllm-mlx@7e454a0. The new regression drives the direct reasoning-parser `delta_msg is None` branch on the finished output and verifies the fallback emits `finish_reason="stop"` with exactly one final usage payload. Focused terminal matrix: 4 passed; full `tests/test_server.py`: 126 passed, 3 deselected. The author can cherry-pick `7e454a0`.

vllm_mlx/server.py
funkymonkeymonk: Thanks @Thump604 — cherry-picked `7e454a0` onto the PR head (commit `f525d2d`), attribution preserved. Verified:

- **RED**: without the usage guard, the new assertion in `test_stream_terminal_finish_reason_when_engine_emits_none` fails `assert 2 == 1` — exactly the double-usage payload you diagnosed; the engine-emits-none path attaches usage to the preceding nonterminal chunk **and** the fallback `stop` chunk.
- **GREEN**: with the guard, both affected tests pass; full `tests/test_server.py` = **126 passed, 3 deselected**, matching your run. `ruff`/`black` clean, no new `mypy` errors at the changed lines.

vllm_mlx/server.py
funkymonkeymonk: Thanks @Thump604 — same cherry-pick as above (`f525d2d`). The new `test_stream_terminal_when_reasoning_parser_swallows_finished_delta` drives the `reasoning_parser` branch where `extract_reasoning_streaming` returns `None` on the finished delta (`continue`), so no normal-path chunk is emitted and the post-loop guard stamps `finish_reason="stop"` with exactly **one** usage payload (`prompt=5, completion=2, total=7`). Passes with the fix; full suite 126 passed, 3 deselected.

funkymonkeymonk:
Applied both review suggestions by cherry-picking Thump604's verified commit `7e454a0` (`f525d2d`, attribution preserved):

1. **EEC-001 (waybarrios)** — duplicate terminal stream usage: the normal-path `usage` guard now only attaches usage when the chunk actually carries a `finish_reason` (`tool_calls_detected or output.finish_reason`), so a swallowed finished output defers usage to the sole fallback `stop` chunk. Reduces double-counting of prompt/completion tokens.
2. **TC-001 (waybarrios)** — added `test_stream_terminal_when_reasoning_parser_swallows_finished_delta` covering the reasoning-parser `delta_msg is None` `continue` branch on the finished delta.

**Verification** (local, arm64 py3.12 via mise, `pip install -e ".[dev]"`):
- RED proven: pre-fix, `test_stream_terminal_finish_reason_when_engine_emits_none` fails `assert 2 == 1` (double usage).
- GREEN: 4 affected tests pass; full `tests/test_server.py` = **126 passed, 3 deselected** (matches @Thump604's run).
- `ruff`/`black` clean; no new `mypy` errors at the changed lines.

waybarrios:
All set. ready to go

PR #676, opened 2026-08-03
feat: add DeepSeek-V4-Flash support

janhilgard:
DeepSeek-V4 ships no Jinja `chat_template` — its `tokenizer_config.json` carries only BOS/EOS/pad — so the prompt has to be built programmatically. Today the model would either raise in `_apply_chat_template` or fall back to naive `"role: content"` concatenation, and its DSML tool calls would go unparsed. This adds the three pieces needed to serve it.

## Prompt encoder

`vllm_mlx/utils/deepseek_v4_encoding.py` ports the reference `encoding_dsv4.py` published with the weights:

```
<｜begin▁of▁sentence｜>{system}<｜User｜>{question}<｜Assistant｜><think>
```

System content is bare text with no wrapper; roles are delimited solely by `<｜User｜>`/`<｜Assistant｜>`. The turn closes on `<think>` in thinking mode or `</think>` in chat mode, which suppresses reasoning. Tool schemas render into the system message, and tool results fold into the preceding user turn as `<tool_result>` blocks — V4 has no `tool` role. `reasoning_effort` is a text prefix on the whole conversation rather than a token or a sampling parameter, and `drop_thinking` is forced off when tools are present, because the model needs to see why it made the earlier calls.

`vllm_mlx/utils/tokenizer.py` installs the encoder by overriding `apply_chat_template` on the tokenizer when `model_type` is `deepseek_v4`. That fixes all three call sites at once — `engine/batched.py`, `engine/simple.py` and `models/llm.py` all reach the template through that one method — without touching any of them. It mirrors what upstream vLLM does in `vllm/tokenizers/deepseek_v4.py`.

## DSML tool parser

V4 emits its own markup rather than JSON:

```
<｜DSML｜tool_calls>
<｜DSML｜invoke name="get_weather">
<｜DSML｜parameter name="city" string="true">Prague</｜DSML｜parameter>
<｜DSML｜parameter name="days" string="false">3</｜DSML｜parameter>
</｜DSML｜invoke>
</｜DSML｜tool_calls>
```

The `string` attribute carries the type — `"true"` is a raw string, `"false"` is JSON. That is why this is a scanner and not a regex over `name=value` pairs: a string parameter may legitimately contain quotes, angle brackets or a JSON-looking payload. Registered as `deepseek_v4`/`dsml` and wired into `AutoToolParser`; the existing `DeepSeekToolParser` handles the V3/R1 `<｜tool▁calls▁begin｜>` format and shares nothing with this one.

## Reasoning parser

Extends the R1 parser, which already tolerates the missing opening `<think>` that the encoder's prompt implies. What V4 adds is that a tool call must follow completed reasoning, so an opening `<｜DSML｜tool_calls>` terminates the reasoning block even when `</think>` never arrives. Without that, the entire DSML payload is swallowed as reasoning and the caller sees no tool call at all.

## Streaming

Both parsers track how much of the accumulated text they have emitted and withhold a tail that could still grow into a marker. `<｜DSML｜tool_calls>` is assembled from several tokens — only the bare `｜DSML｜` has an id of its own — so it always straddles delta boundaries. Two failure modes follow from that, and both are covered by tests: detecting completion against the delta rather than the accumulated text loses the calls entirely, and emitting marker fragments as they arrive leaks markup into the user-visible stream and then repeats the whole marker once it is recognised.

## Scope

Model loading is deliberately not part of this. vllm-mlx defines no model architectures — `models/` only wraps `mlx_lm.load()` — and `deepseek_v4` lands in mlx-lm via [ml-explore/mlx-lm#1189](https://github.com/ml-explore/mlx-lm/pull/1189), which is open and active. Everything here is text processing and starts working the moment that merges. There is precedent for staging it this way: `text_model_from_vlm.py` already reports `"Cannot import mlx_lm TextModel (need PR #990)"`.

Continuous batching will likely want its own patch alongside `patches/gemma4_mllm.py` and `patches/glm4v_moe_mllm.py`, since V4 has MLA with a compressor, a sparse indexer and mHC. That is best written once the model actually runs.

## Testing

`tests/test_deepseek_v4_encoding.py` asserts against prompts frozen from the reference implementation, which cannot be vendored here. During development the port was also run differentially against that reference across 832 combinations of conversation shape, thinking mode, effort level and tool usage — all identical.

Verified against the real model, DeepSeek-V4-Flash-0731 MXFP4 with 167 GB resident on an M3 Ultra, across 23 scenarios: chat, thinking, every `reasoning_effort` level, single and parallel tool calls, tool-result round trips and multi-turn history. Prompts were built by this encoder and the completions run back through these parsers — **127/128 checks pass**. The one failure is a scenario truncated at the token cap before it emitted `</think>`, where treating the output as content is the correct behaviour. The model's DSML matched the specification including the `string="true|false"` type flag, and parallel calls came back intact:

```
get_weather({"city": "Rome", "days": 1})
get_weather({"city": "Paris", "days": 1})
get_weather({"city": "Madrid", "days": 1})
```

The three new test files are added to the CI whitelist in `.github/workflows/ci.yml`, without which they would never run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,6 +94,9 @@ jobs:
             tests/test_gemma4_openai_format.py \
             tests/test_gemma4_streaming_edge.py \
             tests/test_gemma4_tool_parser.py \
+            tests/test_deepseek_v4_encoding.py \
+            tests/test_deepseek_v4_reasoning.py \
+            tests/test_deepseek_v4_tool_parser.py \
             tests/test_minimax_tool_calling.py \
             tests/test_qwen3_xml_parser.py \
             tests/test_qwen3_xml_registration.py \
--- a/README.es.md
+++ b/README.es.md
@@ -46,7 +46,7 @@ claude
 ### APIs
 - **Compatible con OpenAI**: `/v1/chat/completions`, `/v1/completions`, `/v1/embeddings`, `/v1/rerank`, `/v1/responses`
 - **Compatible con Anthropic**: `/v1/messages` (streaming, tool use, system prompts)
-- **MCP Tool Calling**: 12 parsers (OpenAI, Anthropic, Gemini, Qwen, DeepSeek, Gemma y más)
+- **MCP Tool Calling**: 19 parsers (OpenAI, Anthropic, Gemini, Qwen, DeepSeek, Gemma y más)
 - **Salida estructurada**: JSON Schema vía `response_format` (lm-format-enforcer)
 
 ### Throughput y memoria
@@ -64,7 +64,7 @@ claude
 - **STT**: familia Whisper con RTF hasta 197x en M4 Max
 
 ### Razonamiento y avanzado
-- **Extracción de razonamiento**: Qwen3, DeepSeek-R1 (`--reasoning-parser`)
+- **Extracción de razonamiento**: Qwen3, DeepSeek-R1, DeepSeek-V4 (`--reasoning-parser`)
 - **Reducción de expertos MoE**: `--moe-top-k` para +7-16% en Qwen3-30B-A3B
 - **Decodificación especulativa**: `--mtp` para Qwen3-Next
 - **Prefill disperso**: `--spec-prefill` basado en atención para reducir TTFT
--- a/README.fr.md
+++ b/README.fr.md
@@ -46,7 +46,7 @@ claude
 ### APIs
 - **Compatible OpenAI** : `/v1/chat/completions`, `/v1/completions`, `/v1/embeddings`, `/v1/rerank`, `/v1/responses`
 - **Compatible Anthropic** : `/v1/messages` (streaming, tool use, system prompts)
-- **MCP Tool Calling** : 12 parsers (OpenAI, Anthropic, Gemini, Qwen, DeepSeek, Gemma et plus)
+- **MCP Tool Calling** : 19 parsers (OpenAI, Anthropic, Gemini, Qwen, DeepSeek, Gemma et plus)
 - **Sortie structurée** : JSON Schema via `response_format` (lm-format-enforcer)
 
 ### Débit et mémoire
@@ -64,7 +64,7 @@ claude
 - **STT** : famille Whisper avec RTF jusqu'à 197x sur M4 Max
 
 ### Raisonnement et avancé
-- **Extraction du raisonnement** : Qwen3, DeepSeek-R1 (`--reasoning-parser`)
+- **Extraction du raisonnement** : Qwen3, DeepSeek-R1, DeepSeek-V4 (`--reasoning-parser`)
 - **Réduction d'experts MoE** : `--moe-top-k` pour +7-16% sur Qwen3-30B-A3B
 - **Décodage spéculatif** : `--mtp` pour Qwen3-Next
 - **Prefill creux** : `--spec-prefill` basé sur l'attention pour réduire le TTFT
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ claude
 ### APIs
 - **OpenAI-compatible**: `/v1/chat/completions`, `/v1/completions`, `/v1/embeddings`, `/v1/rerank`, `/v1/responses`
 - **Anthropic-compatible**: `/v1/messages` (streaming, tool use, system prompts)
-- **MCP Tool Calling**: 12 parsers (OpenAI, Anthropic, Gemini, Qwen, DeepSeek, Gemma, and more)
+- **MCP Tool Calling**: 19 parsers (OpenAI, Anthropic, Gemini, Qwen, DeepSeek, Gemma, and more)
 - **Structured output**: JSON Schema via `response_format` (lm-format-enforcer)
 
 ### Throughput & memory
@@ -64,7 +64,7 @@ claude
 - **STT**: Whisper family with RTF up to 197x on M4 Max
 
 ### Reasoning & advanced
-- **Reasoning extraction**: Qwen3, DeepSeek-R1 (`--reasoning-parser`)
+- **Reasoning extraction**: Qwen3, DeepSeek-R1, DeepSeek-V4 (`--reasoning-parser`)
 - **MoE expert reduction**: `--moe-top-k` for +7-16% on Qwen3-30B-A3B
 - **Speculative decoding**: `--mtp` for Qwen3-Next
 - **Sparse prefill**: attention-based `--spec-prefill` for TTFT reduction
--- a/README.zh.md
+++ b/README.zh.md
@@ -46,7 +46,7 @@ claude
 ### API
 - **兼容 OpenAI**：`/v1/chat/completions`、`/v1/completions`、`/v1/embeddings`、`/v1/rerank`、`/v1/responses`
 - **兼容 Anthropic**：`/v1/messages`（流式、工具调用、system prompts）
-- **MCP 工具调用**：12 种解析器（OpenAI、Anthropic、Gemini、Qwen、DeepSeek、Gemma 等）
+- **MCP 工具调用**：19 种解析器（OpenAI、Anthropic、Gemini、Qwen、DeepSeek、Gemma 等）
 - **结构化输出**：通过 `response_format` 的 JSON Schema（基于 lm-format-enforcer）
 
 ### 吞吐与内存
@@ -64,7 +64,7 @@ claude
 - **STT**：Whisper 系列，M4 Max 上 RTF 最高可达 197 倍
 
 ### 推理与高级功能
-- **思维链提取**：Qwen3、DeepSeek-R1（`--reasoning-parser`）
+- **思维链提取**：Qwen3、DeepSeek-R1、DeepSeek-V4（`--reasoning-parser`）
 - **MoE 专家裁剪**：`--moe-top-k`，Qwen3-30B-A3B 上 +7-16%
 - **投机解码**：`--mtp`，用于 Qwen3-Next
 - **稀疏 prefill**：基于注意力的 `--spec-prefill`，降低 TTFT
--- a/benchmarks/bench_deepseek_v4.py
+++ b/benchmarks/bench_deepseek_v4.py
@@ -1,0 +1,214 @@
+"""Benchmark: DeepSeek-V4 prompt encoding and streaming parser overhead.
+
+Covers both serving paths. Single-stream is one sequence decoded token by
+token, where the parser sits directly in the latency path. Batch is N
+concurrent sequences interleaved, the shape continuous batching produces: each
+request carries its own parser state, so per-token cost is paid N times per
+decode step and any per-token work that scales with output length compounds.
+
+The thing to watch for is quadratic scaling. A parser that rescans the whole
+accumulated text on every delta is O(N²) over a generation, which is invisible
+on short replies and dominant on long ones — so the ms/tok column is the one
+that matters, not the totals. It is currently flat for the DSML tool parser and
+grows for the reasoning path; see the notes the script prints at the end.
+
+Usage:
+    python benchmarks/bench_deepseek_v4.py
+"""
+
+import time
+
+from vllm_mlx.reasoning.deepseek_v4_parser import DeepSeekV4ReasoningParser
+from vllm_mlx.tool_parsers.deepseek_v4_tool_parser import DeepSeekV4ToolParser
+from vllm_mlx.utils.deepseek_v4_encoding import apply_chat_template
+
+D = "｜DSML｜"
+
+TOOLS = [
+    {
+        "type": "function",
+        "function": {
+            "name": "get_weather",
+            "description": "Get the weather for a city.",
+            "parameters": {
+                "type": "object",
+                "properties": {
+                    "city": {"type": "string"},
+                    "days": {"type": "integer"},
+                },
+                "required": ["city"],
+            },
+        },
+    }
+]
+
+
+def reasoning_tokens(n: int) -> list[str]:
+    """A plain thinking turn: N reasoning tokens, then a short answer."""
+    return (
+        [f"step{i} " for i in range(n)]
+        + ["</think>"]
+        + [f"word{i} " for i in range(10)]
+    )
+
+
+def tool_call_tokens(n: int) -> list[str]:
+    """A tool-calling turn: N reasoning tokens, then DSML markup.
+
+    The markup is split the way the tokenizer splits it — ｜DSML｜ has an id of
+    its own, the surrounding punctuation does not — so the parser sees the same
+    fragmentation it sees in production.
+    """
+    tokens = [f"step{i} " for i in range(n)] + ["</think>", "\n\n"]
+    tokens += ["<", D, "tool_calls", ">", "\n"]
+    tokens += ["<", D, "invoke", ' name="get_weather"', ">", "\n"]
+    tokens += [
+        "<",
+        D,
+        "parameter",
+        ' name="city"',
+        ' string="true"',
+        ">",
+        "Prague",
+        "</",
+        D,
+        "parameter",
+        ">",
+        "\n",
+    ]
+    tokens += ["</", D, "invoke", ">", "\n"]
+    tokens += ["</", D, "tool_calls", ">"]
+    return tokens
+
+
+def bench_stream(make_tokens, n_tokens: int, streams: int) -> tuple[float, int]:
+    """Interleave `streams` concurrent sequences, one delta each per step.
+
+    Returns (total ms, total deltas). With streams=1 this is the single-stream
+    latency path; above that it is the batched decode step.
+    """
+    token_lists = [make_tokens(n_tokens) for _ in range(streams)]
+    parsers = []
+    for _ in range(streams):
+        reasoner = DeepSeekV4ReasoningParser()
+        reasoner.reset_state()
+        tools = DeepSeekV4ToolParser()
+        tools.reset()
+        parsers.append((reasoner, tools, {"acc": "", "tool_acc": ""}))
+
+    steps = max(len(t) for t in token_lists)
+    deltas = 0
+    start = time.perf_counter()
+    for step in range(steps):
+        for stream in range(streams):
+            tokens = token_lists[stream]
+            if step >= len(tokens):
+                continue
+            reasoner, tools, state = parsers[stream]
+            delta = tokens[step]
+            previous, state["acc"] = state["acc"], state["acc"] + delta
+            deltas += 1
+
+            message = reasoner.extract_reasoning_streaming(
+                previous, state["acc"], delta
+            )
+            if message is None or not message.content:
+                continue
+            prev_tool = state["tool_acc"]
+            state["tool_acc"] = prev_tool + message.content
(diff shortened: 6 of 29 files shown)


janhilgard:
Pushed an update after auditing how the pieces behave once they are wired into the server rather than exercised on their own. Two integration bugs came out of it, plus a benchmark.

**`SUPPORTS_NATIVE_TOOL_FORMAT` was wrong.** I had it `False`. That makes `extract_multimodal_content` flatten `role="tool"` into `"[Tool Result (id)]: ..."` and assistant `tool_calls` into `"[Calling tool: name(...)]"` *before* the encoder runs, so a multi-turn tool conversation reached the model as text it was never trained on and the encoder's own `<tool_result>`/DSML handling never fired. The encoder consumes both natively, so the flag is now `True`.

**Tool call arguments arrive decoded.** With native format preserved, `api/utils.py` json-loads `arguments` in place. The encoder was loading it again, and `json.loads` on a mapping raises, so every parameter collapsed into a single bogus `arguments` entry — the model saw a malformed call in its own history. It now accepts either form.

Both are covered by regression tests. I also added a test that chains the reasoning parser into the tool parser the way the server chains them, since testing them separately misses exactly the handoff where the markup crosses over.

**Benchmark** (`benchmarks/bench_deepseek_v4.py`), covering both serving paths:

```
Prompt encoding      4 messages ->   0.022 ms      130 messages ->   0.335 ms

Single stream, tool-calling turn
   100 tokens -> 0.0061 ms/tok      5000 tokens -> 0.0380 ms/tok

DSML tool parser alone
   100 tokens -> 0.0018 ms/tok      5000 tokens -> 0.0028 ms/tok

Batched decode, 1000 tokens each
     1 concurrent -> 0.0107 ms/tok    16 concurrent -> 0.0107 ms/tok
```

Batching costs nothing per token — each request holds independent parser state and totals scale linearly to 16 streams. The DSML parser's own cost is flat.

The combined single-stream figure does grow with output length, and it is not this PR's code: `BaseThinkingReasoningParser` searches the accumulated text for its start and end tags on every delta while the reasoning block is open, which is O(N²) over a generation. Isolating the tool parser shows the difference. It is 0.2% of the decode budget at 50 tok/s so nothing is on fire, but it affects `qwen3` and `deepseek_r1` equally and belongs in the base class rather than being worked around per model — happy to open a separate PR for it if that is useful.

Rebased onto current `main`.

janhilgard:
@Thump604 — review requested.

Two notes for reviewing this one. The DSML tool parser is a scanner rather than a regex because the `string="true|false"` attribute means a string parameter may legitimately contain quotes, angle brackets or a JSON-looking payload; `TestParameterTyping::test_string_value_may_contain_markup_like_text` pins that.

The streaming logic is the part most worth a careful look. `<｜DSML｜tool_calls>` has no token id of its own — only the bare `｜DSML｜` does — so the marker always straddles delta boundaries, and two obvious implementations are both wrong: detecting completion against the delta rather than the accumulated text drops the calls entirely, and emitting marker fragments as they arrive leaks markup to the client and then repeats the whole marker. Both are covered by tests at chunk sizes 1 through 128.

Related: #677 fixes two bugs in `BaseThinkingReasoningParser` that this parser inherits from. They do not block this PR, but see the note there about merge order.

janhilgard:
Update after validating this against a live model end to end.

**Which mlx-lm PR to pair this with: [#1192](https://github.com/ml-explore/mlx-lm/pull/1192).** I tried three that add `deepseek_v4` and only that one is usable:

| | DSML tool calls | batched cache |
|---|---|---|
| [#1189](https://github.com/ml-explore/mlx-lm/pull/1189) | ✗ | ✗ |
| [#1195](https://github.com/ml-explore/mlx-lm/pull/1195) | ✗ | ✓ |
| **#1192** | **✓** | **✓** |

#1189 and #1195 emit `<｜DSML｜tool_c天气>` where the marker should be `<｜DSML｜tool_calls>`, so tool calls fail while ordinary prose still reads fine — the sort of thing that looks healthy in a smoke test. It is the model implementation, not this PR's parsing or the batching: input token ids are identical across paths, both tokenizers decode the marker correctly on its own, and layer-by-layer hidden-state norms drift from the reference implementation by 1.7% at layer 0 and past 20% from layer 5, once the MoE router starts choosing different experts. #1192 is closed but mergeable, and oMLX ships it, which is how I found it.

**Verified against #1192** on DeepSeek-V4-Flash-0731 MXFP4 (283.8 B params, M3 Ultra), through this branch's encoder and parsers:

```
facts          Paris / 51 / Pacific / No                    4/4
DSML           single, parallel, two tools, integer arg     4/4
determinism    same call three times at temperature 0       1/1

single stream   31 tok/s
aggregate @8    90 tok/s
```

Over HTTP with `--tool-call-parser deepseek_v4 --reasoning-parser deepseek_v4`, a full battery passes 11/11: `finish_reason=tool_calls`, arguments as JSON objects, reasoning split into `reasoning_content`, no DSML in user-visible content, parallel calls intact, and a tool-result round trip where the model uses the returned values.

Nothing in the diff changed — it is the same encoder and parsers as before, now with the model side pinned down.

janhilgard:
Update: getting this model actually serving turned up three engine bugs, now fixed in this branch. Flagging them here since they are not really DeepSeek-specific and reviewers may want them split out.

## 1. Model load and generation ran on different threads

MLX streams exist only in the thread that created them, and an array with pending primitives carries the stream those primitives were built on. `SimpleEngine` spread that work over three threads (load and streaming on the event loop, `_run_blocking_serialized` on `asyncio.to_thread`), and `BatchedEngine` created its own `engine-core` thread but loaded the model on the event loop.

Symptoms differed but the cause was one:

- SimpleEngine returned HTTP 500 on every request with `There is no Stream(gpu, N) in current thread`.
- BatchedEngine looked like a scheduler hang — `running=1`, the step counter climbing into the millions, not one token out. It was not the scheduler: `batch_generator.next()` raised the same error, `engine_core` fell back to model-thread stepping, hit it again there, and since that fallback fires only once the loop then spun on the error.

The scattered `bind_generation_streams()` calls cannot fix this — rebinding a module-level handle changes a global the already-built buffers never consult. I measured the alternatives on a cache built on one thread and evaluated on another: per-worker rebinding failed 3/3, keeping mlx-lm's import-time `ThreadLocalStream` failed 3/3, pinning passed 0/3. The SimpleEngine half is also up standalone as #679.

Side effect worth knowing: generation no longer blocks the event loop, so genuinely concurrent requests now reach admission control where the default `fail_fast` rejects them. That was previously masked. `VLLM_MLX_SIMPLE_ENGINE_LOCK_ADMISSION=wait` restores queueing.

## 2. The prefix cache could never hit on this architecture

The scheduler stored one entry keyed by `prompt + output`. Every later query is *shorter* than such a key, so reuse requires trimming the generated tail — impossible here. `RotatingKVCache.is_trimmable()` is `offset < max_size` and `sliding_window` is 128, so past 128 tokens the ring buffer has physically overwritten the older KV; `PoolingCache` merges tokens into windows (`compress_ratios` alternate 4 and 128) and cannot split one. Zero hits were guaranteed by construction, and the code even relied on the supersequence path that `memory_cache.py` disables for non-trimmable caches.

Fixed by also snapshotting the cache while it still covers exactly the prompt — before the first generated token is appended — and storing that under the prompt tokens. Such an entry is reusable by strict-prefix match with no trimming at all. Two traps: mlx-lm attaches `prompt_cache` only to the response carrying a `finish_reason`, so the per-sequence cache is pulled from the live batch via `extract_cache(idx)`; and the snapshot must be a real copy, because both cache types write into their buffers in place.

Exact matches are deliberately skipped. The scheduler re-feeds `prompt[-1:]` on those, which duplicates that token in the KV cache when the entry covers the whole key — measured as the same prompt returning a different answer.

## 3. `--chunked-prefill-tokens` was silently dropped

It was only applied by monkey-patching `BatchGenerator._process_prompts` / `active_batch`, neither of which exists on current mlx-lm, so the scheduler logged "Chunked prefill disabled" and prefilled in one go. The warning was also wrong about the consequence: mlx-lm chunks the prompt natively — `PromptProcessingBatch` consumes at most `prefill_step_size` per step — so the budget just needed mapping onto that. Only the patch's extra (mid-prefill cache saves) is genuinely unavailable.

## Verification

DeepSeek-V4-Flash-0731, 283.8B MXFP4, M3 Ultra:

| | |
|---|---|
| cache correctness, 5 scenarios × 2 turns | **10/10 character-identical** cold vs warm |
| non-stream suite | 30/30, 0 stream errors |
| streaming suite | 18/18 |
| tool calls | 6/6 non-stream, 3/3 streaming |
| concurrency | 4/4, **54.3 tok/s** aggregate vs ~31 single-stream |
| long shared prefix | 4.19s → **1.09s** |
| chunked prefill, 8413-token prefill + short request 1s later | 1024/chunk: short 27.3s · 256/chunk: short **12.1s** |
| repo suite | 2412 passed |

janhilgard:
Follow-up: the post-prefill snapshot I added above had a memory bug that only showed under a real agentic client, now fixed.

It was stored with `evict_prefixes=False`, but in an agentic loop each turn's prompt *extends* the previous one, so every turn added another full-length KV copy rather than replacing the entry it supersedes — 45 entries of a ~46k-token cache. Each copy is hundreds of arrays (43 layers, `CacheList` of three caches each), and Metal runs out of **buffers** long before the cache's byte budget is reached, so the memory accounting never noticed:

```
RuntimeError: [metal::malloc] Resource limit (499000) exceeded.
[generation_error_recovery] aborted 1 running requests, batch generator closed
```

The failure lands inside the model during a generation step, the request is aborted, and the client just sees the connection stall.

Two changes:

- store the snapshot with `evict_prefixes=True`, so a conversation keeps one entry instead of one per turn
- skip the completion-time `prompt + output` entry when `can_trim_prompt_cache()` is false. Such an entry can never be reused — every later query is shorter than its key, so reuse needs a trim — making it pure memory pressure on exactly the architectures that can least afford it

Verified with an 8-turn tool-calling conversation over a ~33k-token context:

| | before | after |
|---|---|---|
| cache entries | 45 and climbing | **1** |
| Metal errors | 7 | **0** |
| cache hits | — | **7/8**, 84 tokens prefilled instead of 33k |
| tool calls parsed | — | **8/8** |
| steady-state turn latency | — | **3.7–5.2s** |

Worth noting for reviewers: the byte-based limit in `memory_cache.py` is not sufficient on its own, because buffer count is the binding resource for caches split across many small per-layer arrays.

Thump604 (changes requested):
The DeepSeek encoder/parser work is valuable, but the current branch is no longer reviewable or safe to merge as one feature: 26 files / +3,089 lines now combine DeepSeek prompt+DSML support with SimpleEngine and BatchedEngine thread ownership, prefix-cache snapshot/accounting policy, and chunked-prefill behavior. The embedded SimpleEngine lifecycle is also the earlier implementation superseded by #679's stop/drain corrections, and #648 already carries the narrow chunked-prefill API update.

Please split this into independently reviewable changes:

1. DeepSeek V4 encoder, DSML tool/reasoning parsers, registration/CLI, goldens, and their focused tests.
2. SimpleEngine ownership via #679; remove the duplicate implementation here.
3. A separate BatchedEngine owner-thread/error-handling PR with its own reproduction and lifecycle tests.
4. Separate prefix-cache changes, with explicit coverage for non-trimmable/rotating caches and memory ownership.
5. Reuse/rebase onto #648 for chunked prefill rather than carrying another copy.

Then rebase the DeepSeek slice after #677 and rerun the combined streaming/tool replay. Finally, update the dependency story: the PR currently identifies mlx-lm #1192 as the usable implementation, but #1192 is closed and unmerged, so the feature needs an installable upstream version/active dependency plus an end-to-end reproduction before merge. CI being green on this combined branch does not close those scope and dependency blockers.

janhilgard:
Rebased onto current `main` — the branch is mergeable again.

One conflict, in `server.py`: `#562` added the gpt-oss harmony markers to `_STREAMING_TOOL_MARKERS` while this branch added the DSML one. Independent additions to the same tuple, so I kept both. Repo suite at the new head: **2443 passed**, and I checked the gate directly rather than trusting the merge — `<|channel|>commentary`, `<|call|>`, `｜DSML｜`, plus a partial `｜DSML｜tool_c` delta all still trip it, 6/6.

That is only the mechanical blocker. On the substance of your review:

**The split is fair and I am not going to argue it.** The branch grew engine changes because that was the only way to get the model serving at all, but you are right that "DeepSeek encoder + DSML parsers" and "engine thread ownership + prefix-cache policy + chunked prefill" are four different review problems wearing one hat, and the SimpleEngine copy here is genuinely superseded by #679's stop/drain work. I will restructure along your lines: reuse #679 for SimpleEngine, rebase chunked prefill onto your #648 rather than carrying a second copy, and take BatchedEngine's owner thread and the prefix-cache snapshot out as their own PRs with their own reproductions. The DeepSeek slice then rebases after #677 and gets the combined streaming/tool replay re-run.

**The dependency blocker is real and I do not have an answer for it yet.** I re-checked upstream today: no mlx-lm PR adding `deepseek_v4` has merged. #1189, #1201, #1067 and #1337 are open, #1192, #1193 and #1190 are closed, and released mlx-lm (0.31.3) has no `models/deepseek_v4.py`. So the honest position is that the DeepSeek slice cannot merge on an installable dependency today, whatever its own quality. I would rather it sit behind that than ship something whose only working model implementation is a closed PR people have to apply by hand.

Given that, my preference is to land the parts that stand on their own — they are useful independently of whether DeepSeek ever arrives — and hold the DeepSeek slice until upstream settles. If you would rather I keep it open as a tracking branch or close it until then, say which and I will do that; I will not close it on my own.

janhilgard:
Split done, along the lines you set out. This PR is now the DeepSeek slice alone: **14 files, +2333** instead of 26 files / +3089.

| your item | where it went |
|---|---|
| 1. DeepSeek encoder, DSML parsers, registration/CLI, goldens, tests | **this PR**, nothing else left in it |
| 2. SimpleEngine ownership via #679 | removed here — #679 is the only copy |
| 3. BatchedEngine owner thread, own repro + lifecycle tests | **#684** |
| 4. Prefix cache, explicit non-trimmable/rotating + memory-ownership coverage | **#683** |
| 5. Reuse #648 for chunked prefill | dropped here entirely |

Notes on three of them.

**#684 is stacked on #679**, because it needs `run_blocking_startup_work(executor=...)` from it. Its own content is 64 lines across `batched.py` and `engine_core.py`; the diff cleans up once #679 merges. Say the word if you would rather I carry the twelve-line `base.py` change in #684 instead and keep the two independent — I went with stacking specifically to avoid the duplicate implementation you objected to here.

**On #648: I dropped my chunked-prefill change rather than rebasing onto yours, because yours is strictly better.** Mine mapped the budget onto `prefill_step_size` at construction and left the legacy path behind; `_configure_chunked_prefill` dispatches between the legacy and native APIs and sets `batch_gen.prefill_step_size = budget` on the native one. Same effect on current mlx-lm, and it keeps working on older layouts. Nothing of mine was worth carrying forward.

**#683 and #684 both got the coverage you asked for**, and writing #684's tests caught me repeating the exact mistake waybarrios flagged on #679: my first version replicated the engine's load call in the test body, so reverting `start()` to the unpinned pool still passed. Both PRs now drive the real entry points, and I mutation-tested each — reverting the behaviour while keeping the API fails a specific test in every case.

**Merge order** for this PR: it depends on nothing, but #677 changes the reasoning base class `DeepSeekV4ReasoningParser` inherits from, so I checked it *merged with* #677 rather than only alongside it — 307 targeted tests and 2479 repo-wide, clean. On this branch alone: 2439 passed.

**The dependency blocker stands unchanged**, and I am not going to pretend otherwise: no mlx-lm PR adding `deepseek_v4` has merged, and released mlx-lm 0.31.3 has no `models/deepseek_v4.py`. So this slice still cannot merge on an installable dependency, however clean it is now. #683 and #684 do not have that problem — they are about the engine and are useful regardless of whether DeepSeek ever arrives — which is another reason splitting was the right call.

Thump604 (changes requested):
The split is good. This is reviewable now, and the engine, cache, and chunked-prefill scope is gone. Two blockers remain. The only model implementation validated end to end is closed and unmerged mlx-lm #1192, while released mlx-lm and current main still have no DeepSeek V4 model. Please update the PR body and rerun end to end once there is an installable dependency. Also, buffered streaming drops text before a tool call when the text and complete DSML block arrive in one delta; the parser returns only tool_calls. Please cover that case and preserve both. Keeping changes requested until those are fixed.

Thump604:
Keep it open as the tracking branch. The split was the right move, and the standalone engine/cache work can move independently. This one should stay held until there is an installable DeepSeek V4 dependency and the buffered text-before-tool-call case is fixed.

janhilgard:
Understood on keeping it open as the tracking branch — that matches how I read the dependency situation too.

Fixed the second blocker. Reproduced it first:

```
text + DSML in ONE delta  ->  text=''             calls=1     <- text lost
same content, 4-char chunks -> text='Checking.\n\n' calls=1
non-streaming reference     -> content='Checking.'
```

So the response depended on nothing but how the model's output happened to be chunked, which is the worst shape for a bug like this to have — token-level streaming hides it, and it only surfaces when a whole response lands at once.

The cause was a deliberate choice of mine, and the comment in the code even argued for it: *"only a whole response arriving as a single delta would force the choice, and there the calls win."* That was wrong. The text is user-visible assistant content and dropping it silently is data loss, not a tie-break.

It also could not be fixed in the parser alone. The premise I had built on was real — `server.py` did `suppress = result is None or "tool_calls" in result`, so a delta carrying both had its content discarded — and the same assumption appeared in three places, spelled three different ways, behind all 19 tool parsers. **That half is #690**, against `main`, since nothing about it is DeepSeek-specific: any parser that buffers across a block hits it whenever the block arrives whole.

Here, the parser now flushes the buffered head with the calls when the block opens and closes in the same delta, and ahead of them otherwise. New `TestTextAndCallsInOneDelta` asserts the single-delta case keeps both, that the result is identical at chunk sizes 1, 3, 4, 17, 64 and 4096, and that the head is never emitted twice. Mutation-checked — removing the head from the formatted result fails seven of them.

Repo suite on this branch: 2447 passed.

That leaves only the dependency blocker, which is unchanged: still no merged mlx-lm PR adding `deepseek_v4`, and released mlx-lm has no `models/deepseek_v4.py`. I will rerun the end-to-end battery and update the PR body when there is something installable to run it against.

waybarrios:
Pushed follow-up patch `f19475d` addressing the review feedback:

- V4 and auto parsers now receive every delta through a request-local lifecycle.
- Responses routes reasoning content through tool parsing before emitting text.
- EOS finalization preserves partial markers, truncated DSML, and terminal metadata.
- Reasoning-effort mapping now distinguishes the preview and official 0731 profiles.
- Added the requested Chat, Responses, one-character streaming, truncation, and effort regressions.
- Added the V4 test modules to Linux CI and documented the parser contract.

Local verification: 502 compatibility tests and 172 dedicated V4 tests passed. CI is running.

waybarrios:
Pushed `295fda3` to fix the Apple CI failure. Empty final deltas now safely flush truncated DSML in Chat and Anthropic streams, with regressions for both Anthropic paths. CI is rerunning.

PR #677, opened 2026-08-03
fix(reasoning): stop leaking split think tags and rescanning output

janhilgard:
Two problems in `BaseThinkingReasoningParser`'s streaming path. Both are invisible for models whose tokenizer has the think tags as single tokens — which is why they have gone unnoticed — and both are unavoidable for models that spell the tags out in ordinary tokens.

## Split tags leaked into the reasoning stream

A tag arriving across delta boundaries was detected against the accumulated text, but by then the fragments had already gone out: `delta_text.find(end_token)` returns -1 for a partial tag, so the code fell through to treating the whole delta as reasoning. Streaming `weighing it</think>Answer.` five characters at a time produced:

```
reasoning: "weighing it</thi"
content:   "nk>Answer."
```

## Per-token cost grew with output length

While the reasoning block was open, every delta ran `start_token in current_text` and `end_token in current_text` across the whole accumulated output. That is O(N²) over a generation — despite the docstring stating the state machine avoided exactly that. Measured at 50x the tokens:

| parser | before |
|---|---|
| `deepseek_r1` | 1023x |
| `qwen3` | ~1170x |

Perfectly linear would be 50x.

## One change fixes both

A delta whose tail could still grow into a tag is withheld rather than emitted. That makes a completing tag always present in the text at hand, so there is nothing left to search the accumulated output for — the leak and the rescan disappear together.

Withholding is applied to what is about to be emitted rather than to the raw delta. Text that follows a tag completing in the same delta is then covered too; otherwise a fragment is stranded when the phase moves on and resurfaces out of order at the end of the stream. `finalize_stream()` flushes anything still withheld, so a generation that stops on a bare `<` does not drop it.

`DeepSeekR1ReasoningParser` no longer overrides the streaming method. It existed to catch an end token arriving without a start token, which the base class handles itself now; the override only duplicated that while scanning the accumulated text for the start token on every delta — which is what kept `deepseek_r1` quadratic even after the base class was fixed. Its complete-output override, the actual leniency this parser is known for, is untouched.

After: **63x** for `deepseek_r1` and **64x** for `qwen3` at 50x the tokens.

## Verification

Behaviour is unchanged on real traffic. I replayed four completions captured from a running Qwen3.6-27B server through both the old and new parsers, split on the model's own token boundaries rather than arbitrary character chunks: **identical reasoning and content in all eight parser/sample combinations**, and 1.3x faster at 1198 tokens. The gain scales with output length — at 1200 tokens the quadratic term is still small.

The new tests are checked both ways: **32 of 40 fail against the previous implementation** and all 40 pass against this one. They cover split start and end tags at every chunk size from 1 to 16, truncated tag prefixes at end of stream, and a scaling assertion that fails loudly on quadratic growth.

`tests/test_tool_call_promotion.py::test_stream_large_chunks` caught a first attempt that withheld only the think tags and not `<tool_call>`, which broke promotion when an explicit `<think>` handed the remainder of a delta to the thinking phase. That marker is withheld too now.

Full CI test list passes; the only failures in my environment are four pre-existing ones that need `mlx`, identical on a clean checkout of `main`.

### DIFF
--- a/tests/test_chat_template_kwargs.py
+++ b/tests/test_chat_template_kwargs.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 """Tests for chat template kwargs forwarding."""
 
+import json
 from types import SimpleNamespace
 from unittest.mock import MagicMock, patch
 
@@ -435,3 +436,60 @@ async def test_stream_anthropic_skips_reasoning_parser_when_thinking_disabled():
     assert '"type": "thinking"' not in body
     assert "HEL" in body and "LO" in body
     assert '"type": "text_delta"' in body
+
+
+@pytest.mark.anyio
+async def test_stream_anthropic_flushes_partial_reasoning_marker(monkeypatch):
+    async def fake_stream_chat(messages, **kwargs):
+        yield SimpleNamespace(
+            new_text="thinking</thi",
+            prompt_tokens=3,
+            completion_tokens=2,
+            finished=True,
+            finish_reason="stop",
+        )
+
+    engine = MagicMock(stream_chat=fake_stream_chat, tokenizer=None)
+    messages = [{"role": "user", "content": "Think"}]
+    openai_request = srv.ChatCompletionRequest(
+        model="test-model",
+        messages=[srv.Message(**messages[0])],
+        max_tokens=8,
+    )
+    anthropic_request = srv.AnthropicRequest(
+        model="test-model",
+        max_tokens=8,
+        messages=messages,
+    )
+    prepared = srv.PreparedChatInvocation(
+        messages=messages,
+        chat_kwargs={},
+        response_format=None,
+        json_logits_processor=None,
+    )
+    monkeypatch.setattr(srv, "_reasoning_parser_name", "qwen3")
+    monkeypatch.setattr(srv, "_reasoning_parser", None)
+    monkeypatch.setattr(srv, "_model_name", "test-model")
+
+    body = "".join(
+        [
+            chunk
+            async for chunk in srv._stream_anthropic_messages(
+                engine, openai_request, anthropic_request, prepared
+            )
+        ]
+    )
+    events = [
+        json.loads(line.removeprefix("data: "))
+        for line in body.splitlines()
+        if line.startswith("data: ")
+    ]
+    thinking = "".join(
+        event["delta"]["thinking"]
+        for event in events
+        if event["type"] == "content_block_delta"
+        and event["delta"]["type"] == "thinking_delta"
+    )
+
+    assert thinking == "thinking</thi"
+    assert body.index("thinking</thi") < body.index("message_stop")
--- a/tests/test_reasoning_parser.py
+++ b/tests/test_reasoning_parser.py
@@ -1533,3 +1533,113 @@ class TestGemma4DegenerateCycling:
         # Channel tokens must not leak
         assert "<channel|>" not in full_content
         assert "<|channel>" not in full_content
+
+
+class TestSplitTagStreaming:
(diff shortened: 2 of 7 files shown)


janhilgard:
@Thump604 — review requested. One thing worth flagging before this lands.

This PR and #676 touch the same class hierarchy and the merge order matters. #676 adds `DeepSeekV4ReasoningParser`, which subclasses `DeepSeekR1ReasoningParser` and does its own withholding for the DSML marker. An earlier revision of the fix here stranded withheld text when a tag completed in the same delta, which broke that parser on real model output while every unit test still passed — I only caught it by replaying captured DeepSeek-V4 completions with both branches applied together.

The current revision is verified against that combination: 145 DeepSeek-V4 tests plus a pipeline replay over 23 real completions, clean. But if this merges before #676, the V4 branch will need a rebase and that combined check re-run rather than just a green CI.

The scaling assertion in `TestStreamingCostIsFlat` has a deliberately loose bound (200x for a 20x token increase). It is there to fail on quadratic growth, which measured over 1000x, not to police small regressions on a loaded runner. Tighten it if you would rather it be strict.

Thump604 (approved):
Reviewed the current head as a standalone parser fix. The pending-fragment state keeps split `<think>` / `</think>` markers out of emitted channels without rescanning accumulated output; tool-call buffering and end-of-stream flush behavior remain covered. I also ran `tests/test_reasoning_parser.py` at the exact head: 162 passed. No blocking findings. This approval is for #677 only; #676 still needs a rebase and combined DeepSeek replay after this lands.

waybarrios:
Nice work on the parser. I added a small follow-up because `finalize_stream()` existed, but the server streaming paths were not calling it.

Buffered marker prefixes are now flushed with the final output across Chat Completions, Anthropic Messages, and Responses.

This matters when generation ends with something like `thinking</thi` or a bare `<`. Those final characters could be dropped even though the stream completed normally.

I also added regression tests for all three API paths.

janhilgard:
Rebased on \`main\` — #679 and #684 landing put this back into conflict.

Six conflicts in \`server.py\`, three in \`tests/test_server.py\`. The server ones split cleanly: the new streaming helpers you added (\`_prepare_streaming_reasoning_parser\`, \`_request_tool_definitions\`, \`_streaming_json_fence_stripper\`) and this PR's \`_extract_streaming_reasoning_delta\` are independent additions, so both are kept; the remaining five are the same guard in different functions, where this branch's version is a superset (it adds the \`use_reasoning\`/\`output_finished\` pair so a finished output can still flush). Verified with an AST pass that both names are assigned in every function that reads them — the guard appears in three separate functions and a missed definition would only surface at runtime.

The test file needed more than a merge. Resolving it mechanically spliced one test's signature into the middle of another's body, which pytest surfaced as "async def functions are not natively supported" rather than as a syntax error. I rebuilt it from \`main\`'s copy and re-applied only what this branch actually adds.

That turned out to be one test, not two: \`test_build_tool_parser_returns_request_local_instances\` is the same test you kept as \`test_build_tool_parser_returns_a_fresh_instance_per_stream\`, so it is dropped rather than duplicated. The remaining one, \`test_reasoning_stream_flushes_partial_marker_before_finish\`, is mutation-checked — stubbing out \`finalize_stream\` fails it with \`assert 'thinking' == 'thinking</thi'\`, which is the leak this PR exists to prevent.

Repo suite: 2555 passed, 4 failed — the same four that fail on clean \`upstream/main\` (\`test_shutdown_canceled_prepare_error_unwinds_to_unloaded\`, the two \`_SpecPrefillCancelled\` cases, and the ffmpeg smoke test, which needs a binary this machine lacks). Lint and black clean under the CI invocation.

PR #679, opened 2026-08-03
fix: pin SimpleEngine model load and generation to one thread

janhilgard:
## Problem

MLX streams exist only in the thread that created them, and an array with pending primitives carries the stream those primitives were built on. A model loaded on one thread therefore cannot be driven from another — generation dies inside `mx.eval` with:

```
RuntimeError: There is no Stream(gpu, N) in current thread.
```

`SimpleEngine` spread that work across three different threads:

| route | thread |
|---|---|
| `prepare_for_start()` (model load) | event loop |
| streaming (`_stream_generate_impl`, `_stream_chat_impl`) | event loop |
| non-stream (`_run_blocking_serialized`) | `asyncio.to_thread`, any pool worker |

This held together only while all three happened to agree. As soon as a request took a different route than the one that built the prompt cache, that request and every one after it returned HTTP 500.

The three `bind_generation_streams()` calls scattered across these routes were an attempt to paper over the mismatch. They cannot work in principle: rebinding a module-level handle changes a global that the already-built buffers do not consult.

## Why pinning, and not a better rebind

I built a cache on one thread and evaluated it on another, then measured each candidate:

| approach | result |
|---|---|
| `bind_generation_streams()` per worker (current) | 3/3 fail |
| keep mlx-lm's import-time `ThreadLocalStream`, no rebinding | 3/3 fail |
| **pin load + generation to one thread** | **0/3 fail** |

The middle row is worth noting: `mx.new_thread_local_stream()` resolves per thread, but on the loading thread it resolves to a concrete stream that the pending primitives then carry, so it fails the same way. `mx.set_default_stream()` also rejects a `ThreadLocalStream` outright. There is no rebinding that fixes this — load and generation simply have to share a thread.

## Changes

- `_generation_worker()`: one `ThreadPoolExecutor(max_workers=1)` per engine, shut down in `stop()`. This mirrors what `BatchedEngine` already does in `engine_core.py`.
- Model load moved onto that worker, since the loaded buffers carry its stream.
- Both streaming routes are pumped chunk by chunk through it with `run_in_executor`, using a `_STREAM_DONE` sentinel because `StopIteration` cannot cross a thread boundary. Generator `close()` is routed there too, since cleanup touches MLX.
- Streams are bound once on that thread instead of per request.
- `bind_generation_streams()` docstring corrected — it gives the calling thread a stream; it does not make state portable between threads.

As a side effect, token generation no longer runs on the event loop thread.

## Testing

Reproduced and verified against **DeepSeek-V4-Flash-0731** (283.8B, MXFP4) on an M3 Ultra, where this failed on every single request:

- **30/30 requests pass**, 0 stream errors — multi-turn conversations on a shared prefix, single-shot, and concurrent
- **6/6 tool calls** parsed correctly, plus a `role: tool` round-trip
- **decode holds at 30.4 tok/s** (194 tokens in 6.37 s), so the per-token executor hop costs nothing measurable

Repo suite: **403 passed, 2 skipped**. The new `tests/test_simple_engine_thread_pinning.py` fails **3/3** against `main` and passes with the fix; it is added to the CI whitelist.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,6 +83,7 @@ jobs:
             tests/test_anthropic_models.py \
             tests/test_anthropic_adapter.py \
             tests/test_dependency_floors.py \
+            tests/test_simple_engine_thread_pinning.py \
             tests/test_harmony_parsers.py \
             tests/test_endpoint_model_policies.py \
             tests/test_truncation.py \
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -3571,7 +3571,7 @@ class TestChatCompletionStreamingModeSwitching:
 
         bound_thread = {"id": None}
 
-        def fake_bind_generation_streams():
+        def fake_bind_generation_streams(*_args, **_kwargs):
             bound_thread["id"] = threading.get_ident()
 
         class FakeLLMModel:
@@ -3779,7 +3779,7 @@ class TestChatCompletionStreamingModeSwitching:
 
         bound_thread = {"id": None}
 
-        def fake_bind_generation_streams():
+        def fake_bind_generation_streams(*_args, **_kwargs):
             bound_thread["id"] = threading.get_ident()
 
         class FakeLLMModel:
@@ -3896,7 +3896,7 @@ class TestChatCompletionStreamingModeSwitching:
         bound_thread = {"id": None}
         parser_init_threads: list[int] = []
 
-        def fake_bind_generation_streams():
+        def fake_bind_generation_streams(*_args, **_kwargs):
             bound_thread["id"] = threading.get_ident()
 
         class FakeParser:
--- a/tests/test_simple_engine.py
+++ b/tests/test_simple_engine.py
@@ -1508,7 +1508,10 @@ class TestSimpleEngineConcurrency:
         engine._loaded = True
         engine._text_model = MagicMock()
         engine._text_tokenizer = MagicMock()
-        engine._model = FakeMllmModel()
+        # lifecycle._prepare_engine_start runs prepare_for_start on the
+        # generation worker, so the model's owner thread is that worker rather
+        # than the event loop. Build the fake there to match.
+        engine._model = engine._generation_worker().submit(FakeMllmModel).result()
         engine._run_blocking_serialized = fail_if_called  # type: ignore[method-assign]
 
         outputs = [
@@ -1562,7 +1565,9 @@ class TestSimpleEngineConcurrency:
         )
         engine._loaded = True
         engine._text_model = MagicMock()
-        engine._model = FakeMllmModel()
+        # Same ownership rule as the media route: the model is built on the
+        # generation worker, so that is the thread stream_chat must run on.
+        engine._model = engine._generation_worker().submit(FakeMllmModel).result()
 
         outputs = [
             chunk
@@ -1784,9 +1789,14 @@ class TestSimpleEngineConcurrency:
     ):
         """MLLM text-only non-stream path must keep stream_chat on model thread.
 
-        Regression: aggregate_stream_chat -> stream_chat used _run_blocking_serialized,
-        which moved mlx_vlm stream generation to a worker thread and could raise
+        Regression: aggregate_stream_chat -> stream_chat moved mlx_vlm stream
+        generation off the thread that owns the model and could raise
         "There is no Stream(gpu, N) in current thread".
+
+        The owner is now the engine's pinned generation worker rather than the
+        caller's thread, so the model is built there — which is what ``start()``
+        does. The invariant under test is unchanged: whichever thread builds the
+        model must be the one that generates from it.
         """
         from types import SimpleNamespace
 
@@ -1808,7 +1818,10 @@ class TestSimpleEngineConcurrency:
         engine = SimpleEngine("test-model", force_mllm=True, mtp=False)
         engine._loaded = True
         engine._text_model = None
-        engine._model = FakeMllmModel()
+        loop = asyncio.get_running_loop()
+        engine._model = await loop.run_in_executor(
+            engine._generation_worker(), FakeMllmModel
+        )
 
         output = await engine.chat(
             messages=[{"role": "user", "content": "Count: one, two, three"}],
@@ -1876,13 +1889,23 @@ class TestSimpleEngineConcurrency:
 
     @pytest.mark.anyio
     async def test_requests_complete_in_order(self, mock_model):
-        """Test that concurrent requests complete (may be in any order due to lock)."""
+        """Test that concurrent requests complete (may be in any order due to lock).
+
+        Uses "wait" admission explicitly. This test is about queued requests all
+        completing, not about the admission policy — the default "fail_fast"
+        rejects the second and third outright, which
+        ``test_default_admission_rejects_second_serialized_request`` covers.
+        Before generation was pinned to its own thread, the distinction was
+        invisible here: generation blocked the event loop, so the later requests
+        could not reach admission control until the first had already finished.
+        """
         from vllm_mlx.engine.simple import SimpleEngine
 
         with patch("vllm_mlx.engine.simple.is_mllm_model", return_value=False):
             engine = SimpleEngine("test-model")
             engine._model = mock_model
             engine._loaded = True
+            engine._generation_lock_admission = "wait"
 
             # Launch multiple concurrent generate calls
             results = await asyncio.gather(
--- a/tests/test_simple_engine_thread_pinning.py
+++ b/tests/test_simple_engine_thread_pinning.py
@@ -1,0 +1,553 @@
+# SPDX-License-Identifier: Apache-2.0
+"""SimpleEngine must run model load and every generation route on one thread.
+
+MLX streams exist only in the thread that created them, and an array with
+pending primitives carries the stream those primitives were built on. So a
+model loaded on one thread cannot be driven from another: generation dies in
+``mx.eval`` with "There is no Stream(gpu, N) in current thread".
+
+These tests record the thread of each call rather than touching MLX, so they
+run anywhere. Fakes follow the convention in ``test_gemma4_mllm_patch.py``.
+"""
+
+import asyncio
+import sys
+import threading
+import time
+import types
+from typing import Any
+
+import pytest
+
+
+def _install_mlx_stubs() -> None:
+    """Stub mlx.core only when the real one is missing.
+
+    The stub must carry a real __spec__: transformers probes packages with
+    importlib.util.find_spec, which raises ValueError on a spec-less module and
+    would break every other test sharing this pytest process.
+    """
+    try:  # pragma: no cover - depends on the environment
+        import mlx.core  # noqa: F401
+
+        return
+    except Exception:
+        pass
+
+    import importlib.machinery
+
+    core = types.ModuleType("mlx.core")
+    core.__spec__ = importlib.machinery.ModuleSpec("mlx.core", loader=None)
+
+    class _Stream:
+        def __init__(self, idx: int = 0) -> None:
+            self.idx = idx
+
+    class _Array:
+        """Must be a class, not a factory function.
+
+        The engine annotates parameters as ``mx.array | None``, and PEP 604
+        unions are evaluated at definition time, so a plain function here makes
+        importing the module raise "unsupported operand type(s) for |".
+        """
+
+        def __init__(self, *args: Any, **kwargs: Any) -> None:
+            pass
+
+    core.Stream = _Stream
+    core.array = _Array
+    core.default_device = lambda: "gpu"
+    core.new_stream = lambda *a, **k: _Stream()
+    core.set_default_stream = lambda *a, **k: None
+    core.clear_cache = lambda: None
+    core.metal = types.SimpleNamespace(
+        get_active_memory=lambda: 0,
+        get_peak_memory=lambda: 0,
+        get_cache_memory=lambda: 0,
+    )
+    mlx = types.ModuleType("mlx")
+    mlx.__spec__ = importlib.machinery.ModuleSpec("mlx", loader=None, is_package=True)
+    mlx.__path__ = []
+    mlx.core = core
+    sys.modules.setdefault("mlx", mlx)
+    sys.modules["mlx.core"] = core
+
+
+_install_mlx_stubs()
+
+
+class _Chunk:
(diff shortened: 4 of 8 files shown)


waybarrios:
Hi @janhilgard, really nice PR. The writeup about stream ownership is super clear and  verification speaks for itself. I went through the change set and found a couple of things I wanted to bring up. Not blockers, but I think they're worth a look.

**The big one: stop() and in-flight requests don't play nice together**

Look at this part of stop():

```python
# vllm_mlx/engine/simple.py:650-654
if self._generation_executor is not None:
    # Streams and any cache built on that thread die with it.
    self._generation_executor.shutdown(wait=True)
    self._generation_executor = None
```

`shutdown(wait=True)` blocks the event-loop thread until the worker finishes whatever it's doing. And here's the thing: the worker gets whole-generation units of work. The non-streaming chat route runs through `_run_blocking_serialized(self._model.chat, ..., on_cancel=None)`, so there's no abort path at all, and the streaming routes' abort_event can't fire while the loop is frozen. So if stop() lands in the middle of a long request (server default max_tokens is 32768...), the loop just stalls for minutes. Health checks, timers, cancels: all dead.

Then there's the second half of the problem. The pump captures the worker at entry:

```python
# vllm_mlx/engine/simple.py:1001
worker = self._generation_worker()
...
# vllm_mlx/engine/simple.py:1029
chunk = await loop.run_in_executor(worker, _next_chunk)
```

But stop() shuts the executor down and nulls it. So the next chunk submit on the captured worker raises `RuntimeError("cannot schedule new futures after shutdown")` right in the middle of a client stream, which is a 500 mid-response. And the finally block:

```python
# vllm_mlx/engine/simple.py:1093
await loop.run_in_executor(worker, iterator.close)
```

...fails the same way and gets swallowed by the except Exception at debug level, so the generator's MLX cleanup never runs on its owning thread either.

In the baseline, stop() just nulled a few attributes and streams were pumped on the loop thread, so anything in flight survived a stop cleanly. I can hit this from plain API usage, any task calling `await engine.stop()` while another one is streaming, and also from the server shutdown path (`server.py:1504`, `lifecycle.py:474`) and the model-registry unload paths (`model_registry.py:646,663`).

My position: I'd set a stopping flag the pumps check between chunks, then `shutdown(wait=False, cancel_futures=True)` instead of waiting, since the finally already tolerates a dead executor. Or drain through the existing abort mechanisms and move the wait off the loop with `await asyncio.to_thread(executor.shutdown, True)`. And I'd give the non-streaming chat route an on_cancel/abort hook so stop stays bounded even for long generations. If a submit still lands after shutdown, better to raise a defined EngineStopped error than leak a raw RuntimeError.

For tests, I'd add to test_simple_engine_thread_pinning.py: a slow fake generator plus `await asyncio.wait_for(engine.stop(), timeout=1.0)` asserting stop returns fast and the loop stays alive, a pump crossing stop() asserting it ends with a clean error instead of the raw RuntimeError, and a stop-mid-stream then start() then new-stream restart test to lock the executor recreation path.

**Second, smaller one: the new test doesn't actually drive the engine's stream route**

```python
# tests/test_simple_engine_thread_pinning.py:164-169
it = model.stream_generate()
while True:
    chunk = await loop.run_in_executor(
        worker, lambda: next(it, engine_module._STREAM_DONE)
    )
```

The test re-implements the pump in the test body instead of going through the engine. So the actual pumping code, the _next_chunk closure, the _STREAM_DONE protocol inside _stream_generate_impl, the iterator.close routing, is never executed on Linux CI (this file is the only one covering the pinning behavior and it runs in the no-MLX job). It fails against main only mechanically, via AttributeError on the new API. Meaning: if someone reverts _stream_generate_impl to iterate on the loop thread, which is the exact bug this PR fixes, the suite would still go green. That's the one regression you'd really want caught.

I'd drive the real route with the same scaffolding already in the file: set _loaded = True, keep the fake _acquire_generation_slot, use _ThreadRecordingModel, then `async for chunk in engine._stream_generate_impl(...)` and assert the recorded thread matches the load thread.

Small extra note: the new _generation_worker branch in lifecycle.py doesn't have a direct test either, since the lifecycle tests all use fakes without that attribute. Totally optional though.

Anyway, genuinely nice work. Pinning load and generation to one thread is clearly the right fix, and the fact that the per-token executor hop is free at 30 tok/s is great evidence. Happy to help with any of this if useful.

janhilgard:
Thanks — both findings are real, and the first one is a regression I introduced. Pushed a fix for both.

## stop() and in-flight requests

You are right that `shutdown(wait=True)` on the loop thread is not acceptable, and right about the second half too: the pump captures the worker at entry, so the next submit after teardown raises `cannot schedule new futures after shutdown` mid-response, and the `finally` fails the same way so the generator's MLX cleanup never runs on its owning thread. The baseline tolerated a `stop()` mid-flight; my version did not.

I took your first suggestion — stopping flag plus a non-blocking teardown — with one change I could not talk myself out of. `shutdown(wait=False, cancel_futures=True)` still kills the thread the in-flight route needs for `iterator.close()`, so the raw `RuntimeError` goes away but the off-thread cleanup does not. I tested that shape first and it left `close` running on `MainThread`.

So a busy worker is **detached but left alive**, and ownership passes to the last user:

- `stop()` sets `_stopping`, drives the `on_cancel` hooks that blocking routes already register (specprefill, the cache-aware path), then waits on an idle event with a bounded timeout — `STOP_DRAIN_TIMEOUT_S`, 5 s. This is `asyncio.wait_for`, so the loop stays live throughout.
- If the wait resolves, the executor is shut down normally. If it times out, the executor is moved to `_draining_executors` **without** being shut down, and `stop()` returns.
- The route still holding it sees `_generation_worker_is_live()` go false between chunks, ends the response with `finish_reason="abort"`, and closes its generator on the owning thread via `during_shutdown=True` — which deliberately ignores the stopping flag, since that close is exactly what `stop()` was waiting for.
- The last user retires the detached executor (`_retire_detached_workers`).
- On restart, the new worker's first queued job is `_join_executors` on anything still parked. `max_workers=1` orders that ahead of the model load, so the new thread waits the old one out before it touches Metal — the wait lands on the worker, never on the loop. Two threads in Metal at once is precisely what the pinning exists to prevent, and a `stop()` + `start()` over a live generation would otherwise do it.
- Late submits raise `EngineStopped` (new, in `engine/base.py`, alongside `EngineBusy`). The cleanup path distinguishes it: expected-during-shutdown at debug, anything else now at **warning** rather than debug, since you are right that the old blanket swallow hid this.

On giving the non-streaming chat route a real abort hook: I did not do that, and I want to be explicit rather than quiet about it. `stop()` is bounded regardless now, because it no longer waits on that work — a monolithic `mlx_lm.generate()` simply finishes in the background on a detached thread, which is what the baseline did too. Making it genuinely interruptible means injecting a raising logits processor, and `simple.py`'s non-stream chat reroutes to `aggregate_stream_chat()` whenever `logits_processors` is present, so that would silently move every non-stream chat onto the streaming path. That is a behaviour and performance change well outside this PR. Happy to do it as a follow-up if you want it; it needs its own measurement.

## The test not driving the engine's stream route

Fair, and the sharpest part of the review — a test that fails on `AttributeError` proves nothing about the behaviour. `test_load_and_generation_share_one_thread` now drives `engine._stream_generate_impl` with the scaffolding already in the file, so the `_next_chunk` closure, the `_STREAM_DONE` protocol and the close routing all execute on the no-MLX job.

I checked that by mutation rather than by assertion — reverting the behaviour while keeping the API:

| mutation | caught by |
|---|---|
| pump iterates on the loop thread again | `test_load_and_generation_share_one_thread` — load/generate on `simple-generate_0`, stream on `MainThread` |
| `stop()` back to `shutdown(wait=True)` | `test_stop_does_not_block_the_event_loop` — `stop() held on to the loop for 10.0s` |
| pump submits blindly after `stop()` | `test_pump_crossing_stop_ends_cleanly` — `finish_reason` `'stop'` instead of `'abort'` |
| `lifecycle.py` ignores `_generation_worker` | `test_lifecycle_prepare_runs_on_the_engine_generation_worker` — load on `asyncio_0`, generation on `simple-generate_0` |

Two notes from writing those. `asyncio.wait_for(engine.stop(), timeout=1.0)` **cannot** catch a blocking `stop()` — the synchronous block holds the loop thread, so `wait_for`'s own timeout callback cannot fire either, and it returns successfully once the block clears. My first version of that test passed against the unfixed code. It now measures elapsed time and counts heartbeat ticks. Separately, a fresh `ThreadPoolExecutor` restarts its thread numbering, so the restart test compares name **and** ident; on name alone two different threads both answer to `simple-generate_0`.

I also added `test_lifecycle_prepare_runs_on_the_engine_generation_worker` for the gap you called optional, and `test_submit_to_a_dead_executor_is_translated`, which covers the `RuntimeError` → `EngineStopped` conversion itself and asserts unrelated `RuntimeError`s from the work still pass through untouched.

## Verification

Repo suite: **2273 passed, 23 skipped**. Four failures locally, all reproduced on this branch's parent before today's changes: three are Python 3.14 issues on pristine `upstream/main` (two specprefill cancellation, one lifecycle shutdown), one is a missing `ffmpeg` in my environment. CI here is green on all nine jobs.

The stop/restart behaviour is checked in-process against a real MLX model rather than only against fakes — 12/12:

```
[4] stop() while a stream is mid-flight
    stop() returned in 0.03s, loop ticked 2x, 2 trailing chunks
  PASS  stop() does not block the event loop            — 0.03s
  PASS  event loop stayed responsive during stop()      — 2 ticks
  PASS  mid-stream stop raises nothing at the client    — None
  PASS  stream terminated with a final chunk            — finish_reason='abort'

[5] restart after the mid-stream stop
  PASS  restart uses a different thread                 — .../6179614720 -> .../20006907904
  PASS  no worker left parked                           — []
  PASS  generation works after restart                  — 66 chunks, 258 chars
  PASS  same greedy answer before and after restart

[6] idle stop(): 0.008s
  PASS  idle stop is immediate
```

Unrelated, but I hit it while setting that up and would rather not sit on it: on a VLM whose `TextModel` fails to build (`Failed to build TextModel from vlm: '>' not supported between instances of 'NoneType' and 'int'`, gemma-4-12B here), `engine.stream_generate(prompt=...)` yields the right number of chunks with **empty** text — 40 chunks, `text == ''`. `chat()` on the same engine is fine, since it routes through `stream_chat`. It reproduces identically on this branch's parent, so it is not from this PR. I can open a separate issue with the repro if you have not seen it.

Thump604 (approved):
Re-reviewed the current head after the stop/in-flight findings. The bounded drain/detached-worker handoff removes the event-loop-blocking shutdown, cleanup remains on the owning thread, restart is serialized behind any detached worker, and the regression now drives the real engine pump. I ran `tests/test_simple_engine_thread_pinning.py tests/test_simple_engine.py tests/test_lifecycle_manager.py` at the exact head: 87 passed; all nine GitHub jobs are green. No blocking findings. Please file the independently observed VLM `stream_generate(prompt=...)` empty-text fallback as a separate issue with its minimal model/config reproduction rather than extending this PR.

janhilgard:
Filed as #685, with the minimal reproduction you asked for. It turned out to be two bugs stacked, not the single fallback I described here: `_import_text_model_classes` matches `"gemma4_text"` exactly and defaults everything else to `qwen3_5.TextModel`, so a `gemma4_unified_text` config dies on `args.num_experts > 0` with `NoneType > int` and `build_text_model` returns None — and the mlx_vlm route it then falls back to sends the prompt with no chat template unless media is attached. Nothing further needed on this PR.

janhilgard:
Rebased on `main` — the branch had gone stale and was blocking merge despite the approval. Two conflicts, both resolved by keeping **both** sides rather than picking one.

**1. `finish_reason` in the text stream loop.** `main` grew a proper ladder while this branch was open:

```python
finish_reason = getattr(chunk, "finish_reason", None)
if finish_reason is None:
    finish_reason = "length" if completion_tokens >= max_tokens else "stop"
```

This branch still had the older single-line `getattr(chunk, "finish_reason", "stop")`. The restructured loop is mine, the ladder is yours; the merged version has both. Same for the tail yield: `main` changed its default from `None` to `"stop"`, and my `None` there was inherited from the merge-base rather than intentional, so it is now `"abort" if aborted_by_stop else "stop"`.

**2. `_acquire_generation_slot` on the MLLM route.** `main` moved admission control (fail-fast `EngineBusy`, waiter accounting) behind that helper. This branch still took `self._generation_lock` directly, which would have silently dropped the fail-fast path for media requests. Now composed the same way the text routes already are.

**Two of your MLLM tests needed updating, and I want to be explicit about why.** `test_mllm_media_stream_stays_on_owner_thread_with_text_route` and its draft twin built the fake model with `FakeMllmModel()` in the test body, so its owner thread was the event loop, and then asserted `stream_chat` runs there. That assertion encodes the ownership this PR changes: `lifecycle._prepare_engine_start` now runs `prepare_for_start` on `engine._generation_worker()`, so the model is owned by the worker. The fakes are now built on that worker; the invariant the tests are named for — generation runs on the thread that owns the model — is unchanged, only the owner moved. Mutation-checked: replacing the `_submit_to_generation_worker` call with a direct `_next_chat_chunk()` still fails them with the original `Stream(gpu, N)` error.

**A real bug the rebase surfaced, now fixed.** Binding the worker's stream once per worker left `_worker_generation_stream` alive across a stop/start cycle, so a *fresh* worker called `set_default_stream` on a stream the retired thread had created — a stream it cannot enter. Instrumenting `mx.new_stream`/`mx.set_default_stream` across the suite showed it directly: two different idents both named `simple-generate_0` (a new pool restarts the numbering) binding the same stream. The stream is now cleared wherever the executor is replaced, and re-pointed rather than reallocated within one worker's life — allocating per request is what pushed stream numbering into the 30s. Same instrumentation now reports zero cross-thread binds over the whole suite.

New regression: `test_new_worker_never_binds_the_retired_workers_stream` drives the real stop/restart path. Both mutations fail it — keeping the stream across `stop()`, and allocating a new one per request.

**Suite parity with `main`.** Full run on this branch: 2414 passed, 4 failed — the *same* 4 that fail on clean `upstream/main`, so this branch adds none. Lint and black clean under the CI invocation.

Separately, and not from this PR: `test_cancelling_specprefill_request_stops_during_scoring`, `..._during_sparse_prefill` and `test_shutdown_canceled_prepare_error_unwinds_to_unloaded` fail on `upstream/main` as of `61c78dc`. The specprefill pair lets `_SpecPrefillCancelled` escape instead of being absorbed by the cancellation path. Happy to open an issue or a fix if nobody is on it.

Issue #678, opened 2026-08-03
Prefix cache trims RotatingKVCache past its window: cross-conversation KV leak and repetition loops on sliding-window models (gemma-4)


mihailovmike:
## Summary

`memory_cache.py` rewinds cached KV state for sliding-window models even when mlx_lm
explicitly reports that the cache cannot be trimmed. On a model that interleaves
sliding and full attention (gemma-4: 50 of 60 layers are `sliding_attention`,
`sliding_window=1024`) this desynchronizes layer offsets and, with kv-cache
quantization enabled, leaves another conversation's KV physically resident in the
ring buffer. The observable result is a reply that continues a *different* user's
earlier conversation mid-sentence, mixes languages and LaTeX fragments, and then
degenerates into repeating a single token until the length cap.

## Root cause

`mlx_lm.models.cache.RotatingKVCache.is_trimmable()` returns `False` once
`offset >= max_size` — past the window there is nothing left to rewind to.

Both trim guards in `memory_cache.py` (supersequence at L794, LCP at L888 on `main`)
only test for the presence of attributes:

```python
has_non_trimmable = any(
    not (hasattr(lc, "offset") and hasattr(lc, "keys"))
    for lc in best_lcp_entry.cache
)
```

A `RotatingKVCache` has both attributes, so it is treated as trimmable and
`_trim_cache_offset` runs. `is_trimmable` does not appear anywhere in the module.
The comment at the LCP branch shows the guard was written for hybrid SSM models only.

Two consequences:

1. `_trim_cache_offset`'s `entries_to_keep <= 0` branch sets `tc.offset = 0` for the
   sliding layers, while plain `KVCache` layers get `offset - trim_by`. The layers
   then disagree about position by thousands of tokens.
2. With `--kv-cache-quantization`, layers are `_QuantizedCacheWrapper` and the first
   branch of `_trim_cache_offset` only does `tc.offset = max(offset - trim_by, 0)` —
   the ring buffer is never trimmed, so the evicted conversation's KV survives.

The LCP path makes this cross-conversation: two unrelated chats share a long common
prefix (system prompt plus a tool catalog is easily thousands of tokens), so the
cache of chat A gets trimmed and handed to chat B.

## Reproduction

Deterministic, no model required:

```python
import mlx.core as mx
from mlx_lm.models.cache import KVCache, RotatingKVCache
from vllm_mlx.memory_cache import _trim_cache_offset

def feed(c, n):
    for _ in range(n):
[...]

PR #680, opened 2026-08-08
fix: guard non-trimmable rotating prefix caches

Thump604:
A saturated `RotatingKVCache` cannot be rewound safely, but the prefix-cache LCP and supersequence paths only checked for `offset` and `keys`. That allowed partial reuse and could desync mixed cache layers.

This checks each layer's actual trimmability before partial reuse. Unsupported cache containers and rotating quantized wrappers fail closed. Exact and normal prefix hits are unchanged.

Tests cover supersequence and LCP misses, exact/prefix hits, `CacheList`, and rotating wrapper metadata.

Verification: 2,303 passed, 24 skipped.

Refs #678

### DIFF
--- a/tests/test_memory_cache_mlx.py
+++ b/tests/test_memory_cache_mlx.py
@@ -282,6 +282,125 @@ class TestTrimCacheOffset:
         assert remaining == [999, 1000, 1001]
 
 
+class TestRotatingCachePartialReuse:
+    """Regression coverage for saturated rotating caches in prefix reuse (#678)."""
+
+    @staticmethod
+    def _full_cache(length):
+        import mlx.core as mx
+        from mlx_lm.models.cache import KVCache
+
+        layer = KVCache()
+        layer.keys = mx.ones((1, 1, length, 4), dtype=mx.float32)
+        layer.values = layer.keys
+        layer.offset = length
+        return layer
+
+    @staticmethod
+    def _saturated_rotating_cache(offset=12, max_size=8):
+        import mlx.core as mx
+        from mlx_lm.models.cache import RotatingKVCache
+
+        layer = RotatingKVCache(max_size=max_size, keep=0)
+        layer.keys = mx.ones((1, 1, max_size, 4), dtype=mx.float32)
+        layer.values = layer.keys
+        layer.offset = offset
+        layer._idx = max_size
+        return layer
+
+    @staticmethod
+    def _prefix_cache():
+        from vllm_mlx.memory_cache import MemoryAwarePrefixCache, MemoryCacheConfig
+
+        return MemoryAwarePrefixCache(
+            MagicMock(),
+            MemoryCacheConfig(max_memory_mb=64, max_entries=10, min_prefix_tokens=1),
+        )
+
+    def test_supersequence_skips_saturated_rotating_cache(self):
+        cache = self._prefix_cache()
+        rotating = self._saturated_rotating_cache()
+        cache.store(
+            [1, 2, 3, 4, 5, 6],
+            [self._full_cache(6), rotating],
+        )
+
+        fetched, remaining = cache.fetch([1, 2, 3, 4])
+
+        assert fetched is None
+        assert remaining == [1, 2, 3, 4]
+        assert rotating.offset == 12
+
+    def test_lcp_skips_saturated_rotating_cache(self):
+        cache = self._prefix_cache()
+        rotating = self._saturated_rotating_cache()
+        cache.store(
+            [1, 2, 3, 4, 5, 6],
+            [self._full_cache(6), rotating],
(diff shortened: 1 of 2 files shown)


janhilgard (approved):
Approving. The diagnosis in #678 is right, the guard is in the right place for the two paths it covers, and the tests pin the behaviour rather than the shape of the code. I reproduced the underlying defect and checked the fix against it.

**The desync is real and starts exactly where `is_trimmable()` flips.** Mixed sliding/full layers, `max_size=8`, both layers at `offset=12`:

```
trim_by= 1: full offset=11  rotating offset=11
trim_by= 5: full offset= 7  rotating offset= 7
trim_by= 8: full offset= 4  rotating offset= 0   <-- desync
trim_by=10: full offset= 2  rotating offset= 0   <-- desync
```

`_trim_cache_offset`'s `entries_to_keep <= 0` branch zeroes the rotating layer while plain `KVCache` layers keep `offset - trim_by`, exactly as the issue describes. With the patch, `_is_cache_layer_trimmable` returns False for that cache and both `fetch()` paths bail out before reaching it.

Three things I checked and liked:

- Keying the container rejection off `.caches` rather than off `CacheList` specifically. The constraint is that `_trim_cache_offset` does not rewind children, not what the container is called, and the comment says so.
- `is_trimmable()` wrapped in try/except returning False. Fail-closed is the only safe default for a predicate that gates KV reuse.
- `test_cache_list_is_rejected_even_when_children_are_trimmable` — that is the case that would otherwise look safe and pass a naive recursive check.

## One gap worth a follow-up, not a blocker

`_trim_cache_offset` has five more call sites that do not consult the new predicate:

```
vllm_mlx/mllm_batch_generator.py:1519   _trim_cache_offset(cached_kv, 1)
vllm_mlx/mllm_batch_generator.py:2002   _trim_cache_offset(extracted, output_count + S)
vllm_mlx/mllm_batch_generator.py:2907   _trim_cache_offset(partial["cache"], S)
vllm_mlx/mllm_batch_generator.py:2992   _trim_cache_offset(cached_kv, 1)
vllm_mlx/scheduler.py:1452              _trim_cache_offset(extracted_cache, 1)
```

The three that trim by a small constant are below `max_size` and, per the measurement above, do not desync. **The store path at 2002 is the one I would look at**: `total_trim = output_count + S` scales with generated tokens, so on gemma-4 (`sliding_window=1024`) any answer longer than the window trims past `max_size` and writes an already-desynced entry into the prefix cache. That entry is then served by the exact and prefix paths, which this PR deliberately leaves unchanged — so the read-side guard would not catch it.

I have not reproduced that end to end through the MLLM generator, only the primitive plus the unguarded call site with an unbounded trim amount, so I am flagging it rather than claiming it. It is a separate site from what this PR set out to fix and should not hold it up.

Two smaller notes:

- The `_QuantizedCacheWrapper` branch rejects on `"max_size" in orig_attrs` regardless of saturation, so a quantized rotating cache that has *not* wrapped is refused even though the unwrapped equivalent would be allowed. Fail-closed and I would ship it, but it costs reuse on exactly the configuration where KV quantization is most attractive, and the asymmetry is worth a comment so nobody later reads it as a bug.
- `_trim_rotating_caches` in `mllm_batch_generator.py` is doing something adjacent — clamping `offset` to `max_size` on restored caches — but as a crash workaround for `_update_in_place` going negative, not as a correctness guard. Now that there is a real predicate, those two ideas could probably be reconciled.

## Production check

Our two production models (Qwen3.6-27B, Qwen3.6-35B-A3B) report `sliding_window: null` and no `sliding_attention` entries, so they are not exposed to this. The prefix cache itself is live and busy on them — 417k tokens saved, three resident entries at the moment — so the guarded code runs on every request; it just never sees a rotating layer. Gemma 4 is a different story and we do keep those checkpoints around, so this would have bitten us the moment one came back online.

Also checked against #683, which changes `estimate_kv_cache_memory` in the same file and adds the complementary half (a post-prefill snapshot keyed so non-trimmable caches get a reusable entry rather than only being refused one). No conflict, and the two compose the way you would want: this PR closes the unsafe partial-reuse paths, that one opens a safe one.

`tests/test_memory_cache_mlx.py` + `tests/test_memory_cache.py` at your head: 68 passed. All nine CI jobs green.

Thump604:
Thanks. I reproduced the separate completion-store path and filed #689 with the exact mixed-cache failure. Keeping this PR unchanged.

waybarrios (approved):
Verified the current head and GitHub's current merge result. The trimmability guard correctly fails closed for saturated rotating layers and cache containers, while exact and normal prefix hits remain unchanged. The focused cache suites passed 68 out of 68 tests on both revisions. This is ready to merge.

PR #681, opened 2026-08-08
fix(simple-engine): preserve streaming finish reasons

Thump604:
Natural generator exhaustion currently emits a final chunk with no finish reason. Strict OpenAI clients treat that as a truncated stream.

This stamps natural exhaustion as `stop`. It also fixes the engine-enforced token-limit fallback so a backend chunk with no reason is reported as `length`, while preserving explicit backend reasons.

Tests cover natural and empty exhaustion, token limits, exceptions, and request cleanup.

This adopts and extends #629 because the required regression patch could not be added to its fork-owned branch. Miguel Vilhena is credited as co-author.

Verification: 2,301 passed, 24 skipped.

Refs #628

### DIFF
--- a/tests/test_simple_engine.py
+++ b/tests/test_simple_engine.py
@@ -2618,6 +2618,85 @@ class TestSimpleEngineConcurrency:
         assert await asyncio.to_thread(prefill_cancelled.wait, 1.0)
 
 
+class TestSimpleEngineNaturalStop:
+    @staticmethod
+    def _engine(stream_generate):
+        from vllm_mlx.engine.simple import SimpleEngine
+
+        with patch("vllm_mlx.engine.simple.is_mllm_model", return_value=False):
+            engine = SimpleEngine("test-model")
+        engine._loaded = True
+        engine._model = MagicMock()
+        engine._model.tokenizer.encode.return_value = [1, 2, 3]
+        engine._model.stream_generate.side_effect = stream_generate
+        return engine
+
+    @pytest.mark.anyio
+    async def test_generator_exhaustion_emits_one_stop(self):
+        def generate(**kwargs):
+            yield SimpleNamespace(text="Hel", prompt_tokens=3, finish_reason=None)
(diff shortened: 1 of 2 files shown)


janhilgard (approved):
Approving. Reproduced both halves against `main` and confirmed the fix, driving `_stream_generate_impl` with a backend that exhausts without ever emitting a `finished` chunk — the shape #628 describes.

```
main   natural exhaustion  ->  finish_reason=None   finished=True
main   token limit         ->  finish_reason=None
#681   natural exhaustion  ->  finish_reason='stop'
#681   token limit         ->  finish_reason='length'
```

Both were None on `main`, so a strict client sees every stream as truncated, not just the natural-stop case. The issue title says "length works", which is true at the API level but not here — the `length` case was equally unset at this layer.

The root cause is worth spelling out because it is easy to reintroduce: `getattr(chunk, "finish_reason", "stop")` supplies its default only when the **attribute is missing**, never when it is present and `None`. Backends set it to `None` routinely, so the default was dead code. Splitting the lookup from the fallback is the fix, and deriving `length` from `completion_tokens >= max_tokens` rather than assuming `stop` is the right way round.

Crediting Miguel Vilhena as co-author while adopting #629 is the right handling of a fork-owned branch.

## Merge note

This conflicts with #679, which is approved and touches the same block. The conflict is textual only — #679 re-indents that region into a `_generation_worker_in_use()` context and rewrites the tail's terminal chunk — but the **semantic** resolution needs stating, or a rebase will silently drop one intent:

- #679 yields `finish_reason="abort"` from that tail when a stream is cut short by `stop()`.
- #681 yields `"stop"` from the same tail on natural exhaustion.

Combined, it should be `"abort" if aborted_by_stop else "stop"`. Picking either branch wholesale loses the other. Merges cleanly against #690 with no conflict.

CI green on all nine jobs.

waybarrios:
I carefully reviewed and it is all set. Dont see any major flaw or issue here.

PR #683, opened 2026-08-08
fix(scheduler): make the prefix cache usable on non-trimmable KV caches

janhilgard:
Split out of #676 on review.

## Problem

The scheduler stored one prefix-cache entry per request, keyed by `prompt + output`. Every later query is *shorter* than such a key, so reuse requires trimming the generated tail off first. Models with sliding-window or pooled KV cannot do that: `RotatingKVCache.is_trimmable()` is `offset < max_size`, so once the ring buffer wraps the older KV is physically overwritten, and a pooling cache merges tokens into windows it cannot split.

So on those architectures zero hits were **guaranteed by construction** — and the entries were not free. Each holds a full-length KV copy, and in an agentic loop where every turn extends the previous prompt they accumulate. I measured 45 entries of a ~46k-token cache. Each is hundreds of arrays (43 layers, `CacheList` of three caches each), and Metal runs out of **buffers** long before the cache's byte budget notices:

```
RuntimeError: [metal::malloc] Resource limit (499000) exceeded.
[generation_error_recovery] aborted 1 running requests, batch generator closed
```

The failure lands inside the model during a generation step, so the request is aborted and the client just sees the connection stall.

## Changes

- **Snapshot the cache while it still covers exactly the prompt** — before the first generated token is appended — and store that under the prompt tokens. Such an entry is reusable by strict prefix match with no trimming at all.
- **Skip the completion-time `prompt + output` entry when the cache is not trimmable**, since it can never be reused.

Three things the implementation has to get right, each of which cost me a debugging round:

- mlx-lm attaches `prompt_cache` only to the response carrying a `finish_reason`, so mid-generation it is None. The per-sequence cache is pulled from the live batch via `extract_cache(idx)` instead, which is what that attribute is built from anyway.
- **The snapshot must be a real copy.** Both cache types write into their buffers in place, so a snapshot that aliases them is rewritten by the very generation it is supposed to predate.
- **It is stored with `evict_prefixes=True`.** Without it each turn adds another full-length copy rather than replacing the entry it supersedes — that is the buffer exhaustion above. This one I shipped wrong first and only caught under a real agentic client.

Exact matches are deliberately **skipped** rather than used. The scheduler re-feeds `prompt[-1:]` on those, which duplicates that token in the KV cache when the entry covers the whole key — measured as the same prompt returning a different answer.

`memory_cache.py` could not measure these caches either: `CacheList.state` is nested, and the two-way unpack raised a `ValueError` that was swallowed, so every such entry was accounted as 0 bytes and the byte-based LRU never evicted.

## Verification

DeepSeek-V4-Flash-0731 (283.8B MXFP4, M3 Ultra), 8-turn tool-calling conversation over a ~33k-token context:

| | before | after |
|---|---|---|
| cache entries | 45 and climbing | **1** |
| Metal buffer errors | 7 | **0** |
| cache hits | — | **7/8**, 84 tokens prefilled instead of 33k |
| cold vs warm output | — | **10/10 character-identical** over 5 scenarios |
| long shared prefix | 4.19 s | **1.09 s** |

`tests/test_prefix_cache_untrimmable.py` covers what you asked for specifically: a wrapped `RotatingKVCache` and a mixed `CacheList` are recognised as unable to back a `prompt + output` entry while a plain `KVCache` still gets one, and memory ownership is pinned by mutating the live cache after snapshotting and asserting the snapshot does not move. Confirmed by mutation — aliasing instead of copying fails three of them.

Repo suite: **2303 passed**. The four local failures I see are reproduced on pristine `main` (three Python 3.14 issues, one missing `ffmpeg`).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,6 +140,7 @@ jobs:
             tests/test_mllm.py \
             tests/test_server.py \
             tests/test_paged_cache.py \
+            tests/test_prefix_cache_untrimmable.py \
             tests/test_qwen35_mtp_patch.py \
             tests/test_mllm_continuous_batching.py \
             tests/test_mllm_cache.py \
--- a/tests/test_memory_cache.py
+++ b/tests/test_memory_cache.py
@@ -198,6 +198,36 @@ class TestArrayMemory:
         expected = 2 * (1 * 8 * 100 * 64 * 2)
         assert estimate_kv_cache_memory([layer]) == expected
 
+    def test_estimate_handles_nested_cachelist_state(self):
+        """Regression: DeepSeek-V4's CacheList.state nests three sub-states.
+
+        The old two-way unpack raised ValueError, was swallowed, and the whole
+        entry counted as 0 bytes — so the dashboard's Prefix Cache bar stayed
+        at 0% and byte-based LRU eviction never fired for such models.
+        """
+
+        class NestedStateCache:
+            def __init__(self):
+                rot = (
+                    MockShapeArray(shape=(1, 8, 128, 64), dtype_size=2),
+                    MockShapeArray(shape=(1, 8, 128, 64), dtype_size=2),
+                )
+                pool_a = (
+                    MockShapeArray(shape=(1, 3, 512), dtype_size=2),
+                    MockShapeArray(shape=(1, 3, 128), dtype_size=2),
+                    MockShapeArray(shape=(1, 40, 512), dtype_size=2),
+                )
+                pool_b = (None, None, None)  # empty PoolingCache members
+                self.state = [rot, pool_a, pool_b]
+
+        expected = (
+            2 * (1 * 8 * 128 * 64 * 2)
+            + (1 * 3 * 512 * 2)
+            + (1 * 3 * 128 * 2)
+            + (1 * 40 * 512 * 2)
+        )
+        assert estimate_kv_cache_memory([NestedStateCache()]) == expected
+
 
 class TestEstimateKvCacheMemory:
     """Tests for estimate_kv_cache_memory function."""
--- a/tests/test_prefix_cache_untrimmable.py
+++ b/tests/test_prefix_cache_untrimmable.py
@@ -1,0 +1,413 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Prefix cache behaviour for caches that cannot be trimmed.
+
+A ``prompt + output`` cache key is only reusable via a trim: every later query
+is shorter than such a key, so the generated tail has to come off first. Models
+with sliding-window or pooled KV cannot do that, which makes those entries dead
+weight — and expensive weight, since each one holds a full-length KV copy and
+Metal runs out of *buffers* long before the cache's byte budget is reached.
+
+These tests pin the two halves of the fix: skipping the unusable entry, and
+storing a post-prefill snapshot that is reusable by strict prefix match.
+"""
+
+import pytest
+
+mx = pytest.importorskip("mlx.core")
+cache_mod = pytest.importorskip("mlx_lm.models.cache")
+
+from vllm_mlx.scheduler import Scheduler  # noqa: E402
+
+
+def _rotating_past_its_window() -> list:
+    """A RotatingKVCache that has wrapped, so trimming is impossible."""
+    c = cache_mod.RotatingKVCache(max_size=8)
+    # 12 > max_size: the ring buffer has physically overwritten older KV.
+    c.update_and_fetch(mx.zeros((1, 1, 12, 4)), mx.zeros((1, 1, 12, 4)))
+    return [c]
+
+
+def _plain_kv_cache() -> list:
+    c = cache_mod.KVCache()
+    c.update_and_fetch(mx.zeros((1, 1, 4, 4)), mx.zeros((1, 1, 4, 4)))
+    return [c]
+
+
+class TestPromptOutputEntryIsUseless:
+    def test_wrapped_rotating_cache_cannot_back_a_prompt_output_entry(self):
+        cache = _rotating_past_its_window()
+        assert not cache_mod.can_trim_prompt_cache(cache), "setup: must be untrimmable"
+        assert Scheduler._prompt_output_entry_is_useless(cache) is True
+
+    def test_plain_kv_cache_still_gets_its_entry(self):
+        cache = _plain_kv_cache()
+        assert cache_mod.can_trim_prompt_cache(cache), "setup: must be trimmable"
+        assert Scheduler._prompt_output_entry_is_useless(cache) is False
+
+    def test_cache_list_is_useless_when_any_member_is(self):
+        """CacheList reports one verdict for the whole group."""
+        grouped = cache_mod.CacheList(*_plain_kv_cache(), *_rotating_past_its_window())
+        assert Scheduler._prompt_output_entry_is_useless([grouped]) is True
+
+    def test_unknown_cache_objects_do_not_break_the_scheduler(self):
+        """Anything that cannot answer the question keeps its entry."""
+
+        class Opaque:
+            pass
+
+        assert Scheduler._prompt_output_entry_is_useless([Opaque()]) is False
+
+
+class TestSnapshotOwnsItsMemory:
+    """A snapshot that aliases the live cache is rewritten by the generation
+    it is supposed to predate: RotatingKVCache writes into its ring buffer and
+    PoolingCache into its remainder buffer, both in place."""
+
+    def test_copy_is_not_a_view_of_the_source(self):
+        src = mx.array([1.0, 2.0, 3.0])
+        copy = Scheduler._copy_cache_state(src)
+        mx.eval(copy)
+        assert copy is not src
+        assert mx.array_equal(copy, src)
+
+    def test_copy_survives_in_place_mutation_of_the_source(self):
+        src = mx.zeros((4,))
+        copy = Scheduler._copy_cache_state(src)
+        mx.eval(copy)
+        src[0:4] = mx.array([9.0, 9.0, 9.0, 9.0])
+        mx.eval(src)
+        assert copy.tolist() == [
+            0.0,
+            0.0,
+            0.0,
+            0.0,
+        ], "snapshot aliased the live cache and was overwritten by generation"
+
+    def test_nested_containers_are_copied_through(self):
+        src = [mx.array([1.0]), (mx.array([2.0]), mx.array([3.0])), None, 7]
+        copy = Scheduler._copy_cache_state(src)
+
+        assert isinstance(copy, list) and isinstance(copy[1], tuple)
+        assert copy[0] is not src[0]
+        assert copy[1][0] is not src[1][0]
+        # Non-arrays are metadata, passed through as-is.
+        assert copy[2] is None and copy[3] == 7
+
+    def test_real_rotating_cache_state_is_detached(self):
+        live = cache_mod.RotatingKVCache(max_size=8)
+        live.update_and_fetch(mx.ones((1, 1, 4, 4)), mx.ones((1, 1, 4, 4)))
+        snapshot = Scheduler._copy_cache_state(live.state)
+        mx.eval(snapshot)
+
+        before = [mx.sum(a).item() for a in snapshot if isinstance(a, mx.array)]
+        # Keep generating: the ring buffer is written in place past max_size.
+        for _ in range(6):
+            live.update_and_fetch(mx.ones((1, 1, 4, 4)) * 5, mx.ones((1, 1, 4, 4)) * 5)
+        after = [mx.sum(a).item() for a in snapshot if isinstance(a, mx.array)]
+
+        assert before == after, "continued generation mutated the stored snapshot"
+
+
+class _FakeRequest:
+    def __init__(self, prompt_token_ids):
+        self.prompt_token_ids = list(prompt_token_ids)
+        self.request_id = "req-0123456789"
+        self.cached_tokens = 0
+        self.num_output_tokens = 0
+
+
+class _FakeResponse:
+    def __init__(self, token=None, uid=0):
+        self.token = token
+        self.uid = uid
+
+
+def _bare_scheduler():
+    sched = object.__new__(Scheduler)
+    return sched
+
+
+class TestSnapshotKeyAlignment:
+    """The key must name exactly the tokens the snapshot holds.
+
+    The snapshot is taken while processing the response carrying the first
+    generated token, and the batch has already fed that token through the
+    cache: measured ``prompt_len=5, cache_offset=6`` on a real scheduler run.
+    Storing that under ``prompt_token_ids`` leaves warm reuse one token ahead
+    of its key, and trimming the overshoot off is unavailable here — these are
+    the caches that cannot be trimmed.
+    """
+
+    def test_key_is_extended_to_cover_the_generated_token(self):
+        sched = _bare_scheduler()
+        request = _FakeRequest([1, 2, 3, 4, 5])
+        cache = _plain_kv_cache()
+        cache[0].offset = 6  # prompt + first generated token
+
+        key = Scheduler._cache_key_for_snapshot(
+            sched, request, _FakeResponse(token=99), cache
+        )
+
+        assert key == [1, 2, 3, 4, 5, 99]
+        assert len(key) == Scheduler._cache_coverage(
+            cache
+        ), "key length must equal what the cache holds"
+
+    def test_key_is_the_prompt_when_the_cache_stops_there(self):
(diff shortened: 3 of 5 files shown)


Thump604 (changes requested):
Two correctness blockers. First, the prompt snapshot is taken after GenerationBatch.next has already fed the first generated token through the cache, then it is stored under prompt_token_ids. That leaves warm reuse one token ahead. Capture at the prompt-to-generation transition before next, and add a cache/key alignment regression. Second, snapshot reconstruction calls make_prompt_cache without SchedulerConfig.max_kv_size. A rotating live cache can become a plain KVCache destination, fail meta_state assignment, and be silently skipped. Preserve the live cache topology and configured max_kv_size, and test the real store path with rotating KV.

janhilgard:
Both fixed. I measured each before touching it, and one of your two findings sent me somewhere I did not expect.

## 1. Key alignment — you are right, and my earlier verification was not capable of catching it

Instrumented the real scheduler store path on Qwen3-0.6B:

```
prompt_len=5  cache_offset=6  delta=+1  num_output_tokens=0
```

So the entry was stored one token short of what it held, exactly as you said. My "10/10 character-identical cold vs warm" was on DeepSeek and compared full replies, which cannot see a one-token key offset.

I could not take your suggested route of capturing before `next`. The overshoot is already in the cache by the time any response reaches `_process_batch_responses`, and trimming it back off is the one thing unavailable here — these are the non-trimmable caches this PR is about. So I extended the key instead: it now names exactly the tokens the snapshot holds, `prompt + [first generated token]`. That token is the first token of the reply, which the next turn's prompt also contains, so the entry still matches by strict prefix — which is the case the feature exists for. A key that cannot be named exactly means no entry at all, rather than a misaligned one.

Honest note on impact: I could not produce an output difference from the misalignment, only the invariant violation. What visibly changes is the hit accounting — old key 0/1/2 hits across cold/warm/extended, new key 0/0/1. The hit the old key gained on an identical repeat was an exact match, which this code discards anyway (`cache_hit_type == "exact"`), so no usable reuse is lost. The extended-prompt hit, the one that matters, still lands.

## 2. Destination topology — right about the bug, and the fix is not the one either of us proposed

Confirmed from the code: `make_prompt_cache(self.model)` yields plain `KVCache`, assigning a `RotatingKVCache`'s state or meta_state to one raises, and my broad handler turned that into a warning. Silent no-store on exactly the sliding-window configurations this feature targets.

But deriving the destination from `SchedulerConfig.max_kv_size`, as suggested, is also wrong — I only found this because I tested it:

```
max_kv_size=512 configured -> destination RotatingKVCache, live layer KVCache
[cache_store_prompt] layer 0 topology mismatch ... not storing
```

`_create_batch_generator` never passes `max_kv_size` to `BatchGenerator`, so the config does not describe the live batch cache at all. Keying off it just mismatches in the other direction.

So the destination now mirrors the live objects themselves — a shallow copy per layer, which keeps the class and every scalar (`max_size`, `keep`, `step`, `_idx`) while the caller overwrites the arrays, the only part that must not be shared. That is topology-preserving by construction rather than by agreement between two code paths.

Worth flagging separately: **`max_kv_size` not reaching `BatchGenerator` looks like its own bug.** `BatchGenerator.__init__` accepts it, `SchedulerConfig` has it, and the bounded-cache branch in `_prepare_batch_inputs` honours it, but the batch generator is constructed without it. I have not chased it and it is outside this PR.

## Tests

Eight new ones in `tests/test_prefix_cache_untrimmable.py`: key alignment for overshoot 0 and 1, refusal when the cache is short of the prompt or overshoots by more tokens than are known, and destination topology for rotating, mixed, and the state/meta_state assignment that used to raise. Mutation-checked — reverting the key to `prompt_token_ids` fails the alignment test, and going back to `make_prompt_cache` fails all three topology tests.

Re-verified end to end through the real scheduler after both fixes: identical greedy output cold and warm, key length equal to cache coverage, prefix hit on the extended second-turn prompt.

Repo suite: 2310 passed.

janhilgard:
Followed up on the `max_kv_size` oddity I flagged above — it is worse than "not passed to BatchGenerator". **`--max-kv-size` is a complete no-op today**: the scheduler builds the bounded cache, its own `_validate_cache` rejects it (that validator returns False for any cache whose `keys` are still None, which is every freshly created one), and BatchGenerator's fallback is unbounded because it was never given the size. Filed as #688 with the trace and before/after measurements. Independent of this PR.

Thump604 (changes requested):
Thanks. The flat-layer cases are fixed, but CacheList still breaks both invariants. _cache_coverage does not descend into it, so the key falls back to prompt-only, and copy.copy on CacheList shares its child caches, so the snapshot follows live mutation. Please recurse through CacheList for both coverage and destination construction, and add a nested KVCache plus RotatingKVCache regression proving the key length matches coverage and the snapshot stays unchanged after live generation.

janhilgard:
Both fixed. You are right that the flat-layer fix left the container case broken, and it breaks on the architecture this PR was written for — DeepSeek-V4 groups three caches per layer in a `CacheList`.

Confirmed both before touching anything:

```
CacheList has .offset      : False
_cache_coverage([CacheList]): None      (should be 7)
destination shares children : True
```

So on that architecture the key fell straight back to prompt-only — the exact misalignment the previous round fixed for flat layers — and the "snapshot" aliased the live children, so it followed the generation it is supposed to predate. Every flat-layer test still passed throughout, which is the uncomfortable part.

Both now recurse:

- `_cache_coverage` descends through `.caches` until it finds an `offset`.
- `_make_snapshot_destination` rebuilds the container from mirrored children rather than `copy.copy`-ing it, since a shallow copy of a container shares exactly the objects that must not be shared.

Four new tests in `TestNestedCacheListInvariants`, mixed nested `KVCache` + `RotatingKVCache` as you asked: coverage descends, key length equals coverage, the snapshot shares no child with the live cache, and — the one that actually matters — the snapshot's contents are unchanged after six further `update_and_fetch` calls on the live children.

Mutation-checked: making coverage non-recursive fails the coverage and key-length tests; restoring the shallow container copy fails three, including the survives-live-generation one.

Repo suite: 2314 passed.

Separately, I chased the `max_kv_size` oddity I flagged earlier and it turned out to be its own bug — `--max-kv-size` is a complete no-op today. Filed as #688, independent of this PR.

Thump604 (approved):
Rechecked f13ab61. The recursive CacheList coverage and destination fixes address both remaining blockers, and the mixed-cache mutation tests cover the actual failure mode. Focused tests pass and CI is green. Thanks.

waybarrios:
Re-reviewed current head `f13ab61` together with #642. I found two correctness blockers and one CI coverage gap.

### 1. Unknown coverage is stored under the wrong key

In `_cache_key_for_snapshot()`, `covered is None` falls back to `prompt_ids`. Pure `ArraysCache` models have no `offset`, but mlx-lm has already processed the first generated token before the first response is returned. The stored recurrent state therefore covers `prompt + first_generated_token` while its key covers only `prompt`.

A later extended prompt skips only the keyed prompt and processes that first response token again, corrupting cumulative recurrent state. Completion storage is also skipped for `ArraysCache`, so this misaligned snapshot can be the only reusable entry.

Suggested fix: derive the key from BatchGenerator's authoritative per-sequence token list, or pass explicit covered tokens into the snapshot callback. If the covered sequence cannot be established, fail closed and skip the store instead of assuming prompt-only coverage.

Regression: use a real `ArraysCache`, store after the first response, reuse it for the next-turn prefix, and assert key length matches the represented sequence and the continuation matches a cold run.

### 2. The first-response snapshot is also applied to trimmable KV caches

The new path runs for every memory-aware cache. For ordinary `KVCache`, it synchronously copies and evaluates the full context before returning the first token, while the completion path can retain another full entry.

On compatibility paths where the prompt-save callback created the safe `N` key with `N-1` coverage, the new `N+1` snapshot also evicts that entry. An identical `N`-token prompt can then receive a trimmed supersequence cache covering `N`; the scheduler rejects only `exact` hits and replays `prompt[-1]`, duplicating the final prompt token.

Suggested fix: gate the first-response snapshot to cache topologies that actually require non-trimmable reuse. Alternatively, carry explicit cache coverage and a `requires_last_token_replay` flag from the stored entry through `fetch()` to scheduling.

Regression: cold then identical generation with a real plain `KVCache`; assert the consumed cache plus replay covers exactly `N` tokens, greedy output matches the cold run, and only one useful entry remains.

### 3. The new MLX regressions are not executed by CI

`tests/test_prefix_cache_untrimmable.py` requires MLX, but the Apple job explicitly enumerates files and does not include it. None of its 19 key-alignment, topology, ownership, or mutation tests run in the green checks.

Suggested fix: add `tests/test_prefix_cache_untrimmable.py` to the Apple Silicon pytest list.

Thump604 (changes requested):
Updating my earlier approval after the owner rereview: changes are required before merge.

The two correctness blockers are substantive: unknown represented-token coverage must not fall back to prompt_ids for non-trimmable/ArraysCache snapshots, and first-response snapshotting must not create an N+1 supersequence entry for ordinary trimmable KVCache that can be reused as an exact N-token prefix. Please either carry authoritative represented tokens/coverage or skip storage when coverage is unknowable, and gate the first-response snapshot to non-trimmable topologies (or propagate explicit replay coverage through fetch/scheduling). Add the real ArraysCache next-turn parity regression, the identical-prompt KVCache cold-parity/replay regression, and include tests/test_prefix_cache_untrimmable.py in Apple CI.

Once those are green, #642 should rebase on this corrected branch and own detachment/evaluation at the storage boundary.

janhilgard:
All three confirmed and fixed. @waybarrios — thanks for the re-review; the second one in particular is a case where my fix created the very bug this PR was written to prevent, just approached from the other side.

**1. Unknown coverage now fails closed.** Verified `ArraysCache` exposes no `offset`, so `_cache_coverage` returns None and the old fallback stored the state under `prompt_ids` while it actually covered `prompt + first generated token`. Since completion storage is skipped for these models, that misaligned entry was the only one they ever got. It now skips the store and logs which layer types could not be measured. Guessing costs correctness; skipping costs a prefill.

I did not take the "derive the key from BatchGenerator's per-sequence token list" route. It is the better answer in principle, but it means reaching further into mlx-lm internals from the scheduler, and I would rather not add that coupling in a PR whose whole subject is cache-state ownership going wrong. Failing closed is the conservative half of your suggestion and can be upgraded later without changing the contract.

**2. The snapshot is gated to non-trimmable topologies.** Confirmed the over-application:

```
plain KVCache      can_trim=True   _prompt_output_entry_is_useless=False
saturated Rotating can_trim=False  _prompt_output_entry_is_useless=True
```

It ran for both. Gating on `_prompt_output_entry_is_useless` — the predicate this PR already uses to decide whether the completion entry is worth keeping — is exactly the right condition: the snapshot exists for the topologies whose completion entry is unusable, and nowhere else. That also removes the synchronous full-context copy before the first token on the ordinary path, which was pure cost.

**3. `tests/test_prefix_cache_untrimmable.py` added to the Apple job.** You are right and it is worse than it looks: **19 tests, none of them running**, including every mutation-checked one I have been citing as evidence in this thread. I verified the file now sits in `test-apple-silicon` and that the YAML still parses.

Two new regression classes: gating (trimmable left to the completion path, verified through the real store path rather than the predicate alone) and unknown coverage (real `ArraysCache`, flat and nested in a `CacheList`). Mutation-checked — removing the gate and restoring the `prompt_ids` fallback each fail their own tests.

Repo suite: **2320 passed**, lint and black clean under the exact CI invocation.

On the end-to-end parity regressions you asked for — an `ArraysCache` next-turn continuation and an identical-prompt `KVCache` cold-vs-replay comparison — I have not added those. Both need a real model driven through the scheduler to be meaningful, and a fake would assert my own assumptions rather than the behaviour. I would rather say that plainly than ship two tests that look like the ones requested and prove less. If you want them in this PR I will build the harness; otherwise the gate above removes the `KVCache` path entirely and the `ArraysCache` path no longer stores at all, so neither scenario can currently arise.

Happy for #642 to rebase on this once it is green.

waybarrios:
@janhilgard all good!

Thump604 (approved):
Rechecked `ff31276c` after the owner follow-up. The corrected implementation closes the two unsafe states conservatively: unknown recurrent coverage now skips storage instead of inventing a prompt-only key, and first-response snapshots are restricted to the non-trimmable topology that requires them, leaving ordinary KVCache on its existing safe path. The previously omitted MLX regression file is now in Apple CI, and the full Linux/Apple/lint/type matrix is green.

The requested end-to-end parity harnesses were not added, but the corresponding unsafe store/reuse paths are now unreachable and the new mutation-checked regressions cover those gates directly. No remaining blocker from my review. Thanks for the careful correction.

waybarrios (approved):
ready to ship! ⛵

Thump604:
Correction after merge: #683 is now merged, so we will build and validate these as a small tests-only follow-up PR against current `main`, not as a cherry-pick onto Jan’s branch.

The follow-up harness will cover:

1. A real `ArraysCache` model run through the scheduler to the first response and next-turn continuation, compared against a cold run, with represented-token/key coverage asserted explicitly.
2. A real plain `KVCache` model with cold generation followed by an identical-prompt reuse attempt, asserting cache coverage plus replay totals exactly `N`, greedy output matches cold, and no unsafe `N+1` first-response snapshot/supersequence entry is retained.

We will keep the merged direct gate regressions and exercise real cache-writing model behavior at the scheduler boundary rather than faking cache semantics.

Thump604:
Follow-up delivered as #702. The scheduler-level real-cache harness found a remaining pre-existing interaction after #683: an identical prompt could reuse the completion entry as a trimmed `supersequence`, then `_schedule_waiting()` replayed `prompt[-1]` into state already covering all `N` tokens. The deterministic cache-writing model produced divergent output. #702 adds both requested ArraysCache/KVCache parity regressions and narrowly fails closed for full-coverage `supersequence` matches alongside the existing `exact` guard. Focused result: 48 passed.

janhilgard:
Harness added: `tests/test_prefix_cache_real_models.py`, at the scheduler boundary, real models, no fake cache semantics. It is in the Apple CI list alongside the unit file.

**Case 2 (plain `KVCache`)** — `mlx-community/Qwen3-0.6B-8bit`, every layer a `KVCache`. Cold run, then an identical prompt: output matches cold, and crucially **no entry of exactly `N+1`**. Mutation-checked — removing the gate produces `Entries: [321, 344]` for a 320-token prompt and the test names the 321 as the first-response snapshot.

Worth correcting something in my own first attempt: I initially asserted "no entry longer than N", which failed on the legitimate `prompt+output` entry at `N+24`. That entry is the useful safe one for a trimmable cache. The assertion now targets `N+1` specifically.

**Case 1 (recurrent)** — `LiquidAI/LFM2.5-350M-MLX-8bit`, which gives **10 `ArraysCache` + 6 `KVCache`** layers, so it exercises the mixed topology rather than a synthetic one. First turn, next-turn prompt built from the reply, continuation compared against a cold run, plus an explicit assertion that no entry is keyed by exactly the prompt length. Mutation-checked — reverting the key to `prompt_ids` fails it with "keyed by exactly the 309 prompt tokens … one token short".

## Two things you should know before re-reviewing

**The fail-closed branch is not covered end to end, and I could not make it so.** On a *hybrid* model `_cache_coverage` reads an offset from the attention layers, so `covered is None` never happens — the mutation restoring the `prompt_ids` fallback passes this harness. Reaching it needs a **pure** recurrent model, and no published mamba2 MLX checkpoint loads: `mlx-community/mamba2-370m-hf`, `mamba2-2.7b-4bit` and `-8bit` all omit `intermediate_size`, which `mlx_lm.models.mamba2.ModelArgs` requires, so `load()` raises. I could derive it locally as `expand * hidden_size`, but then the test would depend on a hand-patched checkpoint that CI does not have, which is worse than an honest gap. The branch stays covered by the direct unit tests (`TestUnknownCoverageFailsClosed`, real `ArraysCache`, flat and nested).

That mlx-lm cannot load the published mamba2 MLX models looks like its own upstream bug; happy to file it separately.

**One of my own tests was passing for the wrong reason and I nearly shipped it.** With the short prompt I first used, `MemoryCacheConfig.min_prefix_tokens` (128) meant *nothing was ever stored*, so both comparisons were `"" == ""`. I added `test_harness_actually_generates_and_caches`, which asserts generation produced text, the prompt clears the threshold, and the cache is non-empty — it failed with `assert []` and is what caught it. The prompt is now 320 tokens with a comment saying why.

Behaviour without the models: **3 skipped**, so an offline checkout does not fail. Models are overridable via `VLLM_MLX_TEST_HYBRID_MODEL` / `VLLM_MLX_TEST_KV_MODEL`.

Repo suite: **2323 passed**. Lint, black and YAML parse clean.

waybarrios:
following the discussion here: https://github.com/waybarrios/vllm-mlx/pull/702

PR #684, opened 2026-08-08
fix(batched): load the model on the thread that steps it

janhilgard:
Split out of #676 on review.

> ⚠️ **Stacked on #679.** This needs `run_blocking_startup_work(executor=...)` from that PR, so the branch is based on it and the diff shows its commit until it merges. **Only `vllm_mlx/engine/batched.py` and `vllm_mlx/engine_core.py` (plus the new test) are this PR** — 64 lines. It cleans up by itself once #679 lands.

## Problem

MLX streams exist only in the thread that created them, and `BatchGenerator` captures `generation_stream` into `self._stream` when it is built. So the thread that loads the model has to be the thread that drives `scheduler.step`.

`BatchedEngine` loaded inline on the event loop (issue #407) while stepping ran on `engine_core`'s own worker, so the two never matched.

**The symptom does not look like a stream problem, which is why this took so long to find.** `batch_generator.next()` raises `There is no Stream(gpu, N) in current thread`, `EngineCore` catches it and falls back to stepping on the model thread, hits the same error there, and — because that fallback fires only once — then spins on the error. What an operator sees is:

```
running=1, steps=4_182_339, tokens=0
```

I diagnosed that as a scheduler hang and disabled continuous batching over it. It was not the scheduler: the same weights run fine through `BatchGenerator` directly at 67 tok/s @ C=4.

## Changes

`BatchedEngine` owns a single `engine-core` thread, loads through it, and hands it to `AsyncEngineCore`. `EngineCore` steps on a supplied worker and **does not shut it down** — that thread owns the loaded model and outlives the engine loop, and closing it would take the model's streams with it while the engine still holds the model. Without a supplied worker it still creates and retires its own, unchanged.

`getattr` rather than attribute access for `_external_generation_worker`, because tests and older callers build `EngineCore` without going through `__init__`.

## Verification

DeepSeek-V4-Flash-0731 (283.8B MXFP4, M3 Ultra), which could not emit a single token through the batched path before this: **4/4 concurrent requests, 54.3 tok/s aggregate** against ~31 single-stream.

`tests/test_batched_engine_owner_thread.py` drives the engine's real `start()` rather than replicating its load call in the test body. That distinction matters and I got it wrong first: my initial version replicated the load, and reverting `start()` to the unpinned pool still passed. Confirmed by mutation on the current version — dropping `executor=` from the load, and shutting down a caller-supplied worker, are each caught by a distinct test.

Repo suite: **2280 passed**. The four local failures I see are reproduced on pristine `main` (three Python 3.14 issues, one missing `ffmpeg`).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,6 +83,7 @@ jobs:
             tests/test_anthropic_models.py \
             tests/test_anthropic_adapter.py \
             tests/test_dependency_floors.py \
+            tests/test_simple_engine_thread_pinning.py \
             tests/test_harmony_parsers.py \
             tests/test_endpoint_model_policies.py \
             tests/test_truncation.py \
--- a/tests/test_batched_engine_owner_thread.py
+++ b/tests/test_batched_engine_owner_thread.py
@@ -1,0 +1,332 @@
+# SPDX-License-Identifier: Apache-2.0
+"""BatchedEngine must load the model on the thread that later steps it.
+
+MLX streams exist only in the thread that created them, and ``BatchGenerator``
+captures ``generation_stream`` into ``self._stream`` when it is built. So a
+model loaded on one thread cannot be stepped from another: the first batched
+step raises "There is no Stream(gpu, N) in current thread".
+
+That failure does not look like a stream problem from the outside. ``EngineCore``
+catches it, falls back to stepping on the model thread, hits the same error
+there, and — because that fallback fires only once — then spins on the error.
+What an operator sees is ``running=1``, the step counter climbing into the
+millions, and not one token emitted.
+
+These tests record threads rather than touching MLX, so they run anywhere.
+Fakes follow the convention in ``test_engine_core_thread_streams.py``.
+"""
+
+import asyncio
+import threading
+from concurrent.futures import ThreadPoolExecutor
+from types import SimpleNamespace
+
+import pytest
+
+
+async def _noop(marker: list) -> None:
+    marker.append(True)
+
+
+class _SchedulerOutput:
+    outputs = []
+    finished_request_ids = []
+
+
+class _FakeScheduler:
+    batch_generator = SimpleNamespace(_partial=None)
+
+    def __init__(self, engine, steps: int = 3):
+        self._engine = engine
+        self._steps = steps
+        self.calls = 0
+        self.step_threads: list[int] = []
+
+    def has_requests(self):
+        return self.calls < self._steps
+
+    def step(self):
+        self.step_threads.append(threading.get_ident())
+        self.calls += 1
+        if self.calls == self._steps:
+            self._engine._running = False
+        return _SchedulerOutput()
+
+    def _close_batch_generator(self):
+        pass
+
+
+def _bare_engine_core(monkeypatch, worker=None):
+    from vllm_mlx.engine_core import EngineConfig, EngineCore
+
+    engine = object.__new__(EngineCore)
+    engine.config = EngineConfig(step_interval=0, stream_interval=1)
+    engine._running = True
+    engine._steps_executed = 0
+    engine._output_collectors = {}
+    engine._stream_states = {}
+    engine._finished_events = {}
+    if worker is not None:
+        engine._external_generation_worker = worker
+    engine.scheduler = _FakeScheduler(engine)
+    monkeypatch.setattr(
+        "vllm_mlx.engine_core.bind_generation_streams", lambda *a, **k: None
+    )
+    return engine
+
+
+def test_generation_worker_is_one_reused_thread():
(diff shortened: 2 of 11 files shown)


Thump604 (changes requested):
One blocker: this pins every BatchedEngine load to the generation worker, including MLLM, but MLLMScheduler still calls step on the event-loop thread and explicitly relies on MLLM arrays being loaded there. ResidencyManager now selects the worker too, so MLLM load and step can land on different MLX stream owners. The new test sets is_mllm false, so it misses this. Please either keep MLLM loads and lifecycle on the event-loop thread or move MLLM stepping and cleanup onto the worker, then add a test proving MLLM load and step use the same thread.

janhilgard:
You were right, and this would have broken the configuration we actually run. Fixed and pushed.

I verified the claim before touching anything rather than taking it on faith, because it decides whether the whole approach is wrong or just the branching:

- `mllm_scheduler.py:919` calls `self.step()` directly in `_process_loop`, no executor hop.
- `test_mllm_scheduler_runs_steps_on_model_load_thread` already pins that: `step_threads[0] == main_thread`.
- My `start()` called `run_blocking_startup_work(..., executor=self._generation_worker())` unconditionally, *before* the `if self._is_mllm:` branch.

Measured on a fake MLLM engine:

```
event loop (where MLLMScheduler.step runs) = 8314968256
MLLM model loaded on                       = 6187855872
```

Which is the exact failure this PR exists to fix, inverted.

**Why I missed it:** my verification ran against DeepSeek-V4, which is not MLLM, so the MLLM path was never exercised — and the test I wrote sets `is_mllm=False`, so it could not have caught it either. Both our production services run `--mllm --continuous-batching`, so this would have landed on us first.

## The fix

I took your first option. `_start_mllm` never touches AsyncEngineCore or the generation worker at all — only `_start_llm` does — so the policy is genuinely per-path rather than a special case:

```python
def _model_load_executor(self) -> ThreadPoolExecutor | None:
    if self._is_mllm:
        return None      # MLLMScheduler steps on the event loop
    return self._generation_worker()
```

`start()` and `ResidencyManager._prepare_engine_start` both read it, so the two start paths cannot disagree — thank you for catching the lifecycle half, I would have fixed only the engine and left that hole open. MLLM with an overridden `prepare_for_start` keeps the old off-loop behaviour, since test doubles may block and do not own MLX state.

Moving MLLM stepping onto the worker instead is the more interesting option and probably the better end state, but it changes MLLMScheduler's threading model and belongs in its own PR with its own evidence.

## Tests

Two new ones, both driving the real entry points:

- `test_mllm_load_and_step_share_the_event_loop_thread` — runs `start()`, then `MLLMScheduler._process_loop`, and asserts the recorded threads are the same set. Also asserts no generation worker is created for a path that never uses one.
- `test_residency_manager_honours_the_per_path_load_thread` — MLLM lands on the loop, non-MLLM does not.

One thing worth flagging from writing them: my first version set `engine.prepare_for_start` as an *instance* attribute, which makes `_uses_default_prepare_for_start()` return False and quietly routes the fake down the "overridden prepare may block" branch instead of the real default one. It failed for the wrong reason and would have passed for the wrong reason just as easily. It now patches the class.

Mutation-checked, each caught by a distinct test:

| mutation | caught by |
|---|---|
| `_model_load_executor` pins MLLM to the worker | both new tests |
| lifecycle ignores `_model_load_executor` | the ResidencyManager test |
| drop `executor=` from the non-MLLM load | `test_model_load_and_stepping_share_one_thread` |
| loop shuts down a caller-supplied worker | `test_supplied_worker_outlives_the_engine_loop` |

Repo suite: 2282 passed. The four failures I see locally reproduce on pristine `main`.

Thump604 (approved):
The blocker is fixed. MLLM load and scheduler stepping now stay on the event-loop thread, non-MLLM load and EngineCore stepping stay on the worker, and ResidencyManager follows the same policy. The new tests drive both real entry points. Exact-head focused suite: 65 passed, 2 deselected. No remaining blocking findings.

janhilgard:
Rebased — the branch was conflicting against `main` and blocking merge despite the approval.

This branch carries #679's commit underneath, so I rebased it onto the resolved #679 rather than replaying the same conflict twice: `git rebase --onto fix/pin-mlx-generation-thread 99c4409`. Its own commit applied cleanly; the conflict resolution and its rationale are described in [#679](https://github.com/waybarrios/vllm-mlx/pull/679#issuecomment-5306665833) (the `finish_reason` length/stop ladder from `main` kept alongside the restructured loop, and `_acquire_generation_slot` kept on the MLLM route so media requests do not lose fail-fast admission).

Full suite on this branch: **2423 passed, 4 failed** — the same 4 that fail on clean `upstream/main` at `61c78dc`, so this adds none. Lint and black clean under the CI invocation.

Merge order: #679 first, then this one.

PR #688, opened 2026-08-08
fix(scheduler): make --max-kv-size actually bound the KV cache

janhilgard:
`--max-kv-size` is a no-op. Every generation runs on an unbounded cache no matter what the flag says, while startup prints `Max KV size: 104856 (RotatingKVCache)`.

Found while fixing #683 — a snapshot destination derived from `SchedulerConfig.max_kv_size` mismatched the live cache, which is how I noticed the config never described it.

## Two independent breaks

Either one was enough on its own.

**1. The scheduler's bounded cache is rejected by its own validator.** It builds the cache and then hands it to `_validate_cache`, which returns False for any layer whose `keys` are still None — true of *every* freshly created cache, bounded or not. That validator exists for entries restored from the prefix cache, which have populated keys. So `cache_to_use` went straight back to None.

**2. `BatchGenerator` was never told.** It builds the cache for any sequence inserted without one, and its `max_kv_size` stayed None, so the fallback was unbounded.

Traced through a real scheduler run with `max_kv_size=512`:

```
make_prompt_cache(max_kv_size=512)  -> RotatingKVCache      # scheduler builds it
_validate_cache(RotatingKVCache)    -> False                # ...and rejects it
make_prompt_cache(max_kv_size=None) -> KVCache              # generator's fallback
BatchGenerator.max_kv_size          -> None
```

## Fix

The generator is given `max_kv_size` directly, and the scheduler no longer builds one itself. That removes a duplicated rule rather than repairing it twice — and the copy that remains is the correct one.

`make_prompt_cache(model, max_kv_size=N)` **silently ignores `max_kv_size` whenever the model defines `make_cache`**, which most current architectures do:

```python
if hasattr(model, "make_cache"):
    return model.make_cache()      # max_kv_size never consulted
```

`BatchGenerator._make_new_cache` post-processes the model's own cache instead, converting `KVCache` layers to `RotatingKVCache`, which bounds those models correctly. Keeping the scheduler's copy would have left the flag broken on exactly the architectures most likely to need it.

## Verification

Measured before and after on Qwen3-0.6B, reading the live cache out of the running batch:

| | `max_kv_size=0` | `max_kv_size=64` |
|---|---|---|
| before | `{'KVCache': None}` | `{'KVCache': None}` |
| after | `{'KVCache': None}` | `{'RotatingKVCache': 64}` |

`tests/test_scheduler_max_kv_size.py` covers the passthrough, that `0` stays unbounded rather than becoming `RotatingKVCache(max_size=0)`, that `_validate_cache` rejects fresh caches regardless of topology (the behaviour that caused this), and that `make_prompt_cache` ignores the bound for `make_cache` models — that last one is written so it fails loudly if upstream ever fixes it, since the scheduler could then stop relying on post-processing.

Mutation-checked: removing the `max_kv_size=` argument fails the passthrough test. Repo suite: **2300 passed** (four local failures reproduce on pristine `main` — three Python 3.14, one missing `ffmpeg`).

## Heads-up for anyone already passing the flag

It starts taking effect. A context longer than the configured size will now rotate — dropping the oldest KV — where it previously grew without bound. That is the flag's purpose, but it is a live behaviour change for any deployment that set it and never saw it apply. Ours are in that position: two servers running `--max-kv-size 104856` and `65536` against `--max-request-tokens 262144`, which have been silently unbounded this whole time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -144,6 +144,7 @@ jobs:
             tests/test_paged_cache.py \
             tests/test_prefix_cache_untrimmable.py \
             tests/test_scheduler_max_kv_size.py \
+            tests/test_ssd_cache.py \
             tests/test_qwen35_mtp_patch.py \
             tests/test_mllm_continuous_batching.py \
             tests/test_mllm_cache.py \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,6 +143,7 @@ jobs:
             tests/test_server.py \
             tests/test_paged_cache.py \
             tests/test_prefix_cache_untrimmable.py \
+            tests/test_scheduler_max_kv_size.py \
             tests/test_qwen35_mtp_patch.py \
             tests/test_mllm_continuous_batching.py \
             tests/test_mllm_cache.py \
--- a/tests/test_batching.py
+++ b/tests/test_batching.py
@@ -29,6 +29,29 @@ from vllm_mlx.scheduler import (
 mlx_generate = importlib.import_module("mlx_lm.generate")
 
 
+class _MergeableFakeCache:
+    """Minimal per-request and batched cache for legacy prefill tests."""
+
+    def __init__(self):
+        self.state = mx.array([0])
+
+    def empty(self):
+        return True
+
+    @classmethod
+    def merge(cls, _caches):
+        return cls()
+
+    def prepare(self, **_kwargs):
+        return None
+
+    def finalize(self):
+        return None
+
+    def extract(self, _idx):
+        return self
+
+
 class TestRequest:
     """Tests for Request class."""
 
@@ -240,17 +263,6 @@ class TestSchedulerBasic:
     def test_chunked_prefill_accepts_prompt_checkpoints(self, monkeypatch):
         """Chunked prefill must match mlx-lm's 7-field prompt tuples."""
 
-        class FakeCacheEntry:
-            def empty(self):
-                return True
-
-        class FakePromptCache:
-            def __init__(self):
-                self.state = mx.array([0])
-
-            def finalize(self):
-                return None
-
         class FakeStats:
             prompt_tokens = 0
             prompt_time = 0.0
@@ -266,7 +278,7 @@ class TestSchedulerBasic:
                         7,
                         [1, 2, 3, 4, 5],
                         16,
-                        [FakeCacheEntry()],
+                        [_MergeableFakeCache()],
                         None,
                         [None],
                         2,
@@ -288,11 +300,6 @@ class TestSchedulerBasic:
             "_left_pad_prompts",
             lambda prompts, max_length=None: mx.array(prompts),
         )
-        monkeypatch.setattr(
-            mlx_generate,
-            "_make_cache",
-            lambda _model, _padding, _max_kv_size=None: [FakePromptCache()],
-        )
 
         batch_gen = FakeBatchGenerator()
         _install_chunked_prefill(batch_gen, budget=4)
@@ -306,20 +313,6 @@ class TestSchedulerBasic:
 
     def test_chunked_prefill_invokes_checkpoint_callback(self, monkeypatch):
         """prompt_checkpoint_callback must fire after finalization."""
-
-        class FakeCacheEntry:
-            def empty(self):
-                return True
-
-        class FakePromptCache:
-            def __init__(self):
-                self.state = mx.array([0])
-
-            def finalize(self):
-                return None
-
-            def extract(self, idx):
-                return self
 
         class FakeStats:
             prompt_tokens = 0
@@ -347,7 +340,7 @@ class TestSchedulerBasic:
                         7,
                         [1, 2, 3],
                         16,
-                        [FakeCacheEntry()],
+                        [_MergeableFakeCache()],
                         None,
                         [None],
                         2,
@@ -379,11 +372,6 @@ class TestSchedulerBasic:
             "_left_pad_prompts",
             lambda prompts, max_length=None: mx.array(prompts),
         )
-        monkeypatch.setattr(
-            mlx_generate,
-            "_make_cache",
-            lambda _model, _padding, _max_kv_size=None: [FakePromptCache()],
-        )
 
         batch_gen = FakeBatchGenerator()
         batch_gen.stop_tokens = {99}
@@ -404,20 +392,6 @@ class TestSchedulerBasic:
 
     def test_chunked_prefill_replays_checkpoint_tail_before_step(self, monkeypatch):
         """checkpoint tails >1 must be replayed after finalize before _step."""
-
-        class FakeCacheEntry:
-            def empty(self):
-                return True
-
-        class FakePromptCache:
-            def __init__(self):
-                self.state = mx.array([0])
-
-            def finalize(self):
-                return None
-
-            def extract(self, idx):
-                return self
 
         class FakeStats:
             prompt_tokens = 0
@@ -447,7 +421,7 @@ class TestSchedulerBasic:
                         7,
                         [1, 2, 3, 4, 5],
                         16,
-                        [FakeCacheEntry()],
+                        [_MergeableFakeCache()],
                         None,
                         [None],
                         2,
@@ -477,11 +451,6 @@ class TestSchedulerBasic:
             "_left_pad_prompts",
             lambda prompts, max_length=None: mx.array(prompts),
         )
-        monkeypatch.setattr(
-            mlx_generate,
-            "_make_cache",
-            lambda _model, _padding, _max_kv_size=None: [FakePromptCache()],
-        )
 
         batch_gen = FakeBatchGenerator()
         _install_chunked_prefill(batch_gen, budget=2)
@@ -505,20 +474,6 @@ class TestSchedulerBasic:
         self, monkeypatch
     ):
         """Chunked prefill should tolerate missing private mlx_lm.generate exports."""
-
-        class FakeCacheEntry:
-            def empty(self):
-                return True
-
-        class FakePromptCache:
-            def __init__(self):
-                self.state = mx.array([0])
-
-            def finalize(self):
-                return None
-
-            def extract(self, idx):
-                return self
 
         class FakeStats:
             prompt_tokens = 0
@@ -544,7 +499,7 @@ class TestSchedulerBasic:
                         7,
                         [1, 2, 3],
                         16,
-                        [FakeCacheEntry()],
+                        [_MergeableFakeCache()],
                         None,
                         [None],
                         2,
@@ -571,15 +526,11 @@ class TestSchedulerBasic:
 
         monkeypatch.delattr(mlx_generate, "Batch", raising=False)
         monkeypatch.delattr(mlx_generate, "_lazy_extract_cache", raising=False)
+        monkeypatch.delattr(mlx_generate, "_make_cache", raising=False)
         monkeypatch.setattr(
             mlx_generate,
             "_left_pad_prompts",
             lambda prompts, max_length=None: mx.array(prompts),
-        )
-        monkeypatch.setattr(
-            mlx_generate,
-            "_make_cache",
-            lambda _model, _padding, _max_kv_size=None: [FakePromptCache()],
         )
 
         batch_gen = FakeBatchGenerator()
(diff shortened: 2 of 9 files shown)


Thump604 (changes requested):
Two blockers at 06f989e. _make_new_cache only converts flat KVCache layers; KVCache children inside CacheList stay unbounded, so the flag is still a no-op on the nested DeepSeek topology that exposed this. Please recurse through CacheList while preserving native non-KV cache types, with a mixed nested regression. Also, max_kv_size or None forwards negative values and creates RotatingKVCache(max_size=-1). Reject negatives or only forward values greater than zero, with a test. The flat cases are good, but these need to be covered before merge.

janhilgard:
Both fixed.

**Nested layers.** You are right, and the correction lands on my reasoning rather than an oversight: I dropped the scheduler's own bounding specifically *because* `BatchGenerator._make_new_cache` looked like the more correct rule. It is more correct than what it replaced, but only for flat topologies — it rewrites `KVCache` layers and stops, so a `KVCache` inside a `CacheList` stays unbounded, and the architectures that nest are the ones that need the bound.

So the scheduler bounds the cache itself now, recursing through containers and leaving anything that is not a plain `KVCache` alone — a rotating, quantized or chunked layer carries its own behaviour and swapping it out would break the model. `BatchGenerator` still gets `max_kv_size` as a backstop for sequences inserted without a cache.

That required fixing the ordering too. The bounded cache is built **after** `_validate_cache`, not before: the validator rejects any layer whose `keys` are still None, which is every freshly created cache, so building first meant it was discarded on every request.

**Negatives.** `self.config.max_kv_size or None` forwarded `-1` intact. Now only a positive value is treated as a bound; `0` and negatives mean unbounded, and a negative logs a warning rather than passing silently.

Measured end to end through the scheduler:

```
max_kv_size=  0 : bg=None  live={'KVCache': None}
max_kv_size= -1 : bg=None  live={'KVCache': None}    + "Ignoring negative max_kv_size=-1"
max_kv_size= 64 : bg=64    live={'RotatingKVCache': 64}
```

Nine new tests: nested plain layers bounded, existing windows not retuned, `QuantizedKVCache`/`ChunkedKVCache` preserved by identity, negatives and zero rejected at both the helper and the BatchGenerator boundary. One of them asserts that mlx-lm's flat rule *would* miss the nested case, so it fails loudly if BatchGenerator ever learns to recurse and the scheduler's copy becomes redundant.

Mutation-checked: disabling the recursion fails two, allowing negatives through fails three.

Repo suite: 2309 passed.

Thump604 (changes requested):
The two reported blockers are fixed, but the reordering introduced a correctness regression in the default path. When a restored cache fails validation and `max_kv_size=0`, `cache_to_use` becomes `None` but the request reset now only runs inside the bounded-cache branch. The scheduler can then insert `remaining_tokens` with no cache and leave `cached_tokens`/`remaining_tokens` stale. I reproduced a four-token prompt inserting only its one-token suffix after validation failed. Please restore the full request reset immediately when validation rejects a cache, independent of `max_kv_size`, and add that regression. Keeping changes requested.

janhilgard:
You are right, and it is a regression I introduced in the last round rather than something pre-existing. Moving the bounded-cache construction below the validation left the four reset lines attached to the *bounded* branch instead of the validation-failure one, so with `max_kv_size=0` a rejected cache cleared `cache_to_use` and nothing else — `cached_tokens` and `remaining_tokens` stayed as the hit had left them, and the scheduler prefilled only the suffix.

The reset is back inside the validation-failure block, where it runs whatever `max_kv_size` is.

The regression test drives `_schedule_waiting` on a real `Scheduler` and a real `Request` rather than replicating the branch, which matters here: the bug was *where* the reset sits, so any test that re-implements the logic would have passed against the broken code. Parametrised over `max_kv_size` 0 and 512, since the whole point is that the two must behave the same.

Mutation-checked — putting the reset back in the bounded branch fails it with `cached_tokens` 3 instead of 0.

Repo suite: 2311 passed, lint and black clean under the exact CI invocation.

I should have caught this myself: I checked that the reordering fixed what it was meant to fix and not what it might have moved. Thanks for reproducing it concretely.

Thump604 (changes requested):
The invalid-cache reset is fixed and the new real Scheduler regression covers both bounded and default paths. One remaining issue: negative `max_kv_size` is still accepted and `_bounded_kv_size()` logs from the scheduling hot path. Two uncached requests produced the same warning three times. Please reject/normalize it once during config or CLI construction, or warn once outside request scheduling, with a multi-request log-count regression. We should not allow a bad startup value to flood production logs per request.

janhilgard:
Right — I fixed the value and created a second problem, which is that a single bad startup flag then logs on every scheduling pass.

Normalization moved to `SchedulerConfig.__post_init__`, so it happens once at construction and covers every path in including the CLI. `_bounded_kv_size()` is now a pure accessor that never logs; it runs per request, so it has no business logging at all. Non-integers are normalized too, since `"512"` from a config file would otherwise reach `RotatingKVCache(max_size="512")`.

```
input     -1 -> config.max_kv_size=0   + one warning
input      0 -> 0
input    512 -> 512
input  '512' -> 0                      + one warning
```

The log-count regression asserts exactly one warning at construction and zero across five scheduling calls.

Worth flagging how that test went, because the first version of it was not doing its job: it counted only warnings whose message contained `max_kv_size`, so a mutation that logged from the hot path with different wording passed it. It now counts **every** record from the scheduling path — a hot-path warning is the same production problem whatever it says. With that, the mutation fails at 5 warnings instead of 0.

Repo suite: 2313 passed, lint and black clean under the exact CI invocation.

Thump604 (approved):
Re-reviewed the updated head. Invalid max_kv_size values are now normalized once at config construction, the hot-path accessor is side-effect free, nested cache containers preserve specialized cache types while bounding plain KVCache children, and rejected restored caches reset to a full prefill independently of the configured bound. Focused local run: 19 passed. CI is green. No remaining blocker.

Thump604 (changes requested):
The fresh-cache path is fixed, but a valid restored prefix cache bypasses the configured bound entirely.

`cache_to_use = request.prompt_cache` is validated and then retained; `_bound_cache_layers(...)` only runs when `cache_to_use is None`. Prefix caches are persisted to disk and can outlive both this upgrade and changes to `--max-kv-size`, so two concrete cases remain incorrect:

- an unbounded `KVCache` restored after upgrading continues unbounded even with `--max-kv-size 64`;
- a `RotatingKVCache(max_size=64)` restored after changing the flag to `0` or another size retains the stale window.

That means the flag can still be a no-op (or apply the wrong bound) on prefix hits. Please make restored-cache compatibility include the effective KV bound/topology, or reject incompatible restored entries and fall back to a full prefill/new cache. A regression should schedule a request with a populated, otherwise-valid restored cache under a mismatched bound and assert it is not reused unchanged.

janhilgard:
Right again, and this one I would not have found on my own — I was thinking about caches the scheduler *creates* and never about the ones it *restores*.

Reproduced all three shapes you describe:

```
KVCache restored under --max-kv-size 64     _validate_cache=True -> used unbounded
Rotating(64) restored at max_kv_size=0      _validate_cache=True -> kept the stale window
Rotating(64) restored at max_kv_size=128    _validate_cache=True -> kept the wrong window
```

I took your second option — reject and fall back — rather than re-bounding the restored entry. Re-bounding a *populated* cache means re-laying its contents into a ring, which is exactly the kind of rewrite that desynchronizes layers; #691 exists because that went wrong elsewhere. Paying one prefill is much cheaper than generating against a window the entry was not built for.

`_restored_cache_matches_kv_bound()` recurses through containers and checks each layer against the effective bound: a `RotatingKVCache` must match `max_size` exactly, a plain `KVCache` is only acceptable when there is no bound, and anything else is left alone since it carries its own behaviour. One bad child rejects the whole entry.

Regression drives `_schedule_waiting` with a populated, otherwise-valid restored cache under a mismatched bound and asserts the request falls back to a full prefill — `prompt_cache` cleared, `cached_tokens` 0, `remaining_tokens` the whole prompt. Six unit cases cover both directions plus the nested mix. Mutation-checked: disabling the compatibility check fails the scheduler regression with `cached_tokens` 3 instead of 0.

Repo suite: 2320 passed, lint and black clean under the exact CI invocation.

Worth noting for whoever picks this up next: persisted entries surviving a flag change is a general hazard, not specific to `max_kv_size`. #691 bumped `_CACHE_PERSIST_VERSION` for a semantics change; a bound change is not a semantics change, so the version does not move and the entries stay loadable — which is why this needed a per-entry check rather than a version gate.

waybarrios:
I found three issues that should be fixed before merging:

1. Native rotating caches are treated as operator-bounded caches:

```python
if isinstance(layer, RotatingKVCache):
    if effective_max_kv_size is None or layer.max_size != effective_max_kv_size:
        return False
```

Models like Llama and Gemma create rotating layers with their own window, so a valid cache can be rejected the next time it is reused.

2. Rejected entries are cleared only from the request:

```python
request.prompt_cache = None
request.prompt_cache_key = None
```

The stale entry remains in the shared prefix cache, so every matching request can repeat the same false cache hit and full prefill.

3. `tests/test_scheduler_max_kv_size.py` is not included in the Apple Silicon CI test list, so its 21 new tests are not currently executed.

waybarrios (changes requested):
Please check my previous comment for more details

janhilgard:
@waybarrios — all three confirmed and fixed. Rebased on `main` now that #683 has landed.

**1. Native rotating caches were treated as operator-bounded.** Reproduced:

```
model creates:                      [RotatingKVCache(1024), RotatingKVCache(1024)]
_bounded_kv_size():                 None
_restored_cache_matches_kv_bound(): False   <- should be True
```

You are right about the shape of the mistake, and it is the same one #691 exists for: conflating the model's own topology with the operator's configuration. A sliding-window model would have had every restored entry rejected on every reuse — a permanent cache miss on exactly the architectures that benefit most.

The comparison is now against **the cache this scheduler would construct for a fresh request** — `make_prompt_cache(model)`, then `_bound_cache_layers` if a bound is set — rather than against `max_kv_size` in isolation. That answers the actual question ("would we build this the same way now?") instead of a proxy for it, and it handles both directions without a special case. Built once and cached, since it only depends on the model and the config.

One thing that fell out of it: a restored entry whose **layer count** differs from the model is now rejected outright. My first version returned "compatible" there, which the older tests silently relied on.

**2. Rejected entries stayed in the shared cache.** Correct — clearing `request.prompt_cache` leaves the entry for every later request with the same prefix, repeating the false hit and full prefill forever, with nothing to remove it. `_evict_incompatible_entry()` now calls `remove()` on the shared cache.

**3. `tests/test_scheduler_max_kv_size.py` not in Apple CI.** Confirmed — added. That is twice now on adjacent PRs; the enumerated list is easy to forget when adding a file.

Ten new tests: native windows accepted without a bound, a *different* native window still rejected, operator bound still applied to plain models, the reference topology built once, and eviction on the real scheduler path.

That last one matters — my first eviction test called the helper directly, and the mutation removing the **call site** passed it. It now drives `_schedule_waiting` and fails when the call is removed. Mutation-checked overall: comparing against `max_kv_size` instead of the topology, and dropping the eviction, each fail their own test.

Repo suite: **2431 passed**, lint and black clean under the exact CI invocation.

Thump604 (approved):
Re-reviewed f88ed861. Restored caches now match the fresh model topology, so native sliding-window caches remain reusable while entries built under a different bound are rejected and evicted. The reset path falls back to a full prefill, and the Apple Silicon job runs the new coverage. Focused local run: 33 passed. Current-head CI is green. No remaining blocker.

Thump604 (dismissed):
Three blockers remain at `f88ed861`:

1. The chunked-prefill path discards the scheduler-supplied bounded cache. A fresh empty cache makes `is_cached` false, then `_make_cache(..., max_kv_size)` rebuilds it through the dependency path that ignores the bound for models with `make_cache`. Please cover a `make_cache` model with `chunked_prefill_tokens`.

2. Restored topology compares only type and `max_size`. Rotating caches with the same window but different `keep` are accepted even though their layouts differ. Compare the topology-defining attributes and add a differing-`keep` regression.

3. Incompatible-entry eviction does not retain the actual matched key. LCP/supersequence hits can leave the stale RAM entry behind, and an SSD-promoted mismatch can be promoted again because only its RAM copy is removed. Track the matched key and evict both tiers.

Current checks do not cover these combinations. Keeping changes requested.

waybarrios (approved):
Re-reviewed the current head. The remaining cache lifecycle blockers are resolved, the Apple Silicon regressions pass on Python 3.11 and 3.13, and the full CI suite is green. Approved.

Issue #689, opened 2026-08-08
MLLM completion cache can store desynchronized rotating KV state


Thump604:
## Summary

`MLLMBatchGenerator._maybe_store_prefix_cache()` rebuilds a prompt-only cache at request completion by trimming generated tokens from the finished cache. That cannot recover a saturated `RotatingKVCache`: generated tokens have already overwritten prompt KV in the ring.

When the trim reaches the window size, `_trim_cache_offset()` clears the rotating layer and resets its offset to 0 while full-attention layers rewind to the prompt length. The mixed cache is still stored under the prompt tokens. Exact and normal prefix fetches return it without a trimmability check.

This is separate from #680, which guards partial reuse while fetching an existing entry. It does not guard creation of an already-invalid entry here.

## Model-free reproduction

Current `main` at `b99877689a559e73be378e4ff3b67240a2740083`:

```python
from unittest.mock import MagicMock
import mlx.core as mx
from mlx_lm.models.cache import KVCache, RotatingKVCache
from vllm_mlx.memory_cache import (
    MemoryAwarePrefixCache,
    MemoryCacheConfig,
    _trim_cache_offset,
)

prompt = [1, 2, 3, 4]
output_count = 10
max_size = 8
total = len(prompt) + output_count

full = KVCache()
full.keys = mx.arange(total).reshape(1, 1, total, 1)
full.values = full.keys
full.offset = total

rotating = RotatingKVCache(max_size=max_size, keep=0)
rotating.keys = mx.arange(max_size).reshape(1, 1, max_size, 1)
rotating.values = rotating.keys
rotating.offset = total
rotating._idx = max_size

prompt_cache = _trim_cache_offset([full, rotating], output_count)
print([layer.offset for layer in prompt_cache])

cache = MemoryAwarePrefixCache(
    MagicMock(),
    MemoryCacheConfig(max_memory_mb=64, max_entries=10, min_prefix_tokens=1),
)
print(cache.store(prompt, prompt_cache))
for request in (prompt, prompt + [5]):
    fetched, remaining = cache.fetch(request)
    print([layer.offset for layer in fetched], remaining)
```

Output:

```
[4, 0]
True
[4, 0] []
[4, 0] [5]
```

Both exact and prefix hits serve the desynchronized cache.

## Suggested fix

[...]

PR #691, opened 2026-08-08
fix(mllm): harden prefix cache ownership and rewind safety

Thump604:
## Summary

MLLM completion caching could rewind saturated sliding-window state after the prompt tokens had already been overwritten, then store that cache under the prompt key. Exact and prefix hits could later reuse desynchronized KV state. Nested `CacheList` wrappers were also copied by reference, so live generation could mutate stored entries.

This change makes cache reuse type-aware, isolated, and fail-closed:

- recursively clones both `mlx_lm` and `mlx_vlm` cache families while sharing immutable MLX tensors;
- preserves type-specific state for `ChunkedKVCache` and rotating caches;
- uses each cache implementation's native trim contract and requires the full rewind;
- rejects saturated or inconsistent rotating state instead of storing or reusing it;
- applies the same policy to canonical and interleaved exact, partial, and completion-store paths;
- preserves absolute rotating offsets while normalizing oversized plain rotating buffers;
- rejects nested `CacheList` writes until `MemoryAwarePrefixCache` can account their memory recursively, while retaining safe reads;
- bumps persisted cache semantics to v4 so pre-fix v3 snapshots are discarded after upgrade.

The hot path copies Python cache wrappers only. MLX arrays remain shared, so this does not duplicate KV tensor storage.

## Validation

- Full suite: 2313 passed, 24 skipped, 23 deselected
- Focused cache/MLLM suite: 156 passed, 4 deselected
- Combined with #680 exact head: 162 passed, 4 deselected
- Black, Ruff, and `git diff --check`: clean
- Independent review: approved with no remaining P0/P1/P2 findings

## Merge gates

- Depends on #680. Merge #680 first, then rebase and rerun the combined cache suite at the final head.
- Nested cache storage remains intentionally disabled until recursive byte accounting is implemented in the shared memory cache.

Refs #689

### DIFF
--- a/tests/test_memory_cache.py
+++ b/tests/test_memory_cache.py
@@ -555,3 +555,16 @@ class TestGetAvailableMemory:
             # Should return 0 when psutil not available
             # Note: This test may not work as expected due to import caching
             pass
+
+
+def test_load_rejects_v3_cache_after_rewind_semantics_change(tmp_path, caplog):
+    """Caches written before safe MLLM rewind must not survive an upgrade."""
+    import json
+
+    (tmp_path / "index.json").write_text(
+        json.dumps({"version": 3, "model_fingerprint": "", "entries": []})
+    )
+    cache = MemoryAwarePrefixCache(MagicMock(), MemoryCacheConfig(max_memory_mb=1))
+
+    assert cache.load_from_disk(str(tmp_path)) == 0
+    assert "version mismatch: disk=3 current=4" in caplog.text
--- a/tests/test_memory_cache.py
+++ b/tests/test_memory_cache.py
@@ -566,5 +566,13 @@ def test_load_rejects_v3_cache_after_rewind_semantics_change(tmp_path, caplog):
     )
     cache = MemoryAwarePrefixCache(MagicMock(), MemoryCacheConfig(max_memory_mb=1))
 
-    assert cache.load_from_disk(str(tmp_path)) == 0
+    with patch.dict(
+        "sys.modules",
+        {
+            "mlx_lm": None,
+            "mlx_lm.models": None,
+            "mlx_lm.models.cache": None,
+        },
+    ):
+        assert cache.load_from_disk(str(tmp_path)) == 0
     assert "version mismatch: disk=3 current=4" in caplog.text
--- a/tests/test_mllm_continuous_batching.py
+++ b/tests/test_mllm_continuous_batching.py
@@ -851,6 +851,72 @@ if __name__ == "__main__":
 
 
 class TestMLLMBatchGeneratorMTPGuards:
+    def test_process_prompts_rejects_unsafe_exact_rotating_hit(self, monkeypatch):
+        from mlx_lm.models.cache import CacheList, KVCache, RotatingKVCache
+
+        from vllm_mlx.mllm_batch_generator import (
+            MLLMBatchGenerator,
+            MLLMBatchRequest,
+            MLLMBatchStats,
+        )
+
+        class FreshCache:
+            def merge(self, caches):
+                return self
+
+        full = KVCache()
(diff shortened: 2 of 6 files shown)


janhilgard (approved):
Approving. I reproduced the defect, checked the fix, and tested it against the two PRs it sits next to.

**The bug is real and reproduces exactly as #689 describes.** On `main`, a four-token prompt with ten generated tokens and `max_size=8`:

```
after trim by 10:  full.offset=4   rotating.offset=0   (prompt=4)
```

The full-attention layer rewinds to the prompt while the rotating layer collapses to zero, and that mixed cache is stored under the prompt key. This is the creation side of what I raised on #680 — that PR guards partial *reuse* of an existing entry, and I flagged the store path as unguarded without chasing it. Good that it became its own issue rather than being folded in.

Things I liked on reading the diff:

- Using each cache implementation's **native trim contract and requiring the full rewind** rather than approximating it. A partial rewind is the failure mode here, so refusing anything less is the only safe reading.
- Failing closed on saturated or inconsistent rotating state instead of storing it. An entry that cannot be trusted is worse than no entry, since the cost of a miss is a prefill and the cost of a bad hit is a wrong answer.
- `_has_empty_rotating_cache` recursing into nested containers, and being applied on both the canonical and interleaved paths. Getting one and not the other is exactly how this class of bug survives.
- Bumping `_CACHE_PERSIST_VERSION` to 4. Snapshots written before the fix can contain desynchronized state, and silently loading them after upgrade would reintroduce the bug on a machine that had merely restarted.
- Moving the `mlx_lm` import below the version check so a stale cache is rejected even where mlx-lm is absent, with a test that pins it.

Verification on my side: focused suites at your head — `test_prefix_cache.py`, `test_memory_cache.py`, `test_mllm_continuous_batching.py`, `test_specprefill_rotating_cache.py` — **141 passed, 4 deselected**. All nine CI jobs green.

## One thing you can probably drop sooner than you expect

> rejects nested `CacheList` writes until `MemoryAwarePrefixCache` can account their memory recursively

That accounting is already fixed in #683, which is approved and mergeable. `estimate_kv_cache_memory` on `main` cannot see through a container — `CacheList.state` is nested, the two-way unpack raises, and the ValueError is swallowed, so every such entry is booked as **0 bytes** and the byte-based LRU never evicts it. #683 replaces that with a recursive walk. Measured, same two filled `KVCache` layers each way:

| | main | with #683 |
|---|---|---|
| flat | 4096 B | 4096 B |
| nested in `CacheList` | **0 B** | 4096 B |
| nested two levels | — | 4096 B |

So the blocker you are guarding against is gone once #683 lands, and re-enabling nested writes becomes a matter of deleting the guard rather than building the accounting. Worth a follow-up issue referencing this so it does not stay disabled by inertia — nested is the DeepSeek-V4 topology, which is where prefix caching earns the most.

Not asking you to change anything here. Disabling the write while the accounting was broken was the right call at the time you wrote it.

## Merge order

Both merges are clean, which I checked rather than assumed: **#691 + #680** and **#691 + #683** each merge without conflict. Your stated gate of merging #680 first still holds; #683 is independent of both and can land in any order.

Thump604:
Rebased this onto current `main` and reconciled it with the recursive cache accounting that landed in #683.

The old nested-`CacheList` storage guard is now removed: the shared estimator recursively accounts the stored topology, and the updated regression verifies that a nested snapshot is stored, has nonzero estimated bytes, and retains aligned rewound child offsets. Saturated/non-rewindable rotating caches still fail closed through the existing rewind checks.

Validation:
- focused cache/MLLM/speculative-prefill suite: 142 passed
- #683 untrimmable-cache modules: 46 passed
- Runtime repository verification: 630 passed plus auxiliary suites
- Ruff, Black, and `git diff --check`: clean

Remaining conservative behavior: recursive accounting may double-count aliased arrays and evict early, but it no longer permits unaccounted nested entries.

waybarrios (approved):
All set to me

PR #692, opened 2026-08-09
Fix integration test opt-in

Thump604:
## What changed

- Treat an explicitly supplied `--server-url` as the integration-suite opt-in.
- Allow integration tests that are also marked slow to run under that opt-in.
- Query actual pytest markers instead of keyword/path membership.

## Root cause

The collection hook unconditionally attached a skip marker to every integration test. It also used `item.keywords`, which can contain parent/path keywords and caused false marker matches.

## Validation

- `pytest -q tests/test_collection_policy.py`
- `pytest -q -m integration --server-url http://localhost:8000 --setup-plan` selects and schedules all five documented tests.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -9,7 +9,7 @@ def pytest_addoption(parser):
     parser.addoption(
         "--server-url",
         action="store",
-        default="http://localhost:8000",
+        default=None,
         help="URL of the vllm-mlx server for integration tests",
     )
     parser.addoption(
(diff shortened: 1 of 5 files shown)


janhilgard (approved):
Approving. The root cause is exactly as stated, and I confirmed the opt-in never worked rather than taking the description on trust.

On `main`, the integration skip is applied unconditionally — the comment says "unless server URL is explicitly provided" but there is no such condition in the code:

```
main + --server-url http://localhost:1237   ->  5 skipped, 2339 deselected
```

Five tests, collected and skipped, no matter what you pass. With your patch, against the same running server:

```
#692 + --server-url http://localhost:1237   ->  4 passed, 1 failed
```

So they actually run now. Making the explicit `--server-url` the opt-in and dropping the `http://localhost:8000` default is the right shape: a default URL means "opt-in" can never be distinguished from "left alone", which is how the condition went missing in the first place without anyone noticing.

`get_closest_marker` over `item.keywords` is correct regardless, though worth noting the false-positive is theoretical in this tree today — `tests/test_streaming_pipeline_integration.py` has no `integration` marker and its 12 tests are **not** skipped on `main`, so nothing is currently mis-matching. Still the right change; it just is not fixing a live symptom.

## What un-skipping immediately surfaced

`TestServerIntegration::test_chat_completion` fails with 404, and it is the test that is wrong, not the server:

```python
"model": "default",
```

Against a server started with `--served-model-name`, that model does not exist, and 404 is the correct answer. Confirmed on our deployment:

```
model=default        -> 404
model=qwen3.6-35b    -> 200
```

The other four pass because `/health` and `/v1/models` do not name a model. So this is a pre-existing bug that was invisible precisely because the suite never ran — which is the best possible argument for this PR, and also means it lands the repo with a failing integration test unless it is fixed here or immediately after. Reading the model id from `/v1/models` would make the test work against any deployment.

CI green on all nine jobs.

Thump604:
Addressed the integration failure surfaced in review. The chat test now queries `/v1/models` and uses the advertised model ID instead of hardcoding `default`; a focused mocked regression pins that behavior. Coordinator verification: `3 passed`, Black clean on the four touched test files, and `git diff --check` clean. Pushed as `e1ce26e`.

PR #693, opened 2026-08-09
Restore paged-cache bulk release API

Thump604:
## What changed

- Restore a callable bulk block-release operation as `free_block_batch()`.
- Add regression coverage for releasing multiple blocks and updating the free count.

## Root cause

`PagedCacheManager.free_blocks(blocks)` was overwritten later in the same class by the integer `free_blocks` property. The method therefore did not exist at runtime.

## Validation

- `pytest -q tests/test_paged_cache.py` — 42 passed.

### DIFF
--- a/tests/test_paged_cache.py
+++ b/tests/test_paged_cache.py
@@ -163,6 +163,18 @@ class TestPagedCacheManager:
         # Block should be back in free queue
         assert manager.free_blocks == initial_free
 
+    def test_free_block_batch_is_callable_and_updates_free_count(self):
+        """Bulk release must not be shadowed by the free-block count property."""
+        from vllm_mlx.paged_cache import PagedCacheManager
+
+        manager = PagedCacheManager(block_size=64, max_blocks=10)
+        blocks = [manager.allocate_block(), manager.allocate_block()]
+        assert manager.free_blocks == 7
+
+        manager.free_block_batch(blocks)
+
+        assert manager.free_blocks == 9
+
     def test_reference_counting(self):
         """Test reference counting."""
         from vllm_mlx.paged_cache import PagedCacheManager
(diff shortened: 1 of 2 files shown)


janhilgard (approved):
Approving. Verified the defect on `main` rather than reading for it:

```
free_blocks is: property
has free_block_batch: False
callable as a method: NO (AttributeError)
```

`free_blocks(blocks)` is defined at `paged_cache.py:667` and the `@property free_blocks` at line 1101 in the same class shadows it, so the method never existed at runtime. Renaming rather than reordering is the right call — the property is the one with callers, and moving it would be the riskier change of the two.

I also checked nothing was calling the shadowed method and finding it missing: no call sites in the tree, so this restores an API rather than fixing a live crash. Worth saying out loud in case that affects how you prioritise it.

At your head: `free_block_batch` is a method, `free_blocks` remains a property, `tests/test_paged_cache.py` 42 passed, CI green. Mutation-checked — renaming it back to `free_blocks` makes the new test fail on the shadowing.

One thought for later, not this PR: nothing prevents the next method from being shadowed the same way. A `__init_subclass__` or a one-line test asserting no name is defined twice in that class would catch it structurally, since the failure mode is silent — Python is perfectly happy to let the second definition win.

waybarrios:
I reviewed and it is all set!  However, what is your current scope?

Thump604:
For this PR, the scope is deliberately narrow: restore the shadowed bulk-release operation under an unambiguous `free_block_batch()` name and add regression coverage, without changing allocation semantics or adding callers.

More broadly, I am focusing on isolated correctness and lifecycle fixes from Runtime validation, not broad refactors. I am happy to mark this ready if that API shape is useful upstream; otherwise I can close it.

waybarrios:
Good to me too

PR #694, opened 2026-08-09
Fail invalid multimodal inputs atomically

Thump604:
## What changed

- Make multimodal preprocessing request-atomic across the direct `MLXMultimodalLM`, shared `MultimodalProcessor`, and continuous-batching paths.
- If any requested image, audio, or video item cannot be resolved, decoded, or prepared, preprocessing raises instead of silently dropping that item and continuing with a partial media set.
- Preserve per-request isolation in continuous batching. `_preprocess_request` is already wrapped by request-level error handling at all three batch-preprocessing call sites. A failed request is removed from the batch and receives an error result; other requests continue. One bad attachment therefore cannot fail or poison the shared batch.
- Remove unconditional Gradio prompt/file debug output.
- Stop logging remote media URLs, including signed query parameters, at normal INFO level.

## Root cause

Media items were previously processed inside broad per-item exception handlers that logged the failure and continued. That could leave a request with only some of its requested attachments, or with no usable media, while inference proceeded without telling the caller that the input had changed.

This change makes preprocessing all-or-error for the affected request. “Atomic” here means that generation does not start with an incomplete media set; it does not promise a particular HTTP status code.

## Batch isolation

The batched scheduler retains its existing per-request protection around `_preprocess_request`:

- failed requests are collected and removed before the batch is formed;
- each failed request receives a request-level error result;
- unrelated requests in the same batch remain eligible to continue.

The change removes the inner per-item catches so an invalid attachment cannot be converted into a plausible answer to a different, silently modified request.

## Observed deployment behavior

On the current batched `--mllm` deployment, corrupt-media probes did not produce a successful partial-media answer. The observed wire results were:

```text
one corrupt image             -> HTTP 200, content=null, finish_reason="error"
one valid + one corrupt image -> HTTP 200, content="",   finish_reason="error"
```

This PR moves media validation ahead of model work, but it does not itself change the existing HTTP status mapping to `4xx`. In particular, it does not claim that every invalid-media request will return a `4xx` response.

The direct and Gradio entry points still need independent end-to-end verification of exception-to-response serialization and any remaining partial-media behavior. The batched observation above does not certify those paths or imply that every route returns the same status and payload.

## Validation

- `pytest -q tests/test_gradio_app.py tests/test_mllm.py tests/test_mllm_continuous_batching.py`
  - `86 passed, 10 deselected`
- Coverage includes invalid image, audio, shared-processor, and batched-preprocessing failures.
- Coverage verifies that Gradio no longer prints private prompt/file details.
- Coverage verifies that signed remote media URLs are absent from normal logs.
- CI: lint, type-check, Python test matrix, Apple Silicon tests, and aggregate tests pass.

## Explicit non-claims

- This PR does not promise a universal `4xx` response for invalid media.
- This PR does not claim that direct or Gradio paths have the same wire-level behavior as the batched deployment.
- This PR does not claim that every route normalizes invalid-media failures to one identical status or payload.
- Valid media requests retain their existing processing behavior.

### DIFF
--- a/tests/test_gradio_app.py
+++ b/tests/test_gradio_app.py
@@ -42,6 +42,21 @@ def test_create_chat_function_uses_default_served_model_name(monkeypatch):
     assert captured["json"]["model"] == "default"
 
 
+def test_chat_function_does_not_print_prompt_or_file_details(monkeypatch, capsys):
+    monkeypatch.setattr(
+        gradio_app.requests, "post", lambda *args, **kwargs: DummyResponse("ok")
+    )
+    chat_fn = gradio_app.create_chat_function(
+        server_url="http://localhost:8000",
+        max_tokens=16,
+        temperature=0.0,
+    )
+
+    assert chat_fn({"text": "private prompt", "files": []}, history=[]) == "ok"
+    captured = capsys.readouterr()
+    assert "private prompt" not in captured.out
+
+
 def test_create_chat_function_uses_configured_served_model_name(monkeypatch):
     captured = {}
 
--- a/tests/test_mllm.py
+++ b/tests/test_mllm.py
@@ -120,6 +120,56 @@ class TestMLLMHelperFunctions:
         assert not is_base64_image("https://example.com/image.jpg")
         assert not is_base64_image("/path/to/image.jpg")
 
+    def test_remote_media_log_does_not_expose_signed_url(self, monkeypatch, caplog):
+        import logging
+        import requests
+
+        from vllm_mlx.models import mllm
+
+        signed_url = "https://example.com/media.png?token=super-secret"
+
+        def fail_request(*args, **kwargs):
+            raise requests.Timeout("offline")
+
+        monkeypatch.setattr(mllm, "_request_with_safe_redirects", fail_request)
+        with caplog.at_level(logging.INFO), pytest.raises(requests.Timeout):
+            mllm._download_media(signed_url, "image", {"png": ".png"}, ".png", 1, 1024)
+
+        assert "super-secret" not in caplog.text
+        assert signed_url not in caplog.text
+
+
+class TestMLLMMediaAndHelperFunctions:
+    def test_invalid_image_fails_request_instead_of_becoming_text_only(
+        self, monkeypatch
+    ):
+        from vllm_mlx.models import mllm
+
+        model = mllm.MLXMultimodalLM.__new__(mllm.MLXMultimodalLM)
+        monkeypatch.setattr(
+            mllm,
+            "process_image_input",
+            lambda image: (_ for _ in ()).throw(ValueError("invalid image")),
+        )
+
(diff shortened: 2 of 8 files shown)


janhilgard (approved):
Approving. The behaviour change is right and the isolation it depends on already exists — I checked that rather than assuming it, because removing per-item error handling in a batched server is exactly where this could go wrong.

**One bad attachment cannot take the batch down.** `_preprocess_request` is called inside a per-request `try/except` at all three call sites in `mllm_batch_generator.py` (1266, 2757, 2926), with a `failed_requests` list and a comment saying so. So dropping the inner per-item handlers moves the failure from "silently degrade this request" to "fail this request", not "fail the batch". That is the property that makes this safe, and it is worth stating explicitly in the PR body — a reader who only sees `try/except` being deleted from a batched path will reasonably worry.

The behaviour being replaced is the worse one. Answering a question about an image that was quietly dropped produces a confident answer to a question the user did not ask, and nothing in the response says an attachment went missing. A request-level error is recoverable; a plausible answer to the wrong input is not.

Removing the `print()` calls is a real fix too, not tidying: `[Chat] Text: ...` and `[Chat] Files: ...` wrote user prompts and file paths to stdout unconditionally, outside the logging framework and therefore outside any level or redaction policy. Redacting remote media URLs in normal logs belongs in the same bucket.

`tests/test_gradio_app.py tests/test_mllm.py tests/test_mllm_continuous_batching.py` at your head: **86 passed, 10 deselected**. CI green on all nine jobs.

## One observation from trying to reproduce the old behaviour

I could not get our deployment to exhibit the partial-media path, which may be worth knowing for how you frame the fix. On `main`, a corrupt attachment — alone or alongside a valid one — comes back as **HTTP 200 with `content: null` and `finish_reason: "error"`**, not as a degraded answer:

```
1 corrupt image            -> 200, content=null,  finish_reason=error
1 valid + 1 corrupt image  -> 200, content='',    finish_reason=error
```

So on this configuration (`--mllm`, batched) something upstream already fails the request; the silent-degradation window is presumably narrower than the description implies, or lives on the direct/gradio paths. That does not weaken the change — a 200 carrying `finish_reason: "error"` is its own problem, and failing before scheduling model work is strictly better than either. I mention it only so the impact claim is not overstated if someone quotes it later.

Thump604:
Updated the PR body to make the existing per-request batch isolation explicit and to record the observed HTTP 200 / `finish_reason="error"` behavior without claiming universal `4xx` handling. I also kept direct and Gradio wire behavior as explicit non-claims pending separate end-to-end verification.

waybarrios:
Fixed in 8c6341e: failed audio requests are now removed from the queue and receive one error, preventing repeated retries during chunked prefill while other requests continue.

Added regression coverage with and without active decoding. All 10 CI checks passed, including both new regression cases on Apple Silicon with Python 3.11 and 3.13. Ready to merge.

PR #695, opened 2026-08-09
feat(embeddings): add operator-configurable max length ceiling and overflow policy

### LINKED ISSUE #671: [Feature]: Add a configurable embedding max length and explicit overflow behavior
### Motivation

I use [GitNexus](https://github.com/abhigyanpatwari/GitNexus) with a local OpenAI-compatible embedding endpoint to index source code using `Qwen3-Embedding-4B`. The model supports a 32K-token context window and produces 2,560-dimensional embeddings, but `EmbeddingEngine.embed()` currently tokenizes with `truncation=True, max_length=512`.

That means inputs above 512 tokens can succeed while silently losing their tail. This is especially difficult to detect in a vector index because the response shape remains valid and the resulting retrieval-quality loss appears later.

Open PR #626 is directly related and would resolve the default truncation length from `model.config.max_position_embeddings` or `tokenizer.model_max_length`. That is a valuable default fix, but the PR explicitly does not add an operator-facing ceiling/override or an explicit overflow policy. This feature request tracks that remaining user-facing control rather than duplicating #626.

This benefits operators serving modern embedding models with contexts above 512 tokens, while still allowing conservative limits for memory-constrained machines and predictable behavior across clients.

### Proposal

Add a server-level embedding input-length setting, for example:

- CLI: `--embedding-max-length <tokens>`
- Optionally an equivalent environment/config-file setting for service deployments
- `auto` or an omitted value should use the model-aware resolution proposed by #626
- An explicit value should be validated as a positive integer and should not exceed the model's declared context length unless an intentional override mechanism is documented

Also make overflow behavior explicit rather than silent. A compatible design could expose an option such as:

- `error`: reject over-limit inputs with a structured 4xx response containing the observed and allowed token counts
- `truncate`: truncate deliberately, but emit an observable warning/metric containing the model, original token count, and effective limit

The effective limit should be applied to `POST /v1/embeddings`, reflected in documentation, and included in relevant status/config output. Long-context batches should remain bounded by a token budget so one padded batch cannot unexpectedly exhaust unified memory.

Suggested acceptance criteria:

1. A Qwen3-Embedding model declaring 32,768 positions is not silently restricted to 512 tokens.
2. Operators can choose a lower deployment limit such as 1,024 or 4,096 tokens.
3. Inputs above the effective limit are either explicitly rejected or observably truncated according to the configured policy.
4. Classic BERT-family models declaring 512 positions retain their existing effective limit.
5. Tests cover model-derived defaults, explicit overrides, invalid values, over-limit requests, and mixed-length batches.

### Surface impact

- CLI / serve command
- OpenAI API endpoints
- Scheduler / batching
- Docs

### Alternatives considered

- **Replace 512 with another fixed constant:** This only moves the problem and can make smaller models invalid or large batches unsafe.
- **Use only the model's declared maximum:** This avoids accidental 512-token truncation but gives operators no way to select a lower, memory-aware service ceiling.
- **Require every client to pre-chunk perfectly:** Clients should still chunk for retrieval quality, but the server owns tokenization and is the only layer that knows the exact token count and model limit.
- **Rely solely on logs after silent truncation:** This is better than no signal but still permits unnoticed data loss unless overflow behavior is configurable.

### Are you planning to send a PR for this?

No, requesting only

### Additional context

- Related implementation: #626
- [Qwen3-Embedding-4B model card](https://huggingface.co/Qwen/Qwen3-Embedding-4B) — 32K context length and up to 2,560 embedding dimensions
- [mlx-embeddings](https://github.com/Blaizzy/mlx-embeddings) exposes a `max_length` argument, so the serving layer can pass an effective model/operator-derived value
- GitNexus currently creates approximately 1,200-character code chunks plus embedding headers/context; symbol-heavy code can exceed 512 tokens even when the character count looks modest

brandy975:
Closes #671.

> **Depends on #626** (open, unmerged). This branch is stacked on top of `feat/embedding-rerank-max-length`, so the diff below includes #626's two commits until it merges — only the commits after #626 (`feat(embeddings): add operator-configurable max length ceiling and overflow policy`) are new here.

## Summary

Issue #671 points out that PR #626 (model-aware truncation length instead of hardcoded 512) deliberately leaves two things unaddressed: operators have no way to impose a *lower* deployment-wide ceiling than a model's own declared context window, and there's no explicit choice for what happens when an input still exceeds the effective limit — today that's always a silent `truncation=True`.

This PR adds both, plus the observability the issue calls for:

- `--embedding-max-length <tokens|auto>`: `auto` (default, or omitting the flag) keeps #626's model-aware resolution uncapped. A positive integer clamps it down — `min(model-aware default, ceiling)` — for memory-constrained deployments. This matters concretely: requests are tokenized with `padding=True`, so every text in a batch is padded to the length of the *longest* text in that batch, and a large-context model (e.g. `Qwen3-Embedding-4B`) left uncapped can use dramatically more memory per batch than the 512-token models this server shipped with historically.
- `--embedding-overflow-policy {truncate,error}` (default `truncate`, backward-compatible): `truncate` keeps today's behavior but is now observable — one aggregated warning per batch (not per input, so a large batch can't flood the log) plus a `vllm_mlx_embedding_truncated_total` Prometheus counter. `error` rejects over-limit inputs instead, with a structured 400 (`error`, `message`, `input_index`, `token_count`, `max_length`).
- Loading a model with an unbounded, large declared context (>4096 tokens, no ceiling set) logs a `WARNING` at startup nudging the operator toward an explicit ceiling.
- The effective `max_length`/`overflow_policy` are reported under `embedding` in `GET /v1/status`.

Classic 512-token BERT-family models are unaffected by any of this.

## Type of change

- Feature

## Areas affected

- CLI / serve command (`--embedding-max-length`, `--embedding-overflow-policy`)
- OpenAI API endpoints (`/v1/embeddings`)
- `GET /v1/status`
- Prometheus metrics (`vllm_mlx_embedding_truncated_total`)
- Docs (`docs/guides/embeddings.md`)

## Out of scope (follow-up)

Token-budget-bounded batching for embeddings (mirroring the reranker's existing `token_budget` packing, so one long outlier can't inflate an entire batch's padding) is not included here — this PR only covers issue #671's enumerated acceptance criteria. The reranker itself is untouched; `resolve_max_length()`'s new `ceiling` param is generic enough for it to adopt later without further resolver changes.

## Test plan

```
pytest tests/test_truncation.py tests/test_embeddings.py tests/test_rerank.py tests/test_cli.py tests/test_lifecycle_cli.py -q
```


waybarrios (changes requested):
The direction is good and CI is green, but I see three blockers before merge:

1. /v1/status reports the configured ceiling (or "auto"), not the effective resolved max length. A 512-token model with a 4096 ceiling is reported as 4096 even though the engine uses 512.

2. overflow_policy="error" relies on _token_length(), which falls back to len(text) // 4 when tokenizer.encode fails. That can silently truncate an over-limit input or reject a valid one with an inaccurate token_count. Policy enforcement needs an exact token count; keep the heuristic only for usage reporting.

3. EmbeddingRequest still accepts an unbounded list and pads the whole batch to its longest input. With an auto-resolved 32K context, one request can amplify memory and compute enough to cause an availability issue. #671 explicitly asks for token-budget-bounded long-context batches; please implement a bound or keep that part tracked and open rather than closing the issue.

I did not find an injection, auth-bypass, sensitive-input logging, or Prometheus-cardinality problem.

Also, merge #626 first and rebase this stacked branch before the final review.

brandy975:
@waybarrios — thanks for the review. Since you flagged merging #626 first as a prerequisite before rebasing this stack: could you merge that one when you get a chance? I'll take a look at the three blockers you listed afterward.

waybarrios:
@brandy975 you need to update this PR. There are some conflicts. and PR #626 was already merged. Please do the changes and then let me know for personally review

brandy975:
@waybarrios — thanks for the detailed review. All three blockers are addressed in f412a0f:

1. **`/v1/status`**: `max_length` now reports the effective resolved value (what `embed()` actually applies), not the raw ceiling. The configured ceiling itself is still available separately under the new `max_length_ceiling` key (`int` or `null` when unset).
2. **`overflow_policy="error"`**: enforcement now uses an exact tokenizer count and propagates a tokenizer failure instead of falling back to the `len(text)//4` heuristic. That heuristic still exists, but only for `count_tokens()` usage reporting, where an approximation is acceptable.
3. **Unbounded batch/padding**: `embed()` now packs a request's texts into sub-batches bounded by a token budget (default 4096, mirroring `RerankEngine.token_budget`) instead of one batch padded to the longest input. Results are reassembled in original input order; a normal small request still runs as a single forward pass, unchanged.

Also added test coverage for two gaps against #671's own acceptance criteria that weren't there before (invalid `--embedding-max-length` values, and a batch mixing under-/over-limit texts).

waybarrios:
I pushed a follow-up that fixes the remaining issues: embedding batches now account for padding, MLX cache is cleared after every batch even when one fails, the standalone server supports the new embedding length options, and the docs and regression tests were updated to match. This should cover the remaining review blockers.

waybarrios (approved):

PR #640, opened 2026-07-16
feat(registry): add CLI override for memory budget

lunarlinger:
## Summary

Adds an explicit `--memory-budget-gb` CLI override for registry-backed
multi-model serving.

The CLI value takes precedence over:

- `manager.memory_budget_gb`
- the legacy `manager.memory_budget`

When the option is omitted, the existing YAML behavior remains unchanged.
The CLI override can also provide the manager budget when the YAML field is
absent.

This implements the CLI override portion of #627.

## Type of change

- New feature
- Documentation

## Surface touched

- Serve CLI
- Multi-model registry configuration
- Registry manager memory-budget configuration
- Tests
- Documentation

## Behavior

- Accepts only positive finite numeric values.
- Rejects zero, negative values, NaN, infinities, and non-numeric values.
- Rejects `--memory-budget-gb` outside `--models-config` registry mode.
- Uses the existing binary conversion of `GB * 1024^3`.
- Preserves backward compatibility when the option is omitted.

The value controls the resident model-weight budget used for registry
admission and eviction. It is not a total runtime-memory limit and does not
guarantee prevention of Metal/MLX OOM.

## Testing

Automated validation:

- `pytest tests/test_cli.py tests/test_model_registry.py -q`
  - 18 passed
- Broader server tests
  - 116 passed, 3 deselected
- Black checks passed
- Ruff checks passed
- Targeted mypy checks passed
- `git diff --check` passed

Manual end-to-end validation:

- YAML budget absent + CLI budget provided: server started successfully.
- YAML budget absent + no CLI budget: startup failed as before.
- Invalid YAML budget + valid CLI budget: CLI value took precedence and the
  server started successfully.
- Invalid CLI values were rejected.
- Non-registry usage was rejected.
- `/v1/models` returned the configured registry model.

## Out of scope

This PR does not implement:

- automatic Metal allocation-ceiling calculation;
- automatic memory-budget clamping;
- KV-cache or activation-memory accounting;
- startup allocation-ceiling warnings;
- changes to registry admission or eviction algorithms.

PR #696, opened 2026-08-09
feat(registry): reconcile manager memory budget with the Metal allocation ceiling

danmackinlay:
## Summary

Implements the startup-validation-plus-documentation half of #627 — the first of
the two PRs @Thump604 proposed in [that thread](https://github.com/waybarrios/vllm-mlx/issues/627#issuecomment-4935833071).

In registry mode the manager decides admission and eviction purely against
`manager.memory_budget_gb`, and never consults the Metal allocation ceiling
installed from `--gpu-memory-utilization`. The two numbers are independent and
unreconciled, so a budget above the ceiling makes the manager keep N models
resident because *its* arithmetic says they fit — and MLX then fails to
allocate instead of the manager evicting. This adds a deterministic startup
report and warning so that becomes a diagnosable message rather than a silent
crash.

Deliberately **not** implemented here, per the review guidance on the issue:
auto-clamping, KV/activation accounting, or any change to the admission and
eviction algorithms.

## Relationship to the other PRs on #627

- #640 adds the `--memory-budget-gb` CLI override (suggested fix 2) and lists
  "startup allocation-ceiling warnings" as out of scope. This PR is that missing
  piece; the two are complementary and touch disjoint code paths.
- #638 is about `--auto-unload-idle-seconds` in registry mode — unrelated to the
  budget/ceiling reconciliation.

## Behavior

At registry startup the server logs the budget and the ceiling side by side,
with the prefix-cache setting reported alongside:

```
INFO Registry memory budget: 68.0 GB of model weights; Metal allocation ceiling
     69.4 GB (62% of 112.0 GB, from serve default); prefix-cache maximum 20.0 GB
     per continuous-batching engine (--cache-memory-mb, 1 of 1 entries)
```

The conflict check compares **only** the weights budget with the Metal ceiling.
Both are process-wide totals and therefore directly comparable:

```
WARNING models-config manager.memory_budget_gb (68.0 GB) exceeds the Metal
        allocation ceiling (56.0 GB). The budget counts model weights only, so
        the manager will keep models resident that MLX cannot allocate, and a
        load can fail with an out-of-memory error instead of evicting. Lower the
        budget below 56.0 GB — further still, since the KV cache and activations
        also come out of the ceiling — or raise --gpu-memory-utilization.
```

`--cache-memory-mb` is **not** subtracted from the ceiling. It is a per-engine
maximum: `_clone_scheduler_config` copies it into each resident
continuous-batching engine and it is allocated lazily, while `SimpleEngine`
never receives a scheduler config at all. Subtracting it once would understate
capacity with one resident model and overstate it with several, so it is
reported next to the ceiling rather than folded into it, annotated with how many
entries are continuous-batching. It is shown only where it can bind —
continuous-batching entries using the memory-aware prefix cache, not under
`--use-paged-cache`.

An oversized cap gets its own separate warning, with no derived negative
quantity anywhere:

```
WARNING --cache-memory-mb (20.0 GB per continuous-batching engine) is at or above
        the Metal allocation ceiling (8.0 GB) on its own, leaving no room for
        model weights. Note this is a per-engine maximum: it is cloned into every
        resident continuous-batching engine, so the aggregate grows with the
        number of resident models.
```

Other details:

- A per-entry `gpu_memory_utilization` override re-installs the process-wide
  Metal limits whenever that model loads, so the check uses the **lowest**
  effective value across the serve default and every registry entry, and names
  where it came from.
- On hosts where MLX cannot report a working-set size, the log says the budget
  could not be reconciled and no warning is emitted (no false positives).

Nothing is clamped; the server starts with whatever budget is configured.

## Scope of the check

The check is **necessary, not sufficient**. It compares the declared weights
budget with the Metal ceiling; the KV cache, prefix cache and activations also
come out of that ceiling and are workload-dependent, so a budget that clears the
check can still be exhausted at runtime. Catching that needs the residency-aware
accounting deferred to later work. Nothing is clamped — the server starts with
whatever budget is configured.

## Type of change

- Bug fix (diagnostic for a silent failure mode)
- Documentation

## Surface touched

- Multi-model registry configuration
- Registry manager memory-budget configuration
- Serve startup logging
- Tests
- Documentation

## Documentation

`docs/guides/model-registry.md`:

- new "Budget vs. the Metal allocation ceiling" section with the invariant,
  what the check covers, and why the cache cap is reported rather than
  subtracted
- `memory_budget_gb` now says up front that it counts model weights only
- "Sizing Rules" records that both estimate paths — summed on-disk
  `.safetensors`/`.gguf` size, and `estimated_memory_gb` — are weight
  estimates, not total runtime memory
- a new entry under "Failure Modes to Expect"

## Testing

- 12 tests in `tests/test_model_registry.py` covering: budget over/under the
  ceiling; the cache cap being reported but never moving the fit-check; its
  suppression for simple-mode entries and under `--use-paged-cache`; the
  tightest per-entry `gpu_memory_utilization` winning; the inconclusive
  no-device-memory path; and the log shapes. Includes a regression test
  asserting no negative quantity can appear in the output for the 8 GiB-ceiling
  reproduction.
- `pytest tests/test_model_registry.py` — 16 passed
- Full suite `pytest tests/` — **2310 passed, 23 skipped, 23 deselected**
- `ruff check vllm_mlx/ tests/ --select E,F,W --ignore E402,E501,E731,F811,F841`
  and `black --check vllm_mlx/ tests/` — clean
- Python 3.10 (test-matrix floor) syntax check — clean
- Mutation-tested the accounting to confirm the tests constrain it:
  reinstating the subtraction in the fit-check fails three of them, and
  removing either the continuous-batching or paged-cache gate fails one each.
- End-to-end on a real 128 GB machine (112 GB reported working set) driving
  `load_model_registry()`; all log output quoted above is from those runs.

## Out of scope

- automatic memory-budget clamping
- KV-cache, prefix-cache or activation memory accounting (including
  residency-aware aggregation of per-engine cache caps)
- the `--memory-budget-gb` CLI override (#640)
- changes to registry admission or eviction algorithms

### DIFF
--- a/vllm_mlx/model_registry.py
+++ b/vllm_mlx/model_registry.py
@@ -299,8 +299,8 @@ class MemoryBudgetReport:
 
     budget_bytes: int
     device_working_set_bytes: int | None
-    gpu_memory_utilization: float
-    gpu_memory_utilization_source: str
+    gpu_memory_utilization: float | None
+    gpu_memory_utilization_source: str | None
     per_engine_cache_limit_bytes: int | None
     per_engine_cache_percent: float | None
     continuous_batching_entries: int
@@ -308,8 +308,13 @@ class MemoryBudgetReport:
 
     @property
     def allocation_ceiling_bytes(self) -> int | None:
-        """Metal soft allocation limit that will be installed at engine start."""
-        if self.device_working_set_bytes is None:
+        """Metal soft allocation limit that will be installed at engine start.
+
+        ``None`` when no ceiling can be attributed: either MLX cannot report a
+        device working set, or no entry will install one (only ``BatchedEngine``
+        calls ``mx.set_memory_limit``).
+        """
+        if self.device_working_set_bytes is None or self.gpu_memory_utilization is None:
             return None
         return int(self.device_working_set_bytes * self.gpu_memory_utilization)
 
@@ -340,30 +345,45 @@ def build_memory_budget_report(
 ) -> MemoryBudgetReport:
     """Reconcile the manager weight budget against the Metal allocation ceiling.
 
-    ``gpu_memory_utilization`` is process-wide but per-entry overrides re-install
-    the Metal limits every time a model loads, so the *lowest* effective value
-    across the registry is the ceiling the manager actually has to live under.
+    The Metal limit is process-wide but is re-installed by every
+    ``BatchedEngine`` start, so the ceiling the manager has to live under is the
+    *lowest* utilization among the entries that actually install one. Only
+    continuous-batching entries qualify: ``SimpleEngine`` never calls
+    ``mx.set_memory_limit`` and is not even given a ``gpu_memory_utilization``.
+    A registry with no continuous-batching entries therefore gets no attributed
+    ceiling rather than one derived from a value nothing installs.
     """
     if device_working_set_bytes is None:
         device_working_set_bytes = _device_working_set_bytes()
 
-    utilization = defaults.gpu_memory_utilization
-    utilization_source = "serve default"
+    # Only BatchedEngine installs a Metal allocation limit — SimpleEngine is not
+    # even constructed with a gpu_memory_utilization — so a simple-mode entry's
+    # override is inert and must not be treated as a ceiling candidate.
+    candidates: list[tuple[float, int, str]] = []
     for name in sorted(registry):
-        override = registry[name].gpu_memory_utilization
-        if override is not None and override < utilization:
-            utilization = override
-            utilization_source = f"models-config entry '{name}'"
-
-    continuous_batching_entries = sum(
-        1
-        for entry in registry.values()
-        if (
+        entry = registry[name]
+        entry_continuous_batching = (
             entry.continuous_batching
             if entry.continuous_batching is not None
             else defaults.continuous_batching
         )
-    )
+        if not entry_continuous_batching:
+            continue
+        if entry.gpu_memory_utilization is not None:
+            # Rank 1: an override is only attributable to the entry declaring it.
+            candidates.append(
+                (entry.gpu_memory_utilization, 1, f"models-config entry '{name}'")
+            )
+        else:
+            # Rank 0: prefer the serve default as the named source on ties.
+            candidates.append((defaults.gpu_memory_utilization, 0, "serve default"))
+
+    continuous_batching_entries = len(candidates)
+
+    utilization: float | None = None
+    utilization_source: str | None = None
+    if candidates:
+        utilization, _, utilization_source = min(candidates)
 
     # cache_memory_mb only binds for continuous-batching engines, and only when
     # the memory-aware prefix cache is the one actually in use.
@@ -404,11 +424,18 @@ def log_memory_budget_report(report: MemoryBudgetReport) -> None:
     ceiling = report.allocation_ceiling_bytes
 
     if ceiling is None:
+        if report.gpu_memory_utilization is None:
+            reason = (
+                "no continuous-batching entries, and --gpu-memory-utilization "
+                "installs a Metal limit only for those"
+            )
+        else:
+            reason = "MLX cannot report a device working set on this host"
         logger.info(
-            "Registry memory budget: %.1f GB of model weights "
-            "(Metal allocation ceiling unavailable on this host, so the budget "
-            "cannot be reconciled with it)",
+            "Registry memory budget: %.1f GB of model weights; no Metal "
+            "allocation ceiling to reconcile it with (%s)",
             report.budget_bytes / gb,
+            reason,
         )
         return
 
--- a/vllm_mlx/model_registry.py
+++ b/vllm_mlx/model_registry.py
@@ -260,6 +260,166 @@ def _safe_available_memory_bytes() -> int:
     return int(psutil.virtual_memory().available)
 
 
+def _device_working_set_bytes() -> int | None:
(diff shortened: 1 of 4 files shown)


vllm_mlx/model_registry.py
```diff
@@ -260,6 +260,166 @@ def _safe_available_memory_bytes() -> int:
...
+            utilization = override
+            utilization_source = f"models-config entry '{name}'"
+
+    scheduler_config = defaults.scheduler_config
+    cache_reservation_bytes: int | None = None
+    cache_reservation_percent: float | None = None
+    if scheduler_config is not None:
+        cache_memory_mb = getattr(scheduler_config, "cache_memory_mb", None)
+        if cache_memory_mb:
+            cache_reservation_bytes = int(cache_memory_mb) * (1024**2)
```
Thump604: `cache_memory_mb` is cloned into each resident continuous-batching engine and allocated lazily; it does not apply to `SimpleEngine` entries. Recording one copy here and subtracting it once from the process-wide Metal ceiling can both overstate and understate available weight capacity while the warning calls the result ‘actually allocatable.’ Use weight budget versus Metal ceiling as the deterministic hard-conflict check, and report this as a per-engine cache maximum unless residency-aware aggregate accounting is added.

vllm_mlx/model_registry.py
```diff
@@ -260,6 +260,166 @@ def _safe_available_memory_bytes() -> int:
...
+            return None
+        return int(self.device_working_set_bytes * self.gpu_memory_utilization)
+
+    @property
+    def weights_headroom_bytes(self) -> int | None:
+        """Ceiling minus any statically known cache reservation."""
+        ceiling = self.allocation_ceiling_bytes
+        if ceiling is None:
+            return None
+        return ceiling - (self.cache_reservation_bytes or 0)
```
Thump604: This can return negative headroom when the explicit cache maximum exceeds the Metal ceiling. I reproduced an 8 GiB ceiling with `cache_memory_mb=20480` reporting -12 GiB, after which the warning recommends a negative maximum model budget. Clamp the display to zero and report that the cache setting alone exceeds the ceiling, or remove the fixed subtraction as part of correcting the per-engine accounting.

Thump604 (changes requested):
The startup diagnostic is a useful and appropriately narrow direction, and the focused test suite passes locally. Two accounting issues need correction before merge: the explicit cache limit is not one process-wide reservation, and subtracting it can produce negative reported weight headroom. I reproduced the latter with an 8 GiB Metal ceiling and a 20 GiB cache limit, yielding -12 GiB and a recommendation to lower the model budget to a negative value. Please keep the deterministic conflict check to quantities that are globally comparable, or add residency-aware accounting for per-engine cache caps.

Thump604 (changes requested):
The current implementation at 95140a3 resolves both cache-accounting findings: the fit check now compares the weights budget only with the process-wide Metal ceiling, the cache limit is reported as a per-continuous-batching-engine lazy cap, SimpleEngine and paged-cache paths are excluded, and an oversized cap gets a separate non-negative warning. I reran the focused suite (103 passed), Ruff, and diff whitespace checks successfully. One remaining correction is needed before approval: please update the PR description, which still says `--cache-memory-mb` is subtracted from the ceiling and still shows the old 49.4 GB recommendation. That now contradicts the implementation and revised guide. Please also update the test-count/behavior summary to match this head; after that description-only correction, this is ready for my approval.

danmackinlay:
Description updated to match `95140a3`: removed the claim that `--cache-memory-mb` is subtracted from the ceiling, replacing it with the per-engine characterisation and its gating; dropped the obsolete 49.4 GB example in favour of log output regenerated from the current head; and corrected the test summary to 16 focused / 2310 full-suite. The scope note now states plainly that the check is necessary but not sufficient, since the KV cache, prefix cache and activations come out of the same ceiling.

Thump604 (approved):
Approved at `95140a3`. The implementation resolves both accounting findings, the focused suite and lint checks passed in my rerun, and the updated PR description now accurately states the per-engine cache semantics, separate oversized-cap warning, necessary-but-not-sufficient scope, and current test results. My two superseded review threads are resolved. Leaving merge ownership to the maintainers.

waybarrios:
Following up on the resolved cache-accounting discussion, I noticed that the cache reporting is now correctly gated by the effective engine mode, but the gpu_memory_utilization selection is not.

Before merging, we should clarify the corresponding Metal-ceiling behavior. build_memory_budget_report() still considers the serve default and overrides from every registry entry, including simple-mode entries, while the Metal limit appears to be configured only through the BatchedEngine path.

This means an all-simple registry, or a mixed registry with a lower utilization override on a simple entry, can derive its reported ceiling from a value that is not installed by that entry. The serve default can also win even when every batched entry overrides it.

Unless there is another path installing this limit for SimpleEngine, the utilization candidates should be limited to the effective values of continuous-batching entries. If there are no such entries, the report should avoid attributing a Metal ceiling to --gpu-memory-utilization.

waybarrios:
Quick follow-up. I pushed commit a73e3c8 to validate each per-entry gpu_memory_utilization before the allocation report is built. Invalid values such as NaN, infinity, zero, negatives, and anything above one are now rejected with a clear configuration error instead of reaching the Metal ceiling calculation and crashing startup. I also added focused regression coverage for every one of those cases. Ready to go!

waybarrios (approved):

PR #698, opened 2026-08-13
Fix missing processors for MLX Whisper checkpoints

### LINKED ISSUE #697: Documented Whisper aliases fail because processor files are missing
## Summary

The documented Whisper aliases can load their MLX weights but fail on the first transcription with current mlx-audio.

## Reproduction

Verified against an untouched main checkout at 5c3bf75 with mlx-audio 0.4.8 and a real generated WAV file:

    curl -F 'file=@speech.wav' \
      'http://localhost:8000/v1/audio/transcriptions?model=whisper-small&language=en&response_format=json'

## Actual behavior

The endpoint returns HTTP 500:

    {"detail":"Transcription failed"}

The server log shows:

    Could not load WhisperProcessor: ... preprocessor_config.json ...
    ValueError: Processor not found. Make sure the model was loaded with a HuggingFace processor.

The documented mlx-community/whisper-small-mlx checkpoint contains MLX weights but not the Hugging Face processor files expected by mlx-audio 0.4.8.

## Expected behavior

Documented Whisper aliases should complete transcription successfully with the supported mlx-audio dependency range.

## Possible solution

When mlx-audio loads a Whisper model without a processor, resolve the matching canonical OpenAI processor and attach it before generation. For example:

    processor = WhisperProcessor.from_pretrained("openai/whisper-small")
    model._processor = processor

The mapping should be explicit for each allowed Whisper alias and should not permit arbitrary request-controlled model loading.

## Regression coverage

- Load a fake Whisper model whose _processor is None.
- Verify the expected canonical processor repository is selected.
- Verify transcription proceeds after the processor is attached.
- Keep Parakeet behavior unchanged.
- Add one real-model integration check on a compatible GPU runner when practical.

## Related response-contract gaps

After manually attaching the canonical processor, mlx-audio returns segment dictionaries. The current duration extraction only checks an end attribute, so duration can remain null. Also, response_format=verbose_json is currently accepted but returned as the same basic JSON shape rather than including segments.

These are related compatibility gaps, but the verified HTTP 500 occurs earlier during processor loading.

## Open PR check

No currently open PR changes the transcription endpoint, Whisper processor loading, audio segment handling, or verbose_json behavior.

## Environment

- vllm-mlx main at 5c3bf75
- mlx-audio 0.4.8
- CUDA 13 container
- NVIDIA GPU

waybarrios:
## Summary

Fixes transcription failures for the documented MLX Whisper checkpoints when `mlx-audio` loads the model weights without a Hugging Face processor.

The transcription endpoint previously returned HTTP 500 because generation received a model with `_processor=None`.

Refs #697.

## Root cause

Some `mlx-community` Whisper checkpoints contain the converted MLX weights but do not include the processor files expected by current `mlx-audio`.

`mlx-audio` catches the processor-loading failure and returns the model with `_processor=None`. The failure then appears during transcription when `generate()` requires a processor.

## Reproduction

```bash
curl -F "file=@speech.wav" \
  "http://localhost:8000/v1/audio/transcriptions?model=whisper-small&language=en&response_format=json"
```

Before this change, the endpoint returned:

```json
{"detail":"Transcription failed"}
```

The server log reported that the checkpoint did not contain `preprocessor_config.json`, followed by a missing processor error.

## Changes

- Add an explicit mapping from documented MLX Whisper checkpoints to their canonical OpenAI processor repositories.
- Load the canonical processor only when the MLX model does not already provide one.
- Preserve processors already bundled with a checkpoint.
- Keep the Parakeet loading path unchanged.
- Publish the loaded model only after processor initialization succeeds.
- Leave the engine retry-safe when the processor download fails.
- Read transcription duration from dictionary segments returned by current `mlx-audio`.
- Run the audio regression suite in the Linux and Apple Silicon CI matrices.
- Exercise canonical processor recovery through transcription and the HTTP endpoint, including failed-download retries and model reuse.

## Supported mappings

| MLX checkpoint | Processor |
| --- | --- |
| `mlx-community/whisper-tiny-mlx` | `openai/whisper-tiny` |
| `mlx-community/whisper-small-mlx` | `openai/whisper-small` |
| `mlx-community/whisper-medium-mlx` | `openai/whisper-medium` |
| `mlx-community/whisper-large-v3-mlx` | `openai/whisper-large-v3` |
| `mlx-community/whisper-large-v3-turbo` | `openai/whisper-large-v3-turbo` |

The mapping is fixed and cannot be replaced by a request-provided repository.

## Compatibility

This does not change:

- The transcription API contract
- Existing model aliases
- Authentication behavior
- Dependency declarations
- Models that already include a processor
- Parakeet behavior
- TTS behavior

## Reproduction environment

- `vllm-mlx 0.4.1`
- `mlx-audio 0.4.8`
- `transformers 5.15.0`
- `huggingface-hub 1.27.0`
- `numpy 2.5.2`
- Python 3.12.3
- Base commit `5c3bf75`

## Original verification

- Real `mlx-community/whisper-small-mlx` model loaded successfully.
- The fallback attached `WhisperProcessor`.
- Real audio transcription completed successfully.
- Dictionary segment duration was returned correctly.
- Focused audio suite passed with 30 tests and 3 expected integration skips.
- Linux CI selection passed with 1,113 tests, 6 skips, and 20 deselections.
- Ruff passed.
- Black passed.
- Compile check passed.
- Final diff check passed.
- Three independent final reviews found no remaining issues.

## Regression coverage

Tests now verify:

- Every documented Whisper checkpoint selects the expected processor.
- Existing processors are never replaced.
- Parakeet never receives a Whisper processor.
- Failed processor downloads do not retain a partially initialized model.
- Retrying after a processor failure does not begin with the previous model attached.
- Dictionary-based Whisper segments provide transcription duration.

## Follow-up verification

- Signed, verified commit: [`7d57801`](https://github.com/waybarrios/vllm-mlx/commit/7d57801726879b5892d808a0cc87f0605008c50d).
- [All 10 fresh CI checks passed](https://github.com/waybarrios/vllm-mlx/actions/runs/35382617679), including both Apple Silicon jobs.
- Python 3.11 and 3.13 on macOS explicitly passed both HTTP recovery cases, all five canonical processor mappings through transcription, and the failure-then-success engine retry. Model loading, processor downloads, and generation are mocked in these regressions.
- Local validation with current main: 1,454 portable tests passed; full Ruff and Black checks passed.
- The six engine recovery cases fail against the pre-fix code, confirming that they detect the missing-processor regression.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,6 +77,7 @@ jobs:
             tests/test_memory_cache.py \
             tests/test_prefix_cache.py \
             tests/test_mllm_cache.py \
+            tests/test_audio.py \
             tests/test_api_models.py \
             tests/test_api_utils.py \
             tests/test_request.py \
--- a/tests/test_audio.py
+++ b/tests/test_audio.py
@@ -9,6 +9,22 @@ import pytest
 import numpy as np
 
 
+def _install_fake_stt_loader(monkeypatch, model=None, load_model=None):
+    """Install a minimal mlx-audio module tree around a fake STT model."""
+    import sys
+    from types import ModuleType
+
+    utils_module = ModuleType("mlx_audio.stt.utils")
+    utils_module.load_model = load_model or (lambda _model_name: model)
+    stt_module = ModuleType("mlx_audio.stt")
+    stt_module.utils = utils_module
+    mlx_audio_module = ModuleType("mlx_audio")
+    mlx_audio_module.stt = stt_module
+    monkeypatch.setitem(sys.modules, "mlx_audio", mlx_audio_module)
+    monkeypatch.setitem(sys.modules, "mlx_audio.stt", stt_module)
+    monkeypatch.setitem(sys.modules, "mlx_audio.stt.utils", utils_module)
+
+
 class TestSTTEngine:
     """Tests for Speech-to-Text engine."""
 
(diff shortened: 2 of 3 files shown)


waybarrios:
A quick look at the implementation:

```python
model = load_model(self.model_name)
processor_repo = _canonical_whisper_processor(self.model_name)

if processor_repo and getattr(model, "_processor", None) is None:
    model._processor = WhisperProcessor.from_pretrained(processor_repo)

self.model = model
self._loaded = True
```

The fallback only uses the fixed checkpoint mapping and preserves processors already included with a model. Assigning `self.model` last also keeps retries clean if the processor download fails.

Current `mlx-audio` returns Whisper segments as dictionaries, so duration now handles that shape directly:

```python
if isinstance(last_seg, dict):
    duration = last_seg.get("end")
```

The regression tests cover all mapped checkpoints, existing processors, Parakeet, failed downloads, retries, and dictionary segment duration.

Thump604 (dismissed):
Changes requested before merge:

1. The fallback tests currently prove only that an arbitrary object is assigned to `model._processor`. They never exercise `generate()`, `transcribe()`, or the `/v1/audio/transcriptions` path that returned HTTP 500. Please add a mocked successful transcription regression that proves the canonical fallback processor reaches generation—ideally through the endpoint—while retaining the existing retry and Parakeet cases.

2. Please change `Closes #697` to `Refs #697` until the issue's served-path acceptance criteria are directly covered. The current upstream lifecycle preflight rejects the closing keyword for that reason.

The implementation itself otherwise looks sound: processor-loading failure is atomic and retry-safe, existing processors are preserved, Parakeet is isolated, the checkpoint mapping is fixed rather than request-controlled, and dictionary/object segment duration is handled.

janhilgard (commented):
The diagnosis is right and the fix works for what it covers — `mlx-audio` swallowing the processor-load failure and leaving `_processor=None` for generation to trip over is a clean root cause. Two things beyond @Thump604's test point, both about scope rather than correctness.

**1. The map covers five checkpoints; the failure mode is not limited to five.** Any other MLX Whisper repo still returns the same HTTP 500, and there are common ones — all of these exist today:

```
mlx-community/whisper-large-v3-mlx-4bit    200
mlx-community/whisper-base-mlx             200
mlx-community/distil-whisper-large-v3      200
```

A quantized `large-v3` needs exactly the `openai/whisper-large-v3` processor the map already knows about, but is not in it. Deriving the processor from the checkpoint name — strip a quantization suffix, drop the `-mlx` marker, map to the `openai/` namespace — would cover the variants without enumerating them, with the explicit map kept as an override for the ones that do not follow the pattern. As written, the next 4-bit conversion someone points at this reopens #697.

**2. `WhisperProcessor.from_pretrained()` is an unguarded network call inside `load()`.** The surrounding `try` only catches `ImportError`, so a transient HF failure or an air-gapped deployment now raises out of model loading. That trades a per-request 500 for a startup failure — arguably worse, because the server previously came up and served everything else. Wrapping it so a failed processor fetch logs and leaves `_processor=None` would put the behaviour back to today's for that case, with the fix applying whenever the fetch succeeds.

Not blocking on my side — @Thump604's request for a regression that actually reaches `generate()` is the one that matters most, and both of these would be naturally covered by the same test if it exercised an unlisted checkpoint and an unreachable processor repo.

Unrelated but worth calling out so it does not get lost in review: the `isinstance(last_seg, dict)` change to duration extraction is a separate bug fix riding along. It looks correct — `mlx-audio` returns dict segments in some paths — but it has no test and nothing in the PR description mentions it.

CI green on all nine.

waybarrios:
@Thump604 both requested changes are addressed in [7d57801](https://github.com/waybarrios/vllm-mlx/commit/7d57801726879b5892d808a0cc87f0605008c50d):

- Added regressions proving canonical processor recovery reaches generation and successful HTTP transcription, including failed-download retries and model reuse. Existing processor-preservation and Parakeet coverage remains intact.
- Changed `Closes #697` to `Refs #697`.

PR #702, opened 2026-08-13
test(cache): cover real scheduler replay parity

Thump604:
## Summary

- add deterministic tiny MLX models that write real `ArraysCache` and `KVCache` state through the production `Scheduler` and mlx-lm `BatchGenerator`
- prove unknown recurrent coverage takes the cold path and its next-turn continuation matches an independent cold run
- prove an identical prompt cannot reuse a trimmed `N+completion` KV supersequence and replay the final prompt token
- fail closed for full-coverage `supersequence` matches in the same way as existing full-coverage `exact` matches

## Root cause

Merged #683 correctly stopped first-response snapshots for ordinary KV caches and stopped guessing keys for unknown recurrent coverage. Its direct gate tests could not expose a pre-existing interaction: the completion path still stores `prompt + output`; `MemoryAwarePrefixCache.fetch()` trims that longer entry to an identical `N`-token prompt and reports `supersequence` with no remaining tokens; `_schedule_waiting()` only rejected full-coverage `exact`, then replayed `prompt[-1]` into a cache already covering all `N` tokens.

### Hosted GitHub CI

The normal CI matrix validates the fast deterministic regression suite. The Apple Silicon jobs explicitly use `-m "not slow"`, so they **do not execute** `tests/test_prefix_cache_real_model_parity.py`.

Fast focused result:

```text
48 passed in 1.89s
```

### Local real-checkpoint qualification

The real-model file is intentionally marked `slow` and opt-in because it requires locally available checkpoints, including a 58 GB Gemma artifact. These results were produced locally through the production `Scheduler` and mlx-lm `BatchGenerator`; they are qualification evidence, not GitHub-hosted CI results:

```text
Mamba 130M pure ArraysCache: 1 passed in 2.96s
Qwen3.5 4B hybrid ArraysCache+KVCache: 1 passed in 4.36s
Gemma 4 31B BF16 attention cache: 1 passed in 36.83s
Qwen3 1.7B plain KVCache: 1 passed in 2.72s
```

The exact artifact identities, hashes, cache-topology assertions, and unpatched-main failure are recorded in the PR comment: https://github.com/waybarrios/vllm-mlx/pull/702#issuecomment-5285733862

On unpatched merged `main`, the real Qwen KV case reaches `cache_hit_type=supersequence`, `cached_tokens=144`, and `remaining_tokens=[]`; this PR forces a cold prefill for that full-coverage state.

Static checks: `python3 -m py_compile` and `git diff --check` pass.

Upstream-local-repro preflight: `/opt/ai-runtime/run/upstream-local-repro-preflight/20260813T194013Z-upstream-local-repro-preflight.json`

### DIFF
--- a/tests/test_prefix_cache_real_model_parity.py
+++ b/tests/test_prefix_cache_real_model_parity.py
@@ -1,0 +1,189 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Opt-in real-checkpoint scheduler parity for prefix-cache safety."""
+
+import os
+from pathlib import Path
+
+import pytest
+
+mx = pytest.importorskip("mlx.core")
+
+from mlx_lm import load  # noqa: E402
+from mlx_lm.models.cache import ArraysCache, KVCache, make_prompt_cache  # noqa: E402
+
+from vllm_mlx.request import Request, SamplingParams  # noqa: E402
+from vllm_mlx.scheduler import Scheduler, SchedulerConfig  # noqa: E402
+
+pytestmark = pytest.mark.slow
+
+
+def _required_model_path(variable):
+    if os.environ.get("VLLM_MLX_REAL_CACHE_MODEL_TESTS") != "1":
+        pytest.skip("set VLLM_MLX_REAL_CACHE_MODEL_TESTS=1 for real checkpoints")
+    value = os.environ.get(variable)
+    if not value or not Path(value).joinpath("config.json").is_file():
+        pytest.fail(f"{variable} must name a complete local model artifact")
+    return value
+
+
+def _scheduler(model, tokenizer):
+    return Scheduler(
+        model,
+        tokenizer,
+        SchedulerConfig(
+            enable_prefix_cache=True,
+            use_memory_aware_cache=True,
+            cache_memory_mb=256,
+            max_num_seqs=1,
+            prefill_batch_size=1,
+            completion_batch_size=1,
+        ),
+    )
+
+
+def _load_vlm_text_model(path):
+    from mlx_vlm import load as load_vlm
+
+    from vllm_mlx.text_model_from_vlm import build_text_model
+
+    vlm_model, processor = load_vlm(path, strict=False)
+    text_model = build_text_model(vlm_model, path)
+    assert text_model is not None
+    tokenizer = getattr(processor, "tokenizer", processor)
+    assert callable(getattr(tokenizer, "encode", None))
+    return text_model, tokenizer
+
+
+def _long_prompt(tokenizer, minimum=144):
+    seed = tokenizer.encode(
+        "The cache must preserve every represented token across scheduler reuse."
+    )
+    assert seed
+    prompt = (seed * ((minimum + len(seed) - 1) // len(seed)))[:minimum]
+    return [int(token) for token in prompt]
+
+
+def _generate(scheduler, request_id, prompt, max_tokens=4):
+    request = Request(
+        request_id=request_id,
+        prompt=list(prompt),
+        sampling_params=SamplingParams(max_tokens=max_tokens, temperature=0.0),
+    )
+    scheduler.add_request(request)
+    emitted = []
+    for _ in range(32):
+        result = scheduler.step()
+        for output in result.outputs:
+            emitted.extend(output.new_token_ids)
+        if request_id in result.finished_request_ids:
+            return emitted, request
+    raise AssertionError(f"request {request_id} did not finish")
+
+
+def test_real_arrays_cache_next_turn_matches_cold():
+    path = _required_model_path("VLLM_MLX_ARRAYS_MODEL")
+    model, tokenizer = load(path)
+    topology = make_prompt_cache(model)
+    assert topology and all(isinstance(layer, ArraysCache) for layer in topology)
+
+    warm = _scheduler(model, tokenizer)
+    prompt = _long_prompt(tokenizer)
+    first_tokens, _ = _generate(warm, "arrays-first", prompt)
+    assert first_tokens
+    assert list(warm.memory_aware_cache._entries) == []
+
+    suffix = tokenizer.encode(" Next turn.")
+    assert suffix
+    next_prompt = prompt + first_tokens + [int(suffix[0])]
+    warm_tokens, warm_request = _generate(warm, "arrays-next", next_prompt)
+    cold_tokens, cold_request = _generate(
+        _scheduler(model, tokenizer), "arrays-cold", next_prompt
+    )
+
+    assert warm_request.cached_tokens == cold_request.cached_tokens == 0
+    assert warm_request.remaining_tokens == cold_request.remaining_tokens == next_prompt
+    assert warm_tokens == cold_tokens
+
+
+def test_real_kv_cache_identical_prompt_matches_cold_without_replay():
+    path = _required_model_path("VLLM_MLX_KV_MODEL")
+    model, tokenizer = load(path)
+    topology = make_prompt_cache(model)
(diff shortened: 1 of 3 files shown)


waybarrios:
@Thump604 could you paste the log from these tests?

Thump604:
Exact local logs below. Important scope note: these tests use the production `Scheduler`, mlx-lm `BatchGenerator`, and real `ArraysCache` / `KVCache` implementations, but the model is a tiny deterministic cache-writing `nn.Module`; this is not a downloaded production checkpoint.

Before the scheduler guard was widened, the new KV test failed as follows:

```text
collected 27 items

tests/test_prefix_cache_scheduler_parity.py .F                           [  7%]
tests/test_prefix_cache_untrimmable.py .........................         [100%]

FAILED test_identical_plain_kv_prompt_never_replays_a_supersequence
E   AssertionError: assert 3 == 0
E    + where 3 = Request(... output_token_ids=[1, 2], ...
E      cached_tokens=3, remaining_tokens=[],
E      cache_hit_type='supersequence').cached_tokens

1 failed, 26 passed in 1.82s
```

The independent cold model produced the original token sequence `[6, 4]`; the trimmed-supersequence path produced `[1, 2]`.

After treating a full-coverage `supersequence` like the existing full-coverage `exact` fail-closed path:

```text
collected 48 items

tests/test_prefix_cache_scheduler_parity.py ..                           [  4%]
tests/test_prefix_cache_untrimmable.py .........................         [ 56%]
tests/test_memory_cache_mlx.py .....................                     [100%]

48 passed in 1.89s
```

Command:

```bash
PYTHONPATH=. /opt/ai-runtime/worktrees/vllm-mlx/specprefill-platform/.venv/bin/python \
  -m pytest -q \
  tests/test_prefix_cache_scheduler_parity.py \
  tests/test_prefix_cache_untrimmable.py \
  tests/test_memory_cache_mlx.py
```

The full GitHub Linux 3.10–3.13, Apple 3.11/3.13, lint, type-check, and aggregate checks are also green on this head. If you want production-checkpoint evidence beyond this deterministic scheduler/cache regression, I will keep that claim separate rather than presenting this fixture as equivalent.

waybarrios:
That's good. However, you should add those tests in the ci file: https://github.com/waybarrios/vllm-mlx/blob/main/.github/workflows/ci.yml

Please do so.

Thump604:
Real-checkpoint scheduler evidence is now pushed in `a13ca1c`. These are opt-in tests using complete local model artifacts, production `Scheduler` / mlx-lm `BatchGenerator`, and model-owned cache topologies. No synthetic model is used in this matrix.

### Artifacts

- Pure recurrent: `mlx-community/mamba-130m-hf-f32`; config SHA-256 `0633e18ba15631cf9943f74fcc758ff3028356fec79be33d692974a88ecccc3b`; model SHA-256 `5b0c2400e3161bf9f12179770ff2c20db33e72f8678717a312e5eb2b2f341cbf`. Asserted every model-owned cache layer is `ArraysCache`.
- Production-relevant hybrid: local `Qwen3.5-4B-VLM-MTP-4bit`; config SHA-256 `10ec6f4aa5340dee9e64bd8a4b51b8db9737aab3d233fd075542c94b7fcde815`; index SHA-256 `48d7eb04aeac56a2fdb8dccef0dfcd2d257395fb5a9e55d00a35e4b29f617369`. Asserted both `ArraysCache` and `KVCache` layers.
- Attention-cache arm: Gemma 4 31B BF16 artifact revision path `842da3794eaa0b77d5f08bae87a17459d91ff475`; config SHA-256 `431b1c1364c7f043612717de5de557069d23d8ee06a532b887224251043ec495`; complete artifact on local storage. Asserted no `ArraysCache` layers.
- Additional plain KV arm: local `Qwen3-1.7B-MLX-8bit`; config SHA-256 `387d47c904acbcdb2d124f25db20aa3ce65e96b40766039245a7db09ccda2a05`; index SHA-256 `1132afd3de21eadad35d248c57a8f4b2021a39fc5df01c551605fad29d00c9b3`. Asserted every layer is `KVCache`.

### Exact results

```text
Mamba pure ArraysCache next-turn warm vs independent cold:
1 passed, 1 warning in 2.96s

Qwen3.5 4B hybrid next-turn prefix reuse vs independent cold:
1 passed, 1 warning in 4.36s
- actual prefix hit
- cached_tokens=145
- cached + remaining exactly equals next-prompt length
- generated tokens equal cold

Gemma 4 31B BF16 identical prompt vs independent cold:
1 passed, 1 warning in 36.83s
- completion entry is prompt+generated
- corrected scheduler cold-prefills instead of replaying into trimmed full coverage
- generated tokens equal cold

Qwen3 1.7B plain KV identical prompt vs independent cold:
1 passed, 1 warning in 2.72s
```

On unpatched merged `main`, the real Qwen KV test fails the coverage assertion with:

```text
cache_hit_type=supersequence
cached_tokens=144
remaining_tokens=[]
AssertionError: assert 144 == 0
```

That is the exact scheduler state #702 corrects. The earlier tiny deterministic tests remain fast CI regressions; this new file is the requested real-model evidence and is opt-in because it requires local checkpoints.

janhilgard (approved):
Approving, and thanks for building the half I could not.

**The hole is mine and it is real.** The guard I added in #683 rejected full-coverage `exact` hits and let `supersequence` fall straight through to `tokens_to_process = prompt[-1:]`. I traced it on merged `main` rather than reading for it:

```
('trim', 24, 28)             # the prompt+output entry, trimmed by max_tokens
('fetch', 320, 'hit', 0)     # full coverage, zero remaining
('validate', 'KVCache', True)
```

So `fetch()` trims the longer completion entry down to exactly the 320-token prompt, reports `supersequence`, validation passes, the cache survives, and the scheduler then replays `prompt[-1]` onto a cache that already holds all 320 tokens. Exactly the mechanism you describe. `{"exact", "supersequence"}` is the right correction — the condition was never about which lookup path found the entry, only about whether the entry already covers the whole key.

**One qualification worth recording, because it argues *for* the fix rather than against it.** In my reproduction the corrupted replay did **not** change the output — `cold == warm` held on Qwen3-0.6B at greedy with a 320-token prompt. The duplicated token is the final one and the argmax survived it. When I measured this for the `exact` case during #683 it *did* diverge. So the symptom is intermittent, which makes it worse rather than better: a bug that corrupts context on every hit gets found, one that corrupts it on some hits ships.

Verified your regression catches it: reverting the guard to `exact` only fails `test_identical_plain_kv_prompt_never_replays_a_supersequence` with `assert 3 == 0`. Fast suite 2 passed at your head, CI green on all nine.

**On the real-checkpoint tier.** Your qualification matrix is better than what I attempted — I got as far as `LiquidAI/LFM2.5-350M-MLX-8bit` (10 `ArraysCache` + 6 `KVCache`) but could not cover the *pure* recurrent case, because every published mamba2 MLX checkpoint I tried (`mlx-community/mamba2-370m-hf`, `mamba2-2.7b-4bit`, `-8bit`) omits `intermediate_size`, which `mlx_lm.models.mamba2.ModelArgs` requires, so `load()` raises. If your Mamba 130M loads unpatched, I would like to know which repo it is — and if it needed a local fix, that gap is worth an upstream mlx-lm issue, which I am happy to file.

My own harness did not make the #683 merge (it landed a few minutes after), so there is nothing to deduplicate. Yours is the better version; I would rather it be the one in the tree.

Marking the real-model file `slow` and saying plainly that the Apple jobs run `-m "not slow"` and therefore do **not** execute it is the right call. I would rather see that stated than have a green check imply coverage it does not have — my file had the same shape of problem in reverse: not marked slow, but skipping silently whenever the checkpoints are absent, which reads as passing.

PR #704, opened 2026-08-13
Add PostgreSQL output to bench-serve

### LINKED ISSUE #475: bench-serve: add Postgres output format
## Summary

bench-serve currently outputs to table, JSON, CSV, SQL (raw text), and SQLite. Users with an existing Postgres database must either parse JSON/CSV into their DB manually or maintain a separate SQLite file alongside their production database.

## Proposal

Add `postgres` as an output format:

```bash
vllm-mlx bench-serve --url http://localhost:8080 --format postgres \
  --output postgresql://user:pass@localhost/benchmarks

# Or via environment variable:
BENCH_SERVE_PG_URL=postgresql://user:pass@localhost/benchmarks \
  vllm-mlx bench-serve --url http://localhost:8080 --format postgres
```

Implementation mirrors the existing SQLite writer:
- Same result schema
- Table auto-creates if absent (`CREATE TABLE IF NOT EXISTS`)
- Connection string via `--output` or `BENCH_SERVE_PG_URL`
- Uses `psycopg2` (widely available) or `psycopg` (async)
- Optional `--pg-table` flag for custom table name (default: `bench_serve_results`)
- Dependency is optional: `--format postgres` fails with a clear message if psycopg2 is not installed

## Schema

Same columns as the SQLite writer, plus:
- `id BIGSERIAL PRIMARY KEY`
- `created_at TIMESTAMPTZ NOT NULL DEFAULT now()`
- `raw_result JSONB` for the full result blob (forward compatibility)

## Notes

- Self-contained, no MLX dependencies (bench-serve is already a pure HTTP client)
- Backward-compatible: existing CLI behavior unchanged when `--format` is not `postgres`

waybarrios:
## Summary

Add PostgreSQL as an optional `bench-serve` output backend for prompt sweeps and declarative workloads.

- accept connection URLs through `--output` or `BENCH_SERVE_PG_URL`
- create mode-specific tables with searchable columns and `raw_result` JSONB
- support validated custom table names through `--pg-table`
- fail before benchmark execution when PostgreSQL configuration or connectivity is invalid
- keep psycopg optional through the `vllm-mlx[postgres]` extra

Closes #475.

## Type of change

- New feature
- Documentation

## Surface touched

- CLI / serve command
- Build / packaging

## Test plan

```bash
pytest tests/test_bench_serve.py tests/test_dependency_floors.py -q
ruff check vllm_mlx/bench_serve.py vllm_mlx/cli.py tests/test_bench_serve.py --select E,F,W --ignore E402,E501,E731,F811,F841
black --check vllm_mlx/ tests/
```

The focused tests use a fake psycopg connection, so they do not require a PostgreSQL service. The final writer was also smoke-tested against PostgreSQL 16.

## Test results

- `pytest`: 110 passed, 2 skipped integration tests
- Ruff: passed
- Black: 223 files unchanged
- PostgreSQL 16 smoke validation: passed for both schemas, transactions, empty preflight, keyword table names, and non-finite numeric normalization

## Performance impact

N/A. PostgreSQL writes occur only when explicitly selected after benchmark measurement. A zero-row schema preflight runs before measurement to fail fast.

## Output parity

N/A. This change does not touch model generation or response handling.

## Risk notes

- Connection URLs may contain credentials, so application messages never print or log them.
- User-selected table names are restricted to lowercase unqualified identifiers and quoted before SQL interpolation.
- All row and JSONB values use psycopg parameter binding.
- Prompt sweeps and workloads use separate default tables because their schemas differ.
- Existing incompatible tables are not migrated automatically.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,7 +63,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install pytest anyio pytest-cov pydantic fastapi jsonschema httpx psutil transformers requests
+          pip install pytest anyio pytest-cov pydantic fastapi jsonschema httpx psutil tabulate transformers requests
 
       - name: Run unit tests (no MLX required)
         run: |
@@ -82,6 +82,7 @@ jobs:
             tests/test_request.py \
             tests/test_anthropic_models.py \
             tests/test_anthropic_adapter.py \
+            tests/test_bench_serve.py \
             tests/test_dependency_floors.py \
             tests/test_harmony_parsers.py \
             tests/test_endpoint_model_policies.py \
--- a/README.md
+++ b/README.md
@@ -200,6 +200,10 @@ vllm-mlx bench-serve --url http://localhost:8000 --workload workload.json --repe
 
 # Append workload rows into SQLite for longitudinal comparisons
 vllm-mlx bench-serve --url http://localhost:8000 --workload workload.json --repetitions 5 --format sqlite --output bench.db
+
+# Append workload rows into PostgreSQL (requires: pip install "vllm-mlx[postgres]")
+export BENCH_SERVE_PG_URL='postgresql://user:password@localhost/benchmarks'
+vllm-mlx bench-serve --url http://localhost:8000 --workload workload.json --repetitions 5 --format postgres
 ```
 
 ### Model acquisition and conversion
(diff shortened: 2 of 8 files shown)


waybarrios:
## Who this feature is for

The PostgreSQL backend is an optional feature aimed mainly at teams or maintainers who run benchmarks repeatedly and want centralized history, cross-machine comparisons, dashboards, or automated reporting.

For an individual user storing results on one computer, SQLite remains the simpler and recommended option. PostgreSQL can also run locally, but its main benefit is allowing multiple authorized machines or users to write to one shared database.

No benchmark data is uploaded automatically. PostgreSQL is used only when the user explicitly:

1. installs `vllm-mlx[postgres]`,
2. selects `--format postgres`, and
3. provides a PostgreSQL connection URL.

Existing defaults and SQLite behavior remain unchanged, so users who do not need centralized storage incur no additional dependency or configuration cost.

PR #707, opened 2026-08-14
Fix MLX prefix cache thread ownership

Teacher-Tayvin:
## Summary

Persisted prefix-cache load/save currently runs outside the single worker used by MLX generation. Because MLX streams are thread-local, a cache created or loaded on another thread can produce:

`There is no Stream(gpu, N) in current thread`

This patch keeps the generation executor on `EngineCore` for its full lifetime and routes text-engine cache persistence and cleanup through that same executor.

## Changes

- Make the `EngineCore` worker executor engine-owned rather than local to `_engine_loop`.
- Bind MLX generation streams once on that worker.
- Run scheduler steps, cache load/save, and worker cleanup on the same single-thread executor.
- Make cache persistence awaitable through `EngineCore`, `AsyncEngineCore`, and `BatchedEngine`.
- Await server lifecycle cache hooks.
- Preserve non-blocking compatibility for legacy synchronous cache hooks.
- Add regression coverage proving cache load/save and generation use one worker thread.

## Validation

Upstream regression tests:

```text
14 passed, 80 deselected
```

Selected coverage includes worker execution, cache persistence, batched-engine restore, lifecycle save/restore, cancellation, and compatibility paths.

In a separate local Bacon AI evaluation using LFM2.5 1.2B MLX on an M4 MacBook Air, 20 fresh server starts produced:

- 20/20 stateful Responses passes
- 0 stream/thread fallbacks
- 0 cache-save failures
- 0 cache-load failures
- 54.25 ms early continuation median
- 59.07 ms late continuation median

The evaluation is external to this repository and is provided as supporting evidence, not as a replacement for upstream tests.

I have limited experience contributing patches upstream, and I used Codex to help inspect the issue, implement the change, and run the local evaluation. I reviewed the resulting diff and am submitting it for maintainer review, especially around lifecycle ordering and multimodal behavior.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,6 +163,7 @@ jobs:
         run: |
           python -c "import platform; assert platform.machine() == 'arm64', 'Not ARM64'"
           python -c "import mlx.core as mx; print(f'MLX OK: {mx.default_device()}')"
+          python -m pip show mlx mlx-lm
 
       - name: Run MLX-dependent tests
         run: |
--- a/tests/test_batched_engine.py
+++ b/tests/test_batched_engine.py
@@ -108,7 +108,8 @@ class TestBatchedEngineCacheRestore:
         engine._is_mllm = True
         return engine
 
-    def test_load_cache_from_disk_bootstraps_mllm_batch_generator(self):
+    @pytest.mark.anyio
+    async def test_load_cache_from_disk_bootstraps_mllm_batch_generator(self):
         engine = self._make_mllm_engine()
 
         prefix_cache = MagicMock()
@@ -122,7 +123,7 @@ class TestBatchedEngineCacheRestore:
         scheduler._ensure_batch_generator.side_effect = ensure_batch_generator
         engine._mllm_scheduler = scheduler
 
-        loaded = engine.load_cache_from_disk("/tmp/cache")
+        loaded = await engine.load_cache_from_disk("/tmp/cache")
 
         assert loaded == 2
         scheduler._ensure_batch_generator.assert_called_once_with()
--- a/tests/test_batched_engine_owner_thread.py
+++ b/tests/test_batched_engine_owner_thread.py
@@ -78,6 +78,25 @@ def _bare_engine_core(monkeypatch, worker=None):
     return engine
 
 
+def _run_executors_deterministically(monkeypatch):
+    """Drive executor callbacks synchronously while retaining worker identity."""
+    loop = asyncio.get_running_loop()
+
+    def run_in_executor(executor, operation, *args):
+        future = loop.create_future()
+        try:
+            if executor is None:
+                result = operation(*args)
+            else:
+                result = executor.submit(operation, *args).result(timeout=5)
+            future.set_result(result)
+        except BaseException as exc:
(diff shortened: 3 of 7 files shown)


Thump604 (changes requested):
The ownership problem is real, but this head is no longer a safe change set to merge or cherry-pick.

At the submitted head 9e20755db418a2eb58a969b63cafeddbbe8d29d6, BatchedEngine.start() calls the default prepare_for_start() inline on the event-loop thread (engine/batched.py:243-274), which loads the LLM model in _prepare_llm_model() (engine/batched.py:479-494). EngineCore.start() then creates a new ThreadPoolExecutor (engine_core.py:120-131), and scheduler.step() runs on that new worker (engine_core.py:260-267). Binding a fresh MLX stream on the worker does not prove that model buffers and the generation stream have the same owner. The submitted tests use fake schedulers/cache operations and do not exercise this model-load/worker boundary.

Current main already contains a different ownership design: BatchedEngine creates a persistent generation executor before LLM preparation (engine/batched.py:314-347), performs the LLM load on that executor (engine/batched.py:275-299), passes the same executor into AsyncEngineCore (engine/batched.py:647-653), and EngineCore uses the caller-supplied worker for stepping and cache I/O (engine_core.py:208-232, 663-673). Please rebase onto current main and do not reintroduce a second standalone worker or duplicate the cache-I/O changes that are already present there.

The rebase should also address the remaining shutdown boundary explicitly: current EngineCore.stop() still has a safety-net call to scheduler._close_batch_generator() on the event-loop thread (engine_core.py:170-187), even though the external generation worker owns the LLM model and the engine-loop cleanup already runs on that worker (engine_core.py:389-404). Prove or correct the owner-thread behavior for cancellation, normal stop, repeated stop, and cache save/load. Keep the MLLM path separate: current BatchedEngine routes MLLMScheduler work through its event-loop/model thread, so a blanket LLM-worker hop must not move MLLM cache operations to the wrong stream.

Please add focused tests with thread IDs and a real EngineCore/BatchedEngine wiring seam (not only fake scheduler dispatch) covering model preparation, scheduler stepping, cache save/load, cancellation, and shutdown cleanup. It is fine to keep the long-running MLX symptom behind the coordinated hardware gate; no inference run is requested for this unit proof. Then rerun current checks and report the exact mlx/mlx-lm versions or commits. This PR is currently CONFLICTING/DIRTY on stale base 7a64e27fdf04997efebb777ad9683998708c3e5e, with no current status checks reported, so its old state cannot qualify a rebased tree.

I am requesting changes; I am not approving or merging this revision.

waybarrios:
Updated this PR to current main. Prefix-cache load/save now stays on the model-owned generation worker, supplied workers are preserved during stream fallback, and cleanup always retires the executor. Added regression coverage and included the BatchedEngine tests in Apple Silicon CI.

waybarrios:
Thanks @Thump604 for the detailed review. @Teacher-Tayvin, I added a follow-up patch that keeps EngineCore and BatchedEngine shutdown cleanup on the MLX owner thread, including cancellation and failure paths, while preserving MLLM event-loop ownership. I also added focused tests for shutdown and cache I/O behavior. CI is awaiting approval to run.

PR #708, opened 2026-08-14
fix(mllm): evaluate batch cache metadata after filter/extend to stop lazy-graph handle leak

rohnach29:
## Summary

Fixes a Metal buffer-handle leak that aborts the server under sustained continuous-batching load with `[metal::malloc] Resource limit (499000) exceeded.` The limit is a per-process **handle count**, not bytes — memory metrics look healthy right up to the abort. Our production server died at this limit after 19h45m / ~9,400 completions.

Root cause: `MLLMBatch.filter()`/`extend()` rebuild per-layer cache metadata (`offset`, `left_padding`, `lengths`) as lazy MLX ops, but the decode loop evaluates only `y`/`logprobs`, and nothing in the decode graph reads `offset` (`make_mask` uses the Python-int `_idx`; `ArraysCache` metadata has no decode-time reader either). Each membership change extends an unevaluated lazy chain that pins every buffer that ever fed it, for the lifetime of the batch — and under continuous traffic the batch never drains.

Fix: `_sync_reshuffle_metadata()` — one batched `mx.eval` of the metadata arrays at the end of every `filter()`/`extend()`, recursing into `CacheList` for hybrid models (Qwen3.5/3.6). This mirrors mlx-lm's own guard on `BatchRotatingKVCache` (`mx.depends(self.keys, (self.left_padding, self.offset))`, `cache.py:1204`), which plain `BatchKVCache` never received. Same disease family as #641/#453/#413/#575.

Repro (no model needed, measured at revision `4d044d9` on base `7a64e27`/v0.4.1): 1000 join/leave churn cycles evaluating only what the decode loop evaluates → **17,000 handles pinned on main vs 0 with this PR**; evaluating the metadata alone releases 100% of them.

Follow-ups (out of scope): the durable cure is upstream in mlx-lm (`mx.depends` guard on `BatchKVCache` + eval in `ArraysCache`), which would also fix the same leak in the text path's `GenerationBatch`.

## Type of change

- Bug fix

## Surface touched

- Scheduler / batching
- KV cache / prefix cache

## Test plan

Model-free — tests use real `mlx_lm.models.cache` objects (verified at head `369469e` with mlx 0.32.2 / mlx-lm 0.31.3, Python 3.12, M-series).

```bash
pytest tests/test_batch_reshuffle_metadata.py -m "not slow" -k "not Integration"
black --check vllm_mlx/ tests/
ruff check vllm_mlx/ tests/ --select E,F,W --ignore E402,E501,E731,F811,F841

13 tests: six public-boundary regressions (filter, extend, hybrid CacheList with ArraysCache, BatchRotatingKVCache, and two sustained-churn tests — one under mx.stream(MLLMBatchGenerator._stream), one driving the real next()/_next() through join/leave churn with a deterministic model-free fixture), three unit tests of the sweep helper (recursion, None slots), and four value-correctness tests asserting the metadata is materialized immediately after each operation (zero pending edges via mx.export_to_dot) and holds correct values after a decode-style y/keys/values evaluation. To see them catch the bug, run the file on main — all 13 fail there.

## Test results

Branch (head 369469e): 13 passed (this file); full apple-silicon CI selection locally: 1005 passed, 12 skipped, 36 deselected plus 27 passed (stream-affinity block); black and ruff clean. On unpatched main (d843a76) all 13 fail — 10 on the leak assertions themselves, e.g.:

AssertionError: filter: BatchKVCache.offset was rewritten but never evaluated
by the public call — its lazy chain stays live and pins buffers

and 3 helper unit tests on the missing method. The boundary tests assert via an mx.eval spy plus a discovery-based pending-graph check that each membership change evaluates its metadata (assertions never touch the arrays before checking — observation would evaluate them and mask the leak). File is registered in the apple-silicon CI job.

## Performance impact

(Measured at revision 4d044d9 on base 7a64e27/v0.4.1; not re-run for the current head — the fix code is unchanged by the rebase.)

- Hardware: Apple Silicon dev machine (probes); Mac Studio M3 Ultra 512GB, Qwen3.6-35B-A3B-8bit (production)
- Benchmark: 200-cycle churn probe, 24 layers, 8 resident sequences, in-flight async_eval
- Delta vs main: no regression — slightly faster (97.1 vs 103.7 ms/cycle), because the growing chain makes filter()'s pre-existing .min().item() sync progressively costlier on main. Added cost is one tiny mx.eval per membership change, not per decode step. Production throughput unchanged (~28-33 tok/s); patched build in production since Jul 25 (pre-rebase revision).

## Output parity

Identical by construction — mx.eval changes when metadata is computed, never its values; no sampling, kernel, or tokenization code touched. The value-correctness tests pin the exact metadata values produced with the eager evaluation in place.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -191,6 +191,7 @@ jobs:
             tests/test_mllm_ssd_spill.py \
             tests/test_ssd_cache.py \
             tests/test_ssd_shutdown_wiring.py \
+            tests/test_batch_reshuffle_metadata.py \
             tests/test_lifecycle_cli.py \
             tests/test_lifecycle_server.py \
             tests/test_normalize_messages.py \
--- a/tests/test_batch_reshuffle_metadata.py
+++ b/tests/test_batch_reshuffle_metadata.py
@@ -1,0 +1,661 @@
+# SPDX-License-Identifier: Apache-2.0
+"""MLX-dependent regression tests for the batch-reshuffle metadata leak fix.
+
+Regression: ``MLLMBatch.filter()``/``extend()`` rebuild per-layer cache
+metadata (``offset``/``left_padding``/``lengths``) as lazy MLX ops, but the
+decode loop only evaluates ``y``/``logprobs`` — and ``offset`` is never
+read by the decode-time forward graph (``make_mask`` passes the Python-int
+``_idx`` as the mask offset; ``ArraysCache`` metadata likewise has no
+decode-time reader on hybrid models).  Left lazy, each membership change
+extends an unevaluated graph chain that transitively retains every buffer
+that ever fed it, for the lifetime of the batch.  Under continuous traffic
+the batch never drains, so retained Metal buffer handles grow until the
+per-process resource limit (499000) aborts the server.  The fix evaluates
+these small arrays after every ``filter()``/``extend()``.
+
+These tests use real ``mlx_lm.models.cache`` objects and run only on the
+apple-silicon CI matrix (and locally on M-series hardware); the Linux
+``test-matrix`` job excludes this file because MLX has no Linux
+distribution.
+"""
+
+import os
+import tempfile
+
+
+def _make_batch(cache):
+    import mlx.core as mx
+
+    from vllm_mlx.mllm_batch_generator import MLLMBatch
+
+    n = None
+    for c in cache:
+        if hasattr(c, "offset"):
+            n = c.offset.size
+            break
+    n = n or 1
+    return MLLMBatch(
+        uids=list(range(n)),
+        request_ids=[f"r{i}" for i in range(n)],
+        y=mx.zeros((n,), mx.int32),
+        logprobs=[mx.zeros((4,)) for _ in range(n)],
+        max_tokens=[16] * n,
+        num_tokens=[0] * n,
+        cache=cache,
+        requests=[None] * n,
+    )
+
+
+def _make_live_batch(n, layers=2, heads=2, seq=4, dim=4):
+    """A batch mid-decode: prefilled ``BatchKVCache`` layers, fully materialized."""
+    import mlx.core as mx
+    from mlx_lm.models.cache import BatchKVCache
+
+    caches = []
+    for _ in range(layers):
+        c = BatchKVCache([0] * n)
+        k = mx.zeros((n, heads, seq, dim), mx.float16)
+        c.update_and_fetch(k, k)
+        caches.append(c)
+    batch = _make_batch(caches)
+    mx.eval(batch.y, *[c.keys for c in caches], *[c.values for c in caches])
+    return batch
+
+
+def _install_eval_spy(monkeypatch):
+    """Route ``mx.eval`` through a spy that records every array it is passed
+    while still evaluating normally.  Lets tests assert *what* a public call
+    evaluated without ever touching the arrays themselves (observing a lazy
+    array evaluates it, which would mask the leak under test)."""
+    import mlx.core as mx
+
+    captured = []
+    real_eval = mx.eval
+
+    def spy(*arrays):
+        captured.extend(arrays)
+        return real_eval(*arrays)
+
+    monkeypatch.setattr(mx, "eval", spy)
+    return captured
+
+
+def _assert_metadata_evaluated(cache_obj, captured, context):
+    """Assert every metadata array on this (possibly nested) cache is among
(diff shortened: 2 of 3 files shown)


Thump604 (changes requested):
The leak mechanism is plausible, but this review cannot proceed on the submitted tree.

I verified the current main line at 22efb47968f96e6e7a79704f4eb0184cda4e0068: MLLMBatch.filter() still calls each cache's filter() without evaluating the rewritten metadata, and MLLMBatch.extend() still calls cache.extend() without an evaluation. Against the declared mlx-lm v0.31.3 API, BatchKVCache.filter()/extend() rebuild offset and left_padding, while ArraysCache.filter()/extend() rebuild left_padding and lengths; CacheList delegates to its children. That supports investigating this as a separate live-batch reshuffle issue, not as the prefix-cache work in #744 or the related #731.

Please:
1. Rebase onto current main and resolve the reported conflict before further maintainer review. The submitted head is 7aab499c237ffaf745c72901734351e512abf6e6 on base 7a64e27fdf04997efebb777ad9683998708c3e5e; the PR is currently CONFLICTING/DIRTY, so its old green checks do not qualify the current tree.
2. Keep the fix scoped to the MLLMBatch.filter()/extend() public boundary. Preserve tests for the actual mlx-lm cache shapes you recurse over: BatchKVCache, BatchRotatingKVCache, ArraysCache, CacheList children, empty/None slots, and the existing hybrid case. Add an assertion that the rewritten metadata is materialized after both operations and remains value-correct after a decode-style evaluation.
3. Add at least one current scheduler/generation-path regression (not only direct helper/spy tests) showing that continuous membership churn does not leave the metadata graph pending after the normal y/keys/values evaluation. Verify the call runs in the scheduler's MLX execution context; do not rely on a model/inference run for the unit proof.
4. Re-run the full current Apple-Silicon test selection, including the new test file, and report the exact mlx/mlx-lm versions or commits used. The long-running Metal-handle production symptom still needs a separately coordinated hardware window; this review does not authorize inference or a production claim.

After the rebase and current checks, please request another review with the updated head. I am not approving or merging this revision.

rohnach29:
@Thump604 I've rebased -- new head is a20ea32, with 4 new tests, making 12 total. All 12 pass on this branch and fail on main. After re-running the full current selection, Main MLX block: 1004 passed, 12 skipped, 36 deselected, Stream-affinity block: 27 passed. Versions: mlx 0.32.2, mlx-lm 0.31.3

Thump604:
Thanks, Rohan. I re-reviewed `a20ea326` and checked the Apple Silicon logs. The conflict is resolved, the new file is registered in CI, and all 12 tests ran on both Python 3.11 and 3.13. The real-cache coverage and pending-graph assertions address the metadata/value concerns from my earlier review.

One part of request 3 remains: [`test_churn_under_generation_stream`](https://github.com/waybarrios/vllm-mlx/blob/a20ea3269720be6a18a77fd4cd02779789a945e5/tests/test_batch_reshuffle_metadata.py#L288-L322) recreates the stream context around direct `extend()`/`filter()` calls. It does not enter the generator path that performs admission and retirement. Please extend that test, or point me to an existing one, that drives real `MLLMBatchGenerator.next()`/`_next()` through join/leave churn and checks pending metadata after the normal decode evaluation. A deterministic model-free fixture is sufficient; no new model benchmark or test framework is requested.

Please also refresh the PR body's older test counts/version details and label the earlier production/performance measurements with their original revision. Those measurements need not be rerun for this reply. I found no new correctness defect in the implementation during this source review; the remaining request is the previously specified call-path coverage.

rohnach29:
@Thump604 I added a new test and updated the PR body -- new head is 369469e, let me know if any other changes are needed!

On the mlx-lm side, I opened a PR but it looks like they were overwhelmed with PRs and closed ~300 out of the 400 to focus on the more critical ones.

Thump604 (approved):
Thanks for following through on this. I reviewed 369469e and the new regression closes my remaining generator-path request: it drives real next()/_next() admission, extension, retirement and filtering with real MLX cache objects, and checks pending graphs before inspecting values could hide the problem.

I also checked the executed Apple Silicon CI log: all 13 reshuffle tests, including the new generator test, passed, and pytest completed normally. All ten checks are green. The updated body clearly separates the historical handle/performance measurements from current-head test evidence.

I have no further changes to request for this scoped fix. No additional broad benchmark is needed from you for my review. This approval does not extend the model-free results into a current-head long-running production claim; I will handle compatibility with my local Runtime carries separately. Ready for maintainer review and merge.

PR #713, opened 2026-08-16
Report MLLM prefix caches in registry memory budgets

### LINKED ISSUE #712: Registry memory report ignores MLLM-path prefix caches under --use-paged-cache
## Summary

The startup memory-budget report added in #696 reports `prefix-cache maximum none
configured` under `--use-paged-cache`, but multimodal (MLLM) engines still build a
`MemoryAwarePrefixCache` sized by `--cache-memory-mb`. The reconciliation applies the
text-path rule to the whole registry, so on a registry of mostly-MLLM entries a real,
per-engine memory consumer is reported as zero.

This is the same accounting hole as #627, pointing the other way: #627 was a budget
that over-committed because it ignored the ceiling; this is a report that
under-states because it ignores a consumer.

## Where the two gates diverge

`build_memory_budget_report` requires paged caching to be **off** before it will
attribute anything to `cache_memory_mb` (`model_registry.py`):

```python
cache_applies = (
    scheduler_config is not None
    and continuous_batching_entries > 0
    and getattr(scheduler_config, "enable_prefix_cache", False)
    and not getattr(scheduler_config, "use_paged_cache", False)      # <-- text-path rule
    and getattr(scheduler_config, "use_memory_aware_cache", False)
)
```

The MLLM scheduler does not consult `use_paged_cache` at all
(`mllm_scheduler.py`):

```python
if self.config.enable_prefix_cache and self.config.use_memory_aware_cache:
    prefix_cache_config = MemoryCacheConfig(
        max_memory_mb=self.config.prefix_cache_memory_mb,
        ...
```

and `prefix_cache_memory_mb` is populated straight from the same flag
(`engine/batched.py`):

```python
prefix_cache_memory_mb = getattr(self._scheduler_config, "cache_memory_mb", None)
```

So with `--use-paged-cache --cache-memory-mb 30720`, text entries take the paged
cache (`--max-cache-blocks`) and correctly ignore `cache_memory_mb`, while every
resident MLLM entry gets a 30 GB memory-aware prefix cache that the report does not
mention.

## Observed

Both lines from the same server start, `v0.4.1`, registry mode,
`--use-paged-cache --cache-memory-mb 30720 --gpu-memory-utilization 0.88`:

```
INFO vllm_mlx.model_registry: Registry memory budget: 68.0 GB of model weights; Metal
     allocation ceiling 98.6 GB (88% of 112.0 GB, from serve default); prefix-cache
     maximum none configured
INFO vllm_mlx.memory_cache:  MemoryAwarePrefixCache initialized: max_memory=30720.0MB,
     max_entries=1000
```

## Why it matters

The cache is per resident engine, so the aggregate scales with residency while the
reported figure stays at zero. On a 9-entry registry where most entries load MLLM
(anything whose config carries a `vision_config` — including models with no
`mllm: true` in the registry, since the loader auto-detects), three resident models
is up to 90 GB of prefix cache against a 98.6 GB ceiling, with the budget line
reading `none configured` throughout.

## Suggested direction

Two separable pieces, happy to send a PR for the first:

1. **Report what is actually built.** The attribution needs to be per resolved entry
   rather than one global rule: an entry that will load MLLM attributes
   `cache_memory_mb` regardless of `use_paged_cache`; a text entry attributes it only
   under the existing condition. That also means the reported figure should be a sum
   over resident-capable entries, or at least state the per-engine number *and* the
   number of entries it can apply to, the way the current line does for
   continuous-batching entries.

2. **A per-entry cache size.** `RegisteredModel` currently exposes
   `gpu_memory_utilization`, `enable_mtp`, `prefill_step_size` and friends, but no
   cache field, and `_clone_scheduler_config` copies the global config verbatim — so
   there is no way to give a small text model a small prefix cache and a large VLM a
   large one. This is the same operator-control gap as #640, one level down.

Point 1 is a straightforward accounting fix. Point 2 is a feature and I am happy to
have it split out or declined.

## Environment

- vllm-mlx 0.4.1, mlx-vlm 0.6.13, mlx-lm 0.31.3, mlx 0.32.0
- macOS on M-series, 128 GB, `iogpu.wired_limit_mb=114688`
- Registry mode, 9 entries, `manager.memory_budget_gb: 68`

waybarrios:
## Summary

Fixes #712 by making the registry startup memory report match the prefix-cache path used by each engine.

Text engines using `--use-paged-cache` remain excluded from memory-aware cache attribution. MLLM engines are included because they continue to construct `MemoryAwarePrefixCache` even when paged cache is enabled.

## What changed

- Count memory-aware prefix-cache consumers per registry entry.
- Preserve per-entry and serve-default `mllm` precedence.
- Detect local MLLM sources from their authoritative `config.json`.
- Require `mllm: true` for unresolved remote IDs instead of using model-name heuristics for capacity warnings.
- Report the per-engine cache limit and applicable entry count.
- Propagate `--cache-memory-percent` through `BatchedEngine` and `MLLMSchedulerConfig` into the MLLM `MemoryCacheConfig`.
- Apply runtime scheduler defaults when an entry enables continuous batching while the global default is disabled.
- Document text, MLLM, paged-cache, and unresolved remote behavior.

The cache maximum remains informational. It is not subtracted from the process-wide model-weight budget because it is a lazy, per-engine maximum whose aggregate depends on model residency.

## Validation

CI passed on the final branch head:

- Ruff and Black
- mypy
- Linux tests on Python 3.10, 3.11, 3.12, and 3.13
- focused Python 3.14 tests
- Apple Silicon tests on Python 3.11 and 3.13
- EngineCore stream-affinity regression tests

The focused regression coverage includes local MLLM detection, explicit remote MLLM declaration, paged text exclusion, cache percentage propagation, entry/default override precedence, entry-level batching defaults, and operator-facing report output.

CI run: https://github.com/waybarrios/vllm-mlx/actions/runs/33978815366

## Runtime impact

The runtime change is limited to honoring the configured cache-memory percentage on the existing MLLM memory-aware prefix cache. Model execution, scheduling order, tokenization, and generated output are unchanged.

Closes #712

### DIFF
--- a/docs/guides/model-registry.md
+++ b/docs/guides/model-registry.md
@@ -126,8 +126,13 @@ together with the prefix-cache setting:
 ```
 Registry memory budget: 68.0 GB of model weights; Metal allocation ceiling
 64.0 GB (50% of 128.0 GB, from serve default); prefix-cache maximum
-20.0 GB per continuous-batching engine (--cache-memory-mb, 2 of 3 entries)
+20.0 GB per memory-aware prefix-cache engine (--cache-memory-mb, 2 of 3 entries)
 ```
+
+The entry count follows the cache path each model will use. With
+`--use-paged-cache`, text engines use the paged cache and are excluded from this
+count. MLLM engines still construct a memory-aware prefix cache, so their
+per-engine `--cache-memory-mb` maximum remains included.
 
 When the weights budget alone does not fit below the ceiling, startup warns:
 
--- a/docs/guides/model-registry.md
+++ b/docs/guides/model-registry.md
@@ -155,6 +155,10 @@ per-engine `--cache-memory-mb` maximum remains included. For an unresolved
 remote model ID, set `mllm: true` on the registry entry so the startup report
 can include its cache without relying on a model-name heuristic.
 
+If an entry enables continuous batching while the global serve default leaves
+it disabled, the report uses the same scheduler defaults as the engine,
+including the default 20% memory-aware cache limit.
+
 When the weights budget alone does not fit below the ceiling, startup warns:
 
 ```
--- a/docs/guides/model-registry.md
+++ b/docs/guides/model-registry.md
@@ -132,7 +132,9 @@ Registry memory budget: 68.0 GB of model weights; Metal allocation ceiling
 The entry count follows the cache path each model will use. With
 `--use-paged-cache`, text engines use the paged cache and are excluded from this
 count. MLLM engines still construct a memory-aware prefix cache, so their
-per-engine `--cache-memory-mb` maximum remains included.
+per-engine `--cache-memory-mb` maximum remains included. For an unresolved
+remote model ID, set `mllm: true` on the registry entry so the startup report
+can include its cache without relying on a model-name heuristic.
 
 When the weights budget alone does not fit below the ceiling, startup warns:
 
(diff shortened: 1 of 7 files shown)


waybarrios:
@danmackinlay could you take a look on this? This is a potential draft to solve the issue you mentioned on #712

PR #714, opened 2026-08-16
fix(cache): probe system prefix from normalized messages so agent tool turns cache

iotashan:
## Summary

On SimpleEngine (i.e. without `--continuous-batching`), the system-KV cache silently stops working for **every conversation turn that follows a tool call**. The first turn caches correctly; each turn after re-prefills the entire system prompt. For an agent client that prompt is ~20k tokens of instructions plus tool schemas, which is **60-80 s per turn** on an M3 Ultra at ~300 tok/s.

Nothing reports this. When the eligibility check fails, `kv_cache_eligible` stays `False` and the request falls through to the uncached path **without logging anything**, and a request that never reaches the cache branch increments no counter in `/v1/cache/stats` — so the hit ratio still looks healthy while describing a shrinking share of traffic.

> **Attribution: this bug was found, diagnosed, and fixed entirely by Claude** (Anthropic), running in Claude Code during an autonomous performance-tuning session. Every measurement below was produced by Claude. Flagging this explicitly at the request of the human operator, who supplied the hardware and the decision to publish, and wrote none of the diagnosis or the patch.

## Reproducing

Send the same multi-turn tool-calling request twice:

```
req1 61.8s
req2 61.5s     <- identical input, zero speedup, and no cache log line at all
```

The discriminating control — same 3-message conversation, same tools, only the content encoding differs:

| Message content encoding | 1st | 2nd | Server log |
|---|---|---|---|
| `tool_use` / `tool_result` blocks | 61.8 s | 61.5 s | *(nothing)* |
| plain strings | 61.4 s | **1.13 s** | `STORED` -> `HIT` |

## Root cause

The probe in `_stream_chat_impl` renders the chat template twice with two different user contents ("Alpha"/"Bravo"), takes the divergence point, and requires the result to be a token prefix of the real prompt. Two independent defects break that for tool turns.

**1. The probe renders a different message list than the prompt.** The prompt is rendered from `safe_messages = normalize_messages_for_chat_template(messages)` (line ~1348); the probe from the raw `messages` (line ~1511). Normalization is not cosmetic — it converts tool-call `arguments` from JSON **strings** into **mappings**. Without it the Qwen template raises `TypeError: Can only get item pairs from a mapping` while rendering a tool call, which the probe's bare `except Exception: pass` swallows, leaving `rendered_a is None`.

**2. The probe appends a synthetic user turn.** Fixing (1) alone still leaves turns at ~65 s. `_with_user` only substitutes content when the last message is a `user` message; a `tool_result` converts to a **`role="tool"`** message, so the helper *appends* a user turn and renders a conversation that does not exist.

Replaying the probe offline against the real tokenizer:

```
converted messages: system, user, assistant, tool    <- last role is "tool"

=== full-conversation probe (before) ===
  full=18682 sys=18680 prefix_valid=False
  probe prefix is a str-prefix of prompt: False
  first token mismatch at index 18678
   full: ' pass\n</tool_response><|im_end|>\n<|im_start|>assistant\n<think>\n'
   sys : ' pass\n</tool_response><|im_end|>\n<|im_start|>user\n'

=== system-only probe (after) ===
  full=18682 sys=18523 prefix_valid=True
  probe prefix is a str-prefix of prompt: True
```

## The fix

Probe the **normalized** messages, restricted to the **system** messages. The system span is also the only part of a conversation stable across turns, so it is the correct thing to key a cross-turn cache on. `messages_for_cache` is used only by the probe, so narrowing it is contained; `safe_messages` is initialised to `None` before the `apply_chat_template` branch so the non-template fallback still works.

34 lines added, 2 removed, most of it explanatory comment.

## Verification

Hardware: **Mac Studio M3 Ultra, 256 GB**, macOS 27.0. Model: `qwen3.8-27b-q8` (Qwen3.5 hybrid, 64 layers). Client: Claude Code against the Anthropic-compatible endpoint.

| Probe | Before | After |
|---|---|---|
| Identical multi-turn `tool_result` request, 3x | 61.8 / 61.5 / 63.4 s | 63.4 / **1.68** / **1.69 s** |
| Claude Code 2-tool write task, median of 3 | 331.7 s | **266.9 s** |
| Task output correctness | 5/5 | 5/5, 5/5, 5/5 |

Log now reads `System KV cache HIT (stream_chat): reusing 18523 tokens, prefilling 159 new` where previously there was no line at all.

`ruff check --select E,F,W --ignore E402,E501,E731,F811,F841` and `black --check` both pass on the changed file.

**On regression coverage:** the probe is inline in a large async generator whose cached path needs a real MLX model to reach, so I did not find a clean seam for a unit test without a refactor larger than the fix. Happy to extract the probe into a testable helper and add coverage if you'd prefer that shape — say the word and I'll push it.

## Scope, honestly

The end-to-end task improved 20%, not 37x, because the bottleneck moves once the cache works: long generations become decode-bound (~15.5 tok/s here), and Claude Code injects per-turn `<system-reminder>` content that changes the merged system block's hash on some turns, causing genuine misses. A cross-turn longest-common-prefix cache would recover those; that is a much larger change and is not attempted here.

## Related findings (happy to file separately)

Found while diagnosing this; not touched by this PR:

- `vllm_mlx_cache_hits` / `cache_type` in `metrics.py` only look for the stats keys `memory_aware_cache` / `paged_cache` / `prefix_cache`. SimpleEngine reports under `system_kv_cache`, so in simple mode the gauge reports `cache_type="none"` and `cache_hits 0` regardless of real behaviour.
- `--continuous-batching` appears to disable this cache entirely: 3 identical streaming requests cost 56.6 / 56.0 / 56.2 s with `hits 0` after 15 requests, versus ~1.1 s on SimpleEngine.
- `--max-request-tokens` bounds output tokens only; there is no prompt-length validation. A ~152k-token prompt prefilled for 300 s+ with no error, and generation continued after the client disconnected, holding the lock 343 s+.
- Setting `--max-kv-size`, `--specprefill` or `--enable-mtp` silently disables the cache via `cache_blocking_controls` — note `--help` currently recommends `--max-kv-size >= 32768` for reasoning models.

Full writeup: https://gist.github.com/iotashan/954befbf6f2c821b16b8d30186694fa6

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,6 +83,10 @@ jobs:
             tests/test_anthropic_models.py \
             tests/test_anthropic_adapter.py \
             tests/test_dependency_floors.py \
+            tests/test_ssd_cache_shutdown.py \
+            tests/test_chat_template_safety.py \
+            tests/test_simple_engine_thread_pinning.py \
+            tests/test_engine_base.py \
             tests/test_harmony_parsers.py \
             tests/test_endpoint_model_policies.py \
             tests/test_truncation.py \
@@ -111,6 +115,28 @@ jobs:
         with:
           file: ./coverage.xml
           fail_ci_if_error: false
+
+  test-python-3-14:
+    # Focused job for the CPython 3.14+ asyncio.shield() behavior change
+    # that shield_task() (vllm_mlx/engine/base.py) works around -- the
+    # regular test-matrix only covers 3.10-3.13, none of which reproduce it.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python 3.14
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.14"
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install pytest anyio pytest-cov
+
+      - name: Run engine base tests
+        run: |
+          pytest tests/test_engine_base.py -v --tb=short
 
   test-apple-silicon:
     runs-on: macos-14
@@ -145,15 +171,22 @@ jobs:
             tests/test_paged_cache.py \
             tests/test_prefix_cache_untrimmable.py \
             tests/test_qwen35_mtp_patch.py \
+            tests/test_batched_engine_mllm_config.py \
             tests/test_mllm_continuous_batching.py \
             tests/test_mllm_cache.py \
             tests/test_optimizations.py \
             tests/test_simple_engine.py \
+            tests/test_simple_engine_prefix_trie_cache.py \
             tests/test_text_model_dispatch.py \
+            tests/test_cli.py \
+            tests/test_model_registry.py \
             tests/test_chat_template_kwargs.py \
             tests/test_batching.py \
             tests/test_continuous_batching.py \
             tests/test_memory_cache_mlx.py \
+            tests/test_mllm_ssd_spill.py \
+            tests/test_ssd_cache.py \
+            tests/test_ssd_shutdown_wiring.py \
             tests/test_lifecycle_cli.py \
             tests/test_lifecycle_server.py \
             tests/test_normalize_messages.py \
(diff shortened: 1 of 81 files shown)


waybarrios:
Resolved the cache-selection interaction introduced by stable system-prefix probing. SimpleEngine now compares reusable prefix lengths before materializing trie state, avoids copying non-winning or untrimmable exact entries, and preserves the empty-trie fast path.

waybarrios:
I rechecked this PR and it found evidence to accept it

PR #720, opened 2026-08-21
Add capacity envelope benchmark artifacts

Thump604:
## Summary

Adds an HTTP-only capacity-envelope mode to `vllm-mlx bench-serve` for measuring sustainable serving capacity on Apple Silicon.

The new mode produces a versioned JSON evidence artifact across configurable concurrency, prompt-size, output-size, and cache-state sweeps. It does not change model loading, scheduling, generation behavior, or serving defaults.

## What it measures

- p50/p95/p99 TTFT, content-chunk gap, and end-to-end latency
- Successful completion-token throughput and requests per second
- Request failures, timeouts, OOMs, and validation failures
- Cold, model-warm, prefix-hit, and prefix-miss behavior
- Sustainable concurrency under configurable SLOs
- Sustainable concurrency/GiB and throughput/GiB
- Server-reported memory and optional MTP, speculative-decoding, queue, and cache telemetry

## Evidence integrity

The artifact:

- Fails closed when authoritative token usage is unavailable
- Does not count failed requests toward successful throughput
- Treats output-limit completions as valid benchmark results
- Represents unavailable telemetry as unavailable rather than zero
- Verifies cache reset state and observed prefix hit/miss behavior
- Records model, runtime, software, hardware, source-commit, and workload provenance
- Includes deterministic workload and provenance hashes
- Labels SSE timing as `chunk_gap_ms`, not token TPOT
- Optionally performs greedy A/B output checks when baseline model identity can be verified

## Usage

```bash
vllm-mlx bench-serve \
  --url http://localhost:8000 \
  --capacity-envelope \
  --capacity-concurrency 1,2,4,8,16 \
  --capacity-prompt-tokens 256,2048 \
  --capacity-output-tokens 128 \
  --capacity-cache-modes cold,warm,prefix-hit,prefix-miss \
  --capacity-max-p95-ttft-ms 1000 \
  --capacity-max-p95-e2e-ms 10000 \
  --output capacity.json
```

## Compatibility

The capacity-envelope schema and runner are additive. Existing prompt-sweep and declarative-workload behavior remains unchanged.

## Verification

- Focused benchmark tests: 121 passed, 2 skipped
- Complete suite: 2,677 passed, 25 skipped, 27 standard deselections
- Black, Ruff, and `git diff --check`: passed
- Independent adversarial review: no remaining P0/P1 findings

## Scope

This PR contains benchmarking and evidence-generation work only. It does not include serving optimizations, Metal kernels, scheduler changes, or competitive benchmark claims.

### DIFF
--- a/docs/guides/capacity-envelope.md
+++ b/docs/guides/capacity-envelope.md
@@ -1,0 +1,44 @@
+# Capacity-envelope benchmarking
+
+`bench-serve --capacity-envelope` measures how much concurrent work a running
+OpenAI-compatible server can sustain under explicit latency and failure SLOs.
+It is an HTTP-only client: it does not load a model or change serving defaults.
+
+```bash
+vllm-mlx bench-serve --url http://localhost:8000 \
+  --capacity-envelope \
+  --capacity-concurrency 1,2,4,8,16 \
+  --capacity-prompt-tokens 256,2048 \
+  --capacity-output-tokens 128 \
+  --capacity-cache-modes cold,warm,prefix-hit,prefix-miss \
+  --capacity-max-p95-ttft-ms 1000 \
+  --capacity-max-p95-e2e-ms 10000 \
+  --output capacity.json
+```
+
+The versioned JSON artifact includes the canonical workload hash, runtime and
+host provenance, request samples, p50/p95/p99 cell summaries, successful-token
+throughput, request failures (including timeout and OOM classifications), cache
+actions, optional server telemetry, sustainable concurrency, and throughput per
+GiB. Missing telemetry is `null`/unavailable; it is never estimated as zero.
+
+## Measurement rules
+
+- A cell fails when any request exceeds the configured failure allowance.
+- Throughput uses successful completion tokens and measured batch wall time.
+- `chunk_gap_ms` measures content-SSE chunk gaps because an HTTP chunk is not
+  necessarily one token. Aggregate token throughput uses server `usage` counts.
+- Every cache case starts with a successful `DELETE /v1/cache`; the run aborts
+  if the server cannot verify that reset. Cold cells do not run a warmup request.
+- Warm cells exercise model-warm/no-prefix-hit behavior by warming a distinct
+  prompt. Prefix-hit cells warm the measured prompt; prefix-miss cells warm a
+  deterministic alternate prefix before measuring the requested prompt.
+- Memory and optional MTP/speculative/cache counters include their observation
+  source. Unsupported or inaccessible endpoints remain unavailable.
+- Optional A/B parity uses greedy generation only. It compares token IDs when
+  the server emits them and otherwise compares a decoded-output SHA-256 hash.
+
+Prompt-token targets are workload buckets, not tokenizer claims. Each sample
+records the server-reported prompt-token count as the authoritative observation.
+Only compare artifacts whose workload and provenance describe compatible model,
+quantization, software, cache, sampling, and hardware conditions.
--- a/docs/guides/capacity-envelope.md
+++ b/docs/guides/capacity-envelope.md
@@ -26,6 +26,9 @@ GiB. Missing telemetry is `null`/unavailable; it is never estimated as zero.
 
 - A cell fails when any request exceeds the configured failure allowance.
 - Throughput uses successful completion tokens and measured batch wall time.
+- A response that reaches the requested output limit is successful. A response
+  without authoritative OpenAI `usage` token counts fails instead of producing
+  zero-token throughput.
 - `chunk_gap_ms` measures content-SSE chunk gaps because an HTTP chunk is not
   necessarily one token. Aggregate token throughput uses server `usage` counts.
 - Every cache case starts with a successful `DELETE /v1/cache`; the run aborts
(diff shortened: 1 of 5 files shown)


janhilgard:
Ran this against a live server rather than only reading it, and the evidence-integrity behaviour holds up — with one portability finding worth deciding on.

## What I ran

An OpenAI-compatible server that is **not** vllm-mlx (oMLX, serving an MLX model on this M3 Ultra), because "HTTP-only" suggested the mode should work anywhere the chat-completions API does.

All four cache modes stop before producing an artifact:

```
RuntimeError: cold cache state could not be verified:
  Client error '404 Not Found' for url 'http://localhost:8000/v1/cache'
```

Same for `warm`, `prefix-hit` and `prefix-miss`.

## The good half

This is exactly the behaviour the PR promises, and I want to be clear that I think it is the right call: the harness refused to emit a capacity artifact whose cache-state provenance it could not substantiate, instead of quietly labelling an unverified run as `cold`. A benchmark that fabricates its own preconditions is worse than no benchmark, and this one does not.

The error also names the endpoint and the mode, so the cause took seconds to find rather than an afternoon.

## The finding

`/v1/cache` is a vllm-mlx endpoint. In practice the capacity-envelope mode is vllm-mlx-only, while "HTTP-only" and the `--url` flag both read as portable. Two ways to close that, and the choice is yours:

1. **Document the constraint** — say in `docs/guides/capacity-envelope.md` that cache-state sweeps require a vllm-mlx server, and fail with that sentence rather than a bare 404.
2. **Degrade instead of refusing** — when the endpoint is absent, run the sweep with cache state recorded as `unverified` in the artifact. That is the same principle the PR already applies elsewhere ("represents unavailable telemetry as unavailable rather than zero"), just extended to cache state, and it would let the harness measure third-party servers for comparison.

I lean towards (2) because cross-runtime comparison is where I would most want this tool — Apple Silicon has several MLX servers now and no shared capacity methodology, and this artifact schema is the closest thing I have seen to one. But (1) is a legitimate scoping decision and cheaper.

## Note on the methodology

Labelling SSE timing `chunk_gap_ms` instead of TPOT, not counting failed requests toward throughput, and treating output-limit completions as valid are all the right calls. I have been burned by the first one specifically: on this machine wall-clock throughput drifts by more than 2x with background load, so anything that is not a paired or provenance-stamped measurement is close to worthless a day later. The provenance hashes in the artifact are the part I would keep even if everything else changed.

Not blocking on my side — the fail-closed behaviour is correct as written, and the portability question is a scoping call rather than a defect.

Thump604:
Thanks for testing this against oMLX. You found a real scope ambiguity. `HTTP-only` means the client does not load MLX or alter model execution locally, but the current controlled cache sweep does require vllm-mlx cache-control endpoints.

For this PR I will keep the fail-closed behavior and document that requirement. I will also replace the raw 404-oriented error with a direct compatibility message naming `DELETE /v1/cache` and the cache-state endpoint.

I do not want the four existing modes to degrade silently. Missing telemetry can be recorded as unavailable. Cache state is different because it is a benchmark precondition. A result labelled `cold`, `warm`, `prefix-hit`, or `prefix-miss` should not be emitted when the harness cannot establish that state.

Cross-runtime comparison is still valuable. A separate explicit `unmanaged` mode is the better follow-up. Its artifact would state that cache state was not controlled and would make no cold or prefix claim. That keeps the existing evidence contract intact while allowing comparisons with other OpenAI-compatible servers.

Thanks as well for checking the SSE timing, failure accounting, output-limit handling, and provenance behavior on a live server.

janhilgard (approved):
Approving. Thanks for the clarification on `HTTP-only` — with that scope stated, the harness does what it claims, and it is additive: `bench_serve.py` plus a `bench-serve` subcommand, nothing on the serving path.

The parts I think earn their weight are the provenance and verification machinery rather than the measurement loop itself — `_capacity_hardware_fingerprint()`, `_capacity_source_commit()`, `_capacity_package_version()` and `_verified_capacity_cache_reset()` are what make a capacity number quotable later. Benchmarks without that get compared across machines and builds and quietly mislead; a cache mode asserted rather than assumed (`_verify_capacity_cache_mode()`) is the difference between a measurement and an anecdote. `compare_capacity_outputs()` gating on `_capacity_baseline_identity()` is the same instinct applied to regressions, and 460 lines of tests for a benchmark harness is unusual in a good way.

One defect worth fixing, small:

`_positive_csv()` raises `ValueError` and nothing catches it. `bench_serve_command()` is dispatched directly from `main()` with no wrapper, so a typo in a flag surfaces as a raw Python traceback:

```
vllm-mlx bench-serve --capacity-envelope --capacity-concurrency 1,x
```

Everywhere else `cli.py` handles bad input as `print("Error: ...")` + `sys.exit(1)` (there are a dozen instances in `validate_serve_args` alone). Either matching that pattern here, or moving the parsing into `type=` callables on the arguments so `argparse` formats the error, would keep it consistent — the second is probably less code.

Same for `--capacity-cache-modes`: the values are lowercased and split but never validated against the accepted set, so an unrecognised mode is carried into `CapacityConfig` and fails later, further from the typo.

Method note: static review. I have not run the harness — this box is memory-committed to another workload right now, and a capacity sweep is precisely the thing that needs the machine to itself to mean anything.

PR #721, opened 2026-08-21
Metal context kernel foundation

Thump604:
## Summary

Introduces the first, deliberately non-serving slice of the Metal Context Engine:

- optional Objective-C++/Metal extension and precompiled metallib build path;
- BF16 paged-decode attention kernel for Qwen-style GQA, head dimension 128, block sizes 16/32, noncontiguous page tables, and partial tail blocks;
- narrow Python backend/capability contract plus NumPy and MLX SDPA correctness oracles;
- `--attention-backend mlx|metal-context|auto`, with `mlx` still the default, `auto` conservatively selecting MLX, and explicit `metal-context` failing closed until an executor is integrated;
- backend status in health/status output;
- source-distribution and Apple Silicon native-build validation.

This does **not** replace the scheduler, own production KV pages, or claim a serving speedup. The native bridge currently copies host buffers and reports `serving_ready=false`; scheduler/page-runtime integration belongs in later PRs.

## Safety and unsupported configurations

The kernel rejects non-Apple-Silicon environments, incompatible ABI, missing/corrupt metallibs, non-native buffer formats, malformed dimensions, invalid page tables, non-finite or overflow-risk inputs, head dimensions other than 128, block sizes other than 16/32, and non-BF16 K/V storage. Existing pure-Python/MLX installations remain functional without compiling the extension.

## Verification

- Full local suite: `2687 passed, 30 skipped, 27 deselected`
- Focused backend/native surface: `32 passed, 5 skipped`
- Ruff, Black, type check, and `git diff --check`: pass
- source distribution contains the native source, shader, build script, and tests
- Objective-C++ host syntax check: pass
- two Luna Max adversarial review cycles plus post-fix verification; final verdict: `SHIP`
- GitHub CI: all jobs green on Python 3.10–3.13, including strict native/metallib compilation on arm64 macOS

GitHub's hosted arm64 macOS workers do not expose a Metal device, so capability probing correctly fails closed there after compilation. Actual GPU numerical dispatch remains a required device/self-hosted qualification before any serving integration or performance claim. Capacity/performance artifacts begin when a serving path exists and will use the capacity-envelope schema from #720 (or its accepted successor).

## Stack

1. This PR: native kernel and ABI foundation
2. Page runtime
3. Dense executor
4. Concurrency and Qwen3 MoE
5. Persistence and tiering
6. Automatic dispatch and packaging qualification

Please do not merge yet: this is a draft for architectural review. Wayner retains final architectural and merge approval.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -176,7 +176,13 @@ jobs:
           assert platform.machine() == "arm64"
           capabilities = _metal_context.capabilities()
           assert capabilities.get("compiled") is True, capabilities
-          assert capabilities.get("available") is True, capabilities
+          # GitHub's arm64 macOS hosts expose the Apple toolchain but may not
+          # expose a Metal device to jobs.  Compilation is mandatory here;
+          # dispatch tests run when a device is actually present and otherwise
+          # must report the precise fail-closed reason.
+          if not capabilities.get("available"):
+              assert capabilities.get("metal_device") is False, capabilities
+              assert "no GPU" in capabilities.get("reason", ""), capabilities
           print("Metal Context native capability:", capabilities)
           PY
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,33 @@ jobs:
       - name: Run black
         run: black --check vllm_mlx/ tests/
 
+      - name: Verify native source distribution manifest
+        run: |
+          pip install build
+          rm -rf /tmp/vllm-mlx-sdist
+          python -m build --sdist --outdir /tmp/vllm-mlx-sdist
+          python - <<'PY'
+          import glob
+          import tarfile
+
+          archive = glob.glob("/tmp/vllm-mlx-sdist/*.tar.gz")
+          assert len(archive) == 1, archive
+          with tarfile.open(archive[0], "r:gz") as source:
+              names = source.getnames()
+          required = {
+              "native/metal_context/kernels/paged_decode.metal",
+              "native/metal_context/src/python_module.mm",
+              "native/metal_context/README.md",
+              "scripts/build_metal_context.py",
+              "tests/test_attention_backend.py",
+              "tests/test_metal_context_native.py",
+          }
+          missing = {
+              path for path in required if not any(name.endswith(path) for name in names)
+          }
+          assert not missing, f"sdist is missing: {sorted(missing)}"
+          PY
+
   type-check:
     runs-on: ubuntu-latest
     steps:
@@ -63,7 +90,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install pytest anyio pytest-cov pydantic fastapi jsonschema httpx psutil transformers requests
+          pip install pytest anyio pytest-cov pydantic fastapi jsonschema httpx psutil transformers requests numpy
 
       - name: Run unit tests (no MLX required)
         run: |
@@ -89,6 +116,7 @@ jobs:
             tests/test_endpoint_model_policies.py \
             tests/test_truncation.py \
             tests/test_cli_arg_types.py \
+            tests/test_attention_backend.py \
             tests/test_gemma4_openai_format.py \
             tests/test_gemma4_streaming_edge.py \
             tests/test_gemma4_tool_parser.py \
@@ -128,6 +156,8 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install project and dependencies
+        env:
+          VLLM_MLX_BUILD_METAL_CONTEXT: "1"
         run: |
           python -m pip install --upgrade pip
           pip install -e ".[dev,vision,harmony]"
@@ -137,10 +167,25 @@ jobs:
           python -c "import platform; assert platform.machine() == 'arm64', 'Not ARM64'"
           python -c "import mlx.core as mx; print(f'MLX OK: {mx.default_device()}')"
 
+      - name: Verify strict native Metal build
+        run: |
+          python - <<'PY'
+          import platform
+          from vllm_mlx import _metal_context
+
+          assert platform.machine() == "arm64"
+          capabilities = _metal_context.capabilities()
+          assert capabilities.get("compiled") is True, capabilities
+          assert capabilities.get("available") is True, capabilities
+          print("Metal Context native capability:", capabilities)
+          PY
+
       - name: Run MLX-dependent tests
         run: |
           pytest \
             tests/test_platform.py \
+            tests/test_attention_backend.py \
+            tests/test_metal_context_native.py \
             tests/test_llm.py \
             tests/test_mllm.py \
             tests/test_server.py \
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,14 @@
+include LICENSE
+include README.md
+include MANIFEST.in
+include pyproject.toml
+include setup.py
+
+# Optional Metal Context Engine source distribution inputs.
+recursive-include native/metal_context *.metal *.mm *.md
+include scripts/build_metal_context.py
+
+# Keep the portable/native foundation regression tests available to downstream
+# source checkouts without broadening the package's installed data files.
+include tests/test_attention_backend.py
+include tests/test_metal_context_native.py
--- a/README.md
+++ b/README.md
@@ -41,6 +41,24 @@ export ANTHROPIC_API_KEY=not-needed
 claude
 ```
 
+### Optional Metal Context foundation
+
+The normal package and editable installs keep the experimental compiled Metal
+Context foundation disabled.  This preserves the portable MLX path and avoids
+silently requiring an Xcode/Metal toolchain.  On an Apple Silicon Mac with
+Xcode installed, opt in explicitly when building from source:
+
+```bash
+VLLM_MLX_BUILD_METAL_CONTEXT=1 python -m pip install -e .
+```
+
+The strict build rejects non-Apple-Silicon hosts or missing Metal tools.  The
+foundation currently provides a validated kernel/oracle surface only; it is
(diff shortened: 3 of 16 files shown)


janhilgard:
Reviewed the diff and ran what I could on real hardware: M3 Ultra, macOS, **Command Line Tools only — no Xcode**. That turns out to be a useful configuration to report, because it is the one where the optional extension cannot be built at all.

## Verified on device

The fail-closed contract behaves exactly as documented:

```
resolve('mlx')           -> mlx,  fallback=False, probing skipped entirely
resolve('auto')          -> mlx,  fallback=True, reason: "automatic Metal Context routing is
                            disabled until the explicit qualification matrix passes"
resolve('metal-context') -> AttentionBackendCapabilityError
```

`discover_capabilities()` reports `available=False, serving_ready=False` with a reason that names every missing piece and how to get it. That is a genuinely good error string — it told me what was wrong without reading the source, which is rare.

The package imports and works with the extension absent, as claimed.

## What I could not verify

The GPU numerical qualification you flag as required. `xcrun metal` is not part of Command Line Tools and `xcodebuild -downloadComponent MetalToolchain` needs full Xcode, so the metallib cannot be compiled here. Nobody has executed this kernel on a Metal device yet, and I could not change that. If a self-hosted runner is not on the cards, I can install Xcode and run the numerical oracles against the MLX SDPA reference — say the word.

## One rough edge in the build path

`scripts/build_metal_context.py` raises an uncaught `RuntimeError` and prints a traceback when the `metal` tool is missing:

```
RuntimeError: xcrun: error: unable to find utility "metal", not a developer tool or in PATH
```

Command Line Tools without Xcode is a common developer setup, and this is the documented build step. The capability path handles the same situation with a helpful sentence; the build script could reuse that tone rather than a stack trace.

## Two notes on the kernel

**1. The `tid >= 128` guard is unsafe in exactly the case its comment anticipates.**

```c
if (request >= p.batch_size || query_head >= p.query_heads || tid >= 128) {
    return;
}
```

`request` and `query_head` are uniform across the threadgroup, so those two are fine. `tid` is not. The comment says the guard "prevents accidental writes if a future caller uses a larger grid" — but with a threadgroup larger than 128, lanes 128+ return before the `threadgroup_barrier` calls that the remaining lanes wait on, which is undefined behaviour rather than a safe rejection. Today it is dead code because the host dispatches exactly 128 threads; the concern is that it reads as protection for a case where it would make things worse. A uniform reject (e.g. validated host-side, or a params field the whole group tests) would hold up.

**2. Parallelism is across `(request, query_head)` only, not across KV blocks.**

Each threadgroup walks the whole sequence one token at a time, with a 7-step barrier reduction per token. For a decode-shaped workload — batch 1, 24 query heads on a Qwen3-class model — that is 24 threadgroups on a GPU with far more cores than that, each serialising `seq_len` iterations with ~9 barriers apiece.

Long-context decode is the case where a paged kernel is supposed to win, and it is the case this structure is weakest on. The usual shape is to split the KV range across threadgroups and combine partial softmax states in a second pass, which trades a small combine step for real occupancy. I realise phase 1 is deliberately about the ABI and correctness rather than speed — but since the stack lists "dense executor" and "concurrency" as later PRs and not "reparallelise the kernel", it seemed worth raising before the shape is load-bearing.

For context on why I care about this particular axis: on this machine the MoE prefill ceiling is a kernel occupancy problem, not a bandwidth one — `gather_qmm` on thin tiles runs at 2.2 TFLOPS against 8.8 for the same kernel on wide tiles, with the GPU at 99 % residency but only 9-23 W. Under-occupied Metal kernels on Apple Silicon fail quietly like that, which is why the threadgroup count stood out to me.

No verdict from me — you asked for architectural review and explicitly not a merge, and I have not run the kernel. Happy to do the device qualification if that would help.

Thump604:
Thanks for testing this and for the detailed review. I checked each point against the current head and agree with all three.

The standalone build script should not print a traceback when the Metal compiler is missing. I will change that to a short error that says full Xcode and the Metal toolchain are required.

The `tid >= 128` guard is also wrong. The host dispatches exactly 128 threads today, so it is not reached in the current path. It is still unsafe as a future guard because some lanes could return before barriers used by the remaining lanes. I will remove it and keep the threadgroup-width contract uniformly validated.

The occupancy point is the important architectural finding. This phase-one kernel walks KV serially within each `(request, query_head)` threadgroup. It can remain a correctness and ABI reference, but it should not become the long-context serving kernel. I will add split-KV attention and its reduction pass as an explicit prerequisite before dense-executor or serving-performance qualification.

The exact PR bytes do not yet have published device numerical qualification. A related standalone derivative has run locally on an M2 Ultra, but that does not qualify this revision. If your offer still stands, an M3 Ultra run after the fixes would be useful. I will provide the exact commands and artifact expectations once the refreshed revision is ready.

Thanks for catching both the unsafe future-dispatch assumption and the missing performance step in the PR stack.

janhilgard (changes requested):
Following up on my earlier review — you agreed with all three points on 26 Aug, but the branch has had no commits since 21 Aug (`6dfb30f`), so they are all still present at head. Re-stating the one that is a correctness issue rather than a preference, because it is the blocker.

## The `tid >= 128` guard is undefined behaviour in exactly the case its comment describes

```c
// The host dispatches exactly one 128-thread group per output vector.  A
// guard remains here because it is cheap and prevents accidental writes if
// a future caller uses a larger grid.
if (request >= p.batch_size || query_head >= p.query_heads || tid >= 128) {
  return;
}
```

The first two terms are uniform across the threadgroup (they derive from `tg_pos`), so those returns are safe. `tid >= 128` is not: with a 256-thread group, lanes 128–255 return while lanes 0–127 go on to `threadgroup_barrier(mem_flags::mem_threadgroup)` at line 94 and then ~10 more per token. MSL requires every thread in the threadgroup to reach a `threadgroup_barrier`; a partial return is UB — in practice a hang or silent corruption.

So the guard does not make an oversized dispatch safe. It makes it worse than no guard: without it, lanes 128+ would reach the barriers, contribute nothing (every write is already guarded by `tid < p.head_dim`, and the reduction by `if (tid < stride)`), and the kernel would still be memory-safe.

Two ways out, both small:

1. Drop `tid >= 128` from the early return and rely on the existing per-write guards. The uniform `p.head_dim != 128` return above already rejects the unsupported shape.
2. Keep the intent but move it host-side — reject a threadgroup size other than 128 at dispatch, where it can be a clean error instead of UB.

I would take (1); it is one term removed and the invariant is already enforced everywhere it matters.

## The serialisation point still stands

Parallelism is only across `(request, query_head)`; the sequence itself is walked one token at a time by a single threadgroup:

```c
for (uint token = 0; token < sequence_length; ++token) {
```

with roughly ten `threadgroup_barrier` calls per iteration (one after the dot product, seven in the 128-wide reduction, two around the softmax update). At a 46k-token context that is ~46k serialised iterations and ~460k barriers per `(request, query_head)` pair. For a "foundation" PR that is a defensible starting point, but it is worth stating in the README as a known bound rather than leaving it to be discovered by the first caller who benchmarks a long context — otherwise the natural next step (wiring this into the decode path) will look like a regression for reasons that are structural, not a bug.

Everything else in the diff — the build qualification, the deployment-target handling, the CI Metal-availability guard — reads well and I have no concerns there.

Method note: static review of the shader. I have not run it; this box is memory-committed to another workload, and the kernel is not yet on a live path anyway.

Thump604:
I corrected the divergent lane guard in 1949e48700306bee50375fe802a859a674ac2986 with a uniform rejection of unsupported threadgroup dimensions before scratch access or barriers. The host’s 128-thread launch and numerical loop are unchanged. I documented the serial sequence loop and its ten barriers per token as a limitation of this correctness/ABI foundation. The source check and offline Metal compilation passed. I have not rerun GPU numerical qualification on this revision.

Thump604:
I’ve now run the compiled native extension on Apple Silicon at `1949e487`. The existing BF16 paged-decode fixtures match NumPy (block size 16) and MLX (block size 32), and the uniform-geometry source regression passes: 3 tests, no skips. This exercises the supported 128-thread dispatch only; invalid geometry is source-checked, not device-tested. This is correctness evidence, not a performance or serving-readiness claim.

janhilgard (approved):
Both points addressed, and the geometry fix is better than either option I offered. Approving.

Reading the actual threadgroup size and rejecting on it is uniform across the group by construction, so every lane returns together — no divergent barrier — while still refusing an oversized dispatch:

```c
uint3 group_size [[threads_per_threadgroup]]) {
  // Reject unsupported actual geometry uniformly: every lane must either
  // return here or participate in all barriers and the 128-element scratch.
  if (group_size.x != 128 || group_size.y != 1 || group_size.z != 1 ||
      p.query_heads == 0) {
    return;
  }
```

I had suggested either dropping the `tid` term and relying on the existing per-write guards, or moving the check host-side. This keeps the shader self-defending against a caller that bypasses host validation — which was the original intent — without the UB, and it also catches `.y`/`.z` geometry that neither of my suggestions would have. The `p.query_heads == 0` guard ahead of the modulo is a nice catch on the way past.

`test_uniform_geometry_guard_precedes_scratch_and_barriers` asserting the guard appears before `threadgroup float partial_dot[128]` and before any `threadgroup_barrier(` is the right shape for this: it pins the ordering property that matters rather than the specific text of the condition, so a future edit that reintroduces a divergent early return fails rather than silently passing. Runs green here.

The README now states the bound explicitly:

> *Parallelism is across `(request, query_head)` only. Each threadgroup walks the sequence serially, one token at a time, with ten threadgroup barriers per…*

That is exactly what I was asking for — my own count was also ten (one after the dot product, seven in the 128-wide reduction, two around the softmax update). Having it written down means the first person to benchmark a long context reads it as a known bound of the phase-1 foundation rather than filing it as a regression.

Method note: static review of the shader plus running the new contract test. The kernel is not on a live path yet, so there is nothing to measure end to end.

PR #77, opened 2026-02-12
Add resumable model download with retry, timeout, and offline mode

janhilgard:
## Summary

- Adds a pre-download step with configurable **retry** (exponential backoff) and **timeout** before `load_model()` is called, so interrupted downloads of large models can be resumed
- New CLI flags for `serve`: `--download-timeout`, `--download-retries`, `--offline`
- New standalone subcommand: `vllm-mlx download <model>` for pre-warming HF caches (useful for CI/CD)
- Replaces direct `snapshot_download()` call in tokenizer fallback path with the new retry-aware wrapper

## Motivation

Addresses [#75](https://github.com/waybarrios/vllm-mlx/discussions/75) — HuggingFace downloads hang or fail around 10GB for large models with no way to resume.

## Usage

```bash
# Download model to cache without starting server
vllm-mlx download mlx-community/Qwen3-Next-80B-A3B-Instruct-6bit

# Serve with custom retry/timeout
vllm-mlx serve <model> --download-timeout 600 --download-retries 5

# Offline mode (only locally cached models)
vllm-mlx serve <model> --offline
```

## Test plan

- [x] 12 unit tests pass (`pytest tests/test_download.py -v`)
- [x] Manual test: `vllm-mlx download mlx-community/Qwen3-0.6B-4bit` succeeds
- [x] Manual test: nonexistent model fails with clear error message after retries
- [x] `ruff check` and `black` pass on all changed files

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #723, opened 2026-08-22
patch: repetition-detection-stop — end degenerate loops at the scheduler stop-check (#77)

TimotejLabsky:
## What

Env-gated (`VLLM_MLX_REPDETECT=1`, default **off**) repetition detector in the BatchedEngine per-token stop-check (`_generation_step`): when a request's generated tail ends in an exact repeating token cycle (period 1–128, ≥3 consecutive exact copies covering ≥48 tokens, 512-token window, checked every 16 tokens past 64 generated), the request finishes as `finish_reason="stop"` instead of burning to `max_tokens`. Logs `[repetition-stop] uid=… period=… repeats=…` and increments `vllm_mlx_repetition_stops_total`.

## Why

Qwen3.8-27B degenerates in agentic opencode sessions into exact token cycles ("items items items…", duplicated tool-call retries, multi-line loops) and burns 16K tokens per degenerate turn. Verified live 2026-08-22 that **no sampler-level lever ends it**: DRY 0.8/1.75/4/8192 + thinking budget 6144 + medium effort floor + qwen3_coder parser + the froggeric fixed chat template were all armed simultaneously and the loop still fired on turn 2 — near-variants in the wall ("its", "itemsThe") show DRY actively forbidding exact repeats while the model stays in the attractor. Penalties forbid forms, not behaviours; `\n` is a canonical DRY breaker so multi-line cycles are invisible to DRY by construction. Detection-as-stop is the only zero-cost mechanism that ends verbatim loops.

## Design

- Direct analogue of vLLM main's `SamplingParams.repetition_detection` (`check_sequence_repetition`, `stop_reason="repetition_detected"`); parameters follow vllm-project/vllm#52673 (ThinkingLoopBreaker proposal).
- **Whole-output**, not think-scoped: the 2026-08-21 audit verified the thinking budget works and the loop continues in the content stream after `</think>`.
- Rotation-invariant (blocks compared from the end) — an ongoing loop triggers wherever the last token lands in the cycle.
- `min_span` scales strictness inversely with period: period-1 needs 48 identical tokens, a 16-token cycle needs 3 copies.
- Generated tokens only; buffers discarded on finish/abort/batch-drain; MTP path (`_mtp_next`, unused in prod) not covered.
- Known limitation (documented + tested): semantic loops that reword each cycle are not caught — nothing token-level can.

## Tests

- `tests/test_repetition_stop.py`: 21 new tests (shapes incl. multi-line + rotation, near-variant limitation, gating, per-uid isolation, env parsing, disabled no-op, perf smoke, wiring smoke) — all pass.
- Regression: `test_batched_engine`, `test_batched_dry`, `test_batched_per_request_sampling`, `test_batched_stop_strings`, `test_batched_flip_enablement`, `test_batched_engine_mllm_config`, `test_metrics` — all green.

## Deploy plan (infra side, after merge)

Add to both Qwen3.8 llama-swap routes alongside the DRY env:
```
VLLM_MLX_REPDETECT=1
```
Watch `vllm_mlx_repetition_stops_total` and `[repetition-stop]` log lines during a real opencode session.

https://claude.ai/code/session_01R8MCVqWLs5wuwWXdr1ChLe


TimotejLabsky:
Apologies — opened against the wrong base (this was meant for our fork). Closing.

Issue #699, opened 2026-08-13
DFlash/DSpark block speculative decoding — planned? PR welcome?


CBribiescas:
**a) Why I want it:** on Apple Silicon, verifying a block of drafted tokens in one forward costs much less per token than one forward per token (measured on Nemotron 3.5 Lightning 6-bit, M5 Max: width-1 8.5 ms, width-2 only 1.59×). A ~4-token draft block projects to roughly 1.5–1.6× faster decode. Lightning's built-in single-token MTP is roughly break-even — one draft per verify can't amortize the verify cost. That makes block-draft speculative decoding (DFlash/DSpark style) a big unlock for MLX serving generally, not just this model.

**b)** Is block-draft spec decode (DFlash/DSpark style) on anyone's roadmap here?

**c)** If not — would a large PR be welcome at this time? I'd keep it process-oriented rather than one big drop:

1. propose an approach skeleton (short design doc / draft PR for comment),
2. validate the core assumptions with a spike (draft-model forward + k-wide verify behind a flag, correctness bar = byte-identical greedy output),
3. land an MVP for one model family,
4. iterate from there (more targets, batching, non-greedy).

Happy to adjust the split to whatever fits the repo's review bandwidth.

waybarrios:
@CBribiescas I am planning to workin on this in the upcoming days. Any extra hand is welcome, if you already have a draft I could take a look.

CBribiescas:
Yes — I have a rough Claude implementation. Acceptance is landing right at your spec (mean 3.20 of 7 drafted, greedy-valid output), but the throughput result turned out more complicated than what I expected when I started this thread, and I'd rather write it up properly than summarize it badly here.

I'll post the details this weekend — the measurements, where the bottleneck actually sits, and the state of the code.

If you're planning to start before then, say so and I'll send what I have early. The short version is that the drafter isn't the hard part, so it's worth a read before you commit time.

Thump604:
[...]

Issue #502, opened 2026-05-05
Add DFlash speculative decoding backend for Qwen 3.5/3.6


Thump604:
## Summary

I would like to explore adding DFlash as a separate speculative decoding backend for Qwen 3.5 / 3.6 models in `vllm-mlx`.

This should not be treated as native MTP. DFlash uses an external block-diffusion draft model plus verification against the target model, so the integration shape should probably be a distinct speculative path rather than another `--enable-mtp` mode.

References:

- DFlash repo: https://github.com/z-lab/dflash
- Paper: https://arxiv.org/abs/2602.06036
- Example draft models on Hugging Face:
  - `z-lab/Qwen3.6-35B-A3B-DFlash`
  - `z-lab/Qwen3.6-27B-DFlash`
  - `z-lab/Qwen3.5-35B-A3B-DFlash`
  - `z-lab/Qwen3.5-122B-A10B-DFlash`

## Proposed direction

A narrow first implementation could add an explicitly configured DFlash backend, for example:

- `speculative_method=dflash`
- `dflash_draft_model=<path-or-repo>`
- `dflash_block_size=<n>`
- `dflash_draft_sliding_window_size=<n>`

The first target should be a non-resident diagnostic path for Qwen 3.6 35B, not production/default serving.

## Why separate from native MTP

Native MTP and DFlash have different contracts:

- native MTP depends on target-model MTP heads / draft logits
- DFlash uses a separate block-diffusion draft model
- DFlash needs draft-model provenance and compatibility checks
- DFlash needs rollback-safe integration with KV/stateful processors
- DFlash performance should be compared against both feature-off baseline and native MTP where native MTP weights are actually present

## Initial validation requirements

Before claiming support, I think the minimum evidence should include:

- model and draft-model identity / revision / config hashes
- route and engine path
- final content present
- semantic correctness
- serving safety
- acceptance / rejection statistics
- TTFT, final-content latency, and total latency
- active/peak Metal memory
- parser/final-content cleanliness
- comparison against matching feature-off baseline

## Non-goals for the first pass

- no resident/defaultability claim
- no broad feature matrix
- no assumption that DFlash is equivalent to native MTP
- no Open WebUI or launch default integration until the diagnostic path is validated

I can begin working on a narrow prototype and compatibility probe once local test capacity is available.

TomLucidor:
[...]

PR #724, opened 2026-08-22
fix(mllm): dispatch assistant-drafter loading by architecture instead of hardcoding Gemma 4

brandy975:
# fix(mllm): dispatch assistant-drafter loading by architecture instead of hardcoding Gemma 4

## Summary

`MLXMultimodalLM._load_draft_model()` calls `load_gemma4_assistant_drafter()` for every `--mllm-draft-model`, regardless of `--mllm-draft-kind`. That function unconditionally imports `mlx_vlm.speculative.drafters.gemma4_assistant` and builds that one architecture. Any assistant drafter that isn't Gemma-4-shaped — e.g. a Qwen3.5-MTP standalone drafter such as `lukaskremla/Qwen3.8-27B-MTP-6bit-MLX` — fails to load, even though mlx-vlm itself already ships a matching `mlx_vlm.speculative.drafters.qwen3_5_mtp` architecture and a generic, quantization-aware loader (`mlx_vlm.utils.load_model`) that already knows how to pick it.

This PR replaces the hand-rolled Gemma-4-only loading with a call to `mlx_vlm.utils.load_model()`, mlx-vlm's own generic loader — the same one it uses for its regular (non-drafter) models — which already resolves the architecture from the checkpoint's `config.json` `model_type` and handles quantization. One function now covers every drafter family mlx-vlm ships, including the Gemma 4 assistant drafters that previously worked; nothing is hardcoded to one architecture any more.

## Problem

Given a target model (`--mllm`) and a standalone Qwen3.5-MTP drafter:

```
vllm-mlx serve <qwen3.8-27B target> --mllm \
  --mllm-draft-model <qwen3.8-27B MTP drafter, 6-bit> \
  --mllm-draft-kind mtp
```

Loading fails:

```
File "vllm_mlx/models/mllm.py", line 1404, in _load_draft_model
    return load_gemma4_assistant_drafter(self.draft_model_path)
File "vllm_mlx/models/mllm.py", line 455, in load_gemma4_assistant_drafter
    model.load_weights(list(weights.items()))
File "mlx/nn/layers/base.py", line 185, in load_weights
    raise ValueError(f"Received {num_extra} parameters not in model: \n{extras}.")
ValueError: Received 31 parameters not in model:
fc.biases, fc.scales, fc.weight,
layers.0.input_layernorm.weight,
layers.0.mlp.down_proj.{biases,scales,weight}, ...
```

(Line numbers as of `origin/main` `b281bee`, 2026-08-22 — the original reproduction was against the released `vllm-mlx==0.4.0` PyPI package, whose line numbers differ slightly; verified the call chain and error message are otherwise identical between the two.)

`load_gemma4_assistant_drafter()` builds `mlx_vlm.speculative.drafters.gemma4_assistant.Model(config)` and calls `load_weights()` on it directly — the drafter's checkpoint uses standard Qwen decoder-layer key names (`layers.N.self_attn.*`, `layers.N.mlp.*`, ...), which don't exist on the Gemma-4 assistant architecture at all, hence "not in model" rather than a shape mismatch.

Even after pointing the loader at the correct architecture (`mlx_vlm.speculative.drafters.qwen3_5_mtp`) by hand, loading still fails for a *quantized* checkpoint (this one is 6-bit): the model is built with plain `Linear` layers, but the checkpoint's tensors include `.scales`/`.biases` (`QuantizedLinear`). `load_gemma4_assistant_drafter()` never calls `nn.quantize()` before `load_weights()` — presumably because Gemma 4 assistant drafters have so far only shipped unquantized (bf16), so this was never exercised.

`mlx_vlm.utils.load_model()` already handles both problems: it selects the architecture class from the checkpoint's own `model_type` and calls `nn.quantize(..., group_size, bits, mode)` before `load_weights()` whenever `config.json` carries a `quantization` block. `_load_draft_model()` just never calls it.

`--mllm-draft-kind` currently only has one valid choice (`mtp`) — the CLI doesn't distinguish drafter *families* at all yet, only whether a drafter is configured.

## Fix

`vllm_mlx/models/mllm.py`:

- `load_gemma4_assistant_drafter(model_path)` renamed to `load_assistant_drafter(model_path)` and rewritten: instead of importing `mlx_vlm.speculative.drafters.gemma4_assistant` directly and hand-rolling `Model(config)` + `load_weights()`, it now just calls `mlx_vlm.utils.load_model(Path(model_path))`. That function already resolves the architecture from the checkpoint's own `config.json` `model_type` against every family mlx-vlm ships under `mlx_vlm.speculative.drafters` (`gemma4_assistant`, `qwen3_5_mtp`, `eagle3`, `deepseek_v4_mtp`, ...), and — unlike the old hand-rolled path — quantizes before `load_weights()` whenever the checkpoint carries a `quantization` block. Existence checks for `config.json`/`*.safetensors` are kept (same `FileNotFoundError`s as before, checked first so the error names the actual missing file rather than surfacing whatever `load_model` would raise). `_load_draft_model()`'s single call site updated to the new name; no other change there — it still only calls this for `draft_kind == "mtp"`, unchanged.
- Deliberately does **not** use `mlx_vlm.utils.load()` (the higher-level function already used for every other `draft_kind` in `_load_draft_model()`'s `else` branch): `load()` additionally loads a processor/tokenizer, which standalone assistant/MTP drafters typically don't ship (they reuse the target model's) — this was presumably why the original code hand-rolled the Gemma-4 path instead of reusing that existing `else` branch.

No change to `--mllm-draft-kind`'s CLI surface, to `cli.py`'s validation, or to the scheduling/accept-reject logic in `vllm_mlx/scheduler.py` — this only changes *which model gets built* for the drafter, not how it's driven afterwards.

## Tests

New file `tests/test_mllm_assistant_drafter_loader.py` (6 tests), following `tests/test_qwen35_mtp_patch.py`'s style — `mlx_vlm.utils.load_model` is monkeypatched via `sys.modules`, so no real model architecture or safetensors content is needed:

- `test_load_assistant_drafter_dispatches_any_model_type_through_load_model` (parametrized `gemma4_assistant` / `qwen3_5_mtp` / `eagle3`) — every `model_type` goes through the same `mlx_vlm.utils.load_model` call with the drafter's own path; nothing is hardcoded.
- `test_load_assistant_drafter_missing_config_raises_file_not_found`
- `test_load_assistant_drafter_missing_weights_raises_file_not_found`
- `test_load_assistant_drafter_missing_mlx_vlm_raises_import_error`

`tests/test_mllm_assistant_drafter.py` (existing chat-time drafter wiring tests, inject `_draft_model` directly and don't exercise the loader) — unaffected, still 7/7 passing.

## Test results

Full suite: `pytest tests/` — 2679 passed, 24 skipped, 27 deselected, 0 failed.

`black --check`, `ruff check --select E,F,W --ignore E402,E501,E731,F811,F841`, and `mypy vllm_mlx/models/mllm.py` all pass on the changed file. mypy's error set on `mllm.py` actually *shrinks* by two (both were on the old hand-rolled `weights: dict = {}` / `weights.update(mx.load(...))` code this PR removes; mypy couldn't infer its types).

## Risk notes

Scope is intentionally narrow: only the drafter-loading dispatch changes. Every currently-working drafter (Gemma 4 assistant drafters) now goes through `mlx_vlm.utils.load_model` too instead of the hand-rolled loader — confirmed equivalent by inspection (same `Model(config)` + `sanitize` + `load_weights` + `mx.eval` + `.eval()` sequence, plus quantization handling the old code never had), but not diffed against a real Gemma 4 assistant drafter checkpoint (none available in this environment) the way the Qwen3.5-MTP path was.

## Related work

#669 ("fix: enable configured MLLM assistant drafters by default") touches the same file but a disjoint region — `__init__`/`_draft_generation_kwargs` (the default-on/opt-out behavior for an *already-loaded* drafter) vs. this PR's `load_gemma4_assistant_drafter`/`_load_draft_model` (which architecture gets *loaded* in the first place). Verified no overlap with a `git merge-tree` dry-run against #669's branch: 0 conflicts. The two are complementary, not competing — #669 fixes a real gap this PR doesn't touch (a configured drafter currently still needs a private per-request `mllm_draft=true` to actually engage).

The dependency runs one way, not both: #669 is independently useful today for its own tested case (Gemma 4 assistant drafters, which already load fine without this PR) and doesn't need this PR to land. This PR, conversely, doesn't need #669 either — the per-request `mllm_draft=true` opt-in it fixes still works as the engage mechanism regardless of whether #669 exists. Where the two actually interact: for a *non*-Gemma-4 drafter family (Qwen3.5-MTP and any other family under `mlx_vlm.speculative.drafters`), `--default-mllm-draft` is currently unreachable — the server fails to load the drafter before that flag would ever matter. This PR is what makes #669's flag meaningful for those families; it doesn't make #669 itself faster, safer, or more correct for the case it was written and verified against. I support #669 landing on its own merits, independently of this one.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,6 +99,7 @@ jobs:
             tests/test_qwen3_xml_registration.py \
             tests/test_qwen3_xml_streaming_chunks.py \
             tests/test_qwen35_mtp_hidden_state_mode.py \
+            tests/test_mllm_assistant_drafter_loader.py \
             tests/test_streaming_fence_stripper.py \
             tests/test_streaming_think_router.py \
             tests/test_streaming_tool_filter.py \
--- a/tests/test_mllm_assistant_drafter_loader.py
+++ b/tests/test_mllm_assistant_drafter_loader.py
@@ -1,0 +1,85 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Regression tests for load_assistant_drafter's config-driven architecture
+dispatch (ex load_gemma4_assistant_drafter — see fix/mllm-draft-loader-
+architecture-dispatch).
+
+These test the dispatch logic itself (config.json -> mlx_vlm.utils.load_model),
+not real weight loading — mlx_vlm.utils.load_model is monkeypatched so no
+actual model architecture or safetensors content is needed. The existing
+tests in test_mllm_assistant_drafter.py cover chat-time drafter wiring and
+inject `_draft_model` directly, bypassing this loader entirely; these tests
+cover the loader itself, which nothing previously exercised.
+"""
+
+import json
+import sys
+from types import SimpleNamespace
+
+import pytest
+
+
+def _write_drafter_checkpoint(tmp_path, model_type: str):
+    (tmp_path / "config.json").write_text(json.dumps({"model_type": model_type}))
+    (tmp_path / "model.safetensors").write_bytes(b"")
+    return tmp_path
+
+
+@pytest.mark.parametrize("model_type", ["gemma4_assistant", "qwen3_5_mtp", "eagle3"])
+def test_load_assistant_drafter_dispatches_any_model_type_through_load_model(
+    tmp_path, monkeypatch, model_type
+):
+    """No architecture is hardcoded: every model_type goes through the same
+    mlx_vlm.utils.load_model call, just with a different config.json."""
+    from vllm_mlx.models.mllm import load_assistant_drafter
+
+    drafter_path = _write_drafter_checkpoint(tmp_path, model_type)
+    captured = {}
+
+    class FakeDrafterModel:
+        def eval(self):
+            captured["eval_called"] = True
+            return self
+
+    def fake_load_model(path):
+        captured["path"] = path
+        return FakeDrafterModel()
+
+    monkeypatch.setitem(
+        sys.modules, "mlx_vlm.utils", SimpleNamespace(load_model=fake_load_model)
+    )
+
+    result = load_assistant_drafter(str(drafter_path))
+
+    assert captured["path"] == drafter_path
+    assert captured["eval_called"] is True
+    assert isinstance(result, FakeDrafterModel)
+
+
+def test_load_assistant_drafter_missing_config_raises_file_not_found(tmp_path):
+    from vllm_mlx.models.mllm import load_assistant_drafter
+
+    (tmp_path / "model.safetensors").write_bytes(b"")
+
+    with pytest.raises(FileNotFoundError, match="config"):
+        load_assistant_drafter(str(tmp_path))
+
+
+def test_load_assistant_drafter_missing_weights_raises_file_not_found(tmp_path):
+    from vllm_mlx.models.mllm import load_assistant_drafter
+
+    (tmp_path / "config.json").write_text(json.dumps({"model_type": "qwen3_5_mtp"}))
+
+    with pytest.raises(FileNotFoundError, match="weights"):
+        load_assistant_drafter(str(tmp_path))
+
+
+def test_load_assistant_drafter_missing_mlx_vlm_raises_import_error(
+    tmp_path, monkeypatch
+):
+    from vllm_mlx.models.mllm import load_assistant_drafter
+
+    drafter_path = _write_drafter_checkpoint(tmp_path, "qwen3_5_mtp")
+    monkeypatch.setitem(sys.modules, "mlx_vlm.utils", None)
+
+    with pytest.raises(ImportError, match="mlx-vlm"):
+        load_assistant_drafter(str(drafter_path))
--- a/tests/test_mllm_assistant_drafter_loader.py
+++ b/tests/test_mllm_assistant_drafter_loader.py
@@ -24,12 +24,19 @@ def _write_drafter_checkpoint(tmp_path, model_type: str):
     return tmp_path
 
 
-@pytest.mark.parametrize("model_type", ["gemma4_assistant", "qwen3_5_mtp", "eagle3"])
-def test_load_assistant_drafter_dispatches_any_model_type_through_load_model(
+@pytest.mark.parametrize(
+    "model_type",
+    [
+        "gemma4_assistant",
+        "gemma4_unified_assistant",
+        "qwen3_5_mtp",
+        "deepseek_v4_mtp",
+    ],
+)
+def test_load_assistant_drafter_dispatches_mtp_model_type_through_load_model(
     tmp_path, monkeypatch, model_type
 ):
-    """No architecture is hardcoded: every model_type goes through the same
-    mlx_vlm.utils.load_model call, just with a different config.json."""
+    """Every supported MTP model type uses the generic mlx-vlm loader."""
     from vllm_mlx.models.mllm import load_assistant_drafter
 
     drafter_path = _write_drafter_checkpoint(tmp_path, model_type)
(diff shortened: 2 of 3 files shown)


Thump604 (commented):
The dispatch direction is sound on the submitted head: the call at `vllm_mlx/models/mllm.py:1404-1411` now routes the `mtp` drafter through `load_assistant_drafter()`, and the replacement loader at `:417-458` delegates to `mlx_vlm.utils.load_model` after retaining the existing config/weights checks. The minimum declared dependency (`mlx-vlm>=0.6.5`) already resolves `qwen3_5_mtp` through `mlx_vlm.speculative.drafters` and applies checkpoint quantization in `load_model`; the referenced Qwen3.8 artifact declares `model_type=qwen3_5_mtp` with 6-bit quantization. I found no static correctness blocker in this path.

One qualification gap remains before treating this as fully evidenced. All six new tests replace `mlx_vlm.utils.load_model` with a fake, so they do not exercise the integration this change relies on: config-to-`Qwen3_5MTPConfig` construction, quantized `load_weights()`/sanitization, or the existing Gemma assistant checkpoint path. Please add a focused dependency-compatibility fixture or loader smoke that exercises those real seams under a pinned mlx-vlm version (and keeps the missing-file/error cases); if the real checkpoint cannot be loaded in CI, document the exact deferred Apple-Silicon gate and artifact/version matrix. Keep this separate from #669: its merged default-on behavior is complementary and touches the request-time wiring, while this PR changes only loader selection.

This is a COMMENT, not an approval or merge recommendation. Rebase/current-main verification and the focused loader evidence are the remaining maintainer-action items; no MLX/inference run was performed here.

waybarrios:
I pushed follow-up commit `44ea821` to keep this PR aligned with the runtime support it actually provides:

- keep architecture-aware and quantization-aware loading for Gemma, Qwen3.5, DeepSeek, and other MTP drafters
- resolve the checkpoint's required draft kind through mlx-vlm
- reject Eagle3 and DFlash checkpoints before loading because vllm-mlx does not expose those generation paths yet
- perform local checkpoint checks before importing mlx-vlm
- run the loader regression tests in the required Linux CI matrix

The rejection regression was verified red before the implementation and green afterward. All 10 GitHub checks now pass, including both Apple Silicon jobs, and commit `44ea821` is Verified.

Eagle3 runtime support is tracked separately in #762. DFlash remains tracked in #502 and #699.

Thump604:
I compared the current `44ea821` head directly with #752 at `8dd9413`. The two changes replace the same `_load_draft_model()` call site and should not both land.

The requested architecture and CI corrections here are valid, but #752 now covers the same dispatch with the stricter contract: local paths and repo IDs, nested/config conflict handling, preserved Gemma loading, registered `load_drafter()` return validation, target/drafter compatibility, and real-call-site error preservation. This head still delegates to generic `load_model()`, has local-path-only admission, and fakes the loader in its regressions.

To avoid sending the contributor through redundant work, I recommend **no further changes on #724** while maintainers choose the winning implementation. My recommendation is to retain #752, which already has Jan’s approval, and close #724 as superseded while crediting this report and its architecture-dispatch diagnosis. If maintainers prefer the smaller #724 implementation instead, #752 should be closed rather than attempting to merge both. No additional model run is requested from this contributor for that decision.

brandy975:
Originally I just wanted to get Qwen 3.8 27B (6-bit or 4-bit quantization) running. Along the way I noticed the performance was very poor, and on top of that the responses weren't OpenAI-compliant. It quickly became clear this was a deep rabbit hole I couldn't solve alone in one step. So I wanted to at least contribute a quick win that lets the draft model be loaded correctly in the first place.

@Thump604 I'm genuinely glad you implemented a more comprehensive solution — what matters to me in the end is that the functionality lands, and that's achieved either way.

brandy975:
@waybarrios My implementation doesn't support Hugging Face repo IDs, only local paths — that was my only tested use case. That's a clear weakness compared to #752, which is why I support it too. Your addition rejecting Eagle3/DFlash checkpoints should already be covered there as well.

PR #725, opened 2026-08-23
fix(reasoning): GLM-4.7 injects <think> in the prompt — stream thinking as reasoning_content, not content

Rune-Humborstad:
## Summary

With `--reasoning-parser glm4`, **GLM-4.7's entire thinking block streams into `content`** instead of `reasoning_content`. Non-streaming responses are unaffected, which makes the bug easy to miss — it only shows up for streaming clients (observed via an agent frontend rendering thinking text as the visible reply).

## Root cause

GLM-4.7's chat template **injects `<think>` at the end of the generation prompt** when thinking is enabled, and a pre-closed `</think>` when disabled:

```
>>> tok.apply_chat_template(msgs, add_generation_prompt=True, tokenize=False)
'...<|assistant|><think>'                        # thinking enabled (default)
'...<|assistant|></think>'                       # enable_thinking=False
```

So with thinking enabled, model output contains **only the closing tag** — but `Glm4ReasoningParser.extract_reasoning_streaming` was built on the opposite assumption ("GLM-4 does NOT inject `<think>` in the prompt") and emits pre-tag deltas as content. Result: every thinking token streams as `content` until `</think>` arrives, at which point only the final pre-tag fragment lands in `reasoning_content`.

The non-streaming path works because `extract_reasoning` sees the closing tag in the full text and takes the only-closing-tag branch.

## Fix

- **`glm4_parser.py`**: drop the content-assuming `pre_think` override and delegate to the base class, whose implicit-reasoning mode (default to reasoning until `</think>`) matches GLM-4.7's template. GLM-4.6V box-tag stripping is retained. Thinking-disabled requests are not misrouted: the server already bypasses reasoning parsing for them (`_thinking_disabled`), which also covers the template's pre-closed `</think>` variant.
- **`think_parser.py`**: fix a latent boundary bug in the `idx == -1` branches (start/end tag completed across a delta boundary): the tag remnant at the head of the delta leaked into content/reasoning (e.g. `ink>` prefixed to content). Now the remnant is dropped by computing the straddle offset from `current_text`/`previous_text`. Remaining cosmetic edge: the tag's head characters from the *prior* delta stay in reasoning; full fix needs tail-holdback buffering, which isn't needed while the tags are single special tokens (they are for GLM/Qwen — the straddle only occurs for models whose tags tokenize into pieces).
- **Tests**: encode the GLM-4.7 contract — tag-less streaming deltas are implicit reasoning; a stream with only `</think>` splits cleanly with nothing leaking across; box-tag stripping asserted on the combined stream.

## Verification

- `pytest tests/test_reasoning_parser.py tests/test_streaming_think_router.py tests/test_thinking_processor.py tests/test_thinking_aware_processor.py` — 174 passed.
- Live on `mlx-community/GLM-4.7-4bit` (M3 Ultra, `--tool-call-parser glm47 --reasoning-parser glm4`), streaming: thinking default → clean separation (806-char reasoning, 20-char content); tools present → clean; `enable_thinking: false` → content only, no reasoning; tool-call turn → reasoning separated, `tool_calls` delta emitted correctly.
- Parser-level replay of the GLM-4.7 output shape at 1/5/7-char chunk sizes: content exactly matches the post-`</think>` text at every granularity.

### DIFF
--- a/tests/test_reasoning_parser.py
+++ b/tests/test_reasoning_parser.py
@@ -1291,13 +1291,34 @@ class TestGlm4Parser:
 
     # Streaming tests
 
-    def test_streaming_no_tags_emits_content(self, parser):
-        """GLM-4 streaming without tags should emit content (not reasoning)."""
+    def test_streaming_no_tags_emits_reasoning(self, parser):
+        """GLM-4.7 injects <think> in the prompt, so tag-less streaming
+        output is implicit reasoning (thinking-disabled requests bypass
+        the parser server-side and are not affected)."""
         parser.reset_state()
         result = parser.extract_reasoning_streaming("", "Hello", "Hello")
         assert result is not None
-        assert result.content == "Hello"
-        assert result.reasoning is None
+        assert result.reasoning == "Hello"
+        assert result.content is None
+
+    def test_streaming_implicit_close_splits_reasoning_and_content(self, parser):
+        """Only </think> in output (GLM-4.7 shape): reasoning before the
+        tag, content after — nothing leaks across."""
+        parser.reset_state()
+        tokens = ["thinking a", "bout it", "</think>", "the answer"]
+        accumulated = ""
+        reasoning_parts, content_parts = [], []
+        for token in tokens:
+            prev = accumulated
+            accumulated += token
+            result = parser.extract_reasoning_streaming(prev, accumulated, token)
+            if result:
+                if result.reasoning:
+                    reasoning_parts.append(result.reasoning)
+                if result.content:
+                    content_parts.append(result.content)
+        assert "".join(reasoning_parts) == "thinking about it"
+        assert "".join(content_parts) == "the answer"
 
     def test_streaming_with_thinking(self, parser):
         """Test streaming with think tags."""
@@ -1327,16 +1348,16 @@ class TestGlm4Parser:
 
         tokens = ["<|begin_of_box|>", "Paris", "<|end_of_box|>"]
         accumulated = ""
-        content_parts = []
+        parts = []
 
         for token in tokens:
             prev = accumulated
             accumulated += token
             result = parser.extract_reasoning_streaming(prev, accumulated, token)
-            if result and result.content:
-                content_parts.append(result.content)
+            if result:
+                parts.append((result.reasoning or "") + (result.content or ""))
 
-        full = "".join(content_parts)
+        full = "".join(parts)
         assert "Paris" in full
         assert "<|begin_of_box|>" not in full
         assert "<|end_of_box|>" not in full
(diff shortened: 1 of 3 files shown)


Thump604 (changes requested):
The GLM-4.7 symptom is plausible, but this revision changes a shared parser contract more broadly than its evidence supports.

The PR changes Glm4ReasoningParser.extract_reasoning_streaming() so every tag-less delta enters BaseThinkingReasoningParser's implicit-reasoning path. The non-streaming Glm4ReasoningParser.extract_reasoning() is unchanged and still returns (None, model_output) when no tag is present. Therefore the same tag-less GLM output is content for non-streaming requests but reasoning for streaming requests. The current tests at tests/test_reasoning_parser.py:1264-1269 and :1294-1300 encode the existing content behavior; changing only the streaming assertion does not establish an API-compatible transition.

The parser registry exposes one glm4 class for the GLM-4.5-Air/GLM-4.6V/GLM-4.7 family (reasoning/__init__.py:76-98), and the CLI selects it explicitly by parser name (cli.py:177-200). This PR's rationale is GLM-4.7 chat-template injection, but there is no model/version-specific mode in the class or server. _thinking_disabled() only bypasses parsing when the request or resolved chat-template kwargs explicitly set enable_thinking=false; it does not protect an enabled parser from a no-tag response, nor does it distinguish GLM-4.7 from the other registered GLM-4 variants. A no-tag response from those variants would be reclassified as reasoning on the streaming path.

Please:
1. Rebase onto current main and resolve the current conflict before review. The submitted head is d05f53ab8937186d03d64dc2f3bb1d4b6377f199 on stale base 4b654c09e61029afa1c061a3684bff0d4c385b3c; GitHub reports CONFLICTING/DIRTY and no current status checks.
2. Define the contract per model/template, rather than making all glm4 streaming output implicit reasoning. Either add an explicit GLM-4.7-only mode selected from verified model/template metadata, or retain tag-less content semantics and handle the injected-prompt case in a request-scoped/server-aware way. Do not rely on an explicit disable flag to cover ordinary no-tag output.
3. Add paired streaming and non-streaming tests for GLM-4.7 thinking enabled (prompt-injected closing tag), thinking disabled, a no-tag fallback, and GLM-4.5/4.6V-compatible tagged/untagged output. Assert the same reasoning/content partition and preserve box-tag stripping.
4. Keep the new split-marker logic, but add boundary tests where both <think> and </think> are split across previous_text/delta_text, including a delta that completes the marker and carries following content. Verify no marker bytes are emitted or dropped.
5. Rerun current unit CI and report the exact model/template evidence and dependency/repository versions used. No inference or MLX run is required for these parser tests; any model behavior claim that requires generation remains separately gated.

I am requesting changes; I am not approving or merging this revision.

PR #727, opened 2026-08-23
fix(cli): wire prefill step into batching

### LINKED ISSUE #711: [Bug]: --prefill-step-size is silently dropped under --continuous-batching (serve_command builds SchedulerConfig without it)
## Summary

`--prefill-step-size` is accepted by the `serve` parser but never reaches the scheduler under `--continuous-batching`: `serve_command` builds its `SchedulerConfig` without passing it. The flag is silently a no-op, prefill falls back to the dataclass default of 2048, and nothing in the logs says so.

This matters because on models where `mx.fast.scaled_dot_product_attention` does not fuse, `prefill_step_size` is one of only two knobs that bound the transient prefill allocation.

## Version

Reproduced on **0.4.1** (latest on PyPI) and **0.3.0**, both from the sdist — not from a patched install.

## Where

`vllm_mlx/cli.py`, in `serve_command` (0.4.1, line 261):

```python
        enable_prefix_cache = args.enable_prefix_cache and not args.disable_prefix_cache

        scheduler_config = SchedulerConfig(
            max_num_seqs=args.max_num_seqs,
            prefill_batch_size=args.prefill_batch_size,
            completion_batch_size=args.completion_batch_size,
            enable_prefix_cache=enable_prefix_cache,
            ...
```

`prefill_step_size` is missing from that call, and only from that one:

- the flag **is** declared on the serve parser (`--prefill-step-size`, `type=int`, `default=2048`);
- the field **does** exist on `SchedulerConfig` (`scheduler.py:65`, `prefill_step_size: int = 2048`) and is consumed at `scheduler.py:1325`;
- two other call sites in the same file **do** pass `prefill_step_size=args.prefill_step_size` (0.3.0 lines 302 and 328 — the SimpleEngine paths).

So the flag works on the non-batched path and is dropped exactly where `--continuous-batching` users land.

## Why it is worth fixing rather than documenting

Measured on `mlx-community/Qwen3.8-27B-8bit` (`head_dim=256`), single 35k-token prompt, `max_tokens: 1`:

| configuration | peak Metal memory |
|---|---|
| defaults | **133.8 GB** (~50 GB transient) |
| `--prefill-batch-size 1` | 39.5 GB |
| `--prefill-batch-size 1` + `--prefill-step-size 512` **(only with the line below added)** | **34.6 GB** |

At `head_dim=256` the fused SDPA kernel does not take over, so the full score matrix is materialized in bf16. Measured in-process (mlx 0.31.2) at `B=1, 24 heads, qL=2048, kL=35000, head_dim=256`: a single SDPA call peaks at **3.44 GB**, exactly `24 × 2048 × 35000 × 2` bytes; the same call at `head_dim=128` (control) costs ~0 GB. The 64-layer forward is one lazy graph containing 16 `full_attention` layers, with a measured accumulation factor of ~4×, giving

```
transient ≈ 4 × prefill_batch_size × 24 × prefill_step_size × context × 2 bytes
```

The 100 GB difference above is enough to evict a second model served on the same machine — which is how we found it.

Throughput cost of the smaller step: none measurable — 175.7 tok/s at `step=2048` vs 172.2 tok/s at `step=512` on the same server (the two prompts were not exactly the same length, so read this as "no meaningful difference", not as an equality).

## Fix

One line, mirroring the two call sites that already do it:

```diff
         scheduler_config = SchedulerConfig(
             max_num_seqs=args.max_num_seqs,
             prefill_batch_size=args.prefill_batch_size,
+            prefill_step_size=args.prefill_step_size,
             completion_batch_size=args.completion_batch_size,
```

Note for whoever patches it: `prefill_batch_size=args.prefill_batch_size,` also appears in `bench_command`, whose parser has no `--prefill-step-size` — patching that second site would raise `AttributeError`.

## Repro

```bash
vllm-mlx serve mlx-community/Qwen3.8-27B-8bit --port 8000 \
  --continuous-batching --prefill-batch-size 1 --prefill-step-size 512
# send a ~35k-token prompt with max_tokens: 1, watch metal.peak_memory_gb on /v1/status
```

Peak stays at the `step=2048` level. Adding the line above brings it to the value in the table.

Possibly related memory reports: #442, #641, #627.

Thump604:
## Summary

- pass `--prefill-step-size` into the continuous-batching scheduler
- preserve the separate MLLM-specific override
- cover the full CLI-to-scheduler wiring with a real-parser regression

## Why

`serve_command` accepted the flag but omitted it when constructing `SchedulerConfig`, so continuous batching silently used the default instead of the requested prefill bound.

## Tests

- `30 passed` across CLI, lifecycle CLI, MLLM config, and scheduler wiring coverage
- Black, Ruff, and `git diff --check`

Fixes #711

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -119,6 +119,40 @@ def test_serve_command_propagates_all_sampling_defaults(monkeypatch):
     assert loaded["kwargs"]["specprefill_backbone_pct"] == 0.25
 
 
+def test_serve_command_wires_prefill_step_size_into_batched_scheduler(monkeypatch):
+    from vllm_mlx import cli, server
+    from vllm_mlx.utils import download
+
+    loaded = {}
+    monkeypatch.setattr(
+        download, "ensure_model_downloaded", lambda *args, **kwargs: "local-test-model"
+    )
+    monkeypatch.setattr(
+        server,
+        "load_model",
+        lambda *args, **kwargs: loaded.update({"args": args, "kwargs": kwargs}),
+    )
+    monkeypatch.setattr("uvicorn.run", lambda *args, **kwargs: None)
+
+    args = cli.create_parser().parse_args(
+        [
+            "serve",
+            "local-test-model",
+            "--continuous-batching",
+            "--prefill-step-size",
+            "512",
+            "--mllm-prefill-step-size",
+            "256",
+            "--offline",
+        ]
+    )
+    cli.serve_command(args)
+
+    scheduler_config = loaded["kwargs"]["scheduler_config"]
+    assert scheduler_config.prefill_step_size == 512
+    assert scheduler_config.mllm_prefill_step_size == 256
+
+
 def test_serve_parser_accepts_registered_step3p5_tool_parser():
     from vllm_mlx import cli
 
--- a/vllm_mlx/cli.py
+++ b/vllm_mlx/cli.py
@@ -299,6 +299,7 @@ def serve_command(args):
         scheduler_config = SchedulerConfig(
             max_num_seqs=args.max_num_seqs,
             prefill_batch_size=args.prefill_batch_size,
+            prefill_step_size=args.prefill_step_size,
             completion_batch_size=args.completion_batch_size,
             enable_prefix_cache=enable_prefix_cache,
             prefix_cache_size=args.prefix_cache_size,


Thump604:
Superseded by #729, which carries the same fix on a ready branch with the focused #711 regression coverage.

PR #728, opened 2026-08-23
fix(mllm): keep media requests out of prefix cache

### LINKED ISSUE #718: [Bug] Multimodal requests reuse a previous image's vision KV via prefix cache (cross-image contamination)
## Summary

With continuous batching + KV prefix caching enabled on an MLLM (vision-language model), sending the **same text prompt with a different image** reuses the KV state of a **previously cached image**. Every request for that prompt returns the content of whatever image was first cached for it — not the image actually sent. In a page-by-page scanned-PDF OCR loop, every page comes back as page 1.

## Environment

- vllm-mlx **0.4.1** (bug also present on `main`)
- `vllm-mlx serve --continuous-batching --mllm ...` (prefix cache enabled by default)
- Model: any VLM with continuous batching (reproduced on `mlx-community/Qwen3.6-35B-A3B-4bit`)

## Minimal reproduction

Two PNGs with clearly different content (e.g. red background labeled "A", blue background labeled "B"). Send `POST /v1/chat/completions` twice with the **same text prompt**, image A then image B:

```python
import base64, requests

URL = "http://localhost:8800/v1/chat/completions"
MODEL = "<model>"

def ask(png_path, prompt):
    b64 = base64.b64encode(open(png_path, "rb").read()).decode()
    r = requests.post(URL, json={
        "model": MODEL,
        "messages": [{"role": "user", "content": [
            {"type": "text", "text": prompt},
            {"type": "image_url", "image_url": {"url": f"data:image/png;base64,{b64}"}},
        ]}],
        "max_tokens": 60,
        "temperature": 0.0,
    })
    return r.json()["choices"][0]["message"]["content"]

PROMPT = "What background color is this image? Answer with the color name only."
print(ask("a.png", PROMPT))  # -> "red"   (correct)
print(ask("b.png", PROMPT))  # -> "red"   (WRONG, should be "blue")
```

Observed output: the second request returns the first image's result. Also, `usage.prompt_tokens` is identical for both requests and reflects only the **text** prompt — the image contributes no tokens to the reported usage (it is computed via `len(tokenizer.encode(text_prompt))` in `mllm_scheduler.py`).

## Root cause

The KV prefix cache (`MemoryAwarePrefixCache` in `mllm_batch_generator.py`) is keyed on the **token sequence only**. For multimodal requests the image placeholder token ID is **content-agnostic** — two requests with the same text but different images produce **identical token sequences**, so the second request hits the prefix cache and reuses the first image's KV (including the vision-encoder output) without re-running the vision encoder.

The existing guard in `_process_prompts` only checks whether the *uncached remainder* contains `image_token_index`. It does not protect the **exact-match** path (`elif cached_kv is not None and not remaining_ids:`), which reuses the cached KV unconditionally.

In addition, `_maybe_store_prefix_cache` stores KV for **all** finished requests including multimodal ones (its docstring says "text-only"), so vision KV pollutes the cache.

## Suggested fix

Restrict the KV prefix cache to text-only requests, on both the store and fetch paths:

```python
# fetch side, in _process_prompts
if self.prefix_cache is not None and req.is_text_only and req.input_ids is not None:

# store side, in _maybe_store_prefix_cache
if req.is_text_only and req.input_ids is not None:
```

Multimodal requests then always run the full VLM forward (correctness), while text-only requests keep prefix caching. This patch was verified locally against 0.4.1 and resolves the reproduction above.

Note: `--disable-prefix-cache` does **not** disable the MLLM-path prefix cache — `_start_mllm` in `engine/batched.py` builds `MLLMSchedulerConfig` without forwarding `enable_prefix_cache`, so it always defaults to `True`.

Thump604:
## Summary

- restrict MLLM prefix-cache fetch and store to text-only requests
- keep image, video, and audio requests out of text-only batching paths
- invalidate legacy disk entries that may already contain media KV

## Verification

- 69 focused no-MLX tests passed; 20 MLX-only cases skipped
- new image/audio regressions are included in the Apple Silicon test job
- formatting, compile, and diff checks passed

The configured prefix-cache flag is already forwarded on current `main`; this fixes the remaining media-isolation defect.

Fixes #718.

### DIFF
--- a/tests/test_memory_cache.py
+++ b/tests/test_memory_cache.py
@@ -564,12 +564,12 @@ class TestGetAvailableMemory:
             pass
 
 
-def test_load_rejects_v3_cache_after_rewind_semantics_change(tmp_path, caplog):
-    """Caches written before safe MLLM rewind must not survive an upgrade."""
+def test_load_rejects_v4_cache_after_media_isolation_change(tmp_path, caplog):
+    """Caches that may contain media KV must not survive the text-only change."""
     import json
 
     (tmp_path / "index.json").write_text(
-        json.dumps({"version": 3, "model_fingerprint": "", "entries": []})
+        json.dumps({"version": 4, "model_fingerprint": "", "entries": []})
     )
     cache = MemoryAwarePrefixCache(MagicMock(), MemoryCacheConfig(max_memory_mb=1))
 
@@ -582,4 +582,4 @@ def test_load_rejects_v3_cache_after_rewind_semantics_change(tmp_path, caplog):
         },
     ):
         assert cache.load_from_disk(str(tmp_path)) == 0
-    assert "version mismatch: disk=3 current=4" in caplog.text
+    assert "version mismatch: disk=4 current=5" in caplog.text
--- a/tests/test_mllm_continuous_batching.py
+++ b/tests/test_mllm_continuous_batching.py
@@ -853,6 +853,68 @@ if __name__ == "__main__":
 
 
 class TestMLLMBatchGeneratorMTPGuards:
+    def test_process_prompts_never_fetches_media_from_token_only_cache(
+        self, monkeypatch
+    ):
+        from vllm_mlx.mllm_batch_generator import (
+            MLLMBatchGenerator,
+            MLLMBatchRequest,
+            MLLMBatchStats,
+        )
+
+        class FakeCache:
+            def merge(self, caches):
+                return self
+
+        monkeypatch.setattr(mx, "stream", lambda stream: nullcontext())
+        monkeypatch.setattr(
+            "mlx_lm.models.cache.make_prompt_cache",
+            lambda model, **kwargs: [FakeCache()],
+        )
+        monkeypatch.setattr(
+            "mlx_lm.sample_utils.make_sampler",
+            lambda **_: MagicMock(return_value=mx.array([1], dtype=mx.uint32)),
+        )
+        monkeypatch.setattr(
+            "mlx_lm.sample_utils.make_logits_processors", lambda **_: []
+        )
+
+        generator = MLLMBatchGenerator.__new__(MLLMBatchGenerator)
+        generator.max_kv_size = 0
+        generator._stats = MLLMBatchStats()
+        generator._pending_error_responses = []
+        generator._aborted_request_ids = set()
+        generator._prefill_progress = {}
+        generator.prefix_cache = MagicMock()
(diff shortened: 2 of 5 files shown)


janhilgard (approved):
The reasoning holds and the comment this replaces was the actively dangerous part — "cached KV from a previous VLM run is valid" is true for the text tokens and silently false for everything the media contributed, which is exactly the kind of claim that survives review because half of it is correct.

Restricting fetch *and* store to `is_text_only`, rather than only the fetch, is the right call: a media request that stores under a token-only key poisons later text-only requests even after the reading side is fixed.

## Checked against real caches

I run four `--mllm` servers with the prefix cache enabled, so I looked at what is actually on disk here: **27 GB** across 8 model directories, all at `version: 3`.

Two things follow. The version bump to 5 does what it says — those entries were already rejected by the existing bump to 4, and anything written at 4 will now be rejected too, so no media-contaminated entry can survive into the fixed code. Good.

The second is a question rather than a finding: rejected entries are *rejected*, not deleted. My 27 GB has been unloadable for two version bumps and is still sitting there, and nothing in the logs suggests it. If the load path already knows an entry is permanently unusable, pruning it — or at least logging the reclaimable size once at startup — would close that loop. Out of scope here; raising it because this PR is the third version bump and the disk cost compounds each time.

## One thing I could not check

`_request_has_media` reads `request.images or request.videos or request.audio`. That matches `MLLMBatchRequest` as constructed today, but it is a positive check for media rather than a check that the request is *only* text — if a future modality lands as a new field, it defaults to cacheable and the failure is silent and data-dependent, which is the worst shape for this class of bug. A single source of truth (deriving `is_text_only` in one place and asserting the two agree) would make the next modality a compile-time problem instead of a correctness one.

Approving — this is a clear improvement on the current behaviour and the version bump makes it safe to deploy.

azamamirza:
Independent verification for this PR. I hit the same defect in production on Aug 7 (Qwen2.5-VL-7B-Instruct-8bit, 67 distinct document pages, one prompt, temperature 0, 2 distinct outputs) and fixed it the same way in my deployment, which is where #768 came from before this review pointed me here. Closed #768 in favor of this one.

Ran this branch on Apple Silicon today: full suite 2697 passed / 24 skipped / 0 failed, and the three regression tests from #768 (media never stored, text-only still stored, is_text_only defaults to media) pass unmodified against this implementation. The v4 rejection test covers a gap my patch missed.

PR #729, opened 2026-08-23
fix(cli): honor --prefill-step-size with continuous batching

Thump604:
Refs #711

I pass `--prefill-step-size` into the continuous-batching scheduler so the configured chunk size is honored. The CLI and `SchedulerConfig` reject nonpositive sizes. Registry overrides are validated when constructing per-model scheduler clones, preserving global and sibling settings.

Verification at `0f97fa4`:
- Eight boundary/precedence regressions fail against the previous `8651193` source.
- All 68 focused CLI, scheduler-config and registry tests pass with the fix.
- Black, Ruff and diff checks pass.

The default remains 2048; the separate MLLM prefill setting is unchanged. This is source regression coverage, not a new live-model memory or performance measurement.

### DIFF
--- a/tests/test_batching.py
+++ b/tests/test_batching.py
@@ -177,6 +177,15 @@ class TestRequestOutput:
 class TestSchedulerConfig:
     """Tests for SchedulerConfig."""
 
+    @pytest.mark.parametrize("value", [0, -1])
+    def test_rejects_nonpositive_prefill_step_size(self, value):
+        with pytest.raises(ValueError, match="prefill_step_size must be > 0"):
+            SchedulerConfig(prefill_step_size=value)
+
+    @pytest.mark.parametrize("value", [1, 512, 2048])
+    def test_accepts_positive_prefill_step_size(self, value):
+        assert SchedulerConfig(prefill_step_size=value).prefill_step_size == value
+
     def test_default_config(self):
         """Test default scheduler config."""
         config = SchedulerConfig()
@@ -185,6 +194,7 @@ class TestSchedulerConfig:
         assert config.policy == SchedulingPolicy.FCFS
         assert config.prefill_batch_size == 8
         assert config.completion_batch_size == 32
+        assert config.prefill_step_size == 2048
 
     def test_custom_config(self):
         """Test custom scheduler config."""
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,4 +1,5 @@
-from types import SimpleNamespace
+import sys
+from types import ModuleType, SimpleNamespace
 
 import pytest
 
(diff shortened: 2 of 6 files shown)


janhilgard (approved):
Confirming this from production, since the report is easy to underestimate as a one-line omission.

Two of my servers run exactly the affected combination:

```
--continuous-batching ... --chunked-prefill-tokens 2048 --prefill-step-size 4096
```

On 0.4.1 (`61c78dc`) `serve_command` builds `SchedulerConfig` at line 262 without `prefill_step_size`, so both have been prefilling at the dataclass default of **2048**, not the 4096 they were configured with. Nothing in the logs says so — I only found out by reading this PR. Any measurement I have taken on that path with a non-default step size was measuring the default.

I checked whether the single-line fix is complete rather than assuming it:

- there are two `SchedulerConfig(` construction sites on `main` (299 and 585) and neither had the field
- the second is in `bench_command`, whose parser does **not** expose `--prefill-step-size`, so passing `args.prefill_step_size` there would raise rather than fix anything

So one hunk is the right scope, and the `serve`-side sites at 399 and 434 already forwarded it — the gap really was specific to the continuous-batching path.

Approving. Worth noting for whoever picks up the neighbouring work: silent flag drops are hard to notice precisely because the server starts fine and serves fine. A one-line "prefill step size: N" in the startup banner, next to the existing chunked-prefill line, would have made this visible on day one.

waybarrios:
Thanks for the focused fix. The main wiring looks correct, but I found two cases to address before merging:

1. `--prefill-step-size` accepts `0` and negative values. Once forwarded to continuous batching, `0` can produce a zero-progress prefill loop and stall requests. Please use the positive-integer parser, validate `SchedulerConfig.prefill_step_size > 0`, and add boundary tests.

2. Registry per-model overrides can be ignored. With global `--prefill-step-size 512` and an entry configured with `prefill_step_size: 2048`, `BatchedEngine` receives `512`. Please apply the resolved entry value to the cloned scheduler config and add a precedence test.

Everything else looks good, and CI is green. Once these cases are covered, this should be ready to merge.

Thump604:
Thanks for catching these. I fixed both in `0f97fa4`: the CLI uses the positive-integer parser, and `SchedulerConfig` rejects zero/negative prefill sizes. The registry now supplies the resolved entry value when constructing each scheduler clone, so it is validated and reaches `BatchedEngine` without mutating the global config or sibling entries.

The eight regressions fail for the expected reasons against `8651193`; with the fix, all 68 focused CLI, scheduler-config and registry tests pass. Black and Ruff pass. The 2048 default and separate MLLM setting are unchanged. [CI is running](https://github.com/waybarrios/vllm-mlx/actions/runs/33888379670); these are source tests, not a new model-memory benchmark.

Issue #730, opened 2026-08-24
[Bug]: Prefix-cache reuse silently disabled for hybrid architectures (Qwen3.5-family) since #691


djacobsmeyer:
### Environment

- Chip: Apple M5 Max
- RAM: 128 GB
- macOS version: 27.0 (26A5416b)
- Python version: 3.13.14
- vllm-mlx version: 0.4.1 (`origin/main` @ `4b654c0`, i.e. immediately after #691/`bf5ccb4` merged — reproduces on current `main` as of this writing)
- mlx / mlx-lm / mlx-vlm / mlx-audio versions: mlx 0.32.1, mlx-lm 0.31.3, mlx-vlm 0.6.15

### Model and quantization

`mlx-community/Qwen3.8-27B-8bit` (Qwen3.5-family hybrid architecture — per its own config, only 16 of 64 layers carry a real KV cache; the rest are SSM/linear-attention state layers). Any hybrid-architecture model should reproduce this — Qwen3.5/3.6/3.8 and siblings all mix `mlx_lm`/`mlx_vlm` `ArraysCache` state leaves with KV-cache leaves.

### Reproduction

```
vllm-mlx serve mlx-community/Qwen3.8-27B-8bit \
  --continuous-batching --enable-prefix-cache --port 8091
```

Send the same request twice, back to back, with a prompt long enough to clear `min_prefix_tokens` (default 128) — a few thousand tokens makes the effect obvious on the wall clock:

```
curl -s http://localhost:8091/v1/chat/completions \
  -H 'Content-Type: application/json' \
  -d '{"model":"mlx-community/Qwen3.8-27B-8bit","messages":[
        {"role":"system","content":"You are a helpful assistant. Answer with one word."},
        {"role":"user","content":"<~24k tokens of stable filler text>"}
      ],"max_tokens":1,"temperature":0}'
# run the exact same request a second time immediately after
```

Watch server stderr for `[chunked_prefill] Starting <uid>: N tokens, step=2048` — this line fires unconditionally on a genuine cache miss above `--prefill-step-size` (2048 tokens by default), from `_run_chunked_text_prefill`'s only call site (the miss branch of `_process_prompts`).

### Expected behavior

The second (byte-identical) request should hit the prefix cache: no `[chunked_prefill]` log lines, and wall time close to just the single-token decode cost (well under a second for a ~24k-token prompt on an M-series Mac).

### Actual behavior

[...]

PR #731, opened 2026-08-24
fix(mllm): make prefix-cache rewind gate aware of non-KV hybrid leaves

djacobsmeyer:
Refs #730

> **Superseded by #744.** Per [review on #730](https://github.com/waybarrios/vllm-mlx/issues/730#issuecomment-5468495295) and [review on this PR](https://github.com/waybarrios/vllm-mlx/pull/731#issuecomment-5468493922), owner PR [#744](https://github.com/waybarrios/vllm-mlx/pull/744) implements detached prompt-boundary ownership/replay for the same hybrid-cache path and supersedes this narrow rewind-gate fix. Kept open for reference until #744 lands; not intended to merge independently. The fix described below remains deployed on the reporter's fork (pin `bd283d8`) in production.

## Summary

#691 (`bf5ccb4`, "harden MLLM prefix cache ownership and rewind safety") introduced `_can_rewind_prefix_cache`/`_rewind_prefix_cache`, requiring every cache leaf to expose real KV semantics before any rewind is permitted. Hybrid architectures (Qwen3.5-family) mix KV leaves with `ArraysCache` (SSM/linear-attention) leaves that have no `.keys` attribute at all and inherit `is_trimmable() -> False` — so the all-or-nothing gate silently rejects every rewind for any hybrid model, both at completion-time store and at exact-match fetch. Verified live: an identical 24k-token prompt resent back-to-back against `mlx-community/Qwen3.8-27B-8bit` showed zero speedup and re-ran the full chunked prefill both times on `origin/main` @ `4b654c0`; the same test against `db90a1c` (the commit immediately before #691) showed a ~58x speedup. The failure is silent — rejection only logs at `DEBUG`.

## What db90a1c did, and why it's the right semantics to restore

`db90a1c`'s `_trim_cache_offset` (`memory_cache.py`, the function #691 replaced) handled recognized KV-family types explicitly and had a catch-all `else: trimmed.append(layer_cache)` for everything else — no rewind attempted, no rejection, because a non-KV leaf isn't a position-indexed buffer and "rewind by N tokens" was never a meaningful operation on it. That catch-all is where `ArraysCache` leaves landed, via a raw shared reference (no copy) — exactly the ownership gap #691 set out to close.

## Changes

- `_can_rewind_prefix_cache`: skip leaves without a `.keys` attribute entirely (`hasattr(cache, "keys")` as the discriminator); leaves that *do* have `.keys` — including an empty/invalid `RotatingKVCache` (`.keys = None`, attribute present) — are still held to the full existing contract and still correctly fail closed.
- `_rewind_prefix_cache`'s post-copy trim loop: same skip, since `_copy_prefix_cache` already gave non-KV leaves a correct isolated copy via `_copy_cache_layer`'s `from_state`/`meta_state` fallback — #691's real improvement over `db90a1c`'s shared reference, preserved here, not reverted.
- No change to `_copy_prefix_cache`/`_prepare_rotating_caches` (the prefix/LCP-match fetch path) — confirmed by dedicated regression test that this path was never affected: `_prepare_rotating_caches` already skips non-`RotatingKVCache` leaves, and `_copy_cache_layer`'s generic fallback already handles `ArraysCache`.

This is the same problem class #642 already named in this file ("hybrid layers (`ArraysCache`) expose their mutable state container via `.state`") — not a new architecture edge case to this codebase.

## Type of change

- Bug fix

## Surface touched

- KV cache / prefix cache

## Test plan

```
pytest tests/test_mllm_continuous_batching.py -v
pytest tests/ -q
```

7 new cases in `TestChunkedPrefillCacheHandling`, built on real `KVCache` + real `ArraysCache` objects (not mocks): both gate functions directly (including that a genuinely invalid KV leaf still fails closed alongside a valid non-KV leaf), an end-to-end store test, and end-to-end `_process_prompts` tests for both exact-match and prefix-extension fetch.

**Verified these are real regression tests, not tautologies:** reverted the fix and reran — the 4 tests targeting the actually-broken paths failed; the 3 tests covering paths that were never broken (bad-KV-leaf rejection, `_copy_prefix_cache`, prefix-extension) correctly passed in both states. Restored the fix, all 7 pass.

## Test results

```
$ pytest tests/ -q
2694 passed, 25 skipped, 27 deselected
```
(`origin/main` baseline: 2687 passed; +7 new, 0 regressed.)

```
$ ruff check vllm_mlx/ tests/ --select E,F,W --ignore E402,E501,E731,F811,F841
All checks passed!
$ black --check vllm_mlx/ tests/
All done! ✨ 🍰 ✨
$ mypy vllm_mlx/mllm_batch_generator.py --ignore-missing-imports --no-error-summary
# zero new errors on the changed lines
```

## Performance impact

- Hardware: Apple M5 Max, 128 GB RAM
- Model and quantization: `mlx-community/Qwen3.8-27B-8bit`
- Benchmark: identical ~24-28k-token prompt sent twice, back to back, `max_tokens=1`, `temperature=0`, measured via `curl` wall time and server logs
- Metrics measured: end-to-end wall time (cold vs. warm), presence/absence of `[chunked_prefill]` log lines
- Delta vs. `main` (i.e. vs. the still-broken state): cold 46.60s → warm **0.42s** (~111x). On `main` as-is (unfixed), warm shows no improvement over cold at all (both ~46s) — this PR is a pure improvement with no metric that could regress, since the broken behavior is "always full re-prefill."

## Output parity

Not applicable in the sense of "different code path, different tokens": this changes *whether* a cache is reused, not how generation is sampled. The KV state served by the fixed path (real KV leaves trimmed by the existing rewind contract, non-KV leaves passed through as an isolated copy) is bit-identical to what the model would have computed from scratch for the same tokens — that's the entire correctness property being restored, not something new to verify separately. No sampling, tokenization, or model code changed.

## Risk notes

Scoped to two `classmethod`s used at exactly the call sites #691 added them for (exact-match fetch, completion-time store). No behavior change for any model where every cache leaf is KV-shaped (pure-transformer models) — the strict contract is untouched for that case, verified by `test_can_rewind_hybrid_cache_still_rejects_bad_kv_leaf` and the pre-existing non-hybrid tests in the same file continuing to pass unchanged.

### DIFF
--- a/tests/test_mllm_continuous_batching.py
+++ b/tests/test_mllm_continuous_batching.py
@@ -2053,6 +2053,230 @@ class TestChunkedPrefillCacheHandling:
             mx.eval(cache.keys, cache.values)
         return cache
 
+    def _make_fake_arrays_cache(self):
+        """Create a real ArraysCache (SSM/hybrid state leaf, no .keys)."""
+        from mlx_lm.models.cache import ArraysCache
+
+        cache = ArraysCache(2)
+        cache[0] = mx.ones((1, 4, 8))
+        cache[1] = mx.full((1, 4, 8), 2.0)
+        mx.eval(*[c for c in cache.state if c is not None])
+        return cache
+
+    def _make_hybrid_cache(self, kv_offset=50):
+        """A cache list mixing a real KVCache leaf with a real ArraysCache
+        leaf, mirroring a hybrid (e.g. Qwen3.5-family) architecture where
+        only some layers carry a real KV cache."""
+        return [
+            self._make_fake_kv_cache(offset=kv_offset),
+            self._make_fake_arrays_cache(),
+        ]
+
+    def test_can_rewind_hybrid_cache_ignores_non_kv_leaf(self):
+        """A non-KV leaf (no .keys) must not block rewind of the KV leaf."""
+        from vllm_mlx.mllm_batch_generator import MLLMBatchGenerator
+
+        hybrid = self._make_hybrid_cache(kv_offset=50)
+        assert MLLMBatchGenerator._can_rewind_prefix_cache(hybrid, 1) is True
+
+    def test_can_rewind_hybrid_cache_still_rejects_bad_kv_leaf(self):
+        """The strict contract must still apply to the KV leaf itself --
+        offset < trim_by must still fail closed even with a non-KV leaf
+        present."""
+        from vllm_mlx.mllm_batch_generator import MLLMBatchGenerator
+
+        hybrid = self._make_hybrid_cache(kv_offset=0)  # offset 0 < trim_by
+        assert MLLMBatchGenerator._can_rewind_prefix_cache(hybrid, 1) is False
+
+    def test_rewind_hybrid_cache_trims_kv_leaf_and_copies_non_kv_leaf(self):
+        """Rewinding a hybrid cache must trim the KV leaf's offset and
+        return an isolated copy of the non-KV leaf (not the same object,
+        not blocked by lacking .trim())."""
+        from vllm_mlx.mllm_batch_generator import MLLMBatchGenerator
+
+        kv, arrays = self._make_hybrid_cache(kv_offset=50)
+        result = MLLMBatchGenerator._rewind_prefix_cache([kv, arrays], 1)
+
+        assert result is not None
+        new_kv, new_arrays = result
+        assert new_kv.offset == 49
+        assert new_arrays is not arrays  # isolated copy, per bf5ccb4's ownership goal
+        assert len(new_arrays.state) == len(arrays.state)
+
+    def test_hybrid_store_prefix_snapshot_succeeds(self):
+        """The completion-store path must not silently skip a hybrid cache
+        the way it did before this fix (only a DEBUG log, no error)."""
+        from vllm_mlx.mllm_batch_generator import MLLMBatchGenerator, MLLMBatchStats
+        from vllm_mlx.memory_cache import MemoryAwarePrefixCache, MemoryCacheConfig
+
+        gen = MLLMBatchGenerator.__new__(MLLMBatchGenerator)
+        gen._stats = MLLMBatchStats()
+        gen.prefix_cache = MemoryAwarePrefixCache(
+            model=object(), config=MemoryCacheConfig(min_prefix_tokens=1)
+        )
+
+        hybrid = self._make_hybrid_cache(kv_offset=50)
+        ok = gen._store_prefix_snapshot(
+            list(range(50)),
+            hybrid,
+            trim_by=1,
+            request_id="hybrid-store",
+            source="completion",
+        )
+
+        assert ok is True
+        assert len(gen.prefix_cache._entries) == 1
+
+    def test_copy_prefix_cache_handles_hybrid_leaves(self):
+        """_copy_prefix_cache (the prefix/LCP-match path's cache prep) was
+        never part of this bug -- it already produces an isolated copy for
+        both KV and non-KV leaves via _copy_cache_layer's from_state
+        fallback. This locks that in as regression coverage."""
+        from vllm_mlx.mllm_batch_generator import MLLMBatchGenerator
+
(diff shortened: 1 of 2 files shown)


djacobsmeyer:
Looked into the `test-matrix (3.11)` / `tests` CI failure on this PR (run 32739507550).

The failure is:

```
FAILED tests/test_simple_engine_thread_pinning.py::test_restart_after_stop_mid_stream_uses_a_fresh_worker
AssertionError: restart must not reuse the old thread, saw ['simple-generate_0/140225380611776']
```

This test isn't touched by this PR's diff (which only changes `vllm_mlx/mllm_batch_generator.py` plus a new test file), and it's asserting that a freshly spawned `ThreadPoolExecutor` worker gets a distinct OS thread id from the one that was just joined — which is a timing-dependent property of the OS thread allocator, not of this PR's code.

I found the same test failing the same way (same job, same assertion, different thread id) on `main` five days ago in run [32336428329](https://github.com/waybarrios/vllm-mlx/actions/runs/32336428329) (commit `7afa61a`), with everything else green. `test-matrix (3.12)` on this run shows as cancelled rather than failed — that's just matrix `fail-fast` (no `fail-fast: false` set in `.github/workflows/ci.yml`) tearing down the sibling job once 3.11 failed, not an independent failure.

I ran this test 30x locally (macOS arm64, Python 3.11.15 via uv) and couldn't reproduce it — consistent with it being a race specific to Linux thread-id recycling under the `ubuntu-latest` runner's scheduler, not something this PR's code path can influence.

Recommendation: re-run the job (should go green), and consider marking `test_restart_after_stop_mid_stream_uses_a_fresh_worker` `flaky`/adding a retry, or asserting thread identity a different way, since it's now failed on both `main` and a PR unrelated to it. Happy to open a separate issue/PR for that if useful.

Thump604:
I checked the exact head `3a53b95f140eb7f32989311d0659f9f47be04a3b` and current main `22efb47968f96e6e7a79704f4eb0184cda4e0068`. Your source diagnosis is consistent with current main: the all-leaves rewind gate still rejects hybrid `ArraysCache` leaves. However, owner PR [#744](https://github.com/waybarrios/vllm-mlx/pull/744) explicitly supersedes this approach and implements detached prompt-boundary ownership/replay for the same hybrid path.

Please do not merge #731 independently of #744. The maintainer should hold or close this duplicate after reviewing #744; #730 should remain open until the replacement is reviewed and its exact current-head tests/qualification satisfy the reported cache-hit behavior. The #731 preflight also flags its `Fixes #730` relationship while acceptance is still pending, so any remaining active relationship should use `Refs #730`.

djacobsmeyer:
Agreed — #744's detached prompt-boundary ownership/replay approach addresses failure modes this PR's narrow rewind-gate fix didn't attempt to: cancellation handling, memory accounting for auxiliary prompt logits, and preserving request-local sampler state across interleaved prefill. This PR was scoped to the `.keys`-based rewind-gate check alone; #744 is the more complete fix for the hybrid-cache path.

Updated this PR's body to `Refs #730` and added a supersession note per the review; leaving it open for reference rather than pushing to merge independently.

I'll run the reporter-side qualification of #744 requested on #730 — same hardware/artifact (`mlx-community/Qwen3.8-27B-8bit`) used for this PR's numbers, current upstream head, identical long prompt sent twice at `max_tokens=1`, plus a control run against `main` without #744 to confirm the regression baseline. Will post results on #730 with a cross-link here.

Thump604 (commented):
I rechecked the supersession at #731 head `3a53b95`. Keeping this narrow rewind-gate PR out of merge while #744 owns detached hybrid-cache replay is correct; #731 is also still based on `4b654c0` and has failing current checks.

The promised reporter-side qualification is now present on #730. It supports #744 for the requested identical-prompt protocol, but it also reports that standard multi-turn history growth still misses through the non-trimmable hybrid path. Please keep #730 open for that separate follow-up and do not merge #731 independently; after #744 is reviewed and accepted, close #731 as superseded.

Issue #522, opened 2026-05-10
tokens_to_prefill always matches prompt_tokens, and caching does not seem to work when using claude code


michael-malandra:
Title: Prefix cache hit rate always 0% despite --enable-prefix-cache and --use-paged-cache flags
Environment

Hardware: Apple M3 Ultra, 256GB unified RAM
OS: macOS
vllm-mlx version: 0.3.0
Model: unsloth/Qwen3.6-35B-A3B-MLX-8bit
Python: 3.12.3

Serve command
bashvllm-mlx serve unsloth/Qwen3.6-35B-A3B-MLX-8bit \
  --served-model-name "qwen" \
  --prefill-batch-size 16384 \
  --prefill-step-size 8192 \
  --enable-prefix-cache \
  --cache-memory-percent 0.85 \
  --kv-cache-quantization \
  --kv-cache-quantization-bits 8 \
  --host 0.0.0.0 \
  --port 8081 \
  --max-tokens 131072 \
  --max-request-tokens 131072 \
  --default-chat-template-kwargs '{"preserve_thinking": true}' \
  --enable-auto-tool-choice \
  --trust-remote-code \
  --continuous-batching \
  --enable-metrics \
  --tool-call-parser qwen3_coder \
  --reasoning-parser qwen3 \
  --use-paged-cache
Describe the bug
Prefix cache hit rate is permanently 0% regardless of configuration. Every request re-encodes the full prompt from position 0, including shared prefixes like system prompts and conversation history. This causes severe prefill latency — 30+ seconds per request on long conversations — because the full context (often 50-99k tokens) is recomputed on every call.
Verified via metrics endpoint:
vllm_mlx_cache_type{cache_type="prefix_cache"} 0.0
Sending identical requests back to back produces no cache hits. The flags --enable-prefix-cache and --use-paged-cache are accepted by the CLI without error but appear to have no effect on actual caching behavior.
Expected behavior
Subsequent requests sharing a common prefix (system prompt, conversation history) should reuse cached KV states, reducing prefill tokens and latency significantly.
Suspected root cause
[...]

PR #732, opened 2026-08-24
feat(api): surface prefix-cache reuse via usage.prompt_tokens_details.cached_tokens

djacobsmeyer:
Refs #522
Depends on #731 for the live "both numbers move together" demonstration below (this PR's plumbing is independently correct and tested without it — see Performance impact).

## Summary

This repo doesn't currently expose per-request prefix-cache hit information anywhere in the API response, for any engine path — the internal `Request.cached_tokens` bookkeeping (present since the shared Phase-4 continuous-batching work) has always fed scheduler logging and `/v1/status` only, never the OpenAI-compatible `usage` object. This PR adds that surface — `usage.prompt_tokens_details.cached_tokens` (OpenAI schema), plus the equivalent Responses API and Anthropic-adapter fields — populated correctly for the `SimpleEngine`, plain-text `BatchedEngine`, and MLLM `BatchedEngine` (continuous-batching) paths.

**Why now:** #522 ("tokens_to_prefill always matches prompt_tokens, and caching does not seem to work when using claude code") has sat open since May with a maintainer reply explicitly naming *"metric reporting mismatch"* as one of several candidate explanations still needing to be ruled out, and noting *"SimpleEngine pure-LLM, MLLM text route, continuous batching, and paged-cache path do not all share the same implementation."* This field answers that question directly, in the API response itself, for every request — no `/v1/cache/stats` before/after diffing needed.

## Changes

- `PromptTokensDetails` / `Usage.prompt_tokens_details` (OpenAI schema), populated on every chat/completions path, streaming and non-streaming; the pre-existing but previously-unpopulated Responses API and Anthropic-adapter cache fields are wired up too.
- Plain-text (`scheduler.py`) and `SimpleEngine`: threads the existing `Request.cached_tokens` / new `peak_cached_tokens` high-water-mark through `RequestOutput`/`GenerationOutput`, correct across both streaming (finished-chunk) and non-streaming (drained+merged) construction, and cache-fallback/retry resets.
- MLLM (`mllm_scheduler.py`/`mllm_batch_generator.py`): same accounting, threaded through `MLLMBatchRequest` → `MLLMBatchResponse` → `MLLMRequest` → `RequestOutput`, covering the prefix/LCP-match, exact-match, and cache-miss branches at both the primary (`_process_prompts`) and interleaved chunked-prefill (`_generation_step`) fetch sites.

**Coverage boundary, disclosed rather than left for review to find:** MLLM requests served through `SimpleEngine` (non-continuous-batching multimodal serving, if/where that combination exists) were not touched or verified — this PR covers `SimpleEngine` for text and `BatchedEngine` for both text and MLLM, which are the three paths #522 and this investigation actually exercised live.

## Type of change

- New feature

## Surface touched

- OpenAI API endpoints
- Anthropic API endpoints
- Scheduler / batching
- KV cache / prefix cache

## Test plan

```
pytest tests/test_cache_token_usage.py tests/test_mllm_continuous_batching.py tests/test_server.py -v
pytest tests/ -q
```

Unit coverage for the data-model plumbing, `RequestOutputCollector` merge semantics (keeps the max reuse count), `get_usage()`, end-to-end streaming/non-streaming chat completions (plain-text path); MLLM-side dataclass defaults and three behavioral tests (exact-hit accounting, miss-reset, scheduler-level propagation from response → request → output, unaccumulated).

## Test results

```
$ pytest tests/ -q
2701 passed, 25 skipped, 27 deselected
```
(`origin/main` baseline 2687; +14 new, 0 regressed.)

```
$ ruff check vllm_mlx/ tests/ --select E,F,W --ignore E402,E501,E731,F811,F841
All checks passed!
$ black --check vllm_mlx/ tests/
All done! ✨ 🍰 ✨
```

## Performance impact

- Hardware: Apple M5 Max, 128 GB RAM
- Model and quantization: `mlx-community/Qwen3.8-27B-8bit`
- This PR only adds counter bookkeeping (integer arithmetic and field assignment on objects already being constructed per-request/per-step) — no additional model inference, no additional allocation on the hot tensor path. Not independently benchmarked in isolation; verified live alongside #731 where the dominant cost (prefill) changed by ~111x for unrelated reasons (that fix), and this PR's own overhead was not separately measurable against that.
- Live demonstration, this PR + #731 together: identical ~28k-token prompt sent twice — cold 46.60s / `cached_tokens: 0`, warm **0.42s** / `cached_tokens: 28062`.

## Output parity

No change to sampling, tokenization, scheduling decisions, or generated content — this PR only adds a new, previously-always-zero field to response metadata. Verified by the existing exact-equality usage-dict assertions in `test_server.py` (two of which needed updating for the new field, included in this PR, since they were added on `main` after this branch's own commits last touched them) continuing to pass for every other field.

## Risk notes

Purely additive: new dataclass fields default to `0`/unset, new response fields default to the same "no reuse" value pre-existing clients already see today. No existing field's value or type changes. The one thing worth a reviewer's attention: this PR's own live proof of "the number is real, not just wired to zero" currently depends on #731 also landing (or testing against a non-hybrid model) — noted explicitly rather than glossed over.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,7 +108,7 @@ jobs:
             tests/test_streaming_tool_filter.py \
             tests/test_tool_call_promotion.py \
             -v --tb=short \
-            -k "not Integration and not InjectJson and not TestMLXMultimodalLMCache" \
+            -k "not Integration and not InjectJson and not TestMLXMultimodalLMCache and not GetUsage and not ChatCompletionCachedTokens" \
             --cov=vllm_mlx \
             --cov-report=term-missing \
             --cov-report=xml
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,6 +77,7 @@ jobs:
             tests/test_memory_cache.py \
             tests/test_prefix_cache.py \
             tests/test_mllm_cache.py \
+            tests/test_cache_token_usage.py \
             tests/test_api_models.py \
             tests/test_api_utils.py \
             tests/test_request.py \
@@ -178,6 +179,7 @@ jobs:
             tests/test_batched_engine_mllm_config.py \
             tests/test_mllm_continuous_batching.py \
             tests/test_mllm_cache.py \
+            tests/test_cache_token_usage.py \
             tests/test_optimizations.py \
             tests/test_simple_engine.py \
             tests/test_simple_engine_prefix_trie_cache.py \
--- a/tests/test_cache_token_usage.py
+++ b/tests/test_cache_token_usage.py
@@ -8,6 +8,8 @@ non-streaming and (the historically broken) streaming chat completion paths.
 """
 
 import json
+import platform
+import sys
 
 import pytest
 
@@ -228,3 +230,75 @@ class TestChatCompletionCachedTokens:
         # Every usage-bearing chunk reports the cache reuse.
         for usage in usages:
             assert usage["prompt_tokens_details"] == {"cached_tokens": 15}
+
+
+@pytest.mark.skipif(
+    sys.platform != "darwin" or platform.machine() != "arm64",
+    reason="Requires MLX (vllm_mlx.mllm_scheduler imports mlx.core)",
+)
+class TestAbortAndReuseRequestId:
+    """Aborting a request must not leak cache accounting into a reused ID.
+
+    The MLLM scheduler pools requests by ``request_id`` in several dicts
+    (``requests``, ``running``, ``request_id_to_uid``/``uid_to_request_id``).
+    A client is free to reuse a request ID once the previous request has
+    finished/aborted; the request-lifetime contract is that ``add_request``
+    always creates a brand-new ``MLLMRequest``, so ``cached_tokens`` and its
+    high-water mark ``peak_cached_tokens`` must start fresh at 0 rather than
+    inheriting whatever the aborted request last reported.
+    """
+
+    def _make_scheduler(self):
+        from vllm_mlx.mllm_scheduler import MLLMScheduler
+
+        class FakeTokenizer:
+            eos_token_id = None
+
+            def encode(self, text):
+                return [1, 2, 3]
+
+        class FakeProcessor:
+            tokenizer = FakeTokenizer()
+
+        class FakeModel:
+            config = None
+
+        return MLLMScheduler(model=FakeModel(), processor=FakeProcessor())
+
+    def test_abort_mid_flight_then_reuse_request_id_gives_fresh_peak(self):
+        from vllm_mlx.request import RequestStatus
+
+        scheduler = self._make_scheduler()
+        request_id = "dup-request-id"
+
+        scheduler.add_request(prompt="hello", request_id=request_id, max_tokens=8)
+        first = scheduler.requests[request_id]
+
+        # Simulate the request having progressed far enough to record a
+        # prefix-cache hit before the client disconnects mid-flight.
+        first.status = RequestStatus.RUNNING
+        first.cached_tokens = 42
+        first.peak_cached_tokens = 42
+        scheduler.waiting.remove(first)
+        scheduler.running[request_id] = first
+        scheduler.request_id_to_uid[request_id] = 7
+        scheduler.uid_to_request_id[7] = request_id
+
+        assert scheduler.abort_request(request_id) is True
+
+        # Abort must fully clear the ID from every bookkeeping structure --
+        # a stale entry in any of these would let the old request's state
+        # leak into the reused ID below.
+        assert request_id not in scheduler.requests
+        assert request_id not in scheduler.running
+        assert request_id not in scheduler.request_id_to_uid
+        assert 7 not in scheduler.uid_to_request_id
+
+        # Client reuses the same request ID for an unrelated new request.
+        scheduler.add_request(prompt="hello again", request_id=request_id, max_tokens=8)
+        second = scheduler.requests[request_id]
+
+        assert second is not first
+        assert second.status == RequestStatus.WAITING
+        assert second.cached_tokens == 0
+        assert second.peak_cached_tokens == 0
--- a/tests/test_cache_token_usage.py
+++ b/tests/test_cache_token_usage.py
@@ -1,0 +1,230 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for surfacing prefix-cache reuse in the OpenAI-compatible usage.
+
+The prefix cache already computes how many prompt tokens are served from the
+KV/prefix cache; these tests lock in that the value is threaded all the way to
+the client under ``usage.prompt_tokens_details.cached_tokens`` for both the
+non-streaming and (the historically broken) streaming chat completion paths.
+"""
+
+import json
+
+import pytest
+
+from vllm_mlx.api.models import PromptTokensDetails, Usage
+from vllm_mlx.engine.base import GenerationOutput
+from vllm_mlx.output_collector import RequestOutputCollector
+from vllm_mlx.request import Request, RequestOutput, SamplingParams
+
+
+class TestCachedTokensDataModel:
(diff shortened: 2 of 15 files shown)


Thump604 (changes requested):
I rechecked the exact head `28772d94afd777d40dbd36554adb76e98f87c2d5` against current `main` (`22efb47968f96e6e7a79704f4eb0184cda4e0068`). The normal cache-token plumbing is useful, but this head has one correctness blocker in the MTP/deferred-response path.

In `vllm_mlx/mllm_scheduler.py#L688-L721`, every response unconditionally assigns `request.cached_tokens = response.cached_tokens` before building the `RequestOutput`. The MTP wrapper then appends deferred draft responses at `vllm_mlx/mllm_batch_generator.py#L2798-L2805` and `#L2822-L2829` without a `cached_tokens` argument; `MLLMBatchResponse.cached_tokens` defaults to 0 at `#L240-L256`. A primary response with a prefix-cache hit is therefore followed by a draft response that overwrites the request value with 0, and the emitted output/usage reports a miss. This is a defect in the submitted change, not merely merge-base drift, and contradicts the head comment that the value is stable on every response.

Please carry the stable cache count through every synthetic/deferred response, including draft stop/length responses, or make the scheduler update it monotonically with an explicit request-lifetime contract. Add a focused MTP cache-hit regression that exercises the wrapper and asserts the cached count on every output, including the final deferred draft, and in the client usage object. The existing single-response scheduler test and normal-path cache tests do not cover this seam.

The branch is based on an older main and has no current check evidence in the metadata sweep; please rebase or otherwise validate the corrected head and rerun the focused tests plus current CI. I did not run MLX or inference; source-level evidence is sufficient for this review.

djacobsmeyer:
Thanks — confirmed, this is real. Traced it to the exact seam you named: `mllm_scheduler.py`'s per-response loop unconditionally did `request.cached_tokens = response.cached_tokens`, and the MTP wrapper's deferred-draft `MLLMBatchResponse` constructors in `_mtp_next` (mllm_batch_generator.py) never pass `cached_tokens`, so it defaults to 0. A decode-step primary response also legitimately reports 0 (no cache lookup happens on decode). Either one landing after the prefill hit clobbered the reported count back to a miss, including on the finished output.

**Chosen fix: monotonic request-lifetime contract**, not threading the value through every deferred-response constructor. I mirrored the pattern already used by the LLM-path `scheduler.py` in this same branch (`peak_cached_tokens`, a high-water mark): `request.cached_tokens` keeps tracking the raw per-response value as before, but the emitted `RequestOutput` now reports `request.peak_cached_tokens = max(request.peak_cached_tokens, request.cached_tokens)`, so a later zero-carrying response can no longer clobber an already-established hit. I checked this scheduler's request lifecycle for the cases you flagged (aborts, retries, reuse) — there's no preemption/recompute path that re-prefills a live `MLLMRequest` here (unlike the paged-cache LLM path, which is exactly why it needed the same pattern), so a legitimate decrease never occurs and the peak is safe for the whole request lifetime. I picked this over threading `cached_tokens` through every draft-response constructor because it's a single change at the one place that already owns the request-lifetime value, doesn't require every future synthetic-response call site to remember to pass it, and matches the codebase's existing idiom instead of adding a second one.

Added `test_mtp_deferred_draft_does_not_clobber_cached_tokens` in `tests/test_mllm_continuous_batching.py`, replaying the exact response shape `_mtp_next` produces — a decode-step primary response followed by the deferred draft that finishes the request — and asserting the cache-hit count on every output and on `RequestOutput.usage["cached_tokens"]`, including the final deferred draft. Also tightened the existing `test_process_batch_responses_surfaces_cached_tokens` to assert `peak_cached_tokens` explicitly.

Rebased onto current `main` (`22efb47968f96e6e7a79704f4eb0184cda4e0068`), clean, no conflicts. Focused tests: `tests/test_mllm_continuous_batching.py` 60 passed, `tests/test_cache_token_usage.py` 11 passed. Full suite: 2968 passed / 25 skipped / 27 deselected (deselected = `slow`/`integration`, unchanged from the pre-fix baseline of 2967 passed on the same rebased head). New head: `b4a9b79f5b15a8ce4aaae0d45b44f803de17f32e`, pushed, CI running.

Thump604 (commented):
I rechecked #732 at `b4a9b79` against main `22efb47`. The monotonic `peak_cached_tokens` fix addresses the reported MTP clobber: the deferred responses at `mllm_batch_generator.py:2798-2805` and `:2822-2829` default to zero, while the scheduler now emits the request-lifetime maximum.

I also checked the lifecycle: each add creates a fresh `MLLMRequest`, abort removes the request/UID maps, failed preprocessing is removed rather than retried, and UIDs are monotonic. I found no live-request preemption/recompute path that would make a lower value legitimate. The 10 reported checks are green, including the Apple job containing the MLLM regression. Please map `tests/test_cache_token_usage.py` to a CI job (and, preferably, add an explicit abort/reuse-of-the-same-request-ID regression), then this is a candidate for maintainer decision.

djacobsmeyer:
Mapped `tests/test_cache_token_usage.py` into CI, and added the abort/reuse-ID regression. Final pushed head: `090fb61` (CI green: https://github.com/waybarrios/vllm-mlx/actions/runs/33337133353).

**CI mapping**: the file is listed in both the no-MLX unit-test job and the Apple-Silicon job, mirroring `test_mllm_cache.py`'s existing dual-job listing. Most of the file's 12 tests are plain dataclass/pydantic-model checks that run fine without MLX, but two classes (`TestGetUsage`, `TestChatCompletionCachedTokens`) import `vllm_mlx.server`, which needs `uvicorn` — not part of the no-MLX job's minimal dependency set — so (mirroring the existing `InjectJson`/`TestMLXMultimodalLMCache` precedent) they're excluded from that job's `-k` filter and run under the Apple job's full install instead. The new regression is `pytest.mark.skipif`-gated to Apple Silicon (it imports `vllm_mlx.mllm_scheduler`, which hard-imports `mlx.core`), so it collects cleanly and skips on the no-MLX runner.

My first push got this mapping wrong — it listed the file in test-matrix without accounting for the uvicorn dependency, and the real CI run failed with `ModuleNotFoundError: No module named 'uvicorn'` on exactly those two classes. Fixed in the second commit; both jobs are now green and the actual mapped test files/classes visibly execute in the run logs (test-matrix runs 10/12 with 2 correctly skipped; test-apple-silicon runs all 12/12, including the new regression, PASSED).

**New regression**: `TestAbortAndReuseRequestId::test_abort_mid_flight_then_reuse_request_id_gives_fresh_peak`. It builds an `MLLMScheduler` with a fake model/processor, adds a request, marks it `RUNNING` with `cached_tokens=42`/`peak_cached_tokens=42` to simulate a cache hit mid-flight, aborts it, asserts the ID is fully cleared from `requests`/`running`/`request_id_to_uid`/`uid_to_request_id`, then calls `add_request` again with the *same* `request_id` and asserts the new `MLLMRequest` is a distinct object with `cached_tokens == 0` and `peak_cached_tokens == 0` — no leakage from the aborted request into the reused ID. Verified it's a real regression guard by reverting the `self.requests.pop(request_id, None)` line in `abort_request` locally: the test fails as expected.

**Local runs** (isolated `uv venv`, no model download/serving): full `test_cache_token_usage.py` 12/12 passed under full deps; mapped no-MLX subset 128 passed/8 deselected; `ruff`/`black` clean.

Thump604 (approved):
I rechecked exact head `090fb616`. The requested follow-up is complete.

`tests/test_cache_token_usage.py` is now mapped into both Linux and Apple CI with the server-dependent classes excluded only from the minimal Linux job. I verified the executed logs: the dependency-light cache-token cases run on Linux, and all 12 cases pass on Apple, including the new abort/reuse-of-the-same-request-ID regression. The workflow is green across every required job.

The production fix remains the previously reviewed monotonic request-lifetime cache-count contract; this follow-up adds only the requested CI mapping and lifecycle regression. I found no remaining blocker on this head.

PR #733, opened 2026-08-24
fix(video): video requests silently dropped, then broken on mlx-vlm 0.6

sbayer2:
Six fixes for a chain of failures that made video requests unusable, each of which returned HTTP 200 (or a fluent answer) rather than an error. Found while serving `mlx-community/Qwen3.8-27B-4bit` and reproduced end to end against an 8.17s 2604x2160 HEVC `.mov` from an iPhone.

The headline symptom: a 22 MB video attachment cost **exactly as many prompt tokens as sending no video at all** — 68 both times, byte-identical replies — and the model politely explained that no video had been attached.

## What was wrong

**1. `is_mllm_model()` guessed capability from the model's name** (`vllm_mlx/api/utils.py`)

`_try_read_config_json()` only reads `config.json` when the argument resolves to a local directory. Given an HF repo ID it falls through to substring matching against `MLLM_PATTERNS`. `mlx-community/Qwen3.8-27B-4bit` matches none of those patterns, despite a config declaring `vision_config`, `image_token_id: 248056` and `video_token_id: 248057` — so it loaded as a text-only LLM (`SimpleEngine loaded: ... (MLLM=False)`) and every media content part was flattened out of the request with no error.

The config is in the local hub cache the whole time. Now: local dir, then `try_to_load_from_cache()` (no network), then the name patterns as the pre-download fallback.

Audited against all 25 models in one hub cache: six flip `False -> True`, all genuinely multimodal (Qwen3.5/3.6/3.8 `ForConditionalGeneration`, Qwen3-Omni, Muse-Glimmer), each carrying `vision_config`. Nothing flips the other way.

**2. The native video path imported a module mlx-vlm had deleted** (`vllm_mlx/models/mllm.py`)

`from mlx_vlm.video_generate import process_vision_info, generate` — that module does not exist in mlx-vlm 0.6.x (layout is `mlx_vlm.generate.*`; `process_vision_info` was removed). Every video request against a model with `video_token_id` raised `ModuleNotFoundError` -> HTTP 500.

Ported to `mlx_vlm.utils.load_video` + `mlx_vlm.generate.generate(video=...)`. Decoding happens here rather than inside `prepare_inputs` so `video_max_frames` is honored (upstream's default cap is 768 frames) and the result can be validated before it reaches the model. Falls back to the frames-as-images path instead of raising, with a warning that the two paths are not token-equivalent.

**3. Media failures were silent** (`mllm.py`, `multimodal_processor.py`)

- `extract_video_frames_smart()` returned whatever it managed to decode, including `[]` — frame seeking can silently skip frames on long-GOP codecs. Now `assert_video_decoded()` rejects empty/degenerate results and logs `Video decode OK: N frames, WxH` on success, so frame count and orientation are visible in the log.
- `MultimodalProcessor.process()` caught every media exception and continued with `logger.warning`. Now raises.
- `decode_base64_video()` built the temp-file suffix as `mime.split("/")[-1]`, giving `.quicktime` for a `.mov` and `.x-matroska` for a `.mkv`. Latent with OpenCV (ffmpeg content-sniffs; verified `.quicktime`/`.mov`/no-suffix all decode identically) but live for any decoder that infers the container from the extension. Replaced with a MIME->container map.

**4. The chat UI could display a chip for a file it had discarded** (`vllm_mlx/gradio_app.py`)

Unknown extensions defaulted to `image/jpeg`; resolution is now extension -> system MIME db -> raise, with `.mov`/`video/quicktime`, `.m4v`, `.mkv`, `.heic` added. Replies now carry a visible warning when attachments were sent but `prompt_tokens` is implausibly low, and startup checks `/health` for a text-only model in multimodal mode. Attachments are base64-encoded once per turn instead of twice.

**5. `enable_thinking` was dropped on the native video path** (`mllm.py`)

The native branch built its chat-template kwargs from `tools` alone. A server started with `--default-chat-template-kwargs '{"enable_thinking": false}'` therefore honored it for text and images and ignored it for video — the model reasoned to the full `max_tokens` and tripped the request timeout with a 504. The Qwen3.5 template renders `<think>\n\n</think>` when false and leaves `<think>` open when true, so the dropped kwarg was the whole difference. Now forwarded, with the same `TypeError` retry the frames path uses.

**6. The chat UI replayed all historical media every turn** (`gradio_app.py`, + tests)

Payload grew 22.3 MB -> 66.3 MB across three turns and the server re-decoded every prior video on each one. History is now text-only; media stays with the turn it belongs to.

## Verification

Same server, same clip, `mlx-community/Qwen3.8-27B-4bit`:

| | prompt_tokens | result |
|---|---|---|
| before | 68 | "I don't see any video" |
| after fix 1 only | — | HTTP 500, `ModuleNotFoundError` |
| after fixes 1+2 | 12,110 | correct description |

With thinking disabled, a video reply is **396 tokens in 60.8s** where it previously ran past the 300s timeout. Regression checks: `mlx-community/Qwen3-8B-4bit` still loads `MLLM=False` and answers normally; text and image requests unaffected; the frames-as-images fallback still decodes 16 frames at correct orientation.

Decode correctness was confirmed by writing frames to PNG and inspecting them — OpenCV 5.0.0 auto-rotates (`CAP_PROP_ORIENTATION_AUTO=1`), so the `rotation=-90` display matrix is applied and no orientation compensation was added.

## Known limitations, not addressed here

- **Multi-video requests use the first clip's sampled fps for all clips.** `mlx_vlm.prepare_inputs` takes a single scalar `fps`, and that value drives Qwen's interleaved timestamp tokens. Fixing it properly means calling the processor directly instead of routing through `mlx_vlm.generate`. Happy to do that in a follow-up if you'd prefer it here.
- **The shutdown segfault workaround has been removed from this PR** and now lives in `fix/shutdown-tls-segfault`, opened separately, with the cleanup-exception defect fixed and lifecycle tests added. It is a process-wide behaviour change and does not belong here.
- **The chat-UI document pipeline** (PDF/DOCX/XLSX attachments) has likewise been split out to `feat/document-attachments`; it is stacked on this branch and will be opened once this merges, to avoid a PR that contains this one's commits.

Each commit is self-contained and reviewable on its own.

---

## Update after review

Rebased on `22efb47`. This PR is now video-only: no `cli.py`, no `server.py`, no document
pipeline. 7 commits, 7 files.

Added `tests/test_video_pipeline.py`, wired into the existing no-MLX CI job. It runs on
`ubuntu-latest` without MLX by injecting a stand-in `mlx_vlm`, which suits what is being
guarded — the two failures behind this work were an upstream module rename and a silently
empty decode, and neither needs a GPU to catch. Verified by running the file with `mlx`
blocked at the import hook: 17 pass, 1 skips.

Coverage, against the list requested:

| asked for | test |
|---|---|
| pinned mlx-vlm API | `test_real_mlx_vlm_still_exposes_the_pinned_api` — asserts `load_video` still takes fps/max_frames/min_frames/frame_factor and `generate()` still accepts `video=`. Skips without MLX, so it needs an Apple Silicon runner to be meaningful. |
| decode failures | `test_zero_frames_raises_and_names_the_layer`, `test_implausible_frame_size_raises`, `test_malformed_frame_shape_raises`, `test_empty_decode_stops_the_request` |
| mixed image/video/audio | `test_mixed_image_video_audio_are_routed_separately` |
| multiple videos, distinct rates | `test_multiple_videos_share_the_first_clips_sampled_rate` |
| frames fallback | `test_missing_upstream_module_is_reported_not_swallowed` |

One of those pins a limitation rather than asserting correctness.
`prepare_inputs` takes a single scalar `fps` and fans it out to every video, so a second
clip's frames are labelled with the first clip's sampled rate — which drives Qwen's
interleaved timestamps. The test says so and asserts the second rate is discarded, so a
future fix changes a test deliberately instead of quietly altering timestamps. Fixing it
means calling the processor directly rather than going through `mlx_vlm.generate`; happy to
do that here if you would rather it not ship as a known gap.

Native-video model output still needs an Apple Silicon window and is not covered by these.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,6 +79,7 @@ jobs:
             tests/test_mllm_cache.py \
             tests/test_api_models.py \
             tests/test_api_utils.py \
+            tests/test_video_pipeline.py \
             tests/test_request.py \
             tests/test_anthropic_models.py \
             tests/test_anthropic_adapter.py \
--- a/tests/test_gradio_app.py
+++ b/tests/test_gradio_app.py
@@ -65,6 +65,89 @@ def test_create_chat_function_uses_configured_served_model_name(monkeypatch):
     assert captured["json"]["model"] == "my-served-model"
 
 
+def test_create_chat_function_does_not_replay_historical_media(monkeypatch):
+    captured_requests = []
+
+    def fake_post(url, json, timeout):
+        captured_requests.append(json)
+        return DummyResponse("ok")
+
+    monkeypatch.setattr(gradio_app.requests, "post", fake_post)
+    monkeypatch.setattr(
+        gradio_app,
+        "build_media_items",
+        lambda files: [
+            {
+                "type": "video_url",
+                "video_url": {"url": "data:video/quicktime;base64,AAAA"},
+            }
+        ],
+    )
+
+    chat_fn = gradio_app.create_chat_function(
+        server_url="http://localhost:8000",
+        max_tokens=128,
+        temperature=0.3,
+    )
+    chat_fn(
+        {"text": "Identify the instrument.", "files": ["clip.mov"]},
+        history=[],
+    )
+    chat_fn(
+        {"text": "What pattern is on the strap?", "files": []},
+        history=[
+            {"role": "user", "content": "Identify the instrument."},
+            {"role": "assistant", "content": "An electric guitar."},
+        ],
+    )
+
+    first_messages = captured_requests[0]["messages"]
+    assert first_messages[0]["content"][1]["type"] == "video_url"
+
+    follow_up_messages = captured_requests[1]["messages"]
+    assert all(isinstance(message["content"], str) for message in follow_up_messages)
+    assert gradio_app.OMITTED_MEDIA_NOTE in follow_up_messages[0]["content"]
+    assert "data:video/" not in str(follow_up_messages)
+
+
+def test_create_chat_function_clears_media_indexes_for_new_chat(monkeypatch):
+    captured_requests = []
+
+    def fake_post(url, json, timeout):
+        captured_requests.append(json)
+        return DummyResponse("ok")
+
+    monkeypatch.setattr(gradio_app.requests, "post", fake_post)
+    monkeypatch.setattr(
+        gradio_app,
+        "build_media_items",
+        lambda files: [
+            {
+                "type": "video_url",
+                "video_url": {"url": "data:video/quicktime;base64,AAAA"},
+            }
+        ],
+    )
+
+    chat_fn = gradio_app.create_chat_function(
+        server_url="http://localhost:8000",
+        max_tokens=128,
+        temperature=0.3,
+    )
+    chat_fn({"text": "Old chat", "files": ["clip.mov"]}, history=[])
+    chat_fn({"text": "New chat", "files": []}, history=[])
+    chat_fn(
+        {"text": "Follow up", "files": []},
+        history=[
+            {"role": "user", "content": "New chat"},
+            {"role": "assistant", "content": "ok"},
+        ],
+    )
+
+    new_chat_follow_up = captured_requests[2]["messages"]
+    assert gradio_app.OMITTED_MEDIA_NOTE not in str(new_chat_follow_up)
+
+
 def test_main_text_only_uses_configured_served_model_name(monkeypatch):
     captured = {}
 
--- a/tests/test_video_pipeline.py
+++ b/tests/test_video_pipeline.py
@@ -121,6 +121,81 @@ def test_name_patterns_remain_the_fallback():
     assert is_mllm_model("some-org/definitely-a-text-model") is False
 
 
+@pytest.fixture
+def no_network(monkeypatch):
+    """Any socket creation fails the test: these paths must stay offline."""
+    import socket
+
+    def _refuse(*args, **kwargs):
+        raise AssertionError("network access attempted in a no-network test")
+
+    monkeypatch.setattr(socket, "socket", _refuse)
+    monkeypatch.setattr(socket, "create_connection", _refuse)
+
+
+@pytest.fixture
+def hub_cache(tmp_path, monkeypatch):
+    """Redirect huggingface_hub's cache lookup to an isolated directory,
+    keeping the real cache-layout parsing in the loop."""
+    import functools
+
+    import huggingface_hub
+
+    real = huggingface_hub.try_to_load_from_cache
+    monkeypatch.setattr(
+        huggingface_hub,
+        "try_to_load_from_cache",
+        functools.partial(real, cache_dir=str(tmp_path)),
+    )
+    return tmp_path
+
+
+def _write_hub_cached_config(cache_root, repo_id: str, config: str) -> None:
+    """Lay out config.json the way huggingface_hub stores a cached repo."""
+    repo_dir = cache_root / f"models--{repo_id.replace('/', '--')}"
+    rev = "0" * 40
+    (repo_dir / "snapshots" / rev).mkdir(parents=True)
+    (repo_dir / "refs").mkdir()
+    (repo_dir / "refs" / "main").write_text(rev)
+    (repo_dir / "snapshots" / rev / "config.json").write_text(config)
+
+
+def test_cached_neutral_id_is_detected_from_hub_config(no_network, hub_cache):
+    """A neutral repo ID (no VL marker) already in the hub cache must be
+    classified by its own config.json - the case that silently dropped video
+    for Qwen3.8 - and without any network traffic."""
+    _write_hub_cached_config(
+        hub_cache,
+        "mlx-community/Qwen-Neutral-Name-4bit",
+        '{"architectures": ["Qwen3_5ForConditionalGeneration"],'
+        ' "vision_config": {}, "video_token_id": 248057}',
+    )
+    assert is_mllm_model("mlx-community/Qwen-Neutral-Name-4bit") is True
+
+
+def test_cached_text_only_id_stays_text_only(no_network, hub_cache):
+    """The cached config governs in both directions: a text-only config keeps
+    a model text-only even if a broad name pattern would have matched."""
+    _write_hub_cached_config(
+        hub_cache,
+        "mlx-community/SomePixtral-Like-Name",
+        '{"architectures": ["Qwen3ForCausalLM"]}',
+    )
+    assert is_mllm_model("mlx-community/SomePixtral-Like-Name") is False
+
+
+def test_cold_neutral_id_falls_back_to_name_patterns_offline(no_network, hub_cache):
+    """A neutral ID with nothing cached is decided by name patterns alone,
+    before any download. This pins the current pre-download behaviour: the
+    CLI's MLLM-versus-LLM routing decision for a cold neutral ID is made from
+    the name, with no network attempted, and misroutes a neutrally-named VLM
+    until its config is cached. Resolving hub metadata before that decision
+    is a known follow-up; this test exists so the limitation is deliberate.
+    """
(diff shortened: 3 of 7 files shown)


Thump604 (changes requested):
Blocking lifecycle issue in the segfault workaround:

At `vllm_mlx/server.py:1609-1618`, shutdown exceptions are stored in `cleanup_exc`, but the `finally` block calls `_exit_without_finalizing()` whenever the flag is enabled. The later `cleanup_exc` propagation at `server.py:1632-1633` is therefore unreachable. The PR enables this for both launch paths (`vllm_mlx/cli.py:493-494` and `vllm_mlx/server.py:6972-6979`), and `cli.py:50-87` performs best-effort cleanup before unconditional `os._exit(0)`. A cache-save, engine-stop, MCP, or model-manager shutdown failure can consequently be reported as a successful process exit, while Python/host finalization is bypassed. This is a process-wide behavior change, not video-specific, and the PR has no lifecycle tests or CI checks.

Please remove this workaround from the video PR or split it into a separately reviewable change with an exit design that preserves cleanup failures/exit status and proves the behavior for clean shutdown, cleanup exceptions, signal/exception shutdown, `VLLM_MLX_CLEAN_EXIT=0`, both launchers, and embedded/library use. Rebase on current `main` (`22efb47968f96e6e7a79704f4eb0184cda4e0068`) before re-review.

For the video portion, please add focused CI/unit coverage for the pinned `mlx-vlm` API (`load_video` plus `generate(video=...)`), decode failures, mixed image/video/audio inputs, multiple videos with distinct sampling rates, and the frames fallback. Native-video/model results still need an Ops-coordinated Apple-Silicon/MLX window; I did not run MLX or inference.

sbayer2:
Thanks — the blocking finding is correct, and it was the more serious of the two problems in that commit.

**On the cleanup path.** Confirmed by reading it back: `except BaseException as exc: cleanup_exc = exc` captured the failure, the `finally` block checked only `primary_exc`, and `os._exit(0)` ran regardless, making the `raise cleanup_exc` below it unreachable. A failed cache-save, engine-stop, MCP or model-manager shutdown reported success and exited 0. That is the same silent-failure shape this whole change set exists to remove, which makes introducing it in the fix for it worse rather than merely unlucky.

## What changed

**#733 is now video-only.** Rebased on `22efb47`. The shutdown workaround is gone from it, along with the `VLLM_MLX_CLEAN_EXIT` doc commit. 7 commits, 7 files, no `cli.py` and no `server.py`.

**#745 carries the shutdown change**, with the defect fixed: finalization is skipped only when *both* phases raised nothing. If either failed, the guard stands aside and the exception propagates with a non-zero exit. A teardown segfault on that path is the right trade — a process that already has an error to report should report it loudly rather than exit silently and clean. `tests/test_shutdown_exit.py` covers clean shutdown, cleanup exception, serving exception, `VLLM_MLX_CLEAN_EXIT=0`, both launchers, and library use where the flag stays off. The regression test discriminates: restoring the old guard fails it.

**Video coverage** is in `tests/test_video_pipeline.py`, wired into the existing no-MLX job in `ci.yml`, covering the pinned API, decode failures, mixed image/video/audio, multiple videos with distinct sampling rates, and the missing-module fallback. It runs on `ubuntu-latest` by injecting a stand-in `mlx_vlm` — appropriate here, since the two failures behind this work were an upstream module rename and a silently empty decode, and neither needs a GPU to catch. Verified by running it with `mlx` blocked at the import hook: 17 pass, 1 skips.

## Three things worth flagging rather than leaving you to find

**The lifecycle tests in #745 cannot run in the ubuntu job.** `vllm_mlx.server` and `vllm_mlx.cli` both import mlx transitively, so they skip there — the same reason `tests/test_lifecycle_server.py` is not in that list. They need an Apple Silicon runner to mean anything. If you would rather they ran in CI, that means extracting the guard's decision into a module that does not pull in the engines; happy to do it, but it is a design change and I did not want to make it unasked.

**One video test pins a limitation rather than asserting correctness.** `prepare_inputs` takes a single scalar `fps` and fans it out to every video, so a second clip's frames are labelled with the first clip's sampled rate, which drives Qwen's interleaved timestamps. `test_multiple_videos_share_the_first_clips_sampled_rate` documents that and asserts the second rate is discarded, so a future fix changes a test deliberately instead of quietly altering timestamps. Fixing it properly means calling the processor directly instead of going through `mlx_vlm.generate` — say the word if you would rather it not ship as a known gap.

**There is a third pile you have not seen yet.** #733 had also accumulated a chat-UI document pipeline — PDF, DOCX, XLSX attachments — which is no more video-specific than the shutdown change was. It is split out to `feat/document-attachments` on my fork. It is stacked on this branch, so opening it now would produce a PR containing this one's commits; I will open it once this merges, unless you would prefer it against `main` immediately.

**Still outstanding, and I have not fixed it:** the MLX bug behind #745 has no standalone reproducer, so nothing is filed upstream. `mx.compile` on a worker thread that exits does not reproduce it, nor does adding a per-thread GPU stream and a surviving array. I would rather say that than file a report that does not reproduce.

Native-video model results still need the Apple Silicon window you mentioned; nothing in these two PRs substitutes for it.

Thump604 (changes requested):
I rechecked #733 at `9fe1119` against main `22efb47`. The video-only split is correct, and the portable decode/routing/validation/API tests belong in the standard no-MLX job; that file is now listed there. The captured head has no GitHub checks, so the portable result still needs a retained CI run.

The multi-video test at `mllm.py:1804-1833` currently proves that only the first decoded FPS is passed to `generate`; it documents a real timestamp correctness limitation, not correct multi-video support. Please either narrow/fail-closed native requests with more than one video or implement per-video timestamp handling and add native coverage. Native model/video output remains a maintainer Apple Silicon gate; unavailable runner capacity is not contributor failure. Keeping PDF/DOCX/XLSX attachments out of this PR is right—open that as a separate PR against current main with its own MIME/size/cleanup/security tests rather than a stacked video branch.

The added hub-cache config lookup improves already-cached neutral HF IDs, but the CLI still makes the MLLM-versus-LLM download-pattern decision before downloading. Please add no-network tests for cached and cold neutral IDs (or resolve metadata before that decision); this is a scoped routing/test follow-up, separate from the native Apple gate.

sbayer2:
Addressed at 263b484. Multi-video native requests now fail closed before any decoding, with the fps-collapse reason and a workaround in the error; the pinned-limitation test became a rejection test (asserting no `load_video` call happens) plus a single-video-unaffected guard. For detection: added no-network tests — cached neutral IDs classified from a real hub-cache layout in both directions (VLM config wins; text-only config wins even against a broad name pattern), and a cold neutral ID falling back to name patterns with socket creation monkeypatched to fail the test if any network is attempted. The cold-ID test pins the pre-download routing decision as a deliberate, documented limitation; resolving hub metadata before that decision is noted as the follow-up. (Note: the detection tests share commit 6c7a895 with the fail-closed guard — same test file; the commit message emphasizes the guard.)

Agreed on documents as a separate PR against current main once this lands. The CI run for this head is queued pending workflow approval (`action_required`) — one approval click and the portable result is retained.

Thump604:
I approved and inspected the withheld exact-head run. The requested no-network routing and multi-video fail-closed work is behaving as intended: the Linux matrix is green, and I independently ran `tests/test_video_pipeline.py` at `263b4841` with 22/22 passing.

The only contributor-side blocker exposed by CI is mechanical: Black 26.5.1 would reformat `vllm_mlx/gradio_app.py`; Ruff is already clean. Please run the repository Black command (or format that file), push only the formatting delta, and let CI rerun. No additional source work, broad test campaign, or model run is requested from you. Native-video model output remains a separate maintainer Apple-Silicon gate.

sbayer2:
Black 26.5.1 formatting delta pushed as requested (be493a8), formatting only: one line in `vllm_mlx/gradio_app.py`. `black --check vllm_mlx/ tests/` passes on 26.5.1. Ready for a CI approval.

PR #734, opened 2026-08-26
test(simple-engine): key stream ownership by Thread, not ident

janhilgard:
Follow-up to 36deafe1, which fixed `test_restart_after_stop_mid_stream_uses_a_fresh_worker` for recycled thread idents. The same flaw is in `test_new_worker_never_binds_the_retired_workers_stream`, which I added in #679 — it keys stream ownership by `threading.get_ident()`.

That test deliberately retires a worker and starts another, so it runs squarely in the window where the OS can hand the new thread the dead one's ident. If that happens, a genuine cross-thread bind compares equal to same-thread and the guard silently stops catching the regression it exists for. The test would keep passing while the invariant was broken, which is the worst failure mode for a guard.

Keyed by `Thread` object now, same as your fix. Idents stay in the failure message, where they help.

Both directions verified:

- dropping `self._worker_generation_stream = None` on worker retirement still fails the test, so the original regression is still caught
- comparing two distinct threads that share a name *and* an ident: identity by ident reads them as the same thread, identity by object does not

Repo suite: 2909 passed, 1 failed — `test_extract_audio_from_video_integration_smoke`, which needs an ffmpeg this machine does not have. Lint and black clean.

### DIFF
--- a/tests/test_simple_engine_thread_pinning.py
+++ b/tests/test_simple_engine_thread_pinning.py
@@ -76,6 +76,11 @@ def _install_mlx_stubs() -> None:
 _install_mlx_stubs()
 
 
+def _thread_label(thread: threading.Thread | None) -> str:
+    """Readable identity for failure messages; None means never bound."""
+    return "—" if thread is None else f"{thread.name}/{thread.ident}"
+
+
 class _Chunk:
     def __init__(self, text: str, finish_reason: str | None = None) -> None:
         self.text = text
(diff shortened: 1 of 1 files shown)


Thump604 (commented):
I reviewed head `c33b9feeaa8e7863fc71c0cd0b79f99e06cea59` against parent/current-main seam `36deafe1a6c8213becdf682c16e014c97672a674`.

This is a targeted and correct hardening of the regression guard. `SimpleEngine._bind_generation_streams_once()` records the stream on the worker that owns it (current `vllm_mlx/engine/simple.py:755-768`), and `stop()` clears that worker stream before a restart (current `:946-952`). The test at `tests/test_simple_engine_thread_pinning.py:507-574` drives that real stop/restart path, but its previous `threading.get_ident()` map could treat a new executor thread as the old owner if the OS recycled the ident. Comparing `Thread` objects is the correct identity check and matches the ownership invariant; retaining name/ident only for diagnostics is appropriate.

I found no production-code blocker. The test is included in the current unit CI file list, and the ten reported checks on this head are green. The ident-reuse condition is not deterministically forced by the test, but that is a coverage-strength follow-up rather than a reason to reject this direct fix: when reuse occurs, this version continues to detect the cross-thread bind. Maintainer decision/merge can proceed after the normal repository checks; this is a COMMENT review, not an approval or merge.

waybarrios:
Carefully checked and all set!

Issue #736, opened 2026-08-27
Scheduler monkey-patch layer targets pre-0.31 mlx-lm internals: chunked prefill and prompt-cache-save are inert, `--enable-mtp` crashes


azamamirza:
## Summary

The scheduler carries a monkey-patch layer over mlx-lm's `BatchGenerator`
(`_install_chunked_prefill`, `_install_prompt_cache_save`, `_install_mtp`)
that targets the pre-0.31 mlx-lm API (`_process_prompts`, `active_batch`,
`_step`, `unprocessed_prompts`, `_next` with `_partial` state). mlx-lm
0.31.x refactored that surface: `BatchGenerator` now exposes
`insert` / `remove` / `next`, prefill lives in `PromptProcessingBatch`, and
decode lives in `GenerationBatch._step`. Against the currently pinned
`mlx-lm>=0.31.3` this has three consequences:

1. **Chunked prefill is silently inert.** The compatibility check
   (`hasattr(bg, "_process_prompts") and hasattr(bg, "active_batch")`)
   fails on 0.31.3, so `--chunked-prefill-tokens` logs a warning and does
   nothing. Long prompts are prefilled whole by mlx-lm's own
   `prefill_step_size` chunking instead of the scheduler's budget.
2. **Prompt-cache-save is silently inert.** The `_process_prompts` hook is
   skipped by the same guard, so prompt-only cache entries are never
   captured; only the finished-request (prompt + output) store in
   `_cleanup_finished` runs. Mid-prefill saves are likewise dead code.
3. **`--enable-mtp` crashes.** `_install_mtp` is not guarded: it reads
   `batch_gen._step`, which does not exist on 0.31.3, so enabling MTP with
   an MTP-capable model raises `AttributeError` at BatchGenerator
   creation — i.e. on the first scheduled request.

## Reproduction (crash path)

```python
from mlx_lm.generate import BatchGenerator
hasattr(BatchGenerator, "_step")             # False on mlx-lm 0.31.3+
hasattr(BatchGenerator, "_process_prompts")  # False
hasattr(BatchGenerator, "active_batch")      # False
```

Any `vllm-mlx serve <mtp-capable model> --enable-mtp` hits
`_orig_step = batch_gen._step` -> `AttributeError`.

## Companion PR

A minimal fix for the crash path only: `_install_mtp` checks for the
`_step` hook and warns + no-ops when absent, mirroring the existing
chunked-prefill compatibility guard. Serving continues without MTP
[...]

PR #737, opened 2026-08-27
fix(scheduler): guard MTP install against mlx-lm >= 0.31 BatchGenerator

azamamirza:
## Problem

`_install_mtp` wraps `BatchGenerator._step`, which was refactored away in
mlx-lm 0.31.x (decode now lives on `GenerationBatch._step`). Against the
pinned `mlx-lm>=0.31.3`, running `vllm-mlx serve <mtp-capable model>
--enable-mtp` raises `AttributeError: 'BatchGenerator' object has no
attribute '_step'` at BatchGenerator creation — i.e. on the first
scheduled request.

The sibling monkey-patches (`_install_chunked_prefill`,
`_install_prompt_cache_save`) already have `hasattr` compatibility guards
and skip with a warning on 0.31.x; the MTP install is the only unguarded
one. See #736 for the full diagnosis of the patch layer vs the 0.31.x API
(chunked prefill and prompt-cache-save are currently silently inert) and
suggested porting seams.

## Change

- `_install_mtp` checks for the `_step` hook and warns + no-ops when it is
  absent, mirroring the chunked-prefill compatibility guard. Generation
  continues without multi-token prediction instead of failing the request
  loop.
- `tests/test_mtp_mlxlm_compat.py`: the guard no-ops on a 0.31-style
  BatchGenerator (no exception, nothing patched, warning logged), plus a
  tripwire test asserting the installed mlx-lm's BatchGenerator has no
  `_step` — if that ever fails, the API came back and the guard/port
  should be revisited.

This deliberately does NOT port the MTP hook to the 0.31.x API: that means
re-pointing the wrap at `GenerationBatch._step` (a per-batch object, so the
install site moves) and needs hardware verification with a real MTP model.
Scoped out to keep this a safe crash fix; follow-up directions are in #736.

## Verification

- `ruff check` and `black --check` clean on changed files.
- New regression tests pass; the guard no-ops cleanly and the tripwire
  confirms `BatchGenerator` has no `_step` on current mlx-lm.

Closes the crash path described in #736.

### DIFF
--- a/tests/test_mtp_mlxlm_compat.py
+++ b/tests/test_mtp_mlxlm_compat.py
@@ -1,0 +1,45 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Regression tests: --enable-mtp must not crash on mlx-lm >= 0.31.
+
+The MTP monkey-patch wraps ``BatchGenerator._step``, which was refactored
+away in mlx-lm 0.31.x (decode moved to ``GenerationBatch._step``).  On an
+incompatible BatchGenerator the install must warn and no-op instead of
+raising ``AttributeError`` at generator creation.
+"""
+
+import logging
+
+from mlx_lm.generate import BatchGenerator
+
+from vllm_mlx.scheduler import _install_mtp
+
+
+class _ModernBatchGenerator:
+    """Mimics the mlx-lm >= 0.31 BatchGenerator surface (no ``_step``)."""
+
+    def insert(self, *args, **kwargs):
+        raise NotImplementedError
+
+    def remove(self, *args, **kwargs):
+        raise NotImplementedError
+
+    def next(self):
+        raise NotImplementedError
+
+
+def test_install_mtp_skips_without_step_hook(caplog):
+    bg = _ModernBatchGenerator()
+
+    with caplog.at_level(logging.WARNING, logger="vllm_mlx.scheduler"):
+        _install_mtp(bg, model=None)
+
+    assert not hasattr(bg, "_step")
+    assert bg.__dict__.get("_next") is None
+    assert any("[MTP] disabled" in rec.message for rec in caplog.records)
+
+
+def test_installed_mlx_lm_batch_generator_lacks_step_hook():
+    # Documents the incompatibility this guard exists for: if this ever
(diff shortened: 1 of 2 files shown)


Thump604 (commented):
COMMENT

I checked the current base `main` (`22efb47968f96e6e7a79704f4eb0184cda4e0068`) against head `209cb64f7406e95fb967c6a7ead33342a4eb6057`. The crash diagnosis is accurate: current `scheduler.py:794-813` dereferences `batch_gen._step` unconditionally, while the native mlx-lm compatibility path at `:736-791` identifies the newer surface through `_generation_batch` and `_next`, without that hook. The new `hasattr` guard is a proportionate fail-closed fix and leaves generation running without MTP.

Before maintainer decision, please add one integration-level test around `Scheduler._create_batch_generator` (or the equivalent real construction path) with `enable_mtp=True` and a 0.31-style generator. It should assert that construction and request scheduling continue, MTP is reported disabled/absent, and the warning is emitted. The current synthetic `_install_mtp` test proves the guard itself but not the scheduler call path or the resulting operator-visible state.

This PR should remain scoped to the crash guard. Porting the hook to `GenerationBatch._step` is a separate change and needs an Ops-coordinated Apple-Silicon/MLX window with the exact mlx-lm version, model artifact, upstream tree, and rollback/qualification record. I did not import MLX or run inference here. This comment is not approval, adoption, or production qualification.

azamamirza:
Added the requested integration-level test. It builds a real Scheduler with enable_mtp=True and calls _create_batch_generator, so the guard is exercised through the actual construction path against the installed mlx-lm BatchGenerator (which lacks _step). 

The model stub deliberately has an MTP head, so it's the guard, not the missing-head check, that disables. 

Asserts: construction survives, no MTP surface is attached (get_mtp_stats absent), the [MTP] disabled warning is emitted, and request scheduling continues afterward. 

Full suite locally on Apple Silicon: 2957 passed / 24 skipped. 

Agreed on scope: the GenerationBatch._step port stays a follow-up tracked in #736.

waybarrios:
Thanks for this work. I reviewed it addressed what you mentioned. Ready to ship into the main

PR #740, opened 2026-08-27
Fix streaming token loss at --stream-interval>1, DSA CacheList SSD spill, bf16 upcast on mlx 0.31.x, and forced-<think> reasoning

anisoptera:
Five bug fixes plus one test, found while bringing up GLM-5.3-Flash (8-bit MLX) on a 512 GB M3 Ultra. Each is independent — **happy to split this into separate PRs if you'd rather review them piecemeal, in whatever grouping suits you.**

# Human note
While the port is fresh, I've been dragging some form or other of these fixes along with me since Qwen-397b. Opus totally could have made some mistakes pulling them in from my previous branch, but it was pulling from code that was already tested across millions of tokens.

## The fixes

**1. Streaming dropped 4 of every 5 tokens when `--stream-interval > 1`** (`160e587`)

The engine forwarded only every Nth per-step `RequestOutput` to the collector and dropped the ones in between. Each step's output carries ~1 token of `new_text`, so most of the text was silently lost — streamed output came out as scrambled fragments while `stream=False` on the same prompt was always correct.

The fix gates *notification* rather than the *put*: the collector already aggregates `new_text` on `put()`, so `put()` gains `notify=False` to accumulate without releasing, and `get()`/`get_nowait()` wait on the ready event. `engine_core` now always puts each step and only notifies on the interval boundary or at finish.

**2. DSA `CacheList` entries dropped or corrupted on SSD spill** (`65d2053`)

A DSA layer is `CacheList(KVCache latent, KVCache indexer)`. `get_serializer_for_layer` dispatched it to `ArraysCacheSerializer` because `CacheList` also exposes a list `.state` — but iterating that state yields per-sub `(keys, values)` tuples, which either raised inside the numpy conversion or silently flattened the sub-caches and lost the structure and offsets.

Adds a dedicated `CacheListSerializer` that walks `layer.caches` directly rather than relying on `.state` semantics, and moves the `.caches` check ahead of the `.state` branch in dispatch. A quantized sub-cache now raises a clear error so the entry drops cleanly instead of being written corrupted.

**3. bf16 upcast unreachable on mlx 0.31.x — wrong exception class** (`ff1390f`)

`_mx_to_numpy_safe`'s comment already says "bf16 buffer-protocol mismatch on mlx >= 0.31", but the handler catches only `RuntimeError`. mlx 0.31.x raises `ValueError`:

```
ValueError: 'bfloat16' is not a valid PEP 3118 buffer format string
```

So the fp32 fallback was dead code and **every bf16 KV spill to the SSD tier failed** on 0.31.x. Now matched on the message rather than the exception class, which isn't stable across mlx versions, and anything else re-raises.

This hid because the existing bf16 tests raise the error by hand and therefore pass under either class — so this also adds a test that drives a real `mx.bfloat16` array through the function.

**4. Forced-`<think>` chain-of-thought streamed as `content`** (`e67ace5`)

GLM-5.2/5.3 chat templates end the rendered prompt with an open `<think>`, so generation starts *inside* the reasoning block and the model only ever emits the closing tag. `Glm4ReasoningParser.extract_reasoning_streaming` overrides the base `pre_think` branch to default untagged output to `content` — correct for autonomous GLM-4.6-style thinking, wrong here. Streaming and non-streaming disagreed on identical output:

```
streaming:      content='84 * 3 = 252\n252 / 2 = 126**126**'   reasoning_content=''
non-streaming:  content='**126**'                              reasoning_content='84 * 3 / 2 = ...'
```

Adds `_detect_implicit_thinking()`, which renders the chat template once per `(model, chat_template_kwargs)` and checks for a trailing `<think>`, wired into `reset_state(implicit_mode=...)` at `_prepare_streaming_reasoning_parser` — the single chokepoint all streaming reasoning sites already share since #654.

Worth noting this **changes no default**: in implicit mode glm4 simply stops overriding and defers to the base state machine, whose no-tags branch already emits reasoning. It stops a subclass from shadowing that. qwen3, deepseek and gemma4 are untouched.

**5. Bound the implicit-thinking probe cache** (`ea31a7c`)

Follow-up to #4. The cache key includes `chat_template_kwargs`, which comes straight off the request, so a varied client could mint unbounded entries. Bounded FIFO at 256, matching the existing `_responses_store` treatment. Eviction is insertion-order rather than LRU — worst case is a redundant template render, not a wrong answer.

**6. Test only: pin the MLX producer-thread materialization invariant** (`91b3b61`)

No production change. MLX streams are thread-local, so a lazy array built on the request thread aborts the process (`std::terminate`, not a catchable exception) if the spill path evaluates it on another thread through numpy's buffer protocol. `_detach_cache_for_storage` already establishes the invariant by batching every leaf into one `mx.eval` on the calling thread; these are black-box regressions over `store()` covering the two shapes with no `KVCache` trim path to hide behind. Both pass on `main` as-is — they just guard against a future refactor silently dropping the batched eval. Drop this commit if you'd rather not carry it.

## Testing

Full suite on the target machine (mlx 0.31.x / mlx_lm 0.31.3 / mlx-vlm 0.6.17):

```
2974 passed, 2 failed, 24 skipped
```

Both failures (`test_mllm_mtp_routing.py::test_mllm_scheduler_exposes_mtp_attempts_and_accepts_on_outputs`, `test_prefix_cache_scheduler_parity.py::test_identical_plain_kv_prompt_never_replays_a_supersequence`) reproduce on clean `22efb47` with none of these commits applied, so they're pre-existing and unrelated. I diffed the sorted failure sets before and after: zero new failures.

Also verified live against a running GLM-5.3-Flash-MLX-8bit server (`--mllm --enable-prefix-cache --continuous-batching --stream-interval 5 --chunked-prefill-tokens 256 --ssd-cache-dir`):

- streaming reasoning/content split correctly, no tag leakage, all chunks intact
- streaming and non-streaming now agree on the same prompt
- tool calling (`--tool-call-parser glm47`) still separates reasoning correctly
- vision requests return correct content and spatial ordering
- prefix cache round-trips through disk and reloads on restart

## Known limitation, not addressed here

On responses truncated mid-reasoning (`finish_reason=length`, no `</think>` ever emitted), streaming and non-streaming still disagree — streaming puts the text in `reasoning_content`, non-streaming in `content`.

**This is pre-existing on `main` and not introduced by these commits**: qwen3 shows the identical split today, in both implicit and autonomous mode, because `BaseThinkingReasoningParser.extract_reasoning()`'s no-tags branch returns `(None, model_output)` while the streaming state machine's no-tags branch defaults to reasoning. Commit #4 makes glm4-in-implicit-mode behave the way qwen3 already behaves.

I left it alone because the fix isn't local: `extract_reasoning()` takes only the output text, the non-streaming path uses the module-level shared `_reasoning_parser`, and implicit-ness is per-request — so it can't be stashed on the shared instance without a race. It wants either an explicit parameter or a per-request parser instance, which felt like a separate change from these. Happy to take it on separately if you'd like it fixed.

---

Written with Claude Opus 5 and tested on my own hardware (M3 Ultra, 512 GB) against a real GLM-5.3-Flash-MLX-8bit deployment.

### DIFF
--- a/tests/test_memory_cache_stream_safety.py
+++ b/tests/test_memory_cache_stream_safety.py
@@ -64,8 +64,7 @@ class TestStoreMaterializesOnProducerThread:
 
     def test_cache_list_survives_numpy_on_another_thread(self):
         """GLM-5.2 DSA shape: CacheList layers, no KVCache trim path at all."""
-        result = _run_subprocess(
-            """
+        result = _run_subprocess("""
             from mlx_lm.models.cache import CacheList
 
             cache_obj = MemoryAwarePrefixCache(
@@ -88,8 +87,7 @@ class TestStoreMaterializesOnProducerThread:
                     assert np.array(inner.keys).sum() > 0
                     assert np.array(inner.values).sum() > 0
             print("SURVIVED")
-            """
-        )
+            """)
         assert result.returncode == 0, (
             f"exit={result.returncode} (134 == SIGABRT)\n"
             f"stdout={result.stdout}\nstderr={result.stderr}"
@@ -98,8 +96,7 @@ class TestStoreMaterializesOnProducerThread:
 
     def test_quantized_cache_survives_numpy_on_another_thread(self):
         """`_quantize_cache` builds fresh lazy arrays AFTER the trim step."""
-        result = _run_subprocess(
-            """
+        result = _run_subprocess("""
             cache_obj = MemoryAwarePrefixCache(
                 model=object(),
                 config=MemoryCacheConfig(
@@ -123,8 +120,7 @@ class TestStoreMaterializesOnProducerThread:
                 for part in (*layer.keys, *layer.values):
                     np.array(part)
             print("SURVIVED")
-            """
-        )
+            """)
         assert result.returncode == 0, (
             f"exit={result.returncode} (134 == SIGABRT)\n"
             f"stdout={result.stdout}\nstderr={result.stderr}"
--- a/tests/test_memory_cache_stream_safety.py
+++ b/tests/test_memory_cache_stream_safety.py
@@ -1,0 +1,132 @@
+# SPDX-License-Identifier: Apache-2.0
+"""MLX thread-locality regression tests for ``MemoryAwarePrefixCache.store``.
+
+MLX streams are owned by the thread that created them, and every thread gets
+its *own* default GPU stream.  A **lazy** array built on thread A aborts the
+process when it is evaluated on thread B::
+
+    libc++abi: terminating due to uncaught exception of type std::runtime_error:
+    There is no Stream(gpu, 1) in current thread.
+
+Through ``mx.eval`` that surfaces as a catchable ``RuntimeError``; through
+numpy's buffer protocol (``np.array(arr)``, which the SSD spill path uses) the
+C++ exception unwinds with no handler and hits ``std::terminate`` -> SIGABRT.
+
+``_evict_lru`` structurally spills the *least-recently-used* entry, i.e. one
+built by an earlier request on a different thread, so any cache left lazy at
+``store()`` time is a latent uncatchable abort.  These tests pin the invariant
+that ``store()`` materializes on the producing thread.
+
+Upstream satisfies this invariant in ``_detach_cache_for_storage``, which
+accumulates every leaf into one ``eval_targets`` list -- recursing into
+``CacheList`` children -- and issues a single ``mx.eval`` on the calling
+thread (``memory_cache.py``, the ``is_root`` branch).  These tests are
+black-box on purpose: they pin the *invariant*, not that mechanism.
+
+The cross-thread cases run in a subprocess because the failure mode is
+``abort()``, which would take the whole pytest run with it.
+"""
+
+import subprocess
+import sys
+import textwrap
+
+PRELUDE = """
+import threading
+import numpy as np
+import mlx.core as mx
+from mlx_lm.models.cache import KVCache
+from vllm_mlx.memory_cache import MemoryAwarePrefixCache, MemoryCacheConfig
+
+mx.eval(mx.ones((2, 2)))  # bind the main thread's gpu stream
+
+def make_kv(lazy_scale):
+    layer = KVCache()
+    # `* lazy_scale` keeps the graph UNEVALUATED, exactly like a real decode
+    layer.keys = mx.ones((1, 2, 8, 64), dtype=mx.float32) * lazy_scale
+    layer.values = mx.ones((1, 2, 8, 64), dtype=mx.float32) * lazy_scale
+    layer.offset = 8
+    return layer
+"""
+
+
+def _run_subprocess(body: str) -> subprocess.CompletedProcess:
+    return subprocess.run(
+        [sys.executable, "-c", PRELUDE + textwrap.dedent(body)],
+        capture_output=True,
+        text=True,
+        timeout=180,
+    )
+
+
+class TestStoreMaterializesOnProducerThread:
+    """``store()`` must leave nothing lazy for a later thread to evaluate."""
+
+    def test_cache_list_survives_numpy_on_another_thread(self):
+        """GLM-5.2 DSA shape: CacheList layers, no KVCache trim path at all."""
+        result = _run_subprocess(
+            """
+            from mlx_lm.models.cache import CacheList
+
+            cache_obj = MemoryAwarePrefixCache(
+                model=object(), config=MemoryCacheConfig(min_prefix_tokens=1)
+            )
+            tokens = list(range(16))
+
+            def producer():
+                layers = [CacheList(make_kv(3.0), make_kv(5.0)) for _ in range(2)]
+                assert cache_obj.store(tokens, layers)
+
+            t = threading.Thread(target=producer)
+            t.start()
+            t.join()
+
+            # The spill path: numpy buffer protocol, on a DIFFERENT thread.
+            entry = cache_obj._entries[tuple(tokens)]
+            for layer in entry.cache:
+                for inner in layer.caches:
+                    assert np.array(inner.keys).sum() > 0
+                    assert np.array(inner.values).sum() > 0
+            print("SURVIVED")
+            """
+        )
+        assert result.returncode == 0, (
+            f"exit={result.returncode} (134 == SIGABRT)\n"
+            f"stdout={result.stdout}\nstderr={result.stderr}"
+        )
+        assert "SURVIVED" in result.stdout
+
+    def test_quantized_cache_survives_numpy_on_another_thread(self):
+        """`_quantize_cache` builds fresh lazy arrays AFTER the trim step."""
+        result = _run_subprocess(
+            """
+            cache_obj = MemoryAwarePrefixCache(
+                model=object(),
+                config=MemoryCacheConfig(
+                    min_prefix_tokens=1,
+                    kv_quantize=True,
+                    kv_min_quantize_tokens=1,
+                    kv_group_size=64,
+                ),
+            )
+            tokens = list(range(16))
+
+            def producer():
+                assert cache_obj.store(tokens, [make_kv(3.0), make_kv(5.0)])
+
+            t = threading.Thread(target=producer)
+            t.start()
+            t.join()
+
+            entry = cache_obj._entries[tuple(tokens)]
+            for layer in entry.cache:
+                for part in (*layer.keys, *layer.values):
+                    np.array(part)
+            print("SURVIVED")
+            """
+        )
+        assert result.returncode == 0, (
+            f"exit={result.returncode} (134 == SIGABRT)\n"
+            f"stdout={result.stdout}\nstderr={result.stderr}"
+        )
+        assert "SURVIVED" in result.stdout
--- a/tests/test_reasoning_parser.py
+++ b/tests/test_reasoning_parser.py
@@ -1643,3 +1643,55 @@ class TestStreamingCostIsFlat:
             f"{small:.2f}ms at 250 tokens, {large:.2f}ms at 5000 — "
             "per-token cost is growing with output length"
         )
+
+
+class TestForcedThinkingStreaming:
+    """GLM-5.3's template injects an open <think>, so the model output begins
+    INSIDE the reasoning block and only ever emits the closing </think>.
+
+    The glm4 streaming parser assumed GLM-4.6-style autonomous thinking and
+    defaulted pre-</think> output to content, so the whole chain-of-thought
+    landed in ``content`` glued to the answer. Live capture on GLM-5.3:
+
+        streaming:     content='84 * 3 = 252\\n252 / 2 = 126**126**'
+                       reasoning_content=''
+        non-streaming: content='**126**'
+                       reasoning_content='84 * 3 / 2 = 252 / 2 = 126'
+    """
+
+    DELTAS = ["84 * 3 = 252\n", "252 / 2 = 126", "</think>", "**126**"]
+
+    def _run(self, parser):
+        reasoning, content = [], []
+        prev = ""
+        for d in self.DELTAS:
+            cur = prev + d
+            msg = parser.extract_reasoning_streaming(prev, cur, d)
+            if msg is not None:
+                if getattr(msg, "reasoning", None):
+                    reasoning.append(msg.reasoning)
+                if getattr(msg, "content", None):
+                    content.append(msg.content)
+            prev = cur
+        return "".join(reasoning), "".join(content)
+
+    def test_implicit_mode_routes_pre_close_text_to_reasoning(self):
+        from vllm_mlx.reasoning import get_parser
+
+        parser = get_parser("glm4")(None)
+        parser.reset_state(implicit_mode=True)
+        reasoning, content = self._run(parser)
+
+        assert reasoning == "84 * 3 = 252\n252 / 2 = 126"
+        assert content == "**126**"
+
+    def test_autonomous_mode_still_defaults_to_content(self):
+        """Without a forced <think>, GLM-4.6-style output is plain content."""
+        from vllm_mlx.reasoning import get_parser
+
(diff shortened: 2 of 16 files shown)


anisoptera:
Pushed one more commit (`c81d254`) after a review pass over this branch. No changes to the five fixes already described above — this only tightens two edges and corrects a comment.

**1. `_detect_implicit_thinking` now probes with the resolved `enable_thinking`.**
The probe rendered the template without it, so `batched._apply_chat_template` fell back to its `"coder" not in model_name` default, while the real prompt render uses `chat_kwargs["enable_thinking"]` — which `_apply_forced_tool_choice` sets to `False` directly on `chat_kwargs`, not inside `chat_template_kwargs`. The two could disagree, and the cached verdict didn't distinguish them. The flag is now forwarded when the request resolved one (absent still means "let the engine decide") and is part of the cache key.

This is not reachable on GLM-5.3 — its chat template has no `enable_thinking` variable and emits `<|assistant|>{{- '<think>' -}}` unconditionally, so both renders agree. I deliberately did **not** also make `_thinking_disabled` treat a top-level `enable_thinking=False` as thinking-off: on GLM-5.3 that would disable the reasoning parser while the template still injects `<think>`, leaking `</think>` into content. There's a regression test pinning that.

**2. `SSDIndex._SCHEMA_VERSION` 1 → 2, and a mismatch now drops every entry.**
The version row was written but never compared. This matters because of the `CacheList` fix in this same PR: before it, a DSA layer was mis-dispatched to `ArraysCacheSerializer`. I'd assumed that always failed, but it's dtype-dependent — confirmed on mlx_lm 0.31.3 that `CacheList.state` is a list of `(keys, values)` tuples, so bf16 raises the PEP 3118 error and the entry is dropped, but **fp16 stacks cleanly and the spill is written**, tagged `layer_type: "ArraysCache"` with the per-sub offsets lost. Since `_entry_hash` is the token hash alone and the cache dir isn't build-namespaced, such an entry still matches after upgrading and reconstructs into an `ArraysCache` the DSA model can't use. `SSDCacheTier` also clears the data directory on migration rather than leaving it to `reconcile()`.

**3. Softened the comment in `glm4_parser`** — it claimed streaming now agrees with `extract_reasoning()`, which contradicts the truncation limitation documented in the PR description above.

Full suite on the target stack (mlx_lm 0.31.3 / mlx 0.32.2): **2982 passed, 2 failed** — the same two failures reproduce on `22efb47`, and the local failure set is byte-identical to the base commit's.

Thump604 (dismissed):
REQUEST CHANGES

I reviewed head `c81d254e1cc077ccdda4f51a16f543a6c672750c` against `main` `22efb47968f96e6e7a79704f4eb0184cda4e0068`. There is one correctness blocker in the implicit-thinking detection:

1. `vllm_mlx/server.py:1178-1235` calls `engine._apply_chat_template()` without `tools`, but `_prepare_chat_completion_invocation()` passes request-specific tools in `:829-834`. The cache key contains only model name, `chat_template_kwargs`, and `enable_thinking`; it does not contain tool presence/content or another template identity. A template whose `<think>` opening depends on tools can therefore cache the no-tool verdict and reuse it for a tool request, misrouting reasoning as content (or content as reasoning). The docstring currently acknowledges that this probe misreads tool-conditional templates, so this is a known correctness limitation in the generic path, not just a missing test.

Please either (a) pass the canonical tool context to the probe and include a stable tool/template signature in the key, or (b) scope this detector to templates known to open `<think>` unconditionally and make that restriction explicit. Add a regression covering no-tools, tools, forced-tool choice, and changed template/kwargs contexts.

The head also combines streaming delivery, SSD CacheList serialization and destructive schema migration, MLX bf16 conversion, reasoning parsing, and cache bounding across 16 files, while GitHub reports no checks. Please split these independent fixes or add focused CI coverage before asking for a maintainer decision. At minimum, the focused tests need to cover consumer lag plus finish delivery, CacheList empty/unsupported sub-cache round trips and migration cleanup, and the parser/tool/no-tool matrix above.

The live GLM/SSD evidence in the description is useful, but I did not run MLX or inference here. Any live confirmation should use an Ops-coordinated Apple-Silicon window with the exact model, dependency versions, and upstream commit recorded. This review is not adoption or production qualification.

anisoptera:
Thanks for the careful read — `78644e7` is pushed and addresses everything. Summary below, with line citations so we don't re-tread the parts that were already covered.

### Blocker: the probe rendered without `tools`

Fixed. `_detect_implicit_thinking` now forwards `chat_kwargs["tools"]` into the probe render and keys the cache on them (`vllm_mlx/server.py:1178`, `_tool_signature`).

The key carries the **sorted tool names**, not the full schemas. That separates "tools" from "no tools" and one tool set from another — which is what a template can plausibly branch on — without minting a fresh cache entry, and a fresh template render, for every argument schema a client happens to send. Tool order is normalized so two requests offering the same tools in a different order share one entry. The residual limitation (a template that opened `<think>` based on a tool's *parameters* rather than its name) is documented in the docstring rather than left implicit.

Verified against the real GLM-5.3 template, not just mocks — both paths agree:

```
no_tools -> '<|user|>hi<|assistant|><think>'  implicit=True
tools    -> '<|user|>hi<|assistant|><think>'  implicit=True
```

That mattered: the probe body is wrapped in `except Exception: return False`, so a template that raised on tool rendering would have silently flipped the verdict to False for exactly the tool-call requests this PR exists to fix.

### Already covered — line citations

Three of the items you flagged already had tests in the branch:

| Item | Test |
|---|---|
| CacheList spill round-trip incl. empty subs | `tests/test_ssd_cache.py:1271` `TestCacheListSpillRoundTrip` |
| Stale data dirs removed on tier migration | `tests/test_ssd_cache.py:1494` |
| Cache bounded against client-supplied template kwargs | `tests/test_server.py:5342` |

### New focused coverage (19 tests, written red first)

| Gap | Tests |
|---|---|
| Tools matrix on the implicit-thinking probe | `tests/test_server.py:5419` — 8 tests, incl. your exact scenario: a tool-conditional template is not served the no-tool verdict |
| Lagging consumer / finish delivery under `--stream-interval` | `tests/test_stream_interval.py:80` — 6 tests via a `_drive_lagging` harness; no text lost, finish always delivered, survives a merge with earlier releases |
| Unpopulated / unsupported sub-caches | `tests/test_ssd_cache.py:1520` `TestUnsupportedSubCaches` — 5 tests |

That last one caught a real latent bug. A `KVCache` leaves `keys`/`values` as `None` until its first `update_and_fetch`; `np.array(None)` is a 0-d **object** array whose `.size` is 1, so it slipped past `serialize_layer`'s empty-array branch and only failed inside `save_file` with `dtype object is not covered` — on the writer thread, where the error was swallowed and the `.tmp` dir orphaned. The guard now lives in `KVCacheSerializer.snapshot_layer`, so one check covers both CacheList subs and plain layers. A regression test pins that a size-0 *materialized* array is still accepted — the dense-mode DSA indexer is empty but real and must round-trip.

### On CI

`check-runs` returns 0 for this PR because fork PRs need maintainer approval to run workflows, not because the branch is untested. Locally against the CI flag set: `ruff check` clean, `black --check` clean tree-wide. Full suite on an M-series box with the real model stack: **3001 passed, 2 failed, 24 skipped** — both failures pre-exist on the base commit and are unrelated to this branch (`test_mllm_mtp_routing.py::test_mllm_scheduler_exposes_mtp_attempts_and_accepts_on_outputs`, `test_prefix_cache_scheduler_parity.py::test_identical_plain_kv_prompt_never_replays_a_supersequence`). I also confirmed the local failure set is byte-identical to a clean base-commit worktree.

One thing worth flagging that the review didn't name: the branch had picked up 7 `black` violations while the base commit is clean on every file touched here. `black --check` would have failed had the workflow run. Fixed in `78644e7`.

### Splitting

Your ask was "split these independent fixes **or** add focused CI coverage" — the coverage now exists, so I've treated that as a genuine choice and I'm happy either way. If you'd still prefer separate PRs, these are clean seams:

1. `160e587` — streaming / stream-interval
2. `65d2053` + `ff1390f` — SSD cache + CacheList serialization
3. `e67ace5` + `ea31a7c` + `78644e7` — reasoning parser / implicit-thinking probe

Say the word and I'll open them in that order; each stands alone against `main`.

Thump604:
I checked the current PR head `c81d254` against `main`. The `enable_thinking` and SSD schema changes are present, but the main tool-context blocker is not resolved on that head. `_detect_implicit_thinking()` still renders without `tools`, and its cache key still has no tool signature. The current tests cover the enable-thinking flag, not a tool-conditional template or tool/no-tool cache separation.

Your comment refers to `78644e7`, but that commit is not the head currently attached to this PR. Please update the PR branch so the claimed implementation and tests are actually present. Once that is done, I can review the tool-context change and test matrix on the correct diff.

I also found that the stated 19-test count does not match the current PR diff, which appears consistent with the same head mismatch.

I understand GitHub CI has not run because workflows for this fork PR require maintainer approval. The reported local suite is useful supporting information, but I cannot independently verify it until the claimed changes are present on the PR branch and either maintainer-approved CI runs or equivalent artifacts are available.

anisoptera:
You were right, and sorry for the runaround — my previous comment described `78644e7` as pushed when it wasn't on this branch. I have two forks of this repo and pushed to the wrong one; both happened to have a branch of the same name, so the push reported success and I didn't check it against the PR's actual head.

It's on the branch now — head is `78644e7`, 8 commits. Everything described in the previous comment (the tool-aware probe, the 19 new tests, the black fixes) is what's there; the substance of that comment stands, it just wasn't visible to you yet. The line citations in it should now resolve.

(And sorry from the actual human driving the clanker here - I should have checked that asking it to push actually resulted in a real push to the PR, etc.)

Thump604:
Thanks for correcting the wrong-fork push. I verified the PR now points to `78644e79`. Tool forwarding and tool-set cache separation are present, along with the requested tool, lag/finish, unsupported-cache, migration, and formatting coverage.

One blocker remains from the original review: the cache key still has no stable identity for the underlying chat template. Different templates using the same model name and request context can therefore reuse a stale implicit-thinking verdict, and there is no changed-template regression. Please include a stable template signature in the key or scope/clear the cache per template, then add one regression proving that a changed template cannot reuse the previous verdict.

anisoptera:
Fixed in `d7c3d10` (PR head is now `d7c3d10`, 9 commits — verified against the PR rather than the push output this time).

### Template identity in the key

`_template_signature()` (`vllm_mlx/server.py:1178`) hashes the chat template into the cache key, so `model_name` is no longer standing in for template identity.

It hashes **both** the processor's and the tokenizer's `chat_template` rather than re-deriving which one `_apply_chat_template` will pick. That selection is engine-private (`_is_mllm` plus processor capability, `vllm_mlx/engine/batched.py:700-710`), and a second copy of it in `server.py` is a place for the two to drift apart. Hashing both costs at most a spurious re-probe if the unused one changes; missing a change costs a stale verdict, which is the failure you're describing.

When no template is reachable, the probe is **not cached at all** — there would be nothing to invalidate against, so a verdict cached under a partial identity could go stale with no way to notice. That path is slower and correct rather than fast and wrong.

### Regression coverage

Five tests, written failing first (4 red against the previous head; the 5th guards the other direction):

| Test | Proves |
|---|---|
| `test_changed_template_cannot_reuse_the_previous_verdict` | Same engine, template swapped underneath: verdict flips **and** `calls == 2`, so it re-probed rather than got lucky |
| `test_same_model_name_different_templates_do_not_collide` | Your scenario directly — two engines, one `model_name`, different templates, two verdicts |
| `test_identical_template_still_shares_one_entry` | The signature doesn't defeat caching in the common case |
| `test_processor_template_is_part_of_the_identity` | The MLLM render path is covered, not just the tokenizer one |
| `test_unidentifiable_template_is_probed_every_time` | No signature ⇒ nothing cached |

The test harness needed a change worth flagging: `_Engine` previously had no `tokenizer` at all, so without a stub every test in that class would have quietly exercised the new uncacheable path and proved nothing. It now carries a `chat_template`, matching what real engines expose.

### Verification on hardware

Checked against a real GLM-5.3 engine, not only mocks — this is the same class of gap as the previous two rounds, so I checked the live path explicitly before commenting:

```
signature: 613a04170a551cc7... | non-None: True
implicit (no tools): True
implicit (tools)   : True
signature changed  : True
```

Both templates are reachable as 10,644-char strings, so production takes the cacheable path rather than silently falling back to probe-every-time.

Full suite on Apple Silicon (M-series, mlx 0.32.2 / mlx-lm 0.31.3 / transformers 5.16.1), applied on top of `78644e7`: **3006 passed, 2 failed, 24 skipped**. The two failures are the same pre-existing ones as before and unrelated to this branch (`test_mllm_mtp_routing.py::test_mllm_scheduler_exposes_mtp_attempts_and_accepts_on_outputs`, `test_prefix_cache_scheduler_parity.py::test_identical_plain_kv_prompt_never_replays_a_supersequence`). `ruff` and `black` clean with the CI flag set.

One honest note from that run: on a first `-x` pass, `test_memory_cache_mlx.py::TestStoreFailClosedAndEviction::test_store_real_extracted_batch_slice_cuts_batch_retention` failed on an MLX buffer-retention assertion. It passes in the full run and on three consecutive isolated runs, so it looks state-dependent on Metal allocator behaviour rather than related to this change — but it's pre-existing flakiness on that box and I'd rather name it than have you find it.

The three-way split offer from earlier still stands if you'd prefer it over one branch.

Thump604:
I verified exact head `d7c3d10e912349bc2c56a5b02a83d2d18e7c9df0`. The template-identity change resolves my remaining source blocker, and my bounded source review found no remaining P0-P2 issue in that delta.

I did not rerun MLX, inference, or the full suite. Formal approval remains pending maintainer-authorized GitHub CI or equivalent verification accepted by the maintainers.

waybarrios:
@Thump604 CI is running

waybarrios:
Hey @Thump604 CI is all green, if you dont have any questions or concerns, could you change your review? I think @anisoptera solved your issue

Thump604 (approved):
lgtm

PR #741, opened 2026-08-28
fix(qwen4-exp): keep text inference on mlx-vlm path

Thump604:
Qwen4-Exp is not compatible with the generic Qwen3.5 TextModel fallback. Keep text requests on the loaded mlx-vlm language model instead of constructing a mechanically compatible model that produces incorrect logits.

Checks:
- 13 focused text-model dispatch tests
- 954 non-integration Apple-Silicon tests
- Ruff and Black

### DIFF
--- a/tests/test_text_model_dispatch.py
+++ b/tests/test_text_model_dispatch.py
@@ -109,3 +109,20 @@ def test_missing_config_is_not_reported_as_a_build_failure(tmp_path, caplog):
     with caplog.at_level(logging.ERROR, logger="vllm_mlx.text_model_from_vlm"):
         assert build_text_model(_Vlm(), tmp_path) is None
     assert not caplog.records
+
+
+@pytest.mark.parametrize("model_type", ["qwen4_exp", "qwen4_exp_text"])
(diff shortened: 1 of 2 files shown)


janhilgard (approved):
Correct fix for the right reason. Returning `None` so SimpleEngine stays on the loaded mlx-vlm model is better than the alternative here: a mechanically compatible skeleton that loads cleanly and emits wrong logits is far harder to diagnose than a model that refuses the fast path, because nothing fails — the output is just quietly worse.

Two details I checked rather than assumed:

- `str.startswith` accepts a tuple, so `_VLM_ONLY_TEXT_MODEL_PREFIXES` works as written and covers `qwen4_exp_text` from the single entry — the parametrised test pins both spellings, which is what makes the prefix form safe to extend later
- the check sits before `_import_text_model_classes`, so an unregistered architecture short-circuits instead of falling through to the `qwen3_5` default that #686 made prefix-based

The `logger.info` naming the model type matters more than it looks: this path silently changes which engine serves text, and without that line the only symptom would be a throughput difference nobody attributes to dispatch.

Approving.

waybarrios (approved):

waybarrios:
all set and ready to go

PR #742, opened 2026-08-28
Fix schema-aware streamed tool arguments

Thump604:
## Summary

- Recover JSON-encoded tool values conservatively using the declared schema.
- Preserve default incremental argument streaming. Explicit `tool_argument_recovery="buffered"` delays argument bytes until generation finishes while emitting established IDs/names and independent content/reasoning.
- Leave nonfinite, overflowing, or incompatible values unconverted rather than introducing invalid JSON numbers.
- Serialize actual dict/list values targeting strings as compact UTF-8 JSON without sorting keys; preserve already-correct string bytes.

## Why

Models can emit arrays or integers as JSON-encoded strings that strict tool clients reject. Safe recovery needs complete arguments. Buffered streaming is therefore opt-in and delays even already-valid arguments; it does not provide early per-call completion. The selector affects streaming only. Non-streaming continues to recover complete arguments. This is conservative type recovery, not full schema enforcement.

## Verification

- At `4054645`, 56 focused tests passed, covering both streaming branches, selector validation, mixed content/reasoning, call identity, finite JSON, and exact string bytes.
- Selected-source suite: 2,975 passed, 17 skipped for optional Harmony/model-artifact prerequisites, and 69 deselected by slow/integration filters.
- The new selectable streaming mode has not been live-qualified or locally adopted. Local test results above are from `4054645`; `cec511f` changes only the reasoning-parser test fixture. All 10 GitHub checks pass on `cec511f`, including both Apple Silicon jobs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,6 +106,7 @@ jobs:
             tests/test_streaming_think_router.py \
             tests/test_streaming_tool_filter.py \
             tests/test_tool_call_promotion.py \
+            tests/test_tool_argument_coercion.py \
             -v --tb=short \
             -k "not Integration and not InjectJson and not TestMLXMultimodalLMCache" \
             --cov=vllm_mlx \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,7 +106,6 @@ jobs:
             tests/test_streaming_think_router.py \
             tests/test_streaming_tool_filter.py \
             tests/test_tool_call_promotion.py \
-            tests/test_tool_argument_coercion.py \
             -v --tb=short \
             -k "not Integration and not InjectJson and not TestMLXMultimodalLMCache" \
             --cov=vllm_mlx \
@@ -197,6 +196,7 @@ jobs:
             tests/test_normalize_messages.py \
             tests/test_responses_api.py \
             tests/test_thinking_aware_processor.py \
+            tests/test_tool_argument_coercion.py \
             tests/test_tool_choice_forced.py \
             tests/test_tool_choice_none.py \
             tests/test_harmony_render.py \
--- a/docs/guides/tool-calling.md
+++ b/docs/guides/tool-calling.md
@@ -45,6 +45,25 @@ if response.choices[0].message.tool_calls:
         print(f"Arguments: {tc.function.arguments}")
 ```
 
+## Streaming argument recovery
+
+Chat Completions streams parser-produced tool argument fragments unchanged by
+default, including when tools declare arrays or other non-string parameters.
+To opt into schema-aware recovery, send the request extension
+`"tool_argument_recovery": "buffered"` (with the OpenAI Python client, use
+`extra_body={"tool_argument_recovery": "buffered"}`). The default is `"none"`;
+other values are rejected by request validation.
+
+Buffered mode emits each call's ID/name when established and continues emitting
+independent content/reasoning. Only argument bytes wait until the entire
+generation finishes, including for already-valid arguments. This does not
+provide early per-call completion. Complete arguments can then recover, for
+example, a JSON-encoded array string into a schema-declared array. Already-correct
+strings are preserved; actual objects/lists targeting strings use compact UTF-8
+JSON without sorting keys. Nonfinite or incompatible values are not repaired.
+This option does not enforce the full tool schema or change non-streaming
+recovery behavior.
+
 ## Supported Parsers
 
 Use `--tool-call-parser` to select a parser for your model family:
--- a/tests/test_tool_argument_coercion.py
+++ b/tests/test_tool_argument_coercion.py
@@ -61,12 +61,30 @@ def test_preserves_value_that_already_matches_any_union_type():
     assert _coerce_tool_arguments(arguments, "terminal", tools) == arguments
 
 
-def test_preserves_existing_object_to_string_normalization():
-    arguments = json.dumps({"content": {"answer": 42}})
+@pytest.mark.parametrize(
+    ("value", "expected"),
+    [
+        (
+            {"answer": 42, "labels": ["café", "ok"]},
+            '{"answer":42,"labels":["café","ok"]}',
+        ),
+        ([{"answer": 42}, "café"], '[{"answer":42},"café"]'),
+    ],
+)
+def test_object_and_array_to_string_normalization_uses_compact_json(value, expected):
+    arguments = json.dumps({"content": value})
     tools = _tool_schema({"content": {"type": "string"}})
 
     normalized = json.loads(_coerce_tool_arguments(arguments, "terminal", tools))
-    assert json.loads(normalized["content"]) == {"answer": 42}
+    assert normalized["content"] == expected
+
+
+def test_preserves_existing_string_bytes():
+    content = '{\n  "answer": 42,\n  "label": "café"\n}'
+    arguments = json.dumps({"content": content})
+    tools = _tool_schema({"content": {"type": "string"}})
+
+    assert _coerce_tool_arguments(arguments, "terminal", tools) == arguments
 
 
 def test_coerces_json_boolean_without_treating_it_as_integer():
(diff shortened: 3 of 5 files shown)


janhilgard:
Read this one closely because argument coercion is the kind of fix that quietly widens: the failure mode of getting it wrong is a tool silently receiving a *different* value rather than an error, and that is worse than the string-instead-of-array it repairs.

The design is conservative in the ways that matter, and I want to name them because they are what makes this safe rather than clever:

- `_coerce_tool_argument_value` returns the value untouched when it already matches any declared type, so a correct call is never rewritten
- `_decode_tool_argument_string` restricts `json.loads` to types where a JSON-encoded string is unambiguous, and returns the original on `JSONDecodeError` — no partial repairs
- coercion is only accepted if the decoded result actually matches the declared type, so `{"timeout_seconds": "abc"}` against `integer` stays a string rather than becoming garbage
- unions fall through each declared type in order and give up rather than guessing

Buffering until the JSON object is complete is the right trade for typed arguments. Per-fragment coercion cannot work — `"[\"prin` is not decodable and never will be — so the choice is buffer or don't coerce.

One small note, not blocking. When a `dict`/`list` arrives for a `string`-typed parameter:

```python
return json.dumps(value, ensure_ascii=False, indent=2)
```

`indent=2` is a readability choice that changes the bytes the tool receives. For a parameter declared `string` whose content the callee parses, hashes, or diffs, compact separators would be the more neutral default — pretty-printing is a presentation decision being made on the callee's behalf. Worth a sentence either way, since the surrounding functions are otherwise scrupulous about not altering values.

I have not run this against a live server. My own functional battery covers the simple case — an `integer` parameter arriving as `3 (int)` rather than `"3"` — which already worked before this PR, so the regression risk I could have checked is the one this does not touch. The `argv` case in the description is the interesting one and I have no independent evidence on it.

No verdict from me on that basis: the code reads correctly and the test count is substantial, but I would rather leave the approval to someone who has exercised the streaming path end to end.

Thump604:
Agreed. Updated the dict/list → string path to compact UTF-8 JSON with no key sorting. Existing strings remain byte-for-byte unchanged.

Added exact-byte coverage for objects, arrays, nested values, Unicode, key order, and existing strings, and added the test file to CI. Follow-up verification: 2,929 passed, 17 skipped, 69 deselected; Ruff, Black, and workflow YAML checks passed.

The live streamed argv evidence remains valid because this change only affects dict/list values targeting string parameters: one invalid call (`timeout_seconds=0`), one corrected call (`timeout_seconds=1`), exact `argv=["/usr/bin/printf", "transport-ok\n"]`, final `transport-ok`, all 9 checks passed with a concurrent peer.

janhilgard (changes requested):
Thanks for the compact-JSON change — the byte-exactness test on the existing string path is exactly what I asked for and it does what it says.

Reading the whole diff though, the streaming half does considerably more than the title suggests, and I think three things need to change before it lands.

## 1. Tool-call streaming is effectively disabled for every realistic toolset

`_requires_buffered_tool_argument_coercion()` returns `True` if **any** property of **any** tool declares a type in `{array, object, integer, number, boolean, null}`. One integer parameter anywhere in the toolset is enough:

```python
typical = [{"function": {"name": "read_file", "parameters": {"properties": {
    "path":   {"type": "string"},
    "offset": {"type": "integer"},
}}}}]
requires_buffer(typical)      # -> True
```

Once it is `True`, every tool-call delta hits `if not output.finished: continue`, so nothing is emitted until generation ends and the client receives all tool calls in a single terminal chunk. For a string-only toolset it stays `False` and streaming is preserved — but string-only toolsets are the rare case, not the common one.

That is a real behavioural regression for streaming clients: incremental `tool_calls` deltas are what lets a UI render a tool invocation as it is produced, and time-to-first-tool-call-byte goes from "as soon as the name is decoded" to "end of generation".

If the buffering is needed for correctness, I would rather see it scoped to the properties that actually require it — stream string-typed arguments incrementally and buffer only the recoverable-typed ones — or made opt-in. As written, the trade is made unconditionally and invisibly.

## 2. Content carried alongside a tool-call delta is silently dropped

In the buffered branch:

```python
if buffer_typed_tool_calls:
    _merge_streaming_tool_call_fragments(buffered_tool_calls, emitted_tool_calls)
    if not output.finished:
        continue
    emitted_tool_calls = _finalize_streaming_tool_calls(buffered_tool_calls, tools_dict)
```

The `continue` skips the chunk that would otherwise have carried `content=tool_result.get("content") or None`. `_finalize_streaming_tool_calls()` returns only tool calls, so that content is never re-emitted — it is lost, not deferred. Only the content on the final delta survives.

This is the same failure mode as #690 (assistant text dropped when a delta also carries tool calls), reintroduced on the buffered path. Accumulating content next to the tool-call fragments and emitting it with the terminal chunk would fix it.

## 3. The new string → number coercion can emit invalid JSON

`_decode_tool_argument_string()` runs `json.loads()` for `number`/`integer` fields, and Python's `json.loads` accepts `NaN`, `Infinity`, and out-of-range literals that JSON itself does not. `_tool_argument_matches_type()` then sees a `float` and accepts it, and `json.dumps()` re-emits it unquoted:

```
model sent "1e999"     -> inf   json.dumps -> {"x": Infinity}    valid JSON: False
model sent "-1e999"    -> -inf  json.dumps -> {"x": -Infinity}   valid JSON: False
model sent "NaN"       -> nan   json.dumps -> {"x": NaN}         valid JSON: False
model sent "Infinity"  -> inf   json.dumps -> {"x": Infinity}    valid JSON: False
```

A spec-compliant client parser rejects the whole arguments payload. Before this PR those strings passed through untouched, so the coercion creates the problem. `math.isfinite()` on the decoded value before accepting a `number`/`integer` match closes it.

## Smaller notes

- `_coerce_tool_argument_value()` handling union types via `any(_tool_argument_matches_type(...))` is the right call and reads well.
- `changed` now also fires on `type(normalized) is not type(value)`, which correctly catches `True` → `1`-style identity changes that `!=` would miss. Good catch.
- The two blank lines added at `_stream_anthropic_messages` (around the `_finalize_streaming_tool_result` calls) look unintentional.
- Note the dict/list → string path output changed from `indent=2` to compact separators. I agree compact is the better default, but it is a visible change for anything consuming those stringified arguments, so it belongs in the PR description rather than only in a review thread.

Method note: the JSON and buffering behaviour above is from executing the PR's own logic in isolation, not from a live server — this box is memory-committed to another workload right now, so I could not run the streaming path end to end.

Thump604:
Thanks, Jan. I’ve made streamed argument recovery opt-in with `tool_argument_recovery="buffered"`; the default now forwards parser argument fragments unchanged. Buffered mode emits call identity and independent content/reasoning while holding argument bytes until generation finishes. I also fixed the sibling-content loss and leave nonfinite or overflowing values unconverted. Compact UTF-8 conversion and already-correct strings are preserved. The 56 focused tests pass; the new selector has not been live-qualified.

janhilgard (approved):
All three points addressed. Approving.

**1. Streaming is preserved by default.** Making it a request field rather than an inference from the schema is a better answer than the scoping I suggested:

```python
tool_argument_recovery: Literal["none", "buffered"] = "none"
```

Existing clients keep incremental `tool_calls` deltas, and anyone who wants lossless typed recovery asks for it and accepts that the calls arrive at the end. The trade is now visible at the call site instead of being derived from whether some tool happens to declare an integer.

**2. Text alongside a call survives.** The `continue` now fires only when there is genuinely nothing to send:

```python
if not emitted_tool_calls and not tool_result.get("content") and not reasoning:
    continue
```

so content and reasoning carried in the same delta reach the chunk rather than being dropped with it. That was the same failure mode as #690, and it is closed here.

**3. Nonfinite recovery rejected.** `allow_nan=False` on all three `json.dumps` sites turns the bad values into a raise instead of invalid output. Verified against the literals I reported:

```
"NaN"       → ValueError: Out of range float values are not JSON compliant: nan
"1e999"     → ValueError: … inf
"Infinity"  → ValueError: … inf
```

so a client-side `JSON.parse` can no longer be handed `{"x": NaN}`.

Test runs on this revision:

```
tests/test_tool_argument_coercion.py           56 passed
streaming / tool suites across tests/         997 passed, 4 skipped
```

The coercion file grew from 3 tests to 56, and the docs note in `tool-calling.md` means the buffering trade-off is written down where a user picking the option will see it.

Method note: static review plus running the branch's tests and the isolated JSON behaviour. No live server run for this revision — this box is committed to another workload.

PR #744, opened 2026-08-30
fix(mllm): make hybrid prefix-cache replay owned and exact

Thump604:
### Summary

This fixes prefix-cache ownership and replay on the `MLLMScheduler` text path
for hybrid models that combine ordinary KV caches with recurrent cache state.

### Problem

Hybrid prompt state cannot be treated like an ordinary rewindable KV cache.
The previous path could retain request-owned lazy arrays, omit auxiliary prompt
logits from memory accounting, publish an entry after cancellation, or
materialize a cache snapshot before the cold request sampled its first token.
Interleaved prefill could also finish with the generator sampler instead of the
request-local sampling configuration.

These failures made ownership and memory use ambiguous and could make enabled
cold, disabled cold, and warm replay produce different output.

### Changes

- Prepare detached, materialized cache entries outside the publication lock,
  then commit them under a cancellation guard.
- Account for owned cache backing and auxiliary prompt logits before admitting
  an entry. Oversized entries fail closed as normal cache misses.
- Store nonrewindable hybrid state at the prompt boundary and clone owned state
  for replay instead of advancing a stored snapshot with the live request.
- Preserve prompt-final logits, request-local mRoPE state, and request-local
  sampling parameters across ordinary and interleaved prefill.
- Sample and evaluate the cold request before snapshot materialization so
  enabling the cache does not change first-token execution order.
- Prevent cancelled prefill from publishing and release partial request state.

### Verification

The complete repository suite on the final three-commit tree passed:

- 2,976 passed
- 25 skipped
- 27 deselected
- Black and Ruff passed on all five changed files

I also qualified the exact path on an M2 Ultra with Qwen3.8-27B 8-bit,
continuous batching, native-target MTP, a 2,048-token prefill chunk, and a
6,144 MB prefix-cache budget. Each fixture generated a fixed greedy response.

| Fixture | Cache disabled cold | Cache enabled cold | Cache enabled warm | Warm reduction |
|---|---:|---:|---:|---:|
| 2K | 12,160.85 ms | 11,915.68 ms | 583.27 ms | 95.11% |
| 6K | 32,116.78 ms | 32,535.50 ms | 589.34 ms | 98.19% |
| 12K | 67,381.15 ms | 68,253.42 ms | 620.24 ms | 99.09% |
| mRoPE, thinking off | 11,751.55 ms | 11,806.67 ms | 432.97 ms | 96.33% |
| mRoPE, thinking on | 14,866.90 ms | 14,723.32 ms | 3,124.39 ms | 78.78% |

For every fixture, cache-enabled cold output matched cache-disabled cold output,
and warm output matched cache-enabled cold output. The thinking comparison
included reasoning content. MTP completed with zero errors. Cancellation did
not leave an extra entry, and explicit clear returned the cache to zero entries
and zero owned bytes.

These are fixed-workload qualification measurements, not a broad throughput or
concurrency claim.

### Scope and fallback

- Prefix caching remains opt-in.
- Unsupported or oversized cache state is rejected and follows the ordinary
  prefill path.
- Media-request caching remains excluded by the existing request guard.
- Growing chat histories whose tokenized form diverges before the stored prompt
  boundary remain cold misses for hybrid recurrent state. Supporting that reuse
  requires a separately owned stable conversation-boundary checkpoint. This pull
  request does not rewind recurrent cache leaves.
- This does not add SSD or restart restoration, change default routing, or add a
  TextModel override for VLM artifacts.
- This does not qualify external-drafter behavior.

### Related work

Package 1 supersedes #731's hybrid rewind-gate approach. It intentionally keeps
strict rewind rejection for recurrent or non-position-indexed leaves and uses
detached prompt-boundary snapshots with final prompt logits for exact reuse.
Merging #731 independently would change that contract, so please treat #731 as
superseded by this implementation.

PRs #726 and #728 address media-request isolation and overlap parts of the same
generator file. Media caching remains excluded here. Those PRs require separate
reconciliation and should not be folded into this ownership patch without their
own tests. PRs #694, #708, #713, and #732 are also open around adjacent MLLM or
cache behavior; they are not claims or dependencies of this pull request.

### DIFF
--- a/tests/test_memory_cache.py
+++ b/tests/test_memory_cache.py
@@ -265,6 +265,14 @@ class TestCacheEntry:
         assert entry.cache is cache
         assert entry.memory_bytes == 200
 
+    def test_create_entry_accounts_for_auxiliary_arrays(self):
+        cache = [MockKVCache(100, 100)]
+        entry = _CacheEntry.create(
+            [1, 2, 3], cache, auxiliary={"last_logits": MockArray(64)}
+        )
+
+        assert entry.memory_bytes == 264
+
 
 class TestMemoryAwarePrefixCache:
     """Tests for MemoryAwarePrefixCache."""
@@ -298,6 +306,35 @@ class TestMemoryAwarePrefixCache:
         assert len(cache) == 0
         assert cache.memory_limit_mb == 100.0
 
+    def test_prepare_store_rejects_auxiliary_over_memory_limit(
+        self, small_cache, mock_kv_cache
+    ):
+        entry = small_cache.prepare_store(
+            [1, 2, 3],
+            mock_kv_cache(512 * 1024),
+            auxiliary={"last_logits": MockArray(600 * 1024)},
+        )
+
+        assert entry is None
+        assert len(small_cache) == 0
+        stats = small_cache.get_stats()
+        assert stats["store_rejections"] == 1
+        assert stats["current_memory_mb"] == 0
+
+    def test_commit_rejects_oversized_prepared_entry(self, small_cache):
+        entry = _CacheEntry(
+            tokens=(1, 2, 3),
+            cache=[],
+            memory_bytes=2 * 1024 * 1024,
+            auxiliary=None,
+        )
+
+        assert small_cache.commit_prepared(entry) is False
+        assert len(small_cache) == 0
+        stats = small_cache.get_stats()
+        assert stats["store_rejections"] == 1
+        assert stats["current_memory_mb"] == 0
+
     def test_store_and_fetch_exact_match(self, small_cache, mock_kv_cache):
         tokens = [1, 2, 3, 4, 5]
         kv = mock_kv_cache(1000)
@@ -315,6 +352,49 @@ class TestMemoryAwarePrefixCache:
         assert result[0].keys is kv[0].keys
         assert result[0].values is kv[0].values
         assert remaining == []
+
+    def test_prepared_kv_entry_is_not_contaminated_by_live_suffix(self, small_cache):
+        mx = pytest.importorskip("mlx.core")
+        KVCache = pytest.importorskip("mlx_lm.models.cache").KVCache
+
+        live = KVCache()
+        first_keys = mx.array([[[[1.0], [2.0], [3.0]]]])
+        first_values = mx.array([[[[11.0], [12.0], [13.0]]]])
+        live.update_and_fetch(first_keys, first_values)
+
+        prepared = small_cache.prepare_store([1, 2, 3], [live])
+        assert prepared is not None
+
+        live.update_and_fetch(
+            mx.array([[[[4.0], [5.0]]]]),
+            mx.array([[[[14.0], [15.0]]]]),
+        )
+        assert small_cache.commit_prepared(prepared)
+
+        stored, remaining = small_cache.fetch([1, 2, 3])
+        assert remaining == []
+        assert stored[0].offset == 3
+        assert stored[0].keys.shape[2] == 3
+        assert stored[0].keys.tolist() == first_keys.tolist()
+        assert stored[0].values.tolist() == first_values.tolist()
+
+    def test_commit_guard_rejects_before_prefix_eviction(
+        self, small_cache, mock_kv_cache
+    ):
+        base = [1, 2]
+        candidate = [1, 2, 3]
+        assert small_cache.store(base, mock_kv_cache(1000))
+
+        prepared = small_cache.prepare_store(candidate, mock_kv_cache(1000))
+        assert prepared is not None
+        assert not small_cache.commit_prepared(
+            prepared,
+            commit_lock=threading.Lock(),
+            commit_guard=lambda: False,
+        )
+
+        assert tuple(base) in small_cache._entries
+        assert tuple(candidate) not in small_cache._entries
 
     def test_short_prefix_reuse_is_rejected(self, model, mock_kv_cache):
         cache = MemoryAwarePrefixCache(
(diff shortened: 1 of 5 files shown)


djacobsmeyer:
Reporter-side qualification of this PR, requested by review on #730/#731: full protocol, results, and log evidence posted on #730 (#730 (comment): https://github.com/waybarrios/vllm-mlx/issues/730#issuecomment-5469311195).

Summary: on the exact requested protocol (identical ~25.8K-token prompt sent twice, `max_tokens=1`, current upstream head `22efb479`, production's pinned package versions), this PR takes the artifact from 0% reuse on `main` (full 13-chunk re-prefill both times) to a 33.00s → 0.16s (~206x) exact-match cache hit — confirms the fix for the literal case asked about. Separately noted a secondary finding: the natural multi-turn continuation shape (previous prompt + assistant reply + new turn) still misses on this PR, via the same non-trimmable-excess rejection path — details and log excerpts on #730.

Thump604:
Thank you for running the control and exact-head comparison. The 33.00 second to 0.16 second result independently confirms the byte-identical long-prompt case from #730 and the fixed-workload behavior reported here.

I also agree that the natural multi-turn result is a useful separate finding. One implementation detail is worth clarifying: on this PR’s MLLMScheduler path, the owned snapshot is keyed at the prompt boundary from `req.input_ids`, with prompt-final logits stored separately. It is not keyed as prompt plus the first generated token. A request whose token IDs are an exact strict extension of that stored key uses the ordinary prefix path without rewinding. In the reported continuation, however, the new tokenized history shares 25,845 tokens with a 25,846-token stored entry and then diverges. Reuse would therefore require rewinding one token of recurrent `ArraysCache` state.

That recurrent state cannot be safely trimmed, so the cold miss is currently intentional fail-closed behavior. I do not want to weaken that guard or combine recurrent and KV state from different token positions. I clarified this limitation in the PR scope. Supporting growing chat histories that diverge before the stored boundary will need a separate stable conversation-boundary checkpoint design with its own ownership, accounting, cancellation, correctness, and performance qualification.

janhilgard (approved):
Approving. Splitting `store()` into `prepare_store()` / `commit_prepared()` is the right decomposition for this problem: it lets a hybrid recurrent cache be snapshotted at the exact prefill boundary — the only point where the state is still reusable — and published only once the rest of prefill has succeeded, without holding `_memory_lock` across the entry-sized detach. The `commit_guard` closure tied to the request's checkpoint identity is a clean way to make an aborted prefill drop its snapshot instead of publishing a half-owned entry.

Correcting the docstring on `fetch()` from "no copy, MLX arrays are immutable and safe to share" to a statement that callers must clone before a mutating replay matters more than its size suggests — that sentence was the justification for the aliasing this PR removes.

Things I checked:

- **Lock ordering.** `commit_prepared()` takes `_memory_lock` and then the caller's `commit_lock`. All four `with self._prefix_checkpoint_lock:` blocks in `mllm_batch_generator.py` (lines 776, 785, 1336, 1355) touch only plain dicts and sets and never call into the prefix cache, so there is no AB/BA path today. It is worth writing that contract into the `commit_prepared()` docstring, though — "`commit_lock` is acquired inside `_memory_lock`; do not call cache methods while holding it" — because the deadlock it prevents is invisible from the call site.
- **LRU refresh on the fast path.** `prepare_store()` returns an existing entry without `move_to_end()`, which looked like a lost refresh, but `commit_prepared()` does it under the lock, and `store()` routes through both, so behaviour is preserved.
- **Stats after prefix eviction.** The old `if to_remove: ...` stats update is gone, but `entry_count` and `current_memory_bytes` are assigned unconditionally at the end of the commit block, so nothing drifts.
- **Auxiliary detach.** `value + 0` under `_copy_lock` followed by `mx.eval` on just the `mx.array` values is the right shape, and folding `nbytes` into `_CacheEntry.create()` means the auxiliary payload is actually accounted rather than being invisible headroom.

One note, non-blocking. `commit_prepared()` holds *both* locks across `_evict_lru()`, and eviction can run a spill hook. `abort_prefill()` (line 776) needs the same `_prefix_checkpoint_lock` and, judging by the sibling comment on `schedule_removal()`, is called from the event loop on client disconnect. So a slow eviction now blocks a disconnect-abort on the event loop thread, where it did not before. Narrow, and only as slow as the spill hook, but if the SSD-spill path can be slow it would be worth doing the eviction before taking `commit_lock`.

I could not re-run the long-prefix measurement for this revision — the box is memory-committed to another workload right now — so I am not restating the 33.00 s → 0.16 s figure as verification of the current head. The 1074 lines of new tests across `test_memory_cache.py`, `test_prefix_cache.py` and `test_mllm_continuous_batching.py` are what I am leaning on instead, and they cover the ownership and exactness claims directly.

Thump604:
Thanks, Jan. I added the lock-order note to `commit_prepared()`: `_memory_lock` is acquired before the caller commit lock, and callers must not call cache methods while holding that lock.

I also agree with your CI and measurement distinction. I am not treating the earlier 33.00 second to 0.16 second measurement as fresh performance verification of the current head. The current head is supported by the ownership and exactness regressions plus the rerun CI, which is green across lint, type checking, Python 3.10 through 3.14, Apple Silicon 3.11 and 3.13, and the aggregate tests job.

I am keeping the SSD-spill delay on disconnect as a separate follow-up. This commit only documents the existing locking contract.

janhilgard (approved):
Re-approving. The only change since my last review is the docstring on `commit_prepared()`:

> `commit_lock` is acquired inside `_memory_lock`. Callers must not call cache methods while holding the caller-supplied `commit_lock`.

That is exactly the contract I asked to have written down, and in the place a future caller will actually read it. The invariant was already true in the code; now it survives someone adding a fifth `with self._prefix_checkpoint_lock:` block.

My other note — that `commit_prepared()` holds both locks across `_evict_idle_lru_for_prefill()`, so a slow SSD spill can delay a disconnect-abort on the event loop thread — is unchanged and remains non-blocking. It is a latency question under an eviction path that may never be slow in practice, not a correctness one.

Everything else stands as reviewed on 2 Sep.

PR #745, opened 2026-08-30
fix(shutdown): exit the serve process before MLX's thread-local teardown segfaults

sbayer2:
Split out of #733 at review request. This is a process-wide lifecycle change and does not belong in a video PR.

## The crash

Stopping the server segfaults **after** it has finished serving. macOS recorded six identical crash reports:

```
EXC_BAD_ACCESS (SIGSEGV), KERN_INVALID_ADDRESS at 0x350
  _pthread_exit -> _pthread_tsd_cleanup
    -> dyld::ThreadLocalVariables::finalizeList
      -> mlx::core::detail::CompileCache::CacheEntry::~CacheEntry()
        -> _Py_Dealloc
```

MLX keeps its compile cache in thread-local storage and the entries own Python objects. When a thread that ran compiled MLX work exits, dyld runs the TLS destructor, which deallocates Python objects with no thread state and no GIL.

It is **not thread-specific**: five reports fault on `asyncio_0`, one on `simple-generate_0` (the pinned MLX worker), so rerouting work between executors does not help. It is **not video-specific** either: `asyncio_0` is the default-executor thread that lifespan shutdown itself creates via `_run_blocking_engine_cache_io()` to persist the prefix cache. Shutdown spawns a thread, touches MLX from it, and lets it die.

This is an upstream MLX bug. vllm-mlx cannot clear that cache from Python — `clear_streams()` covers streams, `clear_cache()` covers memory, nothing covers the compile cache — so the crash can only be outrun.

## The change

Once lifespan shutdown has finished there is nothing left but teardown, so the serve process exits instead of returning into it.

Placement matters and was got wrong first: a guard after `uvicorn.run()` still crashed, because `uvicorn.run` calls `asyncio.run`, which joins the default-executor threads **inside itself**. The exit has to happen at the end of lifespan shutdown, which runs earlier.

**Finalization is skipped only when both phases raised nothing.** An earlier revision of this checked `primary_exc` but not `cleanup_exc`, so a failure in cache-save, engine-stop, MCP or model-manager shutdown was captured and then `os._exit(0)` ran anyway, making `raise cleanup_exc` unreachable — a broken shutdown reporting success. That was caught in review of #733 and is fixed here. If either phase failed, the guard stands aside and the exception propagates with a non-zero exit; a teardown segfault on that path is the better trade, since a process that already has an error to report should report it loudly rather than exit silently and clean.

The one `atexit` handler this package owns (temp-file cleanup) is invoked explicitly, and streams are flushed. Gated on `server._exit_process_after_shutdown`, set only by the two serve entry points, so importing the app as a library never exits the host process. `VLLM_MLX_CLEAN_EXIT=0` restores normal finalization.

## Tests

`tests/test_shutdown_exit.py`, driving `lifespan` directly with the optional subsystems nulled:

| behaviour | test |
|---|---|
| clean shutdown | `test_clean_shutdown_skips_finalization` |
| cleanup exception | `test_cleanup_failure_is_raised_not_swallowed` |
| serving exception | `test_serving_failure_is_raised_not_swallowed` |
| embedded/library use | `test_library_use_never_exits`, `test_flag_defaults_off` |
| both launchers | `test_both_launchers_arm_the_flag` |
| `VLLM_MLX_CLEAN_EXIT=0` | `test_env_var_disables_the_guard` |
| owned atexit work | `test_exit_runs_owned_atexit_work_first`, `test_exit_survives_cleanup_errors` |

The regression test discriminates: restoring the old guard makes `test_cleanup_failure_is_raised_not_swallowed` fail; the fix makes it pass.

These skip without MLX, since `vllm_mlx.server` and `vllm_mlx.cli` both import it transitively — the same reason `tests/test_lifecycle_server.py` is not in the `ubuntu-latest` job. They need an Apple Silicon runner to execute. Say the word if you would rather they were restructured to run without MLX; it would mean extracting the guard's decision into a module that does not pull in the engines.

## Not resolved

No standalone reproducer for the MLX defect yet, so nothing has been filed upstream. `mx.compile` on a worker thread that then exits does not reproduce it on its own, nor does adding a per-thread GPU stream and a surviving array. Reported here with the crash reports rather than guessed at.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,6 +87,7 @@ jobs:
             tests/test_chat_template_safety.py \
             tests/test_simple_engine_thread_pinning.py \
             tests/test_engine_base.py \
+            tests/test_shutdown_decision.py \
             tests/test_harmony_parsers.py \
             tests/test_endpoint_model_policies.py \
             tests/test_truncation.py \
@@ -201,6 +202,8 @@ jobs:
             tests/test_harmony_render.py \
             tests/test_embeddings.py \
             tests/test_rerank.py \
+            tests/test_shutdown_decision.py \
+            tests/test_shutdown_exit.py \
             -v --tb=short \
             -m "not slow" \
             -k "not Integration"
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -126,6 +126,7 @@ Create `mcp.json`:
 | Variable | Description |
 |----------|-------------|
 | `VLLM_MLX_TEST_MODEL` | Default model for tests |
+| `VLLM_MLX_CLEAN_EXIT` | Set to `0` to keep normal interpreter finalization on shutdown. Default `1`: the serve process exits at the end of lifespan shutdown, once every request is served and caches are persisted, because MLX's thread-local compile cache segfaults during CPython teardown. Only affects the `serve` entry points, never library use. |
 | `HF_TOKEN` | HuggingFace authentication token |
 | `OPENAI_API_KEY` | Set to any value for SDK compatibility |
 
--- a/tests/test_shutdown_decision.py
+++ b/tests/test_shutdown_decision.py
@@ -1,0 +1,207 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Portable coverage for the clean-exit policy (vllm_mlx.shutdown).
+
+The policy decides whether the serve process may skip CPython finalization at
+the end of lifespan shutdown (the MLX thread-local teardown segfault
+workaround). The decision module imports no engine code, so these tests run
+in the standard no-MLX CI job on every platform. The os._exit/lifespan wiring
+itself — driving the real server through startup and shutdown — remains in
+tests/test_shutdown_exit.py, which needs MLX and runs on Apple Silicon.
+
+vllm_mlx.server and vllm_mlx.cli cannot be imported without MLX, so the tests
+that pin their call sites read the source instead: an AST assertion fails the
+same way a broken import would, without needing the platform.
+"""
+
+from __future__ import annotations
+
+import ast
+from pathlib import Path
+
+import pytest
+
+from vllm_mlx.shutdown import (
+    clean_exit_enabled,
+    exit_without_finalizing,
+    should_exit_without_finalizing,
+)
+
+PACKAGE = Path(__file__).resolve().parent.parent / "vllm_mlx"
+
+
+# ---------------------------------------------------------------------------
+# The decision predicate
+# ---------------------------------------------------------------------------
+
+
+def test_clean_shutdown_with_optin_exits():
+    assert should_exit_without_finalizing(None, None, True, env={}) is True
+
+
+def test_primary_failure_never_exits():
+    """A failed serving phase must propagate and exit non-zero."""
+    assert (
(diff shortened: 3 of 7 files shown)


Thump604 (changes requested):
I rechecked #745 at `4941cb4` against current main `22efb47`. The `primary_exc`/`cleanup_exc` predicate now preserves the reviewed failure-reporting boundary, and the default-off/env-disabled and two launcher wiring are clear.

One merge gate remains: `tests/test_shutdown_exit.py` imports MLX at collection, and the current CI workflow does not list it in either the Ubuntu or Apple test commands. Please extract the decision into a small module with no engine/MLX import and add standard-CI tests for clean shutdown, primary failure, cleanup failure, `VLLM_MLX_CLEAN_EXIT=0`, both launcher call sites, and embedded/library use. Keep `os._exit`/lifespan wiring as the maintainer-owned Apple Silicon integration gate. If the project has no Apple runner capacity, that is infrastructure rather than a contributor failure; before merge, retain exact local Apple evidence for the real server/CLI shutdown path and exit status.

sbayer2:
Addressed at 0329abc. The decision predicate now lives in `vllm_mlx/shutdown.py` with no engine/MLX imports; `cli._exit_without_finalizing` keeps its name and delegates. `tests/test_shutdown_decision.py` covers clean shutdown, primary failure, cleanup failure, `VLLM_MLX_CLEAN_EXIT=0`, the embedded/library default, exit mechanics, and both launcher call sites plus the module default via AST source assertions (importing server/cli needs MLX, which is what this file must not do). It's wired into the Ubuntu no-MLX job; `test_shutdown_exit.py` joins the Apple job list as the os._exit/lifespan integration gate.

Local Apple evidence (M-series, mlx 0.32.1, branch head): real `vllm-mlx serve` of Qwen3.8-27B-4bit, one completed `/v1/chat/completions`, SIGINT -> "Shutdown complete", exit status 0, no new DiagnosticReports. Happy to paste the full transcript if useful.

The CI runs for this head are queued pending workflow approval (`action_required`) — could you approve so the portable run is retained on the PR?

Thump604 (changes requested):
I approved the withheld workflow and inspected the executed Apple logs on exact head `0329abcb`. The apparent green result is not a valid test pass.

Both Apple jobs terminate the main pytest process at the same point, around 76%, immediately after starting `TestLifecycleFailureHandling::test_registry_idle_reaper_task_starts_and_stops_with_lifespan`. There is no pytest completion summary, and none of the newly listed `tests/test_shutdown_decision.py` or `tests/test_shutdown_exit.py` cases executes. Because the premature termination is `os._exit(0)`, Actions marks the step successful and continues to the separate EngineCore step.

The likely real interleaving is the process-global `_exit_process_after_shutdown` being armed by an earlier launcher test whose mocked `uvicorn.run()` returns, then remaining true when a later in-process lifespan test runs; that path reaches the real exit helper and ends pytest successfully. Please make launcher arming scoped/restored when `uvicorn.run()` returns, while preserving production behavior when the process exits inside lifespan. Keep direct exit mechanics isolated so a regression cannot terminate the surrounding test process.

Then rerun both Apple jobs and verify the main test process reaches a normal pytest summary and that every new shutdown test is explicitly reported as passed. The portable decision suite is already independently `14/14` green; no model run or broader redesign is requested.

Issue #746, opened 2026-08-30
[Bug]: vllm_mlx_engine_steps_executed is never populated for MLLM-routed models


djacobsmeyer:
### Environment

- Chip: Apple M3 Ultra
- RAM: 128 GB
- macOS version: 26.x
- Python version: 3.13
- vllm-mlx version: 0.4.1 (`origin/main` @ `22efb47968f96e6e7a79704f4eb0184cda4e0068`, current `main` as of this writing — line numbers below verified against this exact commit)
- mlx / mlx-lm / mlx-vlm / mlx-audio versions: mlx 0.32.2, mlx-lm 0.31.3, mlx-vlm 0.6.15

### Model and quantization

`mlx-community/Qwen3.8-27B-8bit` (MLLM-routed per its `config.json`'s `vision_config`). Not specific to this checkpoint — see "Additional context": any model that routes through `BatchedEngine`'s MLLM branch (`self._mllm_scheduler` set) reproduces this, registry mode (`--models-config`) or not.

### Reproduction

```
vllm-mlx serve mlx-community/Qwen3.8-27B-8bit \
  --continuous-batching --enable-prefix-cache --enable-metrics --port 8091
```

Send any request to load and warm the model, then scrape:

```
curl -s -X POST http://localhost:8091/v1/chat/completions \
  -H 'Content-Type: application/json' \
  -d '{"model":"mlx-community/Qwen3.8-27B-8bit","messages":[{"role":"user","content":"Say OK."}],"max_tokens":10}'

curl -s http://localhost:8091/metrics | grep vllm_mlx_engine_steps_executed
```

### Expected behavior

With the model resident and having served at least one request, `vllm_mlx_engine_steps_executed` should be a positive, increasing count of scheduler steps run.

### Actual behavior

`vllm_mlx_engine_steps_executed 0.0` — permanently, regardless of request volume. This reproduces independently of #738/#739: it is present before that fix (module-level `_engine` is `None` in registry mode, so every gauge is dead) and it is **also present after** #739's fix lands, once `/metrics` is wired to a real, live, resident engine — because the value is simply never produced for MLLM models in the first place, on any serving path.

### Additional context

**Root cause**, verified by direct source read against `origin/main@22efb47`:

- `vllm_mlx/engine/batched.py:1223-1242` — `BatchedEngine.get_stats()`'s MLLM branch:
  ```python
  if self._mllm_scheduler:
      mllm_stats = self._mllm_scheduler.get_stats()
[...]

PR #749, opened 2026-08-30
fix(mllm): populate vllm_mlx_engine_steps_executed for MLLM-routed models

djacobsmeyer:
Refs #746

Implements the full spec from Thump604's 2026-08-30T19:58:47Z comment on #746: deterministic no-real-inference coverage for both the scheduler counter and the `BatchedEngine` promotion/metrics mapping, plus an exact-head Apple Silicon MLLM run showing a positive, increasing gauge. Kept separate from #739's registry-engine-selection fix.

## Update (2026-09-08): review-driven changes since the original head

Two rounds of review on this PR changed the counter's placement and the CI mapping described below. Current state, as of head `ed8f2ca`:

- **Counter placement**: `_steps_executed` now increments on `step()`'s successful-return path (immediately before its single `return output`), not unconditionally at the top of the method — see the "Changes" section below, which describes the current code. A step that raises partway through (schedule_waiting, the batch generator's forward pass, response processing) is no longer counted; its requests get terminated by `_fail_requests_after_step_error` rather than retried, since retrying a partially mutated batch is unsafe. Regression test: `test_step_that_raises_is_not_counted`.
- **CI mapping**: six scheduler/gauge cases in `tests/test_mllm_steps_executed_stat.py` run in the Linux `test-matrix` job (a dedicated "Run MLX-stubbed scheduler tests" step, `tests/_mlx_stub.py` standing in for the real mlx/mlx_lm package) as well as in `test-apple-silicon` (real mlx). The Apple job additionally runs `tests/test_metrics.py`'s HTTP-layer metrics test (5 cases) — that file stays Apple-Silicon-only, since its import chain also needs `uvicorn`/`prometheus-client`, which the Linux job doesn't install. The one gauge-mapping assertion that specifically needed Linux coverage moved to a dependency-light equivalent, `TestMetricsEngineStepsExecutedGauge` (in `test_mllm_steps_executed_stat.py`), which calls `MetricsCollector` directly and needs neither uvicorn, prometheus-client, nor mlx.
- CI is green on `ed8f2ca`: [run 33983587011](https://github.com/waybarrios/vllm-mlx/actions/runs/33983587011), all 10 checks passing.

The **"Tests" and "Apple Silicon evidence" sections below, and their numbers, are exactly as gathered against this PR's original head, `52c947a`** (before either review round) — kept as-is rather than restated, since they're a record of that specific run, not a living description of current head. See the PR comments from 2026-09-04 and 2026-09-05 for the intermediate heads (`e6c81fe`, `ed8f2ca`) and what changed/was verified at each.

## Problem

`vllm_mlx_engine_steps_executed` is permanently `0.0` for any model routed through `BatchedEngine`'s MLLM branch (`self._mllm_scheduler` set) — registry mode or not, and independent of #738/#739. Root cause (verified against `origin/main@22efb47`, unchanged at push time):

- `vllm_mlx/mllm_scheduler.py`'s `MLLMScheduler.get_stats()` never produced a `"steps_executed"` key.
- `vllm_mlx/engine/batched.py`'s `BatchedEngine.get_stats()` MLLM promotion allowlist omitted that key, so even if present it would never reach the top-level stats dict `metrics.py` reads.

The plain-LLM path already has this working via `AsyncEngineCore._steps_executed` (`engine_core.py`), incremented once per `scheduler.step()` call inside the engine loop, itself gated by `scheduler.has_requests()`.

## Changes

- `vllm_mlx/mllm_scheduler.py`: adds `MLLMScheduler._steps_executed`, incremented once per `step()` call on the **successful-return path** — immediately before `step()`'s single `return output`, after every operation in the method that can raise (`process_pending_removals`, `_schedule_waiting`, the batch generator's `next()`, `_process_batch_responses`) has already completed. This mirrors `AsyncEngineCore._steps_executed`'s placement exactly: `AsyncEngineCore` increments only after `self.scheduler.step()` returns, not before calling it. (The first push had this unconditional at the top of the method instead; moved per review — see "Update" above.) Exposed via `get_stats()["steps_executed"]`.
- `vllm_mlx/engine/batched.py`: adds `"steps_executed"` to the MLLM promotion allowlist tuple so it reaches the top-level stats dict.
- `vllm_mlx/metrics.py`: **no change** — `_update_engine_gauges` already duck-types `stats.get("steps_executed")`; it was only ever missing the value to read.

## Tests

`tests/test_mllm_steps_executed_stat.py` (mirrors the fake/fixture style of `test_mllm_continuous_batching.py`; real `mlx.core` on the Apple job, `tests/_mlx_stub.py`'s stub on the Linux `test-matrix` job — see "Update" above for the current CI mapping), six cases across three classes:

- `TestMLLMSchedulerStepsExecuted`: `get_stats()` reports `0` on a fresh scheduler; increments by exactly 1 per `step()` call (batch-generator construction bypassed via a no-op `_schedule_waiting` stub, since the counter contract doesn't depend on it); and `test_step_that_raises_is_not_counted`, added on review, which makes `step()` fail via a batch generator whose `next()` raises and confirms the counter is untouched, then confirms a subsequent successful step still counts normally.
- `TestBatchedEngineStepsExecutedPromotion`: a fake `_mllm_scheduler.get_stats()` dict with `steps_executed` gets promoted to the top level; one without it is correctly omitted (not defaulted to 0/false-positive).
- `TestMetricsEngineStepsExecutedGauge`: added on review as the dependency-light home for the gauge-mapping assertion originally in `test_metrics.py` (see "Update" above) — calls `MetricsCollector._update_engine_gauges()` directly against a `defaultdict(MagicMock)` standing in for the real Prometheus registry, confirming `get_stats()["steps_executed"]` reaches `vllm_mlx_engine_steps_executed`.

Verified these are real regression guards, not vacuously green: reverted both source changes locally and confirmed 3 of 4 (original count, before the two review-added cases) new tests fail with `KeyError: 'steps_executed'`; separately, reverting the counter-placement fix alone makes `test_step_that_raises_is_not_counted` fail (`assert 1 == 0`).

Also added one assertion to the existing `test_metrics.py::test_metrics_endpoint_scrapes_runtime_stats` (already used a `FakeEngine.get_stats()` with `steps_executed=7` but never asserted on it) confirming the gauge renders `vllm_mlx_engine_steps_executed 7.0`. This file is Apple-Silicon-only (see "Update" above for why, and for where its CI mapping stands now).

### Test results as of the original head (`52c947a`) (isolated `uv venv`, no model download/serving)

- `tests/test_mllm_steps_executed_stat.py`: 4/4 passed (the original two classes; `test_step_that_raises_is_not_counted` and `TestMetricsEngineStepsExecutedGauge` were added in later commits — see "Update" above)
- `tests/test_metrics.py` (new assertion included): 5/5 passed
- Mapped Apple-job subset (`test_mllm_continuous_batching.py`, `test_mllm_cache.py`, `test_mllm_steps_executed_stat.py`, `test_optimizations.py`): 98 passed, 5 deselected
- Full suite vs. `origin/main` baseline: baseline 2965 passed/13 skipped/27 deselected → this branch 2969 passed/13 skipped/27 deselected (exactly the 4 new tests; no regressions)

For the CI-contract fix (the mlx-stub/uvicorn gap and its resolution), see the 2026-09-04/09-05 PR comments and "Update" above; the exact-head CI run (`ed8f2ca`, [run 33983587011](https://github.com/waybarrios/vllm-mlx/actions/runs/33983587011)) is the current source of truth for whether everything passes together, rather than a further round of local-only numbers.

## Apple Silicon evidence, as of the original head (`52c947a`) (hardware: Apple M5 Max, 128GB)

Built into a disposable `git worktree` + isolated `uv venv`, `pip install -e .` — production's actual serving install/pin untouched throughout. Temporary server on `:8093`, `HOME`-override cache isolation (confirmed: prefix-cache path resolved under the scratch `HOME`, and the shared `~/.cache/vllm-mlx/prefix_cache/` on this machine kept its same 3 pre-existing entries, untouched, before/after).

Registry mode (`--models-config`) was tried first per the original plan, but its `/metrics` still reads only the module-level `_engine` (`None` in registry mode) on this base commit — that's exactly #739's still-unmerged fix, unrelated to #746. To get real evidence for *this* fix uncorrelated with that separate gap, I switched to single-model mode with `--mllm` (bare single-model launch skips MLLM auto-detection for this checkpoint; known gotcha), `--continuous-batching --enable-prefix-cache --enable-metrics`, pointed at the already-downloaded local HF cache snapshot (no network fetch: `Fetching 17 files: 100%` from local cache).

Startup log confirms MLLM routing through `BatchedEngine`:
```
INFO:vllm_mlx.engine.batched:BatchedEngine loaded: mlx-community/Qwen3.8-27B-8bit (mllm=True)
```

Gauge before/after (`_steps_executed` was unconditional-at-top-of-`step()` at this head — see "Update" above for the counter-placement fix made in a later commit; that fix doesn't change these particular numbers, since none of these requests hit the exception path it addresses):
```
# before any request
vllm_mlx_model_loaded 1.0
vllm_mlx_engine_steps_executed 0.0

# after request 1 (max_tokens=10, finish_reason=length)
vllm_mlx_engine_steps_executed 10.0

# after request 2 (max_tokens=20, finish_reason=length)
vllm_mlx_engine_steps_executed 30.0
```

Positive and monotonically increasing, as the issue requires. SIGTERM'd cleanly (`Application shutdown complete`, `BatchedEngine stopped`); `:8093` confirmed free afterward. Production `:8091`/`:8082` (same PIDs throughout) unaffected — `:8091` still answering `200` on `/v1/models` after this run, and the shared prefix-cache directory shows no new entry for this model.

## Test plan for reviewer

- [x] `tests/test_mllm_steps_executed_stat.py` passes (original head `52c947a`: 4/4; current head `ed8f2ca`: 6/6, including the exception-path and gauge-mapping cases added on review)
- [x] Mutation-tested: reverting the counter fix fails the new/moved tests (both the original promotion fix and the later successful-return-placement fix)
- [x] Full suite: no regressions vs. `origin/main`, as of original head `52c947a` (2969 passed vs. baseline 2965)
- [x] Live MLLM run on Apple Silicon: gauge positive and increasing, as of original head `52c947a`
- [x] Production serving untouched, throughout every round of this PR
- [x] CI green on current head `ed8f2ca`: [run 33983587011](https://github.com/waybarrios/vllm-mlx/actions/runs/33983587011), all 10 checks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,20 +119,30 @@ jobs:
           file: ./coverage.xml
           fail_ci_if_error: false
 
-      - name: Run MLX-stubbed metrics/scheduler tests
+      - name: Run MLX-stubbed scheduler tests
         # Deliberately its own step (fresh process), not merged into "Run
-        # unit tests (no MLX required)" above: these two files install a
-        # MagicMock-backed mlx/mlx_lm stub (tests/_mlx_stub.py) so their
+        # unit tests (no MLX required)" above: this file installs a
+        # MagicMock-backed mlx/mlx_lm stub (tests/_mlx_stub.py) so its
         # mlx.core-importing dependencies load without the real package.
         # Several files in the step above (e.g. test_memory_cache.py,
         # test_prefix_cache.py) have their own real `except ImportError`
         # mlx-optional handling that a stub in the same sys.modules/process
         # would silently break (isinstance() against a MagicMock raises
         # TypeError instead of behaving like "mlx absent") -- see
-        # tests/_mlx_stub.py's docstring. Refs #746 / PR #749 review.
+        # tests/_mlx_stub.py's docstring.
+        #
+        # test_metrics.py is NOT included here: it needs the full
+        # vllm_mlx.server import chain (uvicorn, prometheus-client), neither
+        # of which this job installs, and mlx-stubbing them the same way
+        # broke fixture setup with ModuleNotFoundError. That file stays
+        # Apple-Silicon-only; the one contract it covers that also needed
+        # Linux coverage (get_stats()["steps_executed"] reaching the
+        # vllm_mlx_engine_steps_executed gauge) has a dependency-light
+        # equivalent below, in TestMetricsEngineStepsExecutedGauge, which
+        # calls MetricsCollector directly and needs neither of those.
+        # Refs #746 / PR #749 review.
         run: |
           pytest \
-            tests/test_metrics.py \
             tests/test_mllm_steps_executed_stat.py \
             -v --tb=short
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,6 +119,23 @@ jobs:
           file: ./coverage.xml
           fail_ci_if_error: false
 
+      - name: Run MLX-stubbed metrics/scheduler tests
+        # Deliberately its own step (fresh process), not merged into "Run
+        # unit tests (no MLX required)" above: these two files install a
+        # MagicMock-backed mlx/mlx_lm stub (tests/_mlx_stub.py) so their
+        # mlx.core-importing dependencies load without the real package.
+        # Several files in the step above (e.g. test_memory_cache.py,
+        # test_prefix_cache.py) have their own real `except ImportError`
+        # mlx-optional handling that a stub in the same sys.modules/process
+        # would silently break (isinstance() against a MagicMock raises
+        # TypeError instead of behaving like "mlx absent") -- see
+        # tests/_mlx_stub.py's docstring. Refs #746 / PR #749 review.
+        run: |
+          pytest \
+            tests/test_metrics.py \
+            tests/test_mllm_steps_executed_stat.py \
+            -v --tb=short
+
   test-python-3-14:
     # Focused job for the CPython 3.14+ asyncio.shield() behavior change
     # that shield_task() (vllm_mlx/engine/base.py) works around -- the
@@ -179,6 +196,7 @@ jobs:
             tests/test_mllm_continuous_batching.py \
             tests/test_mllm_cache.py \
             tests/test_mllm_steps_executed_stat.py \
+            tests/test_metrics.py \
             tests/test_optimizations.py \
             tests/test_simple_engine.py \
             tests/test_simple_engine_prefix_trie_cache.py \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -178,6 +178,7 @@ jobs:
             tests/test_batched_engine_mllm_config.py \
             tests/test_mllm_continuous_batching.py \
             tests/test_mllm_cache.py \
+            tests/test_mllm_steps_executed_stat.py \
             tests/test_optimizations.py \
             tests/test_simple_engine.py \
             tests/test_simple_engine_prefix_trie_cache.py \
--- a/tests/_mlx_stub.py
+++ b/tests/_mlx_stub.py
@@ -1,0 +1,57 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Opt-in mlx/mlx_lm stub for test files that don't need real MLX behavior.
+
+Deliberately NOT wired into conftest.py: conftest.py loads unconditionally
+for every pytest invocation that touches this directory, so a stub
+installed there would leak into ``sys.modules`` for *every* test file
+collected in the same run -- including ones like test_memory_cache.py and
+test_prefix_cache.py, which have their own working ``except ImportError``
+fallback for real mlx_lm absence and break when that import instead
+"succeeds" against a MagicMock (e.g. ``isinstance(x, mock_attr)`` raises
+``TypeError``, not a clean not-mlx result).
+
+Call ``install_if_unavailable()`` at the top of a test file, before any
+``import vllm_mlx...`` line, and keep that file's CI selector in its own
+pytest invocation, separate from files that have their own mlx-optional
+handling (see .github/workflows/ci.yml's dedicated
+"Run MLX-stubbed metrics/scheduler tests" step). A no-op wherever mlx is
+actually installed (the Apple Silicon CI job, or a dev machine with mlx
+present).
+"""
+
+
+def install_if_unavailable() -> None:
+    try:
+        import mlx.core  # noqa: F401
+
+        return
+    except ImportError:
+        pass
+
+    import importlib.machinery
+    import sys
+    from unittest.mock import MagicMock
+
+    for name in (
+        "mlx",
+        "mlx.core",
+        "mlx.nn",
+        "mlx_lm",
+        "mlx_lm.generate",
+        "mlx_lm.models",
+        "mlx_lm.models.cache",
+        "mlx_lm.sample_utils",
+        "mlx_lm.tokenizer_utils",
+    ):
+        if name in sys.modules:
+            continue
+        stub = MagicMock(name=name)
+        # A real ModuleSpec, not None: importlib.util.find_spec() (used by
+        # e.g. transformers' is_mlx_available()) raises ValueError on a
+        # sys.modules entry whose __spec__ is falsy rather than a spec.
+        stub.__spec__ = importlib.machinery.ModuleSpec(name, loader=None)
+        stub.__path__ = []  # let it satisfy submodule imports like a package
+        sys.modules[name] = stub
+        parent, _, child = name.rpartition(".")
+        if parent:
+            setattr(sys.modules[parent], child, stub)
--- a/tests/test_metrics.py
+++ b/tests/test_metrics.py
@@ -170,6 +170,10 @@ class TestMetricsEndpoint:
         assert (
             'vllm_mlx_cache_type{cache_type="memory_aware_cache"} 1.0' in response.text
         )
+        # get_stats()["steps_executed"] must reach the gauge -- the
+        # duck-typed read side of #746 (the producer side, MLLMScheduler /
+        # BatchedEngine, is covered by test_mllm_steps_executed_stat.py).
+        assert "vllm_mlx_engine_steps_executed 7.0" in response.text
 
     def test_metrics_collapse_unmatched_paths(self, metrics_client, monkeypatch):
         client, server, collector = metrics_client
--- a/tests/test_metrics.py
+++ b/tests/test_metrics.py
@@ -1,16 +1,24 @@
 # SPDX-License-Identifier: Apache-2.0
-"""Tests for Prometheus server metrics."""
+"""Tests for Prometheus server metrics.
 
-import platform
-import sys
+Everything here exercises the metrics endpoint against a FastAPI TestClient
+with a fake/monkeypatched engine -- no real MLX operation runs. It used to
+be gated to Apple Silicon only, but that was never actually about needing
+real MLX: importing ``vllm_mlx.server`` just needs ``mlx.core`` importable,
+which ``_mlx_stub.install_if_unavailable()`` below provides on runners
+without the real package. This file previously wasn't selected by any CI
+job at all (see #746 / PR #749's review) -- it now runs in a dedicated
+Linux-matrix step (kept separate from files with their own real
+``except ImportError`` mlx-optional handling, e.g. test_memory_cache.py --
+see _mlx_stub.py's docstring) and in the Apple job, where mlx is real and
+this stub never engages.
+"""
+
+from tests import _mlx_stub
+
+_mlx_stub.install_if_unavailable()
 
 import pytest
-
-# Skip all tests if not on Apple Silicon
-pytestmark = pytest.mark.skipif(
-    sys.platform != "darwin" or platform.machine() != "arm64",
-    reason="Requires Apple Silicon",
-)
 
 
 class FakeEngine:
--- a/tests/test_metrics.py
+++ b/tests/test_metrics.py
@@ -1,24 +1,32 @@
 # SPDX-License-Identifier: Apache-2.0
 """Tests for Prometheus server metrics.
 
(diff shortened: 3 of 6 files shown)


waybarrios (changes requested):
Thanks for the focused fix and Apple hardware validation. I found two issues to address:

1. `MLLMScheduler.step()` increments `_steps_executed` before operations that can raise. Failed steps remain counted, while `AsyncEngineCore` increments only after `scheduler.step()` returns successfully. Please move the increment to the successful return and add an exception-path regression test.

2. Issue #746 requested deterministic no-MLX coverage. The new scheduler and promotion test module skips entirely without MLX and runs only in the Apple Silicon job job. The added metrics assertion in `tests/test_metrics.py` is not selected by any CI job. Please make these contracts run in the Linux matrix, using stubs or MLX-neutral tests, while retaining Apple coverage.

djacobsmeyer:
Thanks for the review -- both confirmed and fixed, pushed to `fix/mllm-steps-executed-stat`.

## 1. Counter placement (`19fa986`)

Confirmed: `_steps_executed += 1` was right after `output = MLLMSchedulerOutput()`, before `process_pending_removals()`, `_schedule_waiting()`, `batch_generator.next()`, and `_process_batch_responses()` -- all of which can raise (and do, via the `_fail_requests_after_step_error` path). Moved it to immediately before `step()`'s single `return output`, matching `AsyncEngineCore`'s `output = self.scheduler.step(); self._steps_executed += 1` placement (increment after the risky call returns, not before).

Exception-path regression test added in `e6c81fe`: `test_step_that_raises_is_not_counted` makes `step()` fail via a `batch_generator.next()` that raises, asserts `_steps_executed` stayed at 0, then confirms a subsequent successful step still counts normally. Reverting the source fix locally makes this test fail (`assert 1 == 0`) -- checked before committing.

## 2. Linux-matrix coverage (`e6c81fe`)

Confirmed both claims by reading `.github/workflows/ci.yml` directly (not assuming): `test_mllm_steps_executed_stat.py` was only in `test-apple-silicon`'s list, and `test_metrics.py` wasn't in *any* job's pytest invocation on either platform -- its `pytestmark = pytest.mark.skipif(sys.platform != "darwin" or platform.machine() != "arm64", ...)` was gating a file that (once I checked) never actually touches a real `mlx.core` operation; it only needs `mlx.core` to be *importable* so `vllm_mlx.server`'s module-level `from .engine import BatchedEngine` chain succeeds.

Added `tests/_mlx_stub.py`: installs a `MagicMock`-backed stub for the ~8 `mlx`/`mlx_lm` submodules that import chain needs, only when the real package isn't importable (a no-op on the Apple job or any dev machine with mlx installed). Both files call it before their `vllm_mlx` imports; `test_metrics.py`'s Apple-only skip marker is removed.

One thing this surfaced that I want to flag explicitly rather than let it be a surprise in review: my first attempt put the stub in `conftest.py`, which seemed cleanest -- but `conftest.py` loads unconditionally for *any* invocation touching `tests/`, so the stub leaked into every file in the same pytest process. That broke `test_memory_cache.py` and `test_prefix_cache.py`, which have their own working `except ImportError` mlx-optional fallback: `isinstance(x, mock_attr)` raises `TypeError` instead of behaving like "mlx absent," so their existing fallback logic silently misfired (25 failures on a full test-matrix-file-list run under a simulated no-mlx environment). Caught this locally before pushing by actually running the existing test-matrix file list under a blocked-mlx-import simulation, not just checking that the two new/touched files pass in isolation.

Fixed by keeping the stub out of `conftest.py` entirely and giving the two files their own isolated CI step instead of merging them into the existing "no MLX required" step:

- **`test-matrix`** (Linux, Python 3.10-3.13): new step **"Run MLX-stubbed metrics/scheduler tests"**, `pytest tests/test_metrics.py tests/test_mllm_steps_executed_stat.py -v --tb=short` -- its own process, separate from "Run unit tests (no MLX required)" so there's no `sys.modules` cross-contamination with files that have their own real mlx-optional handling.
- **`test-apple-silicon`**: added `tests/test_metrics.py` to the existing "Run MLX-dependent tests" step's list (`tests/test_mllm_steps_executed_stat.py` was already there). The stub is inert here since real `mlx.core` imports successfully first.

Local verification before pushing:
- Real mlx (dev machine): full suite `3020 passed, 25 skipped, 27 deselected`, 0 regressions.
- Simulated no-mlx (blocked `mlx`/`mlx_lm` imports, matching a genuine Linux runner): the two stubbed files via plain `pytest tests/test_metrics.py tests/test_mllm_steps_executed_stat.py` -- `10/10 passed`. The pre-existing (untouched) test-matrix file list, run separately under the same simulation -- `1393 passed`, matching its pre-change behavior exactly.
- `ruff check vllm_mlx/ tests/ --select E,F,W --ignore E402,E501,E731,F811,F841` and `black --check vllm_mlx/ tests/` both clean, matching the `lint` job.

Full details and reasoning for the conftest.py-vs-isolated-step decision are in `tests/_mlx_stub.py`'s docstring and the `e6c81fe` commit message.

Thump604:
Thanks for the exact follow-up. I checked head `e6c81fed` and run 33890840271. The counter-placement correction and exception-path test address the first request, and all five new scheduler/promotion cases pass in the Linux step.

The remaining red checks are narrower than the source change: every Linux matrix job reaches the new isolated step, then all five existing `test_metrics.py` cases error at import because that environment does not install `uvicorn`. The Apple jobs are green.

Please fix only that CI contract. The smallest option is to keep the new MLLM scheduler/promotion file in the Linux step and move the one added gauge-mapping assertion into a dependency-light test path, rather than pulling the whole server-oriented metrics module and more server dependencies into the no-MLX job. An explicit install of the already-declared test dependency is also acceptable if that is the maintainer's intended CI shape. No source redesign, model rerun, or additional broad suite is requested. Once the exact head is green, I can close my review.

djacobsmeyer:
Fixed, pushed as `ed8f2ca`. You were right that `_mlx_stub.py` only ever covered the mlx/mlx_lm gap -- it never touched the missing `uvicorn`, which is what actually broke `test_metrics.py`'s fixture setup.

Went with the smallest option: moved the one gauge-mapping assertion, dropped `test_metrics.py` from the Linux step.

**What runs where now:**
- `test-matrix` (Linux, 3.10-3.13), new step "Run MLX-stubbed scheduler tests": `pytest tests/test_mllm_steps_executed_stat.py` only. `test_metrics.py` is no longer in this job at all.
- `test-apple-silicon`: unchanged from the previous push -- `test_metrics.py` and `test_mllm_steps_executed_stat.py` both still in the existing "Run MLX-dependent tests" step's file list, real mlx/uvicorn/prometheus-client all present.
- `test_metrics.py` itself: back to exactly its pre-PR-749 state (Apple-Silicon-only `pytestmark`, no `_mlx_stub` import) plus the one assertion this PR originally added, which stays as real HTTP-layer coverage on Apple.
- The contract that specifically needed Linux coverage -- `get_stats()["steps_executed"]` reaching the `vllm_mlx_engine_steps_executed` gauge -- now has a dependency-light equivalent: `TestMetricsEngineStepsExecutedGauge` in `test_mllm_steps_executed_stat.py`, which calls `MetricsCollector._update_engine_gauges()` directly against a `defaultdict(MagicMock)` standing in for the real Prometheus registry. `vllm_mlx.metrics` has zero mlx dependency, so this needs neither the mlx stub nor a real `prometheus_client`/`uvicorn`.

Also worth flagging on myself: my local verification for the previous push blocked mlx/mlx_lm to simulate "no MLX," but ran it inside this repo's own dev venv, which already has uvicorn and prometheus-client installed -- exactly the gap that masked this from me and let it reach your CI run instead. Re-verified this time by also blocking uvicorn and prometheus_client locally (not fastapi/pydantic, which test-matrix's dependency list does install) before running pytest for real: the isolated file gets 6/6, and the full pre-existing test-matrix list (run separately, untouched by this PR) still gets 1393, unchanged.

CI is running now on `ed8f2ca` -- will confirm here once it's green.

djacobsmeyer:
CI is green on `ed8f2ca` (run [33983587011](https://github.com/waybarrios/vllm-mlx/actions/runs/33983587011)) -- all jobs passed: `lint`, `type-check`, `test-matrix` (3.10/3.11/3.12/3.13, both the existing no-MLX step and the new MLX-stubbed scheduler step), `test-python-3-14`, `test-apple-silicon` (3.11/3.13), and the aggregate `tests` gate.

Thump604:
I checked ed8f2ca and run 33983587011. The six scheduler/gauge cases execute in the Linux matrix, and the Apple jobs cover both the scheduler tests and the HTTP metrics test. All 10 checks are green. This addresses my missing-uvicorn/CI follow-up; I am not asking for another model run for this CI-only change.

One documentation correction remains: the PR body still describes a top-of-step unconditional increment and Apple-only coverage. Please update those to the successful-return counter and current CI mapping, and identify the earlier test/hardware results by their original head rather than calling them this head. That will keep the evidence clear for Wayner’s final review.

djacobsmeyer:
PR body updated per your 2026-09-05T20:35:59Z comment: the "Changes"/"Tests" sections now describe the current successful-return counter placement and the current CI mapping (six cases in the Linux matrix via the dependency-light test, Apple jobs covering both the scheduler tests and `test_metrics.py`'s HTTP metrics test), and the original "Test results"/"Apple Silicon evidence" sections are now explicitly labeled as gathered against the original head `52c947a` rather than implying current head. No code changes.

waybarrios (approved):
The earlier review requests are addressed in ed8f2ca: the counter increments only after successful steps, Linux runs the six scheduler/gauge cases without server-only dependencies, and the PR description distinguishes original-head hardware evidence from later changes.

Fresh local verification: six tests passed, also with MLX and server-only dependencies blocked. Baseline and early-counter mutations triggered the expected regression failures. Changed-file Ruff and Black passed. The September 5 CI run passed all 10 checks on the generated merge revision containing this head.

No new actionable findings. Approved.

PR #751, opened 2026-08-31
chore: remove dead _extract_media_from_messages helper

djacobsmeyer:
## Problem

`vllm_mlx/engine/batched.py` has a complete OpenAI-message media-extraction helper, `_extract_media_from_messages` (images/videos/audio, ~74 lines), with zero call sites anywhere in `vllm_mlx/` or `tests/`:

```
$ git grep -n "_extract_media_from_messages"
vllm_mlx/engine/batched.py:69:def _extract_media_from_messages(messages: list[dict[str, Any]]) -> tuple:
```

(only its own definition — no callers). It's superseded: `BatchedEngine` actually extracts media via `extract_multimodal_content(messages)`, called at `batched.py:1035` and `:1182` (imported from `..api.utils`). This reads as a leftover from an earlier implementation, not a seam that should be wired up.

Surfaced via a structural repomap-go audit (`find_dead_code`) looking for other instances of the "one path instrumented, sibling silently isn't" pattern from #732/#738/#746 — this particular finding is unrelated to that pattern (it's plain dead code), but came out of the same pass.

## Changes

- Delete `_extract_media_from_messages` from `vllm_mlx/engine/batched.py`. No other files reference it, so no other changes are needed.

## Test results

Verified on Apple M5 Max, 128 GB, macOS 27.0, Python 3.12.13, mlx 0.32.2 / mlx-lm 0.31.3 / mlx-vlm 0.6.15, branched from `origin/main` @ `22efb47968f96e6e7a79704f4eb0184cda4e0068`.

- `pytest tests/` before deletion (baseline): `2965 passed, 13 skipped, 27 deselected in 32.19s`
- `pytest tests/` after deletion: `2965 passed, 13 skipped, 27 deselected in 26.24s`
- Identical pass/skip/deselect counts — no behavior change, as expected for removing unreferenced code.
- `ruff check vllm_mlx/ tests/ --select E,F,W --ignore E402,E501,E731,F811,F841`: clean
- `black --check vllm_mlx/ tests/`: `249 files would be left unchanged`


waybarrios (approved):
Clean and nice catch!

PR #752, opened 2026-08-31
fix(mllm): load registered MTP drafters safely

Thump604:
## Summary

- resolve local paths and Hugging Face repo IDs through mlx-vlm before reading drafter metadata
- normalize top-level and nested drafter architecture declarations
- preserve the established Gemma assistant loader while routing registered MTP architectures through mlx-vlm
- validate loader return values and target compatibility without masking architecture-specific errors

Unknown or malformed drafter configurations now fail before serving starts.

## Validation

- focused loader tests: `30 passed`
- full non-slow/non-integration suite: `2937 passed, 17 skipped, 69 deselected`
- Ruff and touched-file Black checks passed
- composed Qwen3.8 Flash-Next serving path passed 64K continuous batching and 128K MTP qualification

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,6 +170,7 @@ jobs:
             tests/test_platform.py \
             tests/test_llm.py \
             tests/test_mllm.py \
+            tests/test_mllm_assistant_drafter.py \
             tests/test_server.py \
             tests/test_paged_cache.py \
             tests/test_prefix_cache_untrimmable.py \
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -33,7 +33,7 @@ vllm-mlx serve --models-config <yaml> [options]
 | `--timeout` | Request timeout in seconds | 300 |
 | `--enable-metrics` | Expose Prometheus metrics on `/metrics` | False |
 | `--continuous-batching` | Enable batching for multi-user | False |
-| `--mllm-draft-model` | Configured mlx-vlm assistant/draft model path. Requires `--mllm`. | None |
+| `--mllm-draft-model` | Local path or Hugging Face repo ID for an mlx-vlm assistant/draft model. Requires `--mllm`. | None |
 | `--mllm-draft-kind` | Assistant drafter kind. Use `mtp` for continuous batching. | None |
 | `--mllm-draft-block-size` | Speculative block size passed to mlx-vlm. | None |
 | `--default-mllm-draft` | Enable the configured assistant by default. Requests can override with `mllm_draft`. | False |
--- a/tests/test_mllm_assistant_drafter.py
+++ b/tests/test_mllm_assistant_drafter.py
@@ -1,10 +1,173 @@
 # SPDX-License-Identifier: Apache-2.0
 """Tests for MLLM assistant-drafter speculative wiring."""
 
+import json
 import sys
 from types import SimpleNamespace
 
 import pytest
+
+
+def _write_drafter_config(tmp_path, config):
+    text = config if isinstance(config, str) else json.dumps(config)
+    (tmp_path / "config.json").write_text(text, encoding="utf-8")
+    return str(tmp_path)
+
+
+def _drafter_module(load_result, validate=lambda *args: None):
+    return SimpleNamespace(
+        load_drafter=lambda *args, **kwargs: load_result,
+        validate_drafter_compatibility=validate,
+    )
+
+
+def _install_fake_mlx_vlm(monkeypatch, drafter_module):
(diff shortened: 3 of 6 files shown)


janhilgard (approved):
Approving. Generalising the MTP drafter loader beyond the Gemma layout is the right move, and the config parsing is more careful than it needed to be — collecting `model_type`, `speculators_model_type` and `text_config.model_type`, then rejecting a set with more than one drafter-shaped entry, is a good way to fail loudly on a mislabelled checkpoint instead of silently picking one.

I verified the two mlx-vlm entry points this depends on exist in the version this repo floors at:

```
mlx-vlm 0.6.6
  mlx_vlm.speculative.drafters.load_drafter                 present
  mlx_vlm.speculative.drafters.validate_drafter_compatibility  present
```

and that `load_drafter` covers the registered architectures the new branch targets (`qwen3_5_mtp`, `deepseek_v4_mtp`, `eagle3`, `qwen3_dflash`, `gemma4_dflash`). The `ImportError` messages naming the resolved `model_type` are genuinely useful when a checkpoint needs a newer mlx-vlm than the box has.

Eight tests including an explicit no-network path for repo-ID resolution — that is the right coverage for this.

Two notes, neither blocking.

**1. `_resolve_mtp_drafter_path()` can start an unbounded download during server startup.**

```python
return Path(get_model_path(path_or_repo))
```

`mlx_vlm.utils.get_model_path()` falls through to `huggingface_hub.snapshot_download()` when the path does not exist locally, with no timeout, no size bound, and no progress surfaced through the server's logger. A drafter is loaded inside `MLXMultimodalLM.load()`, so a mistyped repo ID or a slow/flaky Hub turns into a server that appears hung at startup rather than one that fails. Under a supervisor with restart-on-failure that becomes a download-restart loop.

The feature itself is opt-in and the help text now documents it, so I do not think it blocks. But a log line before the call (`Downloading MTP drafter %s from the Hugging Face Hub`) would make the stall self-explanatory, and a `local_files_only` escape hatch would let operators who pre-stage weights guarantee no network at boot.

**2. Asymmetric `ImportError` handling for `validate_drafter_compatibility`.**

The Gemma branch swallows it (`except ImportError: pass`) to stay compatible with older mlx-vlm; the registered branch converts it to `MTPDrafterLoadError`. I understand the reasoning — the legacy path predates the validator — but the effect is that on an mlx-vlm without the validator, a Gemma drafter loads unvalidated while a registered one refuses. Since the registered branch has already imported `load_drafter` from the same module by that point, the validator is almost certainly present whenever `load_drafter` is; a short comment saying that would save the next reader the same trace I just did.

Method note: static review plus import-level inspection of the installed mlx-vlm. I have not loaded a drafter end to end — this box is memory-committed to another workload right now.

brandy975:
@Thump604 Small side note on testing: `tests/test_mllm_assistant_drafter.py` (which this PR extends) isn't registered in any job in `.github/workflows/ci.yml` — neither `test-matrix` nor `test-apple-silicon`. Or was that intentionally excluded?

Thump604:
Thanks for catching this. It was an omission, not an intentional exclusion. I added `tests/test_mllm_assistant_drafter.py` to the existing Apple Silicon job for Python 3.11 and 3.13 in 22b7698. The registration in the separate #755 was not sufficient for this PR’s own CI. This changes only the workflow test list, not the loader or its tests. All ten checks now pass, and I checked the logs: all 30 drafter tests executed and passed on both Apple jobs. Run: https://github.com/waybarrios/vllm-mlx/actions/runs/34177923864.

janhilgard (approved):
Re-approving. The only change since 2 Sep is `22b7698`, adding `tests/test_mllm_assistant_drafter.py` to the Apple Silicon list:

```yaml
tests/test_mllm.py \
tests/test_mllm_assistant_drafter.py \
tests/test_server.py \
```

Worth doing, and worth doing this way. I ran it exactly as CI invokes it — in the list, with `-m "not slow" -k "not Integration"` — and got 30 passed, so the filters do not silently deselect it the way `-k "not Integration"` deselected an endpoint file by name on #690.

This is the third time in a week a new test file turned out not to be executed by the green checks (mine on #690 and #767, this one). It is a repeatable footgun rather than three coincidences: adding a file to `tests/` looks like adding coverage, and CI reports success either way. A guard that fails when a `tests/*.py` file appears in neither the matrix nor the Apple list would catch it once instead of a reviewer catching it each time — worth a separate issue rather than anything here.

Everything else stands as reviewed on 2 Sep, including the note about `get_model_path()` reaching `snapshot_download()` at server startup for a repo-ID drafter.

Issue #378, opened 2026-04-20
feat: thinking-aware constrained decoding (lm-format-enforcer + reasoning coexistence)


Thump604:
## Problem

Constrained decoding via lm-format-enforcer (#362) and thinking/reasoning mode were mutually exclusive. When both active, the enforcer's FSM saw `<think>` as the first token, rejected it (not valid JSON), got stuck, and disabled itself. The server worked around this by forcing `enable_thinking=false` when constrained decoding was active.

### lm-format-enforcer path: RESOLVED

Three merged PRs fix the lm-format-enforcer + thinking coexistence:

- **#436** — `ThinkingAwareLogitsProcessor` with 4-phase state machine (IDLE/THINKING/TRANSITIONING/CONTENT), `thinking_token_budget` enforcement, and inner processor delegation
- **#458** — Root cause fix for token duplication in `_step()` that caused the enforcer to crash, plus `_ThinkingAwareLogitsProcessor` wrapper for JSON-phase gating after `</think>`
- **#472** — Skip preamble text before activating JSON schema enforcer, with `_json_scan_offset` tracking so `{` inside reasoning is ignored

### Outlines path: BLOCKED on two upstream issues

1. **[dottxt-ai/outlines#1847](https://github.com/dottxt-ai/outlines/issues/1847)** — `TransformerTokenizer.__init__` reads `eos_token_id`, `eos_token`, `all_special_tokens` from the raw `tokenizers.Tokenizer` backend, but modern `transformers` (tokenizers 0.25+) only exposes these on `PreTrainedTokenizer`. Outlines maintainer (RobinPicard) has opened a fix PR.

2. **[ml-explore/mlx-lm#1175](https://github.com/ml-explore/mlx-lm/issues/1175)** — MTP `_step_mtp` calls `_process_and_sample(prev_tokens, mtp_logits)` without the backbone's just-sampled token in `prev_tokens`. Stateful processors (Outlines Guide FSM) advance by the wrong token and enter invalid state. Blocks Outlines structured output when MTP is enabled.

## Acceptance criteria

This issue is done when:
- [x] lm-format-enforcer + thinking mode coexist without forcing `enable_thinking=false`
- [ ] Outlines + thinking mode coexist (blocked on dottxt-ai/outlines#1847)
- [ ] Outlines + MTP coexist (blocked on ml-explore/mlx-lm#1175)

janhilgard:
This is resolved by two merged PRs:

[...]

PR #753, opened 2026-08-31
fix(api): reject unavailable Outlines backend

Thump604:
## Summary

- reject explicit Outlines selectors with an exact 422 before engine acquisition
- cover both Chat Completions and Responses, streaming and non-streaming
- keep unrelated schema values untouched
- preserve the defensive internal rejection if a request bypasses raw-body preflight

This makes the current unsupported state explicit and fail-closed. Issue #378 remains open; this makes no positive-support claim.

## Validation

- focused structured-output and route tests: `150 passed, 2 known skips`
- full non-slow/non-integration suite: `2942 passed, 17 skipped, 69 deselected`
- Ruff and touched-file Black checks passed
- live Chat and Responses requests returned the exact 422 without changing inference counters


janhilgard (approved):
Approving. The detection is correctly scoped and the fail-closed behaviour is right — silently ignoring an explicit `outlines` selector and answering with a different backend is worse than a 422.

Things I checked:

- **Double body read.** `_preflight_response_format_backend()` calls `await raw_request.json()` after FastAPI has already parsed the body into the model. Starlette caches both the raw bytes and the decoded object (`Request.json()` memoises into `self._json`), and FastAPI's own JSON body handling goes through `request.json()`, so this is a dict lookup, not a second parse of a large prompt payload. No hot-path cost.
- **Why the raw payload at all.** `ChatCompletionRequest` sets no `model_config`, so Pydantic's default `extra="ignore"` drops `guided_decoding_backend` and `structured_outputs` before the handler sees them. Reading the raw payload for those two namespaces is the only way to see an explicit selector, and the docstring says so. Agreed with the approach — and I like that it inspects named selector paths rather than recursively searching the request, which would have made schema fields containing the string `"outlines"` trip it.
- **`request.text.format.type` in `create_response()`.** Safe: `ResponsesRequest.text` is `ResponseTextConfig = Field(default_factory=...)` and `ResponseTextConfig.format` is `ResponseTextFormat = Field(default_factory=...)`, so both are always present, and an explicit `"text": null` fails model validation with a 422 before the handler runs. No `AttributeError` path.
- **Case and whitespace handling** via `.strip().casefold()` is consistent across all the selector sites.

One suggestion, non-blocking: `_OUTLINES_BACKEND_UNAVAILABLE_DETAIL` tells the caller what is not available but not what to do. Since `#378` is tracking whether an Outlines backend comes back at all, a detail like `... ; use the default constrained-decoding backend by omitting the selector` would save a round trip for anyone porting a vLLM config across.

The `_preflight_response_format_backend()` call in `create_response()` also inspects chat-style keys (`response_format`, `guided_decoding`) on a Responses payload, where the format lives under `text.format`. Harmless — the explicit check above it covers the real path, and a top-level `guided_decoding_backend` is worth catching on both endpoints — just noting it so it does not read as an oversight later.

Method note: static review plus checking the request-model definitions. I have not exercised the endpoints against a running server — this box is memory-committed to another workload right now.

PR #754, opened 2026-08-31
fix(mllm): reconcile stateful MTP batch lifecycles

Thump604:
## Summary
- Reconcile stateful external drafters after verified tokens and keep row ownership aligned through lifecycle callbacks.
- Honor explicit lifecycle declarations; use the released callable hook pair when the declaration is absent or `None`. This selects reconciliation behavior, not numerical batching support.
- Avoid a second target forward when hidden states are unavailable: stateless fallback samples the existing logits; stateful drafting fails instead of replaying an advanced cache.

## Failure boundary
The MLLM scheduler catches verification errors and attempts to fail every tracked request, including waiting requests. Cleanup can itself raise and end that processing task. This PR does not guarantee general failure atomicity or recovery; a process restart or model reload is not established by these tests.

## Verification
- Local affected suite: 93 passed; four existing model-required tests excluded by the repository defaults. Python 3.12, mlx-vlm 0.7.0rc0; source-bound execution with no weights loaded.
- Tests cover lifecycle selection using released drafter attribute shapes, reconciliation/row filtering, single-forward fallback, and the actual async loop with normal and failing cleanup.
- All 10 GitHub checks pass on `fcd730e`, including both Apple Silicon jobs. No new hardware qualification claimed.

### DIFF
--- a/tests/test_mllm_continuous_batching.py
+++ b/tests/test_mllm_continuous_batching.py
@@ -2394,7 +2394,7 @@ class TestMLLMBatchGeneratorMTPGuards:
         ):
             install_mtp_mllm(generator, object(), draft_model=drafter)
 
-    @pytest.mark.parametrize("marker", [None, "false", 1])
+    @pytest.mark.parametrize("marker", ["false", 1])
     def test_external_mtp_reconciliation_marker_fails_closed(self, marker):
         from vllm_mlx.mllm_batch_generator import install_mtp_mllm
 
@@ -2408,6 +2408,38 @@ class TestMLLMBatchGeneratorMTPGuards:
 
         with pytest.raises(TypeError, match="literal boolean"):
             install_mtp_mllm(generator, object(), draft_model=drafter)
+
+    @pytest.mark.parametrize("declared", ["absent", None, True, False])
+    @pytest.mark.parametrize("hooks", ["absent", "callable", "partial", "noncallable"])
+    def test_external_mtp_selects_released_lifecycle_contract(self, declared, hooks):
+        from vllm_mlx.mllm_batch_generator import install_mtp_mllm
+
+        # v0.6.5/v0.6.6 Gemma has neither hook; Qwen3.5/DeepSeek/Eagle
+        # have both. This is an attribute/lifecycle fixture, not weight loading.
+        drafter = SimpleNamespace(reset=MagicMock())
+        if declared != "absent":
+            drafter.requires_verified_token_reconciliation = declared
+        if hooks != "absent":
+            drafter.accept_verified_tokens_batch = MagicMock()
+        if hooks == "callable":
+            drafter.filter_batch = MagicMock()
+        elif hooks == "noncallable":
+            drafter.filter_batch = None
+        batch = SimpleNamespace(_row_filter_hook=None)
+        generator = SimpleNamespace(
+            model=object(), active_batch=batch, _step=MagicMock(), _next=MagicMock()
+        )
+        invalid = declared is True and hooks != "callable"
+        invalid |= declared in ("absent", None) and hooks in ("partial", "noncallable")
+        if invalid:
+            with pytest.raises(TypeError, match="lifecycle hook"):
+                install_mtp_mllm(generator, object(), draft_model=drafter)
+            drafter.reset.assert_not_called()
+        else:
+            install_mtp_mllm(generator, object(), draft_model=drafter)
+            drafter.reset.assert_called_once_with(generator.model)
+            stateful = declared is not False and hooks == "callable"
+            assert callable(batch._row_filter_hook) is stateful
 
     def test_stateful_external_mtp_filter_failure_clears_lifecycle(self):
         from vllm_mlx.mllm_batch_generator import install_mtp_mllm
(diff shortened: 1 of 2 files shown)


janhilgard (changes requested):
The reconciliation design is sound and the row-lifecycle ownership is the right shape — attaching `_row_filter_hook` on every `_step` rather than only at install time is what makes it work for batches created after installation, and routing `filter()`/`notify_all_rows_removed()` through one owner is much better than the previous implicit coupling.

Two things before this can land.

## 1. The capability gate makes every shipped drafter fail at install

```python
reconciliation_marker = getattr(draft_model, "requires_verified_token_reconciliation", None)
if external_drafter and type(reconciliation_marker) is not bool:
    raise TypeError(
        "Assistant drafter must declare "
        "requires_verified_token_reconciliation as a literal boolean"
    )
```

No drafter in mlx-vlm 0.6.6 declares that attribute. I checked all seven registered drafter modules:

| drafter | `requires_verified_token_reconciliation` | `accept_verified_tokens_batch` | `filter_batch` |
|---|---|---|---|
| `gemma4_assistant` | absent | absent | absent |
| `gemma4_unified_assistant` | absent | absent | absent |
| `gemma4_dflash` | absent | absent | absent |
| `qwen3_dflash` | absent | absent | absent |
| `qwen3_5_mtp` | absent | **present** | **present** |
| `deepseek_v4_mtp` | absent | **present** | **present** |
| `eagle3` | absent | **present** | **present** |

So `type(None) is not bool` holds for all of them and `install_mtp_mllm()` raises `TypeError` before any MTP work happens. This is stricter than #755 — it fires for *any* external drafter, stateful or not, so `--mllm-draft-model ... --mllm-draft-kind mtp` cannot start at all. And the drafters this PR is actually written for (`qwen3_5_mtp`, `deepseek_v4_mtp`, `eagle3`) are exactly the ones it locks out, so the new reconciliation path is unreachable as written.

The table also points at the fix: **the two lifecycle hooks are already a reliable statefulness signal**, and this PR already computes them as `missing_hooks`. Deriving the flag rather than demanding a declaration needs no mlx-vlm change:

```python
declared = getattr(draft_model, "requires_verified_token_reconciliation", None)
if declared is not None and type(declared) is not bool:
    raise TypeError(...)                      # explicit but malformed -> still an error
stateful_external_drafter = (
    declared if isinstance(declared, bool)
    else all(callable(getattr(draft_model, n, None))
             for n in ("accept_verified_tokens_batch", "filter_batch"))
)
```

That keeps the explicit marker authoritative when a future mlx-vlm ships it, and works today. If you would rather keep the declaration mandatory, this PR needs to raise the `mlx-vlm` floor in `pyproject.toml` and land after the corresponding mlx-vlm release — right now the floor is untouched at `>=0.6.5`.

The same root cause affects #755 (`supports_continuous_batching`); I have left the detail there.

## 2. `raise RuntimeError` after `target_verify_started` — what catches it?

```python
if target_verify_started:
    raise RuntimeError(
        "MTP verification failed after the target cache may have "
        "advanced; aborting the batch instead of continuing with "
        "unverified cache state"
    ) from e
```

I agree completely with the reasoning — continuing against a possibly-advanced cache silently corrupts output, and that is worse than failing. My question is about blast radius: this propagates out of `_step`, so it takes down the whole batch, including the co-scheduled requests that had nothing to do with the failing row. Is there a handler that converts this into per-request failures, or does it reach the engine loop? If it reaches the engine loop under a restart-on-failure supervisor, one bad verification round costs every in-flight request plus a model reload.

If there is no such handler, it would be worth saying so explicitly in the message, or draining the other rows first.

## Smaller notes

- Moving `logits = logits[:, -1, :]` above the `hidden_states is None` check, and sampling from the already-computed logits via `_sample_primary_logits()` instead of falling back to `_orig_step()`, looks like a real fix beyond the stated scope: the old fallback re-ran the model over the same input against a cache that had already advanced. Worth calling out in the description — it is a correctness fix, not a refactor.
- The new `hidden_state_unavailable` bypass counter is a good addition; it makes that path visible in `/v1/status` instead of silently degrading.
- `MLLMBatch._row_filter_hook` as a leading-underscore dataclass field means it participates in the generated `__init__`. Harmless, but a `field(default=None, repr=False)` would keep it out of `repr()` output for a batch that can hold many rows.

I checked the un-notified `self.active_batch = None` at `mllm_batch_generator.py:1902` — it is guarded by `num_active == 0`, so `notify_all_rows_removed()` would be a no-op there and nothing leaks. No action needed; noting it so the next reader does not have to re-derive it.

Note on method: this is static analysis plus attribute inspection of the installed mlx-vlm 0.6.6. I could not run it end to end — this machine is memory-committed to another workload — so the acceptance/throughput behaviour of the reconciliation path is unverified by me.

Thump604:
Thanks, Jan. I’ve kept explicit lifecycle declarations authoritative and added the released hook-pair path when the declaration is absent or None. This identifies reconciliation work, not batching support. The MLLM loop catches verification errors and attempts to fail all tracked requests, including waiting requests. Cleanup can still fail and end that task, so I’ve removed the general failure-atomicity claim. The description also calls out the single-forward hidden-state fix. The affected suite passes 93 tests, including both cleanup outcomes; four model-required tests were excluded.

janhilgard (approved):
This resolves the blocker. Approving.

I re-ran the capability decision from this revision against every drafter registered in mlx-vlm 0.6.6 — the version this repo floors at — by executing the branch's own logic over the real classes:

| drafter | outcome |
|---|---|
| `DeepseekV4MTPDraftModel` | OK — stateful=True |
| `Eagle3DraftModel` | OK — stateful=True |
| `Qwen3_5MTPDraftModel` | OK — stateful=True |
| `Gemma4AssistantDraftModel` | OK — stateful=False |
| `DFlashDraftModel` | OK — stateful=False |
| `Gemma4DFlashDraftModel` | OK — stateful=False |

Zero `TypeError`s, and the three drafters this PR is written for now reach the reconciliation path instead of being locked out of it. Previously all six raised at `install_mtp_mllm()`, which made the feature unreachable and, under a restart-on-failure supervisor, a boot loop.

Deriving statefulness from the two lifecycle hooks when the marker is absent is the right shape, and the comment saying why — *"Released drafters predate the declaration. These hooks identify state reconciliation, not numerical continuous-batching support"* — is worth having, because the distinction is not obvious from the names.

The partial-hooks case is a good addition I had not thought of:

```python
elif not all(hook is absent for hook in hooks):
    raise TypeError(
        "Assistant drafter must provide both callable lifecycle hooks "
        "accept_verified_tokens_batch and filter_batch, or neither"
    )
```

A drafter with one hook is a genuine programming error rather than an older release, so failing there while accepting neither-present is the correct asymmetry. Keeping the explicit-marker `TypeError` for a malformed declaration is also right — that one can only be a mistake.

The rest of the PR stands as reviewed: `_row_filter_hook` re-attached each `_step` so batches created after installation are covered, `notify_all_rows_removed()` on the discard paths, and the `hidden_states is None` branch sampling from the already-computed logits instead of re-running `_orig_step()` against an advanced cache.

My open question about the blast radius of the post-`target_verify_started` `RuntimeError` is unchanged — it still propagates out of `_step` and takes the co-scheduled rows with it — but I do not think it blocks. Failing loudly beats continuing against a possibly-advanced cache, and if there is no handler upstream that is a separate discussion.

Method note: static review plus executing the decision logic against the installed mlx-vlm. I have not run the reconciliation path on a live model — this box is committed to another workload.

PR #755, opened 2026-08-31
Report declared drafter batch capability without changing startup

Thump604:
## Summary

- Report literal drafter capability as true/false and missing or invalid declarations as null.
- Preserve scheduler-reported capability instead of overwriting it with true.
- Leave startup admission unchanged, including released Gemma assistant drafters without a declaration.

## Verification

- At `c63e1f8`, 33 focused tests passed with no exclusions.
- The startup regression follows the real draft-loader routing boundary with a marker-absent return shape matching released Gemma. Weight loading is substituted; this is not numerical Gemma qualification.
- Tests retain explicit/unknown capability reporting and scheduler-precedence coverage. All 10 GitHub checks pass on `c63e1f8`, including both Apple Silicon jobs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,6 +170,7 @@ jobs:
             tests/test_platform.py \
             tests/test_llm.py \
             tests/test_mllm.py \
+            tests/test_mllm_assistant_drafter.py \
             tests/test_server.py \
             tests/test_paged_cache.py \
             tests/test_prefix_cache_untrimmable.py \
--- a/docs/guides/server.md
+++ b/docs/guides/server.md
@@ -543,6 +543,7 @@ Response fields:
 | `metal.peak_memory_gb` | Peak Metal GPU memory usage (GB) |
 | `metal.cache_memory_gb` | Metal cache memory usage (GB) |
 | `cache` | Cache statistics (type, entries, hit rate, memory usage) |
+| `mtp.continuous_batching_supported` | Drafter-declared batching capability: `true` (supported), `false` (unsupported), or `null` (undeclared or invalid/unknown). This reporting field does not block startup. |
 | `requests` | List of active requests with per-request details |
 
 Per-request fields in `requests`:
--- a/tests/test_batched_engine_mllm_config.py
+++ b/tests/test_batched_engine_mllm_config.py
@@ -1,6 +1,7 @@
 """BatchedEngine MLLM scheduler configuration wiring tests.
 
-These tests avoid model loading and MLX imports. They validate that CLI-level
+These tests avoid model loading; importing the real loader may import MLX.
+They validate that CLI-level
 SchedulerConfig fields survive the BatchedEngine -> MLLMSchedulerConfig bridge.
 """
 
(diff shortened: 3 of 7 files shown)


janhilgard (changes requested):
This fails closed on a capability that no drafter shipped by mlx-vlm declares, so it makes the supported configuration unstartable rather than safer.

`continuous_batching_capability()` only returns `True` when the drafter object carries a literal `supports_continuous_batching = True`. I enumerated every drafter registered in mlx-vlm 0.6.6 (the version this repo floors at `>=0.6.5`, unchanged by this PR):

| drafter | `supports_continuous_batching` |
|---|---|
| `gemma4_assistant.Gemma4AssistantDraftModel` | absent |
| `gemma4_unified_assistant.Gemma4AssistantDraftModel` | absent |
| `gemma4_dflash.Gemma4DFlashDraftModel` | absent |
| `qwen3_dflash.DFlashDraftModel` | absent |
| `qwen3_5_mtp.Qwen3_5MTPDraftModel` | absent |
| `deepseek_v4_mtp.DeepseekV4MTPDraftModel` | absent |
| `eagle3.Eagle3DraftModel` | absent |

None of them declares it. `_load_draft_model()` with `draft_kind == "mtp"` returns `Gemma4AssistantDraftModel`, so `continuous_batching_capability(loaded_drafter)` is `None` for every real drafter and `BatchedEngine` raises:

```
ValueError: Continuous batching requires an MLLM MTP drafter that
explicitly declares supports_continuous_batching=True
```

Confirmed on the installed runtime:

```python
>>> from mlx_vlm.speculative.drafters import gemma4_assistant as arch
>>> hasattr(arch.Model, "supports_continuous_batching")
False
```

That combination is not an unsupported accident — `cli.py` validates and permits it explicitly, rejecting only a non-mtp kind:

```python
if mllm_draft_model and args.continuous_batching and mllm_draft_kind != "mtp":
    print("Error: --mllm-draft-model with --continuous-batching "
          "requires --mllm-draft-kind mtp")
```

So after this PR, `--mllm --mllm-draft-model <path> --mllm-draft-kind mtp --continuous-batching` dies during engine startup. Under a launchd/systemd supervisor with restart-on-failure that is a restart loop, not a clean refusal.

**Suggested fix**, keeping the fail-closed intent:

1. Attach the declaration where the repo already knows what it loaded — `load_gemma4_assistant_drafter()` constructs the object, so it can set `supports_continuous_batching` on it; or
2. Treat *absent* as "legacy, assume today's behaviour" and reserve the hard failure for an explicit `False`; or
3. Land this together with the mlx-vlm change that adds the attribute and raise the `mlx-vlm` floor in `pyproject.toml` in the same PR.

Reporting `None` instead of a hardcoded `True` in `/v1/status` is a genuine improvement and I would keep that part as-is.

The tests are good — the `SimpleNamespace` parametrisation covers `True`/`False`/`None`/`"false"` cleanly. What they cannot catch is that no real drafter reaches the `True` branch, because every fixture constructs the attribute that production never has. A test that loads (or fakes at the `_load_draft_model()` boundary) the actual `Gemma4AssistantDraftModel` shape would have caught this.

Note: I could not exercise this against a live server — this box is memory-committed to another workload right now — so the above is static analysis plus direct attribute inspection of the installed mlx-vlm, not an end-to-end run.

Thump604:
Thanks, Jan. I’ve narrowed this to capability reporting and removed the new startup rejection. Released Gemma drafters without the declaration keep the existing startup path and report `null`, not `true`. The regression now follows the real draft-loader routing boundary with the released marker-absent shape; it does not load weights. All 33 focused tests pass. Admission policy and new capability declarations are separate work.

janhilgard (approved):
This is the right call. Approving.

The startup `ValueError` is gone and `scheduler_kwargs` is back to its original form, so the supported configuration — `--mllm --mllm-draft-model … --mllm-draft-kind mtp --continuous-batching`, which `cli.py` explicitly validates — starts again. What remains is the part that was always worth having:

```python
"continuous_batching_supported": batch_capability,
```

instead of a hardcoded `True`. `/v1/status` now reports `null` for a drafter that declares nothing, which is honest: no drafter shipped in mlx-vlm 0.6.6 carries `supports_continuous_batching`, so `True` was an assertion the code could not back. The new title matches what the diff does, which was not true of the previous revision.

`explicit_bool` rejecting anything that is not a literal `bool` also means a stray `1` or `"true"` from an MLLM's own stats dict reports as unknown rather than being silently coerced — the conservative direction for a capability flag.

One thing worth keeping in mind rather than changing here: a monitoring consumer doing `if stats["mtp"]["continuous_batching_supported"]:` treats `None` and `False` alike, so "unknown" and "declared unsupported" collapse for anyone not checking `is None`. That is inherent to a tri-state in JSON and I would not add a fourth field for it, but it is the kind of thing worth a line in the status docs if those get written.

The `SimpleNamespace` parametrisation over `True` / `False` / `None` / `"false"` covers the normaliser cleanly.

Method note: static review plus attribute inspection of the installed mlx-vlm 0.6.6; no live server run for this revision.

waybarrios:
@Thump604 This looks good overall. Could you please add `tests/test_mllm_assistant_drafter.py` to Apple Silicon CI and document that `mtp.continuous_batching_supported` may be `true`, `false`, or `null`, with `null` meaning undeclared or unknown and not blocking startup?

Thump604:
I have addressed both points in abb2086. The assistant-drafter test module now runs in both Apple Silicon CI jobs, and the server status table documents true / false / null, with null meaning unknown or undeclared and no change to startup behavior.

The module passed all 17 focused tests, and all 10 checks are green on this head: https://github.com/waybarrios/vllm-mlx/actions/runs/33990258562. This follow-up changes only CI selection and documentation.

Issue #756, opened 2026-08-31
[Bug]: per-request MTP accept/draft counters reach the response for MLLM requests but never for plain-text LLM requests


djacobsmeyer:
### Environment

- Chip: Apple M5 Max
- RAM: 128 GB
- macOS version: 27.0 (BuildVersion 26A5421a)
- Python version: 3.14.6 (host); 3.13.14 in the isolated venv used to verify this
- vllm-mlx version: `origin/main` @ `22efb47968f96e6e7a79704f4eb0184cda4e0068` — every line number below re-verified fresh against this exact commit via `git show origin/main:<file>` (not the local working tree, which has diverged)
- mlx / mlx-lm / mlx-vlm / mlx-audio versions: mlx==0.32.2, mlx-lm==0.31.3, mlx-vlm==0.6.17, mlx-audio==0.5.0

### Model and quantization

Not model-specific — this is a source-level gap in per-request MTP accounting. Reproducible for any model routed through the plain-text (non-MLLM) `Scheduler`/`AsyncEngineCore` path with a native MTP head and `--enable-mtp`. I did not run this live: `_install_mtp` on this path only fires when `model.mtp is not None` (`scheduler.py:1569`, cited below), and getting a real MTP-capable checkpoint requires the `add_mtp_weights_qwen35.py` injection script against a genuine trained base model, not a small public download — see "Additional context."

### Reproduction (source-level; not run live — see above)

```
vllm-mlx serve <mtp-capable-text-model> --continuous-batching --enable-mtp --port 8091
```

Send a completion/chat request, then inspect the per-response `usage` on the OpenAI-compatible response (or the raw `RequestOutput`/`generation_metadata` server-side).

### Expected behavior

For a request served on an MTP-enabled model, the response's MTP accounting should reflect the real per-request draft/accept activity for that specific request — the same as it already does on the MLLM/VLM path.

### Actual behavior

`RequestOutput.mtp_drafts`/`mtp_accepted` (`vllm_mlx/request.py`, `RequestOutput` dataclass, both default `0`) are never set to anything else on the plain-text LLM path, so `has_mtp_activity` is always `False` there.

**Root cause**, verified against `origin/main@22efb47`:

- `vllm_mlx/server.py` builds the client-facing MTP surface generically from whatever `RequestOutput`/similar object it's given:
  ```python
  mtp_drafts = getattr(output, "mtp_drafts", 0) or 0
[...]

PR #757, opened 2026-08-31
fix(mtp): attribute per-request draft/accept counts on the LLM path

djacobsmeyer:
Refs #756

## Problem

`RequestOutput.mtp_drafts`/`mtp_accepted` (consumed generically by `server.py`'s `has_mtp_activity`/usage builder) were wired end-to-end on the MLLM path (`mllm_scheduler.py`) but never set on the plain-text LLM path: `scheduler.py`'s `RequestOutput` construction never passed them, the shared `Request` class had no such fields at all, and the LLM path's own MTP loop (`_install_mtp`/`_mtp_step`/`_mtp_next`) tracked accept/attempt only as **batch-global** counters (`_mtp_stats`), with no per-UID attribution. Net effect: a client hitting the text-completion endpoint on an MTP-enabled model always got `has_mtp_activity == False`, while the MLLM/VLM endpoint reported real per-response numbers.

Went with option (a) from #756 (wire it up) rather than (b) (just document the gap) — the per-UID attribution below is genuinely useful (distinguishing "this request was part of the batch when MTP ran" from "it wasn't"), and mirrors an idiom (`_deferred_drafts`, keyed by UID) already present in the same function.

## Changes

- `vllm_mlx/request.py`: adds `Request.mtp_drafts`/`mtp_accepted` (default `0`), mirroring `MLLMRequest`.
- `vllm_mlx/scheduler.py`, `_install_mtp`:
  - Adds `_mtp_drafts_by_uid`/`_mtp_accepted_by_uid`, closure-scoped dicts keyed by UID the same way `_deferred_drafts` already is.
  - **Drafts**: incremented in `_mtp_step` for every UID in the active batch whenever an attempt happens, regardless of the eventual accept/reject outcome — mirrors `_mtp_stats["attempted"]`'s existing batch-wide semantics, just attributed per-UID instead of only summed globally.
  - **Accepted**: incremented in `_mtp_next`, only when a deferred draft is *actually emitted* as an output token — not merely predicted-accepted at verify time in `_mtp_step`. This mirrors the MLLM side precisely: `MLLMBatchResponse.from_draft` is likewise set at emission time, not at the verify step.
  - Exposed via `batch_gen.drain_mtp_uid_deltas()` (mirrors the existing `batch_gen.get_mtp_stats` pattern), which returns-and-clears the accumulated deltas. Drained exactly once per `Scheduler.step()` call so a later UID reuse (once a request finishes and its UID is recycled) can't inherit a finished request's stale counts.
  - `Scheduler._process_batch_responses` applies each UID's drained delta **at most once per call** — a UID can legitimately appear twice in one step's `responses` list (its own primary response, plus a deferred draft from the *previous* step emitted alongside it) — then copies the running cumulative totals onto the constructed `RequestOutput`, exactly like the MLLM side does.

### An honest caveat, called out in #756 too

The plain-LLM path's accept/reject decision is currently a **single batch-wide boolean per step** (`all_accepted = pred_list == draft_list`, a Python list-equality comparison across every sequence in the active batch), not a per-sequence verify. So "per-request attribution" here means *"was this UID part of the batch when the step's shared outcome was decided,"* not an independently-verified per-request result the way MLLM's genuinely-per-response `mtp_attempted`/`from_draft` flags are. That's still a real improvement (a request joining the batch late no longer inherits an earlier request's count, and gets `0` until it's actually participated), just a coarser signal than MLLM's. Documented in the code comments at both increment sites.

## Tests

New `tests/test_llm_mtp_per_request_counters.py`, mapped into the Apple-Silicon CI job only (100% MLX-gated — needs `mlx.core` to import `vllm_mlx.scheduler` — matching `test_mllm_continuous_batching.py`'s existing single-job precedent for a fully MLX-gated file):

- `TestInstallMtpPerUidDeltas` drives the *real* `_mtp_step`/`_mtp_next` closures (via `_install_mtp`) through a small deterministic 3-token-vocab fake model (no real weights — a hand-crafted logits spike controls accept vs. reject deterministically):
  - an attempt is recorded even when the verify step rejects;
  - an accept is recorded only once the deferred draft is actually emitted in the *next* step, not at verify time;
  - a UID joining mid-stream does not inherit an earlier UID's accepted-draft count, and only starts accumulating drafts from the step it actually joins.
- `TestProcessBatchResponsesAttributesMtpDeltas` verifies `Scheduler._process_batch_responses` attributes drained deltas to the correct `Request`/`RequestOutput` (two concurrent requests, distinct deltas, no cross-contamination), and that a scheduler with no MTP installed leaves counters at their dataclass default of `0`.

Verified these are real regression guards, not vacuously green: reverted both source changes locally and confirmed 4 of 5 new tests fail (`AttributeError: no attribute 'drain_mtp_uid_deltas'` / `'Request' object has no attribute 'mtp_drafts'`).

Also confirmed the consumer side needs no new test: `test_no_final_content_watchdog.py::test_generation_metadata_includes_mllm_mtp_counters_without_thinking_processor` (existing, unrelated to this PR) already proves `server.py`'s `_generation_metadata`/usage builder is generic and duck-typed — it was only ever missing real values to read from the LLM path, which is exactly what this PR supplies.

### Test results (isolated `uv venv`, this branch's head, no model download/serving)

- `tests/test_llm_mtp_per_request_counters.py`: 5/5 passed
- Mapped Apple-job subset (`test_qwen35_mtp_patch.py`, `test_llm_mtp_per_request_counters.py`, `test_batched_engine_mllm_config.py`): 22 passed
- Full suite vs. `origin/main` baseline: baseline 2965 passed/13 skipped/27 deselected → this branch 2970 passed/13 skipped/27 deselected (exactly the 5 new tests; no regressions)

## Live evidence: not possible within bounded effort — unit-level verification only

Investigated briefly per the mission's bound. `_install_mtp` on the plain-LLM path only fires when `model.mtp is not None` (`scheduler.py`) — a **native MTP head baked into the model's weights**. Per this repo's own `vllm_mlx/patches/qwen3_5_mtp.py` docstring, those weights are added via a separate `scripts/add_mtp_weights_qwen35.py` injection step against a genuine trained base checkpoint — they are not part of any stock public release I could find, and no small (<5GB) MTP-native text-only checkpoint is documented anywhere in this repo's README/docs. The only MTP weight-generation this project's own history (`[[qwen38-mtp-and-flag-tests]]`) has ever exercised targeted the *MLLM* checkpoint (`mlx-community/Qwen3.8-27B-8bit`, 849MB additive), not the plain-text path this PR touches. Producing a live run would require either downloading a large (multi-GB, likely >5GB) unquantized base model to run the injection script against, or locating a pre-built MTP-native text checkpoint that doesn't appear to be published — both out of scope for this bound.

**I am not claiming live evidence I don't have.** Verification here is unit-level only, driving the actual production `_mtp_step`/`_mtp_next` closures with a hand-crafted deterministic model (not a mock of my own new code — a fake of the *model* the real code calls), mirroring the semantics #749 established for the sibling MLLM-side fix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -175,6 +175,7 @@ jobs:
             tests/test_prefix_cache_untrimmable.py \
             tests/test_scheduler_max_kv_size.py \
             tests/test_qwen35_mtp_patch.py \
+            tests/test_llm_mtp_per_request_counters.py \
             tests/test_batched_engine_mllm_config.py \
             tests/test_mllm_continuous_batching.py \
             tests/test_mllm_cache.py \
--- a/tests/test_llm_mtp_per_request_counters.py
+++ b/tests/test_llm_mtp_per_request_counters.py
@@ -1,0 +1,253 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Deterministic coverage for per-request MTP counters on the plain-text
+LLM path (scheduler.py) -- the sibling gap to mllm_scheduler.py's
+mtp_drafts/mtp_accepted, which is already wired end-to-end.
+
+Covers:
+- _install_mtp's per-UID attempt/accept delta tracking, driven through the
+  real _mtp_step/_mtp_next closures with a small deterministic fake model
+  (3-token vocab, no real weights) -- including a UID joining mid-stream,
+  to prove drafts attributed to one request don't leak onto another.
+- Scheduler._process_batch_responses correctly attributes drained deltas
+  to the right Request and copies them onto RequestOutput, applying each
+  UID's delta at most once per call.
+"""
+
+from collections import namedtuple
+
+import pytest
+
+try:
+    import mlx.core as mx
+    from mlx_lm.sample_utils import make_sampler
+
+    HAS_MLX = True
+except ImportError:
+    HAS_MLX = False
+
+pytestmark = pytest.mark.skipif(not HAS_MLX, reason="MLX not available")
+
+
+_Response = namedtuple(
+    "_Response", ["uid", "token", "logprobs", "finish_reason", "cache_out"]
+)
+
+
+class _FakeActiveBatch:
+    def __init__(self, uids, cache):
+        self.uids = uids
+        self.cache = cache
+
+    def filter(self, keep_indices):
+        self.uids = [self.uids[i] for i in keep_indices]
+
+
+class _FakeMTPModel:
+    """Deterministic 3-token-vocab model: primary always predicts token 0,
+    the MTP head always drafts token 1. ``accept`` controls whether the
+    verify pass's own prediction agrees with the draft (accept) or not
+    (reject) -- vocab index 1 vs. index 0 at verify position 0.
+    """
+
+    def __init__(self, accept: bool):
+        self._accept = accept
+
+    def __call__(self, tokens, cache, return_hidden=True):
+        batch, seq = tokens.shape
+        if seq == 1:
+            logits = mx.array([[[10.0, 0.0, 0.0]]] * batch)
+            hidden = mx.zeros((batch, 1, 1))
+            return logits, hidden
+        # 2-token verify pass (primary, draft): position 0's argmax is
+        # compared against the draft token to decide accept/reject.
+        row0 = [0.0, 10.0, 0.0] if self._accept else [10.0, 0.0, 0.0]
+        verify_logits = mx.array([[row0, [10.0, 0.0, 0.0]]] * batch)
+        verify_hidden = mx.zeros((batch, 1, 1))
+        return verify_logits, verify_hidden
+
+    def mtp_forward(self, hidden, primary_tokens, mtp_cache=None):
+        batch = primary_tokens.shape[0]
+        return mx.array([[[0.0, 10.0, 0.0]]] * batch)
+
+
+class _FakeBatchGenerator:
+    """Minimal stand-in for mlx_lm's real BatchGenerator: just enough
+    surface for _install_mtp to patch. ``_next`` mimics the real one's
+    contract of advancing one primary token per active UID via
+    ``self._step`` -- which _install_mtp replaces with ``_mtp_step``.
+    """
+
+    Response = _Response
+
+    def __init__(self, uids, cache, stop_tokens=frozenset({1})):
+        self.active_batch = _FakeActiveBatch(list(uids), cache)
+        self.sampler = make_sampler(temp=0.0)
+        self.stop_tokens = stop_tokens
+
+    def _step(self, *args, **kwargs):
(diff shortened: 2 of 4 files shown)


Thump604 (approved):
I checked the exact `5f0de1d` source and the executed Apple Silicon CI log. The five new tests are mapped, collected, and passed; they exercise the real `_install_mtp` closures and scheduler attribution path. The counters are drained per scheduler response cycle, remain UID-scoped, and default to zero when MTP is absent. All required checks are green.

The PR is also appropriately explicit that plain-LLM verification still makes one batch-wide accept/reject decision. Within that existing implementation, these fields mean per-UID participation and emitted accepted drafts, not independent per-sequence verification. Keeping that limitation in the code and public contract is sufficient here; a live model run is not required to merge this observability wiring. Per-sequence verification would be a separate behavioral change.

waybarrios:
Pushed the follow-up fix in [c2600e0](https://github.com/waybarrios/vllm-mlx/commit/c2600e03abd0bb3ce7d543ab48361710f5d5b004). GitHub marks the commit as Verified.

The review found that the scheduler's per-request MTP counters were lost during output aggregation and in both plain-LLM engine adapters. A request with 2 drafts and 1 acceptance reached the API as 0 / 0, causing MTP metadata to be omitted. The original tests stopped before these conversion boundaries.

This follow-up preserves the latest cumulative `mtp_drafts` and `mtp_accepted` in `RequestOutputCollector._merge_outputs()` and forwards both fields through `BatchedEngine.generate()` and `stream_generate()`. It adds ten regression cases covering growing and repeated cumulative totals, zero acceptance, disabled MTP, and propagation through aggregation and both adapters into API metadata.

Validation:

- The local Linux CI test selection passed: 1,395 passed, 19 skipped, and 17 deselected.
- Focused checks passed again before committing: 54 passed, with 11 MLX-dependent cases skipped on Linux.
- All six local boundary cases passed using the real collector and adapters, import-only MLX placeholders, and the production API metadata helper extracted from source. This does not validate native MLX execution.
- Changed-file Ruff, Black, and whitespace checks passed. The patch still exactly matches the independently reviewed snapshot, which had no findings.
- Full-repository Ruff previously reported three confirmed baseline E721 violations outside this patch.

GitHub CI passed for this revision in both [run 35747804093](https://github.com/waybarrios/vllm-mlx/actions/runs/35747804093) and [run 35747806328](https://github.com/waybarrios/vllm-mlx/actions/runs/35747806328), including the Linux test matrix, lint/type checks, and Apple Silicon Python 3.11 and 3.13 jobs. The Apple Silicon logs confirm that all 11 tests in `test_llm_mtp_per_request_counters.py`, including the six new aggregation-to-API cases, passed on both Python versions.

Issue #759, opened 2026-09-01
[Bug]: MLLMScheduler.clear_runtime_caches()/.reset() reference nonexistent self.vision_cache (AttributeError)


djacobsmeyer:
### Environment

- Chip: Apple M5 Max
- RAM: 128 GB
- macOS version: 27.0 (BuildVersion 26A5421a)
- Python version: 3.14.6
- vllm-mlx version: `origin/main` @ `22efb47968f96e6e7a79704f4eb0184cda4e0068` — line numbers re-verified fresh against this exact commit via `git show origin/main:vllm_mlx/mllm_scheduler.py`
- mlx / mlx-lm / mlx-vlm / mlx-audio versions: mlx==0.32.2, mlx-lm==0.31.3, mlx-vlm==0.6.15

### Model and quantization

`mlx-community/Qwen3.8-27B-8bit` (MLLM/`BatchedEngine`-routed). Not model-specific — this is a source-level attribute bug on `MLLMScheduler` itself, reproducible for any MLLM-routed model.

### Reproduction

Reachable two ways:

1. Call `MLLMScheduler.clear_runtime_caches()` or `.reset()` directly (or via the server's `DELETE /v1/cache`, once #758 is fixed so that endpoint actually reaches a real engine — today it can't reach this code path at all in registry mode, which is how this was found: while verifying the fix for #758).
2. Directly, in-process:
   ```python
   from vllm_mlx.mllm_scheduler import MLLMScheduler
   # any live MLLMScheduler instance
   scheduler.clear_runtime_caches()
   ```

### Expected behavior

`clear_runtime_caches()` clears the vision embedding cache (if present) and the prefix cache, returning `{"vision_cache": bool, "prefix_cache": bool}`. `reset()` clears the same as part of a full scheduler reset.

### Actual behavior

```
AttributeError: 'MLLMScheduler' object has no attribute 'vision_cache'
```

Verified live: reproduced via the server's `DELETE /v1/cache` after locally patching the registry-mode engine-resolution gap from #758 —
```
{"status":"cleared","engine_cache":{"error":"'MLLMScheduler' object has no attribute 'vision_cache'"},"error":"Cache clear not available (mlx_vlm not loaded)"}
```
(caught and reported as a soft error by the caller in `server.py`'s `clear_cache()`, rather than propagating — so the HTTP response still says `"status":"cleared"` even though the exception fired and nothing was actually cleared for this engine.)

### Additional context

[...]

Issue #758, opened 2026-09-01
[Bug]: DELETE /v1/cache and DELETE /v1/cache/prefix silently no-op in registry mode (--models-config)


djacobsmeyer:
### Environment

- Chip: Apple M5 Max
- RAM: 128 GB
- macOS version: 27.0 (BuildVersion 26A5421a)
- Python version: 3.14.6
- vllm-mlx version: `origin/main` @ `22efb47968f96e6e7a79704f4eb0184cda4e0068` — every line number below re-verified fresh against this exact commit via `git show origin/main:<file>` (not the local working tree, which has diverged)
- mlx / mlx-lm / mlx-vlm / mlx-audio versions: mlx==0.32.2, mlx-lm==0.31.3, mlx-vlm==0.6.15

### Model and quantization

`mlx-community/Qwen3.8-27B-8bit`, single-entry registry. Not model-specific — same shape as #738/#739, #750: any model served via `--models-config` reproduces this, MLLM or plain-LLM.

### Reproduction

```
vllm-mlx serve --models-config models.yaml --enable-metrics \
  --continuous-batching --enable-prefix-cache --port 8091
```

```yaml
# models.yaml
manager:
  memory_budget_gb: 100
models:
  - name: mlx-community/Qwen3.8-27B-8bit
    path: mlx-community/Qwen3.8-27B-8bit
```

Warm the model and build up some prefix-cache entries (any prompt over `min_prefix_tokens`, 128 tokens by default), confirm the cache is non-empty, then try to clear it:

```
curl -s http://localhost:8091/metrics | grep vllm_mlx_cache_memory_bytes
# vllm_mlx_cache_memory_bytes 7.6913e+08   (nonzero, real entries cached)

curl -s -X DELETE http://localhost:8091/v1/cache/prefix
# {"status":"cleared","rewarm_scheduled":false}   <- looks successful

curl -s http://localhost:8091/metrics | grep vllm_mlx_cache_memory_bytes
# vllm_mlx_cache_memory_bytes 7.6913e+08   <- unchanged. Nothing was cleared.
```

`DELETE /v1/cache` reproduces the same way (checked separately below).

### Expected behavior

[...]

PR #760, opened 2026-09-01
fix(server): route DELETE /v1/cache(/prefix) through _model_manager in registry mode

djacobsmeyer:
Refs #758

## Scope

This PR makes `DELETE /v1/cache/prefix` and `DELETE /v1/cache` actually clear the resident engine's caches in registry mode (`--models-config`), instead of silently returning HTTP 200 while touching nothing.

**Not covered by this PR**: #759, a separate bug this PR's fix incidentally surfaces (`MLLMScheduler.clear_runtime_caches()`/`.reset()` reference a nonexistent `self.vision_cache` — the real attribute lives on `self.batch_generator.vision_cache`). Different root cause, kept out of this PR's diff; `clear_cache()` in this PR catches and reports it as a soft error (`engine_cache: {"error": "..."}`) rather than crashing the request, matching existing behavior for other `clear_runtime_caches()` failures.

**Prerequisite included, not yet upstream**: this branch is built on `origin/main`, which doesn't yet have `ModelManager.get_metrics_engine()` (#738/#739, open/unmerged as of this writing) — the same accessor this fix reuses. Commit 1 here (`dc77dc0`) is #739's exact fix, cherry-picked as a prerequisite so this PR is self-contained and mergeable independently. If #739 merges first, this branch will need a trivial rebase dropping the now-duplicate commit 1; the actual fix (commit 2) doesn't change either way.

## Summary

Both cache-clear endpoints read the module-level `_engine` global (`server.py:191`), which `load_model_registry()` (the `--models-config` path) never populates — it routes everything through `_model_manager` instead, exactly the split #738 already documented for `/metrics`. Verified live before this fix: `DELETE /v1/cache/prefix` against a resident model with ~769MB of real prefix-cache entries returned `{"status":"no_engine"}` and `vllm_mlx_cache_memory_bytes` stayed byte-for-byte unchanged.

## Changes

- `vllm_mlx/model_registry.py`: adds `ModelManager.get_metrics_engine()` (from #739, included here as a prerequisite — see Scope above).
- `vllm_mlx/server.py`:
  - `/metrics` resolves its engine via `_model_manager.get_metrics_engine() if _model_manager is not None else _engine` (from #739).
  - `clear_cache()` (`DELETE /v1/cache`) and `clear_prefix_cache()` (`DELETE /v1/cache/prefix`) now resolve the engine the same way, at the top of each handler, replacing every subsequent read of the bare `_engine` (including the auto-rewarm closure inside `clear_prefix_cache()`, which previously captured the stale global too).
- `tests/test_server_cache_controls.py`: two new regression tests, `test_clear_prefix_cache_routes_through_model_manager_in_registry_mode` and `test_clear_cache_routes_through_model_manager_in_registry_mode`, using the same `FakeModelManager` convention #739 established in `test_metrics.py`. Both assert the endpoint reaches the fake engine's clear method and returns a real `"cleared"` status with `server._engine` left `None` throughout — proving the fix routes through `_model_manager`, not a coincidental fallback.

## Type of change

- Bug fix

## Surface touched

- Model registry / multi-model manager
- Cache control endpoints (`DELETE /v1/cache`, `DELETE /v1/cache/prefix`)

## Test plan

```
pytest tests/test_server_cache_controls.py tests/test_metrics.py tests/test_model_registry.py -v
pytest tests/ -q
```

## Test results

```
$ pytest tests/test_server_cache_controls.py -v
4 passed in 2.37s
```

```
$ pytest tests/ -q
2961 passed, 25 skipped, 27 deselected in 27.14s
```
(`origin/main` baseline @ `22efb47`: 2953 passed per #739's own reported baseline, +6 from #739's cherry-picked commit, +2 new here = 2961; 0 regressed.)

## Live end-to-end verification

- Hardware: Apple M5 Max, 128GB unified memory
- Model: `mlx-community/Qwen3.8-27B-8bit` (MLLM-routed), single-entry registry, `--continuous-batching --enable-prefix-cache --enable-metrics`

Before (registry mode, 3 distinct >128-token prompts warmed into the prefix cache):
```
$ curl -s http://localhost:8091/metrics | grep vllm_mlx_cache_memory_bytes
vllm_mlx_cache_memory_bytes 7.69130496e+08

$ curl -s -X DELETE http://localhost:8091/v1/cache/prefix
{"status":"no_engine"}

$ curl -s http://localhost:8091/metrics | grep vllm_mlx_cache_memory_bytes
vllm_mlx_cache_memory_bytes 7.69130496e+08   # unchanged
```

After (same server/config, this patch applied, restarted, model re-warmed with the same 3 prompts):
```
$ curl -s -X DELETE http://localhost:8091/v1/cache/prefix
{"status":"cleared","rewarm_scheduled":false}

$ curl -s http://localhost:8091/metrics | grep -E "vllm_mlx_cache_memory_bytes|vllm_mlx_cache_entry_count"
vllm_mlx_cache_entry_count 0.0
vllm_mlx_cache_memory_bytes 0.0
```

Also verified `DELETE /v1/cache` now reaches a real engine too (previously always `engine_cache: null`); for this MLLM-routed model it now surfaces #759 instead (see Scope) rather than silently no-op'ing — a strictly more honest failure mode, and outside this PR's diff to fix.

## Risk notes

Scoped to two call-sites plus the shared accessor from #739. No behavior change for the non-registry serving path (`_model_manager is None`, both endpoints fall back to `_engine` exactly as before). No behavior change to `metrics.py`, `model_registry.py`'s existing methods, or request handling.

### DIFF
--- a/tests/test_metrics.py
+++ b/tests/test_metrics.py
@@ -153,6 +153,70 @@ class TestMetricsEndpoint:
 
         assert response.status_code == 404
 
+    def test_metrics_endpoint_reports_dead_gauges_without_engine_or_manager(
+        self, metrics_client
+    ):
+        client, server, collector = metrics_client
+
+        collector.configure(enabled=True)
+        assert server._engine is None
+        assert server._model_manager is None
+
+        response = client.get("/metrics")
+
+        assert response.status_code == 200
+        assert "vllm_mlx_model_loaded 0.0" in response.text
+        assert "vllm_mlx_scheduler_waiting_requests 0.0" in response.text
+
+    def test_metrics_endpoint_scrapes_registry_mode_engine(
+        self, metrics_client, monkeypatch
+    ):
+        client, server, collector = metrics_client
+
+        collector.configure(enabled=True)
+
+        class FakeModelManager:
+            def __init__(self, engine):
+                self._engine = engine
+
+            def get_metrics_engine(self):
+                return self._engine
+
+            async def shutdown(self):
+                return None
+
+        monkeypatch.setattr(server, "_model_manager", FakeModelManager(FakeEngine()))
+
+        response = client.get("/metrics")
+
+        assert response.status_code == 200
+        assert "vllm_mlx_model_loaded 1.0" in response.text
+        assert "vllm_mlx_scheduler_waiting_requests 2.0" in response.text
+        assert (
+            'vllm_mlx_cache_type{cache_type="memory_aware_cache"} 1.0' in response.text
+        )
+
+    def test_metrics_endpoint_registry_mode_dead_gauges_when_idle(
+        self, metrics_client, monkeypatch
+    ):
+        client, server, collector = metrics_client
+
+        collector.configure(enabled=True)
+
+        class FakeModelManager:
+            def get_metrics_engine(self):
+                return None
+
+            async def shutdown(self):
+                return None
+
+        monkeypatch.setattr(server, "_model_manager", FakeModelManager())
+
+        response = client.get("/metrics")
+
+        assert response.status_code == 200
+        assert "vllm_mlx_model_loaded 0.0" in response.text
+
     def test_metrics_endpoint_scrapes_runtime_stats(self, metrics_client, monkeypatch):
         client, server, collector = metrics_client
 
--- a/tests/test_model_registry.py
+++ b/tests/test_model_registry.py
@@ -339,6 +339,65 @@ def test_preempt_policy_cancels_active_request_and_loads_waiting_model(tmp_path)
     asyncio.run(_run())
 
 
+def test_get_metrics_engine_returns_none_when_idle(tmp_path):
+    registry = _registry(tmp_path, {"alpha": 4})
+    manager = ModelManager(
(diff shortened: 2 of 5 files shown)


Thump604 (changes requested):
I checked the current head and the #739 dependency. The single-model registry fix is the right direction, but one mutating-contract blocker remains.

`get_metrics_engine()` intentionally returns one MRU representative because `/metrics` exposes unlabeled scalar gauges. Reusing that observational accessor for `DELETE /v1/cache` (documented as clearing all caches) and `DELETE /v1/cache/prefix` silently clears only one engine when registry mode has multiple loaded models. The response still reports success while the other resident caches remain. The rewarm closure is likewise bound to only that representative.

Please put the mutation on a manager-owned API that either clears every loaded engine or accepts an explicit model target, and add one deterministic two-loaded-engine route test proving the selected contract. Keep #759 separate as proposed. #739 is now approved at `48735ed`, so please also restack/drop the duplicated prerequisite commit; the current head’s only red check is Black on `server.py` and should be cleared by the refreshed diff.

PR #763, opened 2026-09-04
feat(spec-decode): DSpark block speculative decoding for NemotronH (design + spike, refs #699)

CBribiescas:
This is the spike promised in #699 (steps 1 and 2 of the staged plan proposed there: an approach skeleton plus a flag-gated implementation validated against a correctness bar). It is a draft for design review, not a merge request yet.

## Headline

DSpark block speculative decoding for NemotronH (Nemotron 3.5 Lightning 30B-A3B) on the continuous-batching text path. It is correct and reaches NVIDIA's acceptance figure: 3.20 of 7 drafted tokens accepted per round live (NVIDIA reports 3.75; the gap is the metric, see the guide). It does not yet win: 0.83–0.87x plain greedy on code and 0.55–0.60x on prose at k=4 on this tree today (August spike: 87–90 vs 115–134 tok/s, 0.7–0.8x).

The bottleneck is measured, not guessed. A verify forward of width w costs 8.47 / 13.49 / 15.64 / 20.67 / 28.13 ms at w = 1 / 2 / 3 / 5 / 8. Past width 2 every extra verified token costs ~2.5–2.8 ms, about 30% of a full step, and that marginal is the per-token routed-expert read that the current `switch_layers` gather does not batch across positions. It caps any speculative speedup on this stack near 2x. Batching the expert gather across the verify block would put this same code at a projected ~180–230 tok/s, and would speed up plain MoE decode for every model. Full design, measurements, benchmark matrix and the "byte-identical" definition: `docs/guides/speculative-decoding.md`.

## Run data

(M5 Max 128 GB, Lightning 30B-A3B 6-bit, mlx-lm 0.31, single stream, greedy, `--disable-prefix-cache`, 2026-08-13 to 08-15; full write-up in `docs/guides/speculative-decoding.md`)

Verify-width curve — same sequence, warm ~500-token cache, `mx.eval`-timed, 30 iterations:

| verify width | ms / step | × width-1 | marginal ms / extra token |
|---|---|---|---|
| 1 | 8.47 | 1.00 | — |
| 2 | 13.49 | 1.59 | 5.0 |
| 3 | 15.64 | 1.85 | ~2.2 |
| 5 | 20.67 | 2.44 | ~2.5 |
| 8 | 28.13 | 3.32 | ~2.5 |

Throughput:

| config | tok/s |
|---|---|
| plain greedy | 115 whole-request / 134 steady-state |
| DSpark k=7, BF16 draft, full-vocab Markov (first working build) | 66 |
| DSpark k=4, 8-bit draft, gathered top-64 Markov, sliced lm_head | 90 |
| DSpark k=4 + adaptive block length (`margin_tau` 1.5) | 86.6 |
| Lightning's own MTP head, k=1, 74.5 % acceptance (for comparison) | 106.5 |

Acceptance — mean accepted draft tokens per round at k=7, SPEED-Bench real prompts (6 of its 11 categories ship as license placeholders), greedy exact prefix match:

| category | 6-bit target | BF16 target | NVIDIA (BF16, temp 1.0 rejection sampling) |
|---|---|---|---|
| coding | 4.40 | 4.11 | 4.38 |
| qa | 2.90 | 3.45 | 3.36 |
| rag | 4.57 | 5.24 | 4.25 |
| multilingual | 2.13 | 1.96 | 4.55 |
| writing | 2.04 | 1.92 | 2.83 |

The bit-exact native NVFP4 drafter and the BF16→8-bit chain give identical acceptance, so draft precision is eliminated as a cause of the residual gap; target precision is a ~4 % mixed-direction effect.

Per-round arithmetic at k=7: 28 ms verify + ~4 ms draft (6.1 ms before 8-bit quantization) for a mean of ~4.2 kept tokens ≈ 7.6 ms/token, before the partial-accept re-advance, which adds a second wide forward to ~85 % of rounds. Plain decode is 8.5 ms/token. Feeding the target's aux states into the draft KV costs ~0.6 ms/token.

### Room for improvement, by expected payoff

1. **Batch the routed-expert gather across verify positions** (mlx-lm `switch_layers` as used by `nemotron_h`). The 2.5–2.8 ms marginal is ~0.5 GB of expert reads per position (6 of 128 experts × 23 MoE layers) issued one position at a time. Batching them across the block drops the marginal toward the dense floor (~0.9 ms/token), at which point this exact stack projects to ~180–230 tok/s. It also speeds up plain MoE decode for every model here. This is the only item that flips the sign.
2. **The partial-accept re-advance.** ~85 % of rounds pay a second (m+2)-wide forward because Mamba state cannot be rewound. With that floor the ceiling is 1 / (r + (1 − r) / k), which is ≈ 1.15× at the measured r = 0.85 and k = 7 (corrected: an earlier revision said 1.7×, an arithmetic error). The single-forward commit path that would remove the re-advance does not exist for hybrid targets. An attention-only target keeps the accepted prefix in place and skips this cost entirely.
3. **Host syncs per round.** The round has three `mx.eval` points (primary, draft block, verify argmax). Not measured separately for DSpark; in the MTP port the single per-cycle sync accounted for ~0.2 forward-equivalents per cycle, so overlapping the draft with the verify launch is worth measuring.
4. **Adaptive block length** at `margin_tau` 1.5 measured 86.6 vs 90 at fixed k=4 — no gain at this marginal cost. Worth re-testing only after item 1 changes the cost per verified token.
5. **Prefix-cache hits disable speculation** because no aux states exist for cached tokens. Re-deriving them from the cached prefix with one extra forward would let agentic multi-turn traffic use the drafter at all.

### Clean run on this tree (2026-09-04, host otherwise idle)

Same build for every row; served with `--continuous-batching --disable-prefix-cache --max-num-seqs 1`, native NVFP4 drafter, greedy, thinking off, one warm-up then three 512-token completions per prompt; medians. M5 Max 128 GB, mlx-lm 0.31.3.

| config | code tok/s | vs plain | accepted / round (per token) | prose tok/s | vs plain | accepted / round (per token) |
|---|---|---|---|---|---|---|
| plain greedy | 125.6 | — | — | 118.8 | — | — |
| DSpark k=4 | 104.1 | 0.83× | 2.98 of 4 (74.6 %) | 65.2 | 0.55× | 1.51 of 4 (37.7 %) |
| DSpark k=7 | 76.8 | 0.61× | 3.59 of 7 (51.3 %) | 53.6 | 0.45× | 1.88 of 7 (26.9 %) |
| DSpark k=4, `--spec-draft-margin-tau 1.5` | 109.6 | 0.87× | 2.59 of 2.96 drafted (87.5 %) | 71.1 | 0.60× | 1.17 of 2.22 drafted (52.6 %) |

Counters over the whole run: `errors` 0, `context_disables` 0, `bounded_kv_disables` 0, `despeculations` = one per finished request (the rewind at `extract_cache`). Every config's three runs were byte-identical to each other.

Reading: the ratios match the August measurements (0.7–0.8× then, 0.83–0.87× now on code). Wider blocks lose: k=7 accepts more tokens per round (3.59) but the 8-wide verify costs more than they save. The adaptive cut-off is now the best configuration, because on this tree the cost per drafted position is what it trims. Prose is worse than code at every setting, as the acceptance column predicts.

**Fidelity.** Against the plain run, every DSpark config produced the same two divergences, one per prompt, and all three DSpark configs agree with each other:

| prompt | first divergence | plain chose | DSpark chose | plain-forward logits at that position | gap |
|---|---|---|---|---|---|
| code | token 50 of 512 | `\n\n` | `,` | 28.25 vs 28.25 | **0.000** (exact tie) |
| prose | token 2 of 512 | ` difference` | ` distinction` | 24.125 vs 24.25 | **0.125 = one bf16 ULP** at that magnitude |

The gaps were measured offline with the same weights (the server has no logprobs endpoint): a forward over the prompt plus the common prefix, top-3 logits at the divergence position. In the prose case the wide forward's own argmax is the DSpark token, i.e. the 1-wide run is the one that broke the tie the other way. Both divergences therefore satisfy the acceptance bar proposed below: every emitted token is an argmax of the target over the emitted prefix, and any difference from a 1-wide run sits at a position whose top-1/top-2 gap is within bf16 resolution.

Artifacts (server logs, per-run JSONL with the full completions, status snapshots, the driver and the probe) are kept locally and can be attached on request.

## What is in the PR

- `vllm_mlx/patches/nemotron_dspark.py` — the drafter (DFlash context KV from target aux states, parallel block draft, gathered top-64 Markov fix-up) and the aux-state taps on the NemotronH forward. Loads either an MLX-native NVFP4 repack or the BF16 dequant.
- `vllm_mlx/scheduler.py` — `_install_dspark` on `GenerationBatch._step` (mlx-lm >= 0.31): k-wide verify, longest-prefix accept, KV trim plus recurrent-state restore, re-advance of the accepted prefix plus the target's own corrected token. Config fields and `/v1/status` `spec_decode` counters. `_install_mtp` and the #737 guard are untouched.
- `vllm_mlx/spec_utils.py` — acceptance semantics and fp8-e4m3 / fp4-e2m1 decode tables, MLX-free.
- `scripts/prepare_dspark_weights.py`, `cli.py`, `engine/batched.py`, `server.py`, docs, README, CI test list.

Flags: `--spec-draft dspark`, `--spec-num-draft-tokens` (default 7), `--spec-draft-margin-tau` (adaptive block length). Requires `--continuous-batching`; rejected with `--mllm` or `--enable-mtp`.

Degradation: engages only for temperature-0 single-sequence requests. Multi-sequence steps, logits processors, prefix-cache hits (no aux states for cached tokens), a KV cache past `--max-kv-size`, a missing drafter, or any exception fall back to plain decode for that call or request, logged and counted. Because the cache runs ahead of the emitted tokens inside a round, the hook rewinds it to the emitted prefix before any batch merge, cache extraction (finish/abort) or error fallback, so stock code never sees verified-but-unemitted tokens.

## Tests

- `tests/test_dspark_spec_decode.py` (in CI): acceptance semantics, e4m3/e2m1 tables, config validation, greedy gate, no-op paths through the real `Scheduler._create_batch_generator`.
- `tests/test_dspark_scheduler_step.py`: the real mlx-lm `BatchGenerator` over a tiny random Mamba+attention+MLP NemotronH with scripted oracle / garbage / half-right drafters; the spec path must reproduce plain greedy (tie-tolerant per the definition below; the seed has no ties). Also covers a second request merging mid-round, an exception injected after the verify forward, and the cache handed back at finish, all on real hybrid caches.

## Not in this PR

NemotronH MTP head port, non-greedy (rejection-sampling) verification, re-deriving aux states from a prefix-cache hit, the batched expert-gather kernel.

## Status

Unit and e2e tests pass locally. The clean run above (this tree, idle host) covers load, engagement, throughput against plain on the same build, the counters, and the fidelity bar. Nothing is owed on the measurement side; the open items are the design questions below.

## Questions

1. Flag naming: `--spec-draft` / `--spec-num-draft-tokens` vs. folding into the `--mllm-draft-*` family.
2. Would you take the scheduler hook plus the drafter API (`block_size`, `append_context`, `draft`) behind the flag now, with the DSpark drafter itself as a follow-up, or hold everything until it wins?
3. Is batching the routed-expert gather across the verify block something the project wants?

https://claude.ai/code/session_01AQAzGiCRR1VocVNa3ngyjw

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,6 +102,7 @@ jobs:
             tests/test_qwen3_xml_registration.py \
             tests/test_qwen3_xml_streaming_chunks.py \
             tests/test_qwen35_mtp_hidden_state_mode.py \
+            tests/test_dspark_spec_decode.py \
             tests/test_streaming_fence_stripper.py \
             tests/test_streaming_think_router.py \
             tests/test_streaming_tool_filter.py \
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ claude
 ### Reasoning & advanced
 - **Reasoning extraction**: Qwen3, DeepSeek-R1, DeepSeek-V4 (`--reasoning-parser`)
 - **MoE expert reduction**: `--moe-top-k` for +7-16% on Qwen3-30B-A3B
-- **Speculative decoding**: `--mtp` for Qwen3-Next
+- **Speculative decoding**: `--mtp` for Qwen3-Next; `--spec-draft dspark` block drafting for Nemotron 3.5 Lightning ([design + measurements](docs/guides/speculative-decoding.md))
 - **Sparse prefill**: attention-based `--spec-prefill` for TTFT reduction
 
 ### Observability
--- a/docs/guides/speculative-decoding.md
+++ b/docs/guides/speculative-decoding.md
@@ -48,6 +48,30 @@ prefix. Against a 1-wide greedy run of the same prompt, 199 positions were compa
 divergence, at a position where the 1-wide run's top-1/top-2 logit gap was exactly 0.0000
 (a bf16 tie that a different verify width legitimately breaks the other way).
 
+## Clean run on the rebased tree (2026-09-04, host otherwise idle)
+
+Same build for every row; served with `--continuous-batching --disable-prefix-cache --max-num-seqs 1`, native NVFP4 drafter, greedy, thinking off, one warm-up then three 512-token completions per prompt; medians. M5 Max 128 GB, mlx-lm 0.31.3.
+
+| config | code tok/s | vs plain | accepted / round (per token) | prose tok/s | vs plain | accepted / round (per token) |
+|---|---|---|---|---|---|---|
+| plain greedy | 125.6 | — | — | 118.8 | — | — |
+| DSpark k=4 | 104.1 | 0.83× | 2.98 of 4 (74.6 %) | 65.2 | 0.55× | 1.51 of 4 (37.7 %) |
+| DSpark k=7 | 76.8 | 0.61× | 3.59 of 7 (51.3 %) | 53.6 | 0.45× | 1.88 of 7 (26.9 %) |
+| DSpark k=4, `--spec-draft-margin-tau 1.5` | 109.6 | 0.87× | 2.59 of 2.96 drafted (87.5 %) | 71.1 | 0.60× | 1.17 of 2.22 drafted (52.6 %) |
+
+Counters over the whole run: `errors` 0, `context_disables` 0, `bounded_kv_disables` 0, `despeculations` = one per finished request (the rewind at `extract_cache`). Every config's three runs were byte-identical to each other.
+
+Reading: the ratios match the August measurements (0.7–0.8× then, 0.83–0.87× now on code). Wider blocks lose: k=7 accepts more tokens per round (3.59) but the 8-wide verify costs more than they save. The adaptive cut-off is now the best configuration, because on this tree the cost per drafted position is what it trims. Prose is worse than code at every setting, as the acceptance column predicts.
+
+**Fidelity.** Against the plain run, every DSpark config produced the same two divergences, one per prompt, and all three DSpark configs agree with each other:
+
+| prompt | first divergence | plain chose | DSpark chose | plain-forward logits at that position | gap |
+|---|---|---|---|---|---|
+| code | token 50 of 512 | `\n\n` | `,` | 28.25 vs 28.25 | **0.000** (exact tie) |
+| prose | token 2 of 512 | ` difference` | ` distinction` | 24.125 vs 24.25 | **0.125 = one bf16 ULP** at that magnitude |
+
+The gaps were measured offline with the same weights (the server has no logprobs endpoint): a forward over the prompt plus the common prefix, top-3 logits at the divergence position. In the prose case the wide forward's own argmax is the DSpark token, i.e. the 1-wide run is the one that broke the tie the other way. Both divergences therefore satisfy the acceptance bar proposed in *Greedy versus non-greedy correctness*: every emitted token is an argmax of the target over the emitted prefix, and any difference from a 1-wide run sits at a position whose top-1/top-2 gap is within bf16 resolution.
+
 ## Why it does not win yet: the verify-width curve
 
 Speculative decoding pays for a `(k+1)`-wide verify forward instead of `k+1` single-token
@@ -267,7 +291,7 @@ is a follow-up; the verify forward already produces the full distributions it ne
 | `rounds` | draft/verify rounds attempted |
 | `drafted` / `accepted` | draft tokens proposed / accepted |
 | `acceptance_rate` | `accepted / drafted` (per-token) |
-| `mean_accept_len` | `accepted / rounds`; tokens per round is this + 1 (bonus or corrected token) |
+| `mean_accept_len` | `accepted / rounds`; a round also emits the target's own token after the accepted prefix, so tokens per round is this + 1 after a full block and this + 2 after a partial one (the correction plus the next sample) |
 | `rounds_full` | rounds where the entire block was accepted |
 | `context_disables` | requests where spec turned itself off at the context check |
 | `bounded_kv_disables` | requests where spec turned itself off because the KV cache was no longer trimmable |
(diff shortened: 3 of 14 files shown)


Thump604 (changes requested):
Thanks for the unusually clear spike and measurements. I reviewed exact head `b162745` against current `main`.

My recommendation is to hold this combined feature rather than merge the scheduler hook/API or drafter now:

- The measured product result is negative: the best code result is 0.87x plain decode, and prose is 0.45-0.60x. The 180-230 tok/s figure is a projection, not evidence from an implemented expert-gather path.
- The fallback contract is not yet fail-safe. In `_install_dspark`, both target verify and partial re-advance record `st["advanced"]` only after the model call returns. A later layer can mutate cache and raise before that assignment, leaving the exception path with zero recorded advancement. If rollback itself raises, that error is logged and suppressed and execution still continues through `_finish_step`. Plain decode must not continue unless rollback is proven complete; otherwise the request/generator should fail closed.
- The real hybrid scheduler regression `tests/test_dspark_scheduler_step.py` is not selected by the Apple workflow. The green jobs execute the MLX-free helper file, but the explicit Apple test list omits the scheduler file.

Answers to the design questions:

1. Defer flag naming until there is a qualifying implementation.
2. Hold the scheduler hook/API and drafter together until the path is at least non-inferior on the declared workloads.
3. Yes, the routed-expert path is the useful next experiment, but it should be an independent `mlx-lm` change. Prove exact output parity plus plain Nemotron decode and verify-width improvement there first. Only then restack this spike and remeasure DSpark.

I am not asking for another broad DSpark test or model campaign now. The smallest useful next step is the isolated expert-gather optimization; if it does not materially improve the measured width curve, this spike should remain design evidence rather than shipping code.

PR #764, opened 2026-09-04
fix(mllm): key the prefix cache on the message boundary, not the think suffix

CBribiescas:
## The bug

`_compute_think_suffix_len()` measures the generation-prompt suffix **once at startup**, using `enable_thinking=True`. But callers choose the thinking mode **per request** via `chat_template_kwargs`, and the two suffixes are different lengths:

| mode | generation prompt | tokens |
|---|---|---|
| `enable_thinking=True` | `<think>\n` | 2 |
| `enable_thinking=False` | `<think>\n\n</think>\n\n` | 4 |

So for a request rendered with `enable_thinking=False`, `input_ids_list[:-S]` strips 2 of 4 and leaves `<think>\n\n` on the stored key. That entry is 2 tokens longer than the prefix it shares with the next turn, so it falls out of the strict-PREFIX path into LCP — and LCP reuse is correctly refused for hybrid / linear-attention caches, because recurrent state cannot be rewound.

**The failure is silent.** No error, no warning — the prefix cache simply stops hitting for multi-turn clients, and every turn re-prefills the entire conversation.

## Reproduction (tokenizer only, no server)

On `Qwen3.6-35B-A3B`, comparing turn 1 against turn 2 of the same conversation:

```
_think_suffix_len (measured at startup, enable_thinking=True) = 2

enable_thinking=True
  turn-1 prompt         216 tokens
  shared with turn-2    214
  generation prompt     2 tokens  ['<think>', '\n']
  key with -S strip     214   excess over shared = 0   OK
  key at last <|im_end|> 210   excess over shared = -4   OK

enable_thinking=False
  turn-1 prompt         218 tokens
  shared with turn-2    214
  generation prompt     4 tokens  ['<think>', '\n\n', '</think>', '\n\n']
  key with -S strip     216   excess over shared = 2   <-- NOT a strict prefix: reuse lost
  key at last <|im_end|> 210   excess over shared = -4   OK
```

Any thinking-capable ChatML model served to a client that sends `enable_thinking: false` hits this.

## The fix

Anchor the key on the **last `<|im_end|>`** — the end of the last complete message — instead of on a startup-measured suffix length. That point is independent of the thinking mode and of the template's generation-prompt shape, and is a strict prefix of every later turn of the same conversation by construction. The cost is the few tokens of the trailing `<|im_start|>assistant` header.

`_think_suffix_len` is retained as the fallback for non-ChatML templates, so nothing regresses for models without `<|im_end|>`.

Applied at all four sites, so keys and lookups agree — they must, or the miss is silent again: canonical store/fetch and chunked-prefill store/fetch.

## Scope of the standalone claim

On upstream alone this PR fixes the **key**: for rewindable cache types (plain KV, attention-only models) a stored prompt entry now survives as a strict prefix of the next turn in both thinking modes. It does **not** by itself make hybrid (recurrent-state) models reuse prefixes on upstream: `_store_prefix_snapshot()` still rejects the positive trim for non-rewindable leaves via `_can_rewind_prefix_cache()`. Hybrid reuse is #744's job; once it lands this will be rebased onto it and the hybrid claim re-verified on that tree, without duplicating #744's ownership or publication logic.

## Testing

- The reproduction above is deterministic and needs no server.
- `TestPromptBoundaryKeying` (in `tests/test_mllm_continuous_batching.py`, run by the Apple Silicon job) drives all four call sites with `enable_thinking=true` and `false` token layouts and asserts the key and the lookup cut at the last `<|im_end|>`: canonical fetch through `_process_prompts`, canonical store through `_maybe_store_prefix_cache`, and the chunked fetch and store inside `install_chunked_prefill_mllm` (driven to completion with a 4-token budget). It also pins that the turn-1 key is a strict prefix of turn 2, and that the pre-fix startup-suffix strip leaves two extra tokens on the thinking-off key.
- Runtime evidence, for context only: on a fork that additionally carries a hybrid store, this change took a multi-turn agentic client against `Qwen3.6-35B-A3B-6bit` from a full re-prefill every turn to reusing the prefix (**40.3s cold → 0.51s on the following turn**), with a needle 9,000 tokens inside the cached region retrieved intact in both thinking modes. That result is not reproducible on this branch alone; see *Scope* above.

### DIFF
--- a/tests/test_mllm_continuous_batching.py
+++ b/tests/test_mllm_continuous_batching.py
@@ -2356,3 +2356,206 @@ def test_scheduler_step_error_fails_every_request_once():
     assert "_fail_requests_after_step_error(e)" in inspect.getsource(
         MLLMScheduler._process_loop
     )
+
+
+class TestPromptBoundaryKeying:
+    """Prefix-cache keys and lookups must cut at the last complete message,
+    independent of the per-request thinking mode, at all four call sites.
+
+    ``_think_suffix_len`` is measured once at startup with
+    ``enable_thinking=True`` (``<think>\\n`` = 2 tokens). A request rendered
+    with ``enable_thinking=False`` carries ``<think>\\n\\n</think>\\n\\n``
+    (4 tokens), so a fixed strip leaves two generation-prompt tokens on the
+    key and the entry is never a strict prefix of the next turn.
+    """
+
+    IM_START, IM_END, ASSISTANT = 6, 7, 8
+    THINK, NL, NLNL, END_THINK = 9, 10, 11, 12
+    BODY = [1, 2, 3, IM_END, 4, 5, IM_END]  # two complete messages
+    SUFFIX_ON = [IM_START, ASSISTANT, THINK, NL]  # enable_thinking=True
+    SUFFIX_OFF = [IM_START, ASSISTANT, THINK, NLNL, END_THINK, NLNL]  # False
+    STARTUP_SUFFIX_LEN = 2  # what _compute_think_suffix_len() measures
+
+    @classmethod
+    def _ids(cls, thinking: bool):
+        return cls.BODY + (cls.SUFFIX_ON if thinking else cls.SUFFIX_OFF)
+
+    @classmethod
+    def _gen(cls):
+        from vllm_mlx.mllm_batch_generator import MLLMBatchGenerator, MLLMBatchStats
+
+        gen = MLLMBatchGenerator.__new__(MLLMBatchGenerator)
+        gen._stats = MLLMBatchStats()
+        gen._pending_error_responses = []
+        gen._aborted_request_ids = set()
+        gen._prefill_progress = {}
+        gen.active_batch = None
+        gen.stop_tokens = set()
+        gen.unprocessed_requests = []
+        gen.max_kv_size = 0
+        gen._think_suffix_len = cls.STARTUP_SUFFIX_LEN
+        gen._im_end_id = cls.IM_END
+        gen._has_empty_rotating_cache = lambda cache: False
+        return gen
+
+    BOUNDARY = len(BODY)
+
+    @pytest.mark.parametrize("thinking", [True, False])
+    def test_boundary_is_thinking_mode_independent(self, thinking):
+        gen = self._gen()
+        ids = self._ids(thinking)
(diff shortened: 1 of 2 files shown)


Thump604 (changes requested):
Thanks for isolating the per-request thinking-mode mismatch. The last-`<|im_end|>` boundary is the right stable cut point for standard ChatML.

At exact head `15c55eee`, this does not yet publish reusable hybrid/Qwen3.5 entries on upstream by itself: both store paths still pass a positive trim through `_store_prefix_snapshot()`, and `_can_rewind_prefix_cache()` rejects recurrent leaves. That matches the PR disclosure that the measured hybrid result came from a fork with additional changes.

Please take one of these narrow paths:

1. Keep this standalone and scope the claim to rewindable cache types, with deterministic tests through both canonical and chunked store/fetch call sites for `enable_thinking=true` and `false`; or
2. Coordinate the hybrid claim with #744's exact full-prompt checkpoint/auxiliary-logit path after that contract lands, without duplicating its ownership/publication logic.

No additional hardware campaign is needed for this revision. The missing gate is source-level call-site coverage and a truthful standalone capability claim.

CBribiescas:
Thanks — option 2. The workload this is for is a hybrid target (Qwen3.6-35B-A3B) serving an agent client that sends the whole conversation every turn with `enable_thinking: false`, so the hybrid claim only makes sense on top of #744's store path. Plan:

- Once #744 lands I'll rebase this onto it and re-verify the multi-turn hit on that tree (no fork changes), and the hybrid claim will refer to #744's contract rather than duplicate any of its ownership or publication logic.
- The source-level coverage doesn't depend on that, so it is in now (497cbf8): `TestPromptBoundaryKeying` drives all four call sites — canonical fetch (`_process_prompts`), canonical store (`_maybe_store_prefix_cache`), and the chunked fetch and store inside `install_chunked_prefill_mllm` — with `enable_thinking=true` and `false` token layouts, and asserts the key and the lookup cut at the last `<|im_end|>` (plus the turn-1 key being a strict prefix of turn 2, and the pre-fix suffix strip being two tokens long for thinking-off).
- The PR description now scopes the standalone claim to rewindable cache types and states that the measured hybrid result came from a tree with the hybrid store, pending #744.

PR #767, opened 2026-09-04
fix(anthropic): canonicalize system prompts with the shared stripper

janhilgard:
Follow-up to #524, requested in [that thread](https://github.com/waybarrios/vllm-mlx/issues/524). Replaces the Anthropic adapter's inline billing-header regex with the shared `canonicalize_system_prompt()` from #528.

## The defect

`anthropic_to_openai()` carried its own pattern:

```python
system_text = re.sub(r"x-anthropic-billing-header:[^\n]*\n?", "", system_text)
```

It is unanchored, so it matches anywhere on a line — including the middle of a user's own sentence. And it runs *before* `_prepare_chat_messages()` calls `canonicalize_system_messages()`, so the shared pass cannot restore what it already deleted:

```
in:                    "Explain what x-anthropic-billing-header: means in HTTP terms."
after the adapter:     "Explain what "
after the shared pass: "Explain what "
without the adapter:   "Explain what x-anthropic-billing-header: means in HTTP terms."
```

The two also disagreed on case, in the opposite direction: the shared stripper is `(?im)`-anchored and removes a capitalised standalone header; the inline one did not.

| input | inline only | inline → shared | shared only |
|---|---|---|---|
| `x-anthropic-billing-header: …` on its own line | removed | removed | removed |
| `X-Anthropic-Billing-Header: …` on its own line | **kept** | removed | removed |
| header text mid-sentence | **removed** | **removed** | kept |

Only the second row was a cache-hit question, and on current `main` the shared pass already covers it, so that half is not a live defect. The third row is user-visible data loss, and it is the reason for this change.

## The fix

```python
system_text = canonicalize_system_prompt(system_text) or ""
```

One definition of the pattern instead of two, which is what the registry in #528 was for: the Anthropic path now picks up any future stripper without a second edit. The `re` import is no longer needed in the adapter and is dropped.

## Tests

`tests/test_anthropic_adapter_canonicalization.py`, 15 cases across four groups:

- **standalone header removed** — lower, mixed and upper case; header as the entire prompt; trailing header with no newline
- **user text preserved** — the mid-line mention, asserted both after the adapter alone and after the second pass
- **unrelated content untouched** — plain prompts, ISO timestamps, an unrelated `Authorization:` header, prose mentioning billing headers
- **both stages agree** — the shared pass is a no-op on whatever the adapter returns, so the two can no longer drift apart silently

Every case drives the real `anthropic_to_openai()` and, where the two-stage behaviour matters, the real `canonicalize_system_messages()` — no reimplementation of either.

Mutation-checked. Restoring the inline regex fails 5 of the 15:

```
FAILED …::test_case_insensitive_on_its_own_line[mixed]
FAILED …::test_case_insensitive_on_its_own_line[upper]
FAILED …::test_mid_line_mention_survives
FAILED …::test_mid_line_mention_survives_the_second_pass_too
FAILED …::test_second_pass_is_a_no_op[mixed]
```

## Verification

```
tests/test_anthropic_adapter_canonicalization.py     15 passed
anthropic / canonical / prefix / server suites      554 passed, 3 skipped
```

`black` clean. No benchmark here: this is a correctness fix on a code path, and the cache-hit claim I originally made for it did not survive checking.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,6 +82,7 @@ jobs:
             tests/test_request.py \
             tests/test_anthropic_models.py \
             tests/test_anthropic_adapter.py \
+            tests/test_anthropic_adapter_canonicalization.py \
             tests/test_dependency_floors.py \
             tests/test_ssd_cache_shutdown.py \
             tests/test_chat_template_safety.py \
--- a/tests/test_anthropic_adapter_canonicalization.py
+++ b/tests/test_anthropic_adapter_canonicalization.py
@@ -1,0 +1,112 @@
+# SPDX-License-Identifier: Apache-2.0
+"""The Anthropic adapter must canonicalize system prompts the shared way.
+
+`anthropic_to_openai()` used to carry its own unanchored `re.sub` for the
+billing header. It runs *before* `canonicalize_system_messages()`, so anything
+it removed was gone by the time the shared, line-anchored stripper saw the
+messages — including the middle of a user's own sentence. These tests pin both
+stages: what the adapter alone does, and what the prepared-message path
+produces end to end.
+"""
+
+import pytest
+
+from vllm_mlx.api.anthropic_adapter import anthropic_to_openai
+from vllm_mlx.api.anthropic_models import AnthropicRequest
+from vllm_mlx.api.prompt_canonicalize import canonicalize_system_messages
+
+HEADER = "x-anthropic-billing-header: cch=9f2a1b;"
+BODY = "You are helpful.\nBe concise."
+
+
+def _system_text(system) -> str:
+    request = AnthropicRequest(
+        model="test-model",
+        max_tokens=16,
+        system=system,
+        messages=[{"role": "user", "content": "hi"}],
+    )
+    converted = anthropic_to_openai(request)
+    systems = [m for m in converted.messages if m.role == "system"]
+    assert len(systems) == 1, f"expected one system message, got {len(systems)}"
+    return systems[0].content
+
+
+def _two_stage(system) -> str:
+    """Adapter output after the shared pass the server applies next."""
+    text = _system_text(system)
+    canonicalized = canonicalize_system_messages([{"role": "system", "content": text}])
+    return canonicalized[0]["content"]
+
+
+class TestStandaloneHeaderIsRemoved:
+    @pytest.mark.parametrize(
+        "header",
+        [
+            "x-anthropic-billing-header: cch=9f2a1b;",
+            "X-Anthropic-Billing-Header: cch=9f2a1b;",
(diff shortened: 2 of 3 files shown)


Thump604 (commented):
The source change at `77ae9065` matches the narrow fix agreed in #524: reuse the shared line-anchored canonicalizer instead of deleting mid-sentence prompt text. I checked the real adapter, shared helper and server preparation call path; I found no further source defect in this change. The new tests exercise the real adapter and shared helper, without adding speculative strippers or a benchmark.

One CI wiring item before landing: `tests/test_anthropic_adapter_canonicalization.py` is absent from the workflow's explicit test lists. I checked [run 33886240337](https://github.com/waybarrios/vllm-mlx/actions/runs/33886240337), including the Linux 3.12 and Apple 3.11 logs: the existing adapter tests execute, but these 15 new cases do not. The green checks therefore do not cover them yet.

Please add that filename beside `tests/test_anthropic_adapter.py` in the existing no-MLX unit-test invocation. That is the only remaining request from me; no additional behavior, framework or model benchmark.

waybarrios:
@janhilgard just added a missing test into the CI settings. Now it is ready to merge.

janhilgard:
Thanks both — and thanks @waybarrios for wiring it in directly rather than bouncing it back.

I verified the wiring executes rather than assuming the green checks meant it, since that is precisely what the checks did *not* mean before `505c9a19`. From the run-33890662441 job logs, all 15 cases run in every matrix job:

```
test-matrix (3.10): 15
test-matrix (3.11): 15
test-matrix (3.12): 15
test-matrix (3.13): 15
```

with individual ids present, e.g. `test_case_insensitive_on_its_own_line[lower|mixed|upper]`, `test_mid_line_mention_survives`, `test_second_pass_is_a_no_op[mixed]` — including the five that fail when the inline regex is restored, so the coverage is live and not merely collected. Job total 1408 passed, 19 skipped.

@Thump604 — you caught the same gap on #690, and both times it was a new test file that CI never saw. That is a repeatable footgun rather than two coincidences: a green PR proves nothing about a file nobody listed. I have started running the exact CI invocation locally before pushing rather than only `black`, which is also how the `F541` here would have been caught before it reached the runner instead of after.

Nothing further from me on this one.

Thump604 (approved):
Re-reviewed exact head `505c9a19`. The source remains the narrow shared-canonicalizer fix, and the follow-up commit adds the new test file to the existing Linux matrix invocation. I independently checked run 33890662441: all 15 canonicalization cases execute and pass on Python 3.10, 3.11, 3.12, and 3.13; all ten checks are green.

The prior CI-wiring request is resolved. I found no remaining blocker in this PR.

PR #768, opened 2026-09-04
fix(mllm): media requests must not use the prompt prefix cache

azamamirza:
**Symptom.** Serving a vision model with `--mllm --continuous-batching`
(prefix cache enabled, its default), sending N different images with the
same prompt text at temperature 0 returns the FIRST image's transcription
for all N requests. Reproduced with Qwen2.5-VL-7B-Instruct-8bit: 67
distinct document pages, one transcription prompt, only 2 distinct
outputs. A degenerate garbage response on the first request after model
load comes from the same path.

**Cause.** The prompt prefix cache is keyed on token IDs alone. Every
image occupies identical placeholder token IDs, so for media requests
"same prompt text" is a full-prefix hit and the engine replays another
image's KV wholesale. The existing safeguard in `_process_prompts` only
scans the *remaining* (uncached) ids for `image_token_index`, and a
full-prefix hit leaves `remaining_ids` empty, so it never fires. The
store side (`_maybe_store_prefix_cache`) has a docstring saying
"text-only requests" but no code enforcing it. Two text-only interleave
selections also admit audio requests (`not r.images and not r.videos`),
which walks the same bug in an audio costume.

**Fix.** `is_text_only`, whose field comment has always said "eligible
for prefix cache", now gates every prefix-cache fetch and store, and the
interleave selections exclude audio. Text-only requests keep the cache
unchanged. Repeated identical images still benefit from
`VisionEmbeddingCache`, which hashes image bytes correctly. The default
`is_text_only=False` fails safe: an unpreprocessed request can never be
served another request's KV.

An alternative considered and deliberately not done here: salting the
cache key with an images hash (e.g. `compute_images_hash`) would let
identical-image requests share prompt KV, but pseudo-token keys break the
cache's offset==key-length invariant. Happy to discuss that as a
follow-up.

**Tests.** `tests/test_mllm_prefix_cache_media.py`: media requests are
never stored; text-only requests still are; `is_text_only` defaults to
the safe direction. Full suite on the rebased branch (2026-09-04, on top
of f2d3ad3): 3016 passed / 24 skipped.

Fixes the media half of the behavior confirmed by review in #736.

### DIFF
--- a/tests/test_mllm_prefix_cache_media.py
+++ b/tests/test_mllm_prefix_cache_media.py
@@ -1,0 +1,76 @@
+# SPDX-License-Identifier: Apache-2.0
+"""D15 regression: media requests must never touch the prompt prefix cache.
+
+The prefix cache is keyed on token IDs alone, and every image/audio clip
+occupies identical placeholder IDs — so a media request stored or fetched
+under a token-only key replays a DIFFERENT image's KV for any later request
+with the same prompt text. Found live: 67 distinct document pages through
+an MLLM serve produced 2 distinct transcriptions.
+
+These tests pin the fixed contract at the seams:
+  1. ``_maybe_store_prefix_cache`` stores text-only requests and skips media.
+  2. ``MLLMBatchRequest.is_text_only`` defaults False — an unpreprocessed request
+     is treated as media, so the safe direction is cache bypass.
+"""
+
+from unittest.mock import MagicMock
+
+import mlx.core as mx
+
+from vllm_mlx.mllm_batch_generator import MLLMBatchGenerator, MLLMBatchRequest
+
+
+def _bare_generator(prefix_cache) -> MLLMBatchGenerator:
+    gen = object.__new__(MLLMBatchGenerator)
+    gen.prefix_cache = prefix_cache
+    gen._think_suffix_len = 0
+    return gen
+
+
+def _finished_batch(req: MLLMBatchRequest):
+    batch = MagicMock()
+    batch.requests = [req]
+    batch.num_tokens = [4]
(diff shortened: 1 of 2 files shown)


Thump604 (changes requested):
I reviewed exact head `2bc595e2`. The core diagnosis is valid, but two correctness gaps remain:

1. The fetch/store guards use `getattr(req, "is_text_only", True)`. That is fail-open for request-like objects that lack the field and contradicts the dataclass's safe `False` default. Default missing state to non-text/ineligible (or require a validated field), and bind that behavior at the real fetch/store seam.

2. Current persisted cache format v4 may already contain token-keyed media KV. After this change, a text-only request with the same token IDs can still load one of those legacy entries. Invalidate the old disk format, or provide an equivalent fail-closed migration, with one focused legacy-entry rejection test.

Please coordinate lineage with #728, which already carries these two pieces, rather than landing competing implementations. The separate order-sensitive multi-image VisionEmbeddingCache key is covered by #726 and need not broaden this PR. No additional model benchmark or broad test campaign is requested.

azamamirza:
Both points are correct. The getattr default of True is fail-open for request-like objects without the field and contradicts the dataclass's safe default, and this PR does nothing about v4 disk entries that may already hold media KV. #728 carries both pieces, plus a regression for the full-prefix-hit fetch path, so I am closing this in favor of #728 rather than landing a competing implementation.

For lineage: I hit this defect in production on Aug 7 (67 distinct scanned pages, one transcription prompt at temperature 0, only 2 distinct outputs) and have been running essentially this fix since, so #718's diagnosis matches what I saw exactly. I verified #728 locally today; details on that PR.

PR #769, opened 2026-09-05
fix(mllm): make generator request ownership transactional

Thump604:
I add the generator-side UID and transaction substrate for MLLM request ownership.

Batch filtering now stages row and cache mutations before publication. The generator gains an exact-UID prefill-abort path, failed deferred removals remain observable and retryable, successful removal retires owned abort/progress state, and insertion rolls back without consuming ownership after metadata or live-UID collision failures.

This is intentionally the first of two scoped changes. Scheduler call-site adoption and response/cleanup serialization will follow in a dependent PR; this PR does not claim end-to-end scheduler ownership or cancellation completion. Admission policy, stream shutdown, and cache policy are also outside this change.

Validation on Apple Silicon:
- 12 focused generator ownership cases passed
- 56 adjacent MLLM continuous-batching cases passed; 4 existing slow real-model cases were deselected
- Black, Ruff, and diff checks passed

### DIFF
--- a/tests/test_mllm_generator_ownership.py
+++ b/tests/test_mllm_generator_ownership.py
@@ -1,0 +1,572 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Focused generator request-ownership regressions."""
+
+import threading
+from types import SimpleNamespace
+from unittest.mock import MagicMock
+
+import pytest
+
+try:
+    import mlx.core as mx
+
+    HAS_MLX = True
+except ImportError:
+    HAS_MLX = False
+
+
+pytestmark = pytest.mark.skipif(not HAS_MLX, reason="MLX not available")
+
+
+def _assert_same_error(action, error, *args):
+    with pytest.raises(type(error)) as exc_info:
+        action(*args)
+    assert exc_info.value is error
+
+
+def _make_insert_generator():
+    from vllm_mlx.mllm_batch_generator import MLLMBatchGenerator
+
+    generator = MLLMBatchGenerator.__new__(MLLMBatchGenerator)
+    generator.uid_counter = 0
+    generator.unprocessed_requests = []
+    return generator
+
+
+def _make_chunked_generator():
+    from vllm_mlx.mllm_batch_generator import MLLMBatchGenerator, MLLMBatchStats
+
+    generator = MLLMBatchGenerator.__new__(MLLMBatchGenerator)
+    generator._next = lambda: []
+    generator._pending_error_responses = []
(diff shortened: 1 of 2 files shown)


janhilgard (approved):
Approving. Making `filter()` stage-then-commit is the right shape: the previous version mutated `uids`, `request_ids`, `logprobs`, `max_tokens`, `num_tokens`, `requests`, then `y`, then the caches, so a failure part-way through left the batch describing one set of rows while its KV described another — the kind of corruption that surfaces later as a wrong answer rather than an exception.

## The part I checked rather than took on trust

`_copy_cache_for_filter()` shallow-copies the cache and its mutable containers, then filters the copy. That is only safe if `filter()` rebinds attributes rather than mutating array storage — otherwise the staged filter would corrupt the original it is supposed to protect, and the transaction would be worse than no transaction.

Every implementation that can reach here rebinds:

```python
# BatchKVCache / BatchRotatingKVCache
self.keys = self.keys[batch_indices]
self.values = self.values[batch_indices]
self.offset = self.offset[batch_indices]

# ArraysCache
self.cache = [c[batch_indices] if c is not None else None for c in self.cache]
self.left_padding = self.left_padding[batch_indices]

# CacheList
for c in self.caches:
    c.filter(batch_indices)
```

So `copy.copy(cache)` plus rebinding on the copy leaves the original intact, and the special-casing of the `caches` attribute is exactly what `CacheList` needs — it filters through the list, so copying the list identity without recursing into its members would have staged nothing.

The docstring is precise about the cost, too: *"Copy mutable Python containers without copying MLX array storage."* The staging is Python-object-shaped, not memory-shaped, which matters when a batch holds tens of GB of KV.

`_copy_cache_container` preserving namedtuple and tuple-subclass types via `_fields` is the kind of detail that is easy to get wrong and painful to debug when a cache container quietly becomes a plain tuple.

## Tests

`test_filter_is_atomic_across_nested_cache_failure` is the one that matters — it fails the nested cache mid-filter and asserts the batch is unchanged, which is the property the PR exists for. The surrounding coverage on insert atomicity, abort-marker retirement and partial-removal ownership fills in the same discipline across the other mutation points.

```
tests/test_mllm_generator_ownership.py     12 passed
mllm / batch suites                       400 passed
```

The single failure in that wider run, `test_mllm_av_fusion.py::test_extract_audio_from_video_integration_smoke`, is `ffmpeg` dying with SIGABRT on this box. I checked it against clean `upstream/main` — it fails there too, so it is environmental and not yours.

PR #771, opened 2026-09-05
Preserve assistant reasoning when chat history is replayed

Thump604:
## Summary

I found that replaying an assistant response through Chat Completions drops
its `reasoning_content` during request validation. The response schema already
exposes this field, but the input `Message` schema does not retain it.

This adds the optional field to `Message`, using the same `reasoning` alias
as the response schema. Messages without reasoning remain unchanged. The
tests cover both input names, an assistant-response round trip, and omission
when no reasoning is supplied.

Losing the field can change the rendered history for templates that preserve
reasoning. This PR fixes the schema loss only; it does not change templates,
cache matching, or hybrid-cache rewind behavior, and it does not resolve all
multi-turn cache misses in #730.

## Verification

- Reproduced the dropped field on main `ec8e493` without loading a model.
- All 87 API-model tests pass on this branch.
- Ruff and diff checks pass.

### DIFF
--- a/tests/test_api_models.py
+++ b/tests/test_api_models.py
@@ -159,6 +159,35 @@ class TestMessage:
         msg = Message(role="assistant", content=None)
         assert msg.content is None
 
+    @pytest.mark.parametrize("field", ["reasoning_content", "reasoning"])
+    def test_assistant_reasoning_survives_request_validation(self, field):
+        request = ChatCompletionRequest(
+            model="test-model",
+            messages=[
+                {"role": "user", "content": "Question"},
+                {"role": "assistant", "content": "Answer", field: "Exact thought\n"},
+                {"role": "user", "content": "Continue"},
+            ],
+        )
+        history = request.messages[1].model_dump(exclude_none=True)
+        assert history == {
+            "role": "assistant",
+            "content": "Answer",
+            "reasoning_content": "Exact thought\n",
+        }
+
+    def test_returned_assistant_message_can_be_replayed(self):
+        returned = AssistantMessage(content="Answer", reasoning_content="Thought")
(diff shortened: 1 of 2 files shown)


janhilgard (approved):
Approving. Small, and it closes a real asymmetry: the server returns `reasoning_content` on assistant messages but `Message` dropped it on the way back in, so replayed history lost the reasoning the model itself had produced.

Checked the thing that worried me — whether adding a field to `Message` would push `reasoning_content: None` into every message handed to a chat template. It does not: `server.py:349` dumps with `exclude_none=True`.

```
bez exclude_none : {'role': 'user', 'content': 'ahoj', 'reasoning_content': None, …}
s exclude_none   : {'role': 'user', 'content': 'ahoj'}
```

So templates see the key only when the client actually sent reasoning, and a template that iterates message keys is unaffected for ordinary turns.

`AliasChoices("reasoning_content", "reasoning")` accepting both spellings is the right call rather than picking one: `reasoning` is what some OpenAI-compatible servers emit (vLLM's Metal plugin among them), so a client replaying history captured from one of those would otherwise silently lose it. Both normalise to `reasoning_content` on dump, which the round-trip test pins:

```python
replayed = Message.model_validate(returned.model_dump())
assert replayed.model_dump(exclude_none=True) == returned.model_dump()
```

The same alias pattern is already used at `models.py:229` and `:541`, so this makes `Message` consistent with its siblings rather than introducing a new convention.

`test_absent_reasoning_does_not_add_template_history` is the test I would have asked for — it pins the no-pollution property directly instead of leaving it implied.

Local run: `test_api_models.py` and `test_normalize_messages.py` green.

waybarrios (approved):
ready to merge!

PR #772, opened 2026-09-05
Forward chat reasoning effort to the template

Thump604:
## Summary

I found that a top-level `reasoning_effort` on Chat Completions is silently
dropped by request validation. Clients sending that field therefore do not
override the server's template default.

This retains the optional string and forwards it through the shared chat
preparation path. Precedence is:

1. Server template defaults.
2. A non-null top-level request `reasoning_effort`.
3. Explicit request `chat_template_kwargs`.

Absent or null effort preserves existing behavior. The value is forwarded
unchanged; this does not add model-specific mappings, toggle thinking, or
change the Anthropic/Responses routes. Tests cover the HTTP forwarding path
and the precedence rules.

## Verification

- Reproduced the dropped field on main `ec8e493` without loading a model.
- 274 API-model, template and server tests pass; three existing tests are
  deselected by the repository configuration.
- Ruff and diff checks pass.

### DIFF
--- a/docs/guides/server.md
+++ b/docs/guides/server.md
@@ -45,6 +45,13 @@ Clients route requests by setting the OpenAI `model` field to one of the configu
 Set server defaults for chat template kwargs. Request-level `chat_template_kwargs`
 values still win per key.
 
+For chat requests, top-level `reasoning_effort` is forwarded verbatim to the
+model's chat template. It overrides the server default; an explicit
+`chat_template_kwargs.reasoning_effort` overrides both. Accepted values and
+fallback behavior are template-specific: OpenAI's `low`/`medium`/`high` values
+are not portable across models. Check the selected model's template before
+choosing a value; a template may ignore the field or use its own fallback.
+
 ```bash
 vllm-mlx serve mlx-community/Qwen3-8B-4bit \
   --reasoning-parser qwen3 \
--- a/tests/test_chat_template_kwargs.py
+++ b/tests/test_chat_template_kwargs.py
@@ -105,6 +105,7 @@ def test_chat_completion_endpoint_forwards_chat_template_kwargs():
                 "model": "test-model",
                 "messages": [{"role": "user", "content": "Reply with ORBIT."}],
                 "max_tokens": 8,
+                "reasoning_effort": "medium",
                 "chat_template_kwargs": {"enable_thinking": False},
             },
         )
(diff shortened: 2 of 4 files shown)


janhilgard (approved):
Approving. The precedence — server default < `reasoning_effort` < explicit `chat_template_kwargs` — is the right ordering: the top-level field is the convenient path, and a caller who spells it out in `chat_template_kwargs` clearly means it. The parametrised test pins all three tiers rather than just the happy path.

I checked whether this actually reaches anything, since a forwarded kwarg that no template reads would be inert. It is not — on the models I have locally:

| model | `reasoning_effort` in template |
|---|---|
| `GLM-5.3-Flash-MLX-mixed-4_8bit` | 2 uses |
| `Qwen3.8-Flash-Next-oQ5e-mtp` | 5 uses |
| `Qwen3.6-35B-A3B-UD-MLX-4bit` | none |

So the feature does real work, and the PR is worth having.

**One thing worth documenting rather than changing.** The templates use model-specific vocabularies, and neither matches OpenAI's:

```jinja
{# GLM-5.3-Flash #}
reasoning_effort = reasoning_effort if reasoning_effort is defined
                   and reasoning_effort in ['low', 'high'] else 'max'

{# Qwen3.8-Flash-Next #}
reasoning_effort = reasoning_effort|default('xhigh')
… reasoning_effort == 'low' … reasoning_effort == 'xhigh'
```

GLM accepts `low`/`high` and falls back to `max`; Qwen speaks `low`/`xhigh`. So a client sending OpenAI's `"medium"` — the most natural value to send — falls through GLM's membership test and lands on **`max`**, the opposite end of the scale from what was asked for. Silently, since nothing rejects it.

That is not this PR's doing and I do not think it should try to fix it. It does mean a line in the docs would earn its keep: `reasoning_effort` is forwarded verbatim, the accepted values are the model template's, and OpenAI's `low`/`medium`/`high` are not portable across models.

It also means keeping the field as a plain `str | None` is correct. My first instinct was to suggest `Literal["low", "medium", "high"]`; that would have been wrong, because it would reject `xhigh` and every other model-specific level. Good that the PR left it open.

Local run: `test_chat_template_kwargs.py` 22 passed.

Thump604:
Thanks, Jan. I added the template-specific caveat in `docs/guides/server.md` at `4764c3a`: the value is forwarded verbatim, accepted levels and fallbacks belong to the model template, and OpenAI effort labels are not portable across models. The precedence is documented alongside it. This follow-up changes documentation only.

janhilgard (approved):
Re-approving. The docs note in `server.md` says the part that was missing:

> Accepted values and fallback behavior are template-specific: OpenAI's `low`/`medium`/`high` values are not portable across models. Check the selected model's template before choosing a value; a template may ignore the field or use its own fallback.

"a template may ignore the field" is worth having in there too — of the three models I have locally, `Qwen3.6-35B-A3B-UD-MLX-4bit` has no `reasoning_effort` in its template at all, so a caller setting it there gets silence rather than an error. Someone reading only the OpenAI spec would not expect either that or GLM's `medium` → `max` fallback.

Nothing else changed since my approval on 5 Sep; the precedence chain and its parametrised test are unmodified.

PR #773, opened 2026-09-05
Report request-owned cached prompt tokens in Chat usage

Thump604:
## Summary

I traced cached-prompt reporting through the schedulers and found that the
request's reused prefix count was not carried into Chat usage. This adds the
optional OpenAI-compatible `prompt_tokens_details.cached_tokens` field to
non-streaming responses and the final streaming usage chunk.

- Preserve the full `prompt_tokens` and `total_tokens` values.
- Forward the Text scheduler's existing request count. Track MLLM partial
  reuse from the validated lookup; exact hits report N-1 because the last
  prompt token is replayed.
- Report zero for a known miss and omit the optional details when accounting
  is unavailable. Aggregate cache lookup counters are not used.

This changes reporting, not cache policy, model computation, or generation
settings. It is based directly on main and contains no persistence or cache
ownership prerequisites.

## Verification

The new request-usage regression fails on the clean base. Tests cover partial
reuse, exact-hit rewind, unsafe-cache fallback, scheduler/collector/engine
propagation, unknown-field omission, and streaming/non-streaming usage.

- Focused: 407 passed, 7 deselected.
- Full repository: 3,069 passed, 25 skipped, 27 deselected.
- Black, Ruff, and diff checks pass.

These are source/API contract checks, not a performance benchmark or a
model-wide cache qualification.

### DIFF
--- a/tests/test_api_models.py
+++ b/tests/test_api_models.py
@@ -41,6 +41,7 @@ from vllm_mlx.api.models import (
     Message,
     ModelInfo,
     ModelsResponse,
+    PromptTokensDetails,
     ResponseFormat,
     ResponseFormatJsonSchema,
     StreamOptions,
@@ -370,15 +371,26 @@ class TestChatCompletion:
         assert choice.finish_reason == "stop"
 
     def test_usage(self):
-        usage = Usage(prompt_tokens=10, completion_tokens=20, total_tokens=30)
+        usage = Usage(
+            prompt_tokens=10,
+            completion_tokens=20,
+            total_tokens=30,
+            prompt_tokens_details=PromptTokensDetails(cached_tokens=4),
+        )
         assert usage.prompt_tokens == 10
         assert usage.total_tokens == 30
+        assert usage.model_dump()["prompt_tokens_details"] == {"cached_tokens": 4}
+        assert usage.model_dump(include={"prompt_tokens"}) == {"prompt_tokens": 10}
+        assert "prompt_tokens_details" not in usage.model_dump(
+            exclude={"prompt_tokens_details"}
+        )
 
     def test_usage_defaults(self):
         usage = Usage()
         assert usage.prompt_tokens == 0
         assert usage.completion_tokens == 0
         assert usage.total_tokens == 0
+        assert "prompt_tokens_details" not in usage.model_dump()
 
     def test_chat_completion_response(self):
         resp = ChatCompletionResponse(
--- a/tests/test_batched_engine.py
+++ b/tests/test_batched_engine.py
@@ -30,6 +30,7 @@ class TestBatchedEngineGenerate:
         prompt_tokens=10,
         completion_tokens=3,
         finish_reason="stop",
+        cached_tokens=None,
     ):
         """Build a mock RequestOutput (as returned by AsyncEngineCore)."""
         mock = MagicMock()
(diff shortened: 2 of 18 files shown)


janhilgard (approved):
Approving. Reporting `prompt_tokens_details.cached_tokens` is worth having on its own, and the tri-state discipline is what makes it trustworthy: `None` means "this engine does not account for it", `0` means "accounted, no hit". Collapsing those two would have made the field actively misleading — a client seeing `0` could not tell whether the cache missed or whether nobody was counting.

The serializer earns its place for the same reason:

```python
@model_serializer(mode="wrap")
def _serialize(self, handler) -> dict:
    data = handler(self)
    if self.prompt_tokens_details is None:
        data.pop("prompt_tokens_details", None)
    return data
```

Absent rather than null. A client that does not know the field is unaffected, and one that does can distinguish "unknown" from "zero" without a sentinel.

`type(cached_tokens) is not int` rather than `isinstance` also correctly rejects `bool`, which `isinstance` would have let through as `1`/`0`. Same discipline as #755.

Folding the non-streaming path onto `get_usage(output)` instead of building `Usage` inline removes a place where the two paths could drift, which is how the streaming and non-streaming answers diverge in the first place.

## One thing I checked rather than assumed

Six of the ten `MLLMBatchResponse(...)` construction sites omit `cached_tokens`, so they default to `None` — and the scheduler overwrites unconditionally:

```python
cached_tokens = getattr(response, "cached_tokens", None)
if type(cached_tokens) is not int:
    cached_tokens = None
request.cached_tokens = cached_tokens
```

which looked like a single late response could erase the value, and in the non-streaming path (`RequestOutputCollector` merges with `cached_tokens=new.cached_tokens`) that would drop `prompt_tokens_details` from the response entirely.

It does not. All six omitting sites are error/abort paths appending to `_pending_error_responses` — failed requests, `PrefillAbortedError`, preprocessing failures, aborts — and the scheduler handles `finish_reason == "error"` in an earlier branch that sets `cached_tokens=0` and `continue`s before reaching the overwrite. The four normal token paths (2077, 2797, 2822, 2973) all carry it, and `MLLMBatchRequest.cached_tokens` is a persistent field, so the value is stable across tokens rather than being re-derived per response.

Worth a one-line comment at that overwrite saying error responses never reach it, because the 6-of-10 ratio reads as a bug until you trace the earlier branch. Not blocking.

Local run: `test_api_models.py`, `test_request.py`, `test_server.py` — 303 passed.

Timely, incidentally: I spent this morning chasing a `cache_efficiency: 0.0` on another MLX server that turned out to be honest reporting of a cache that genuinely never engaged. Having the per-request number in `usage` is exactly what would have answered that in one request instead of a log dig.

Thump604:
Thanks, Jan. I added the comment at the overwrite in `6967ce9`, noting that error responses continue above and normal token responses carry cache usage. The Python AST is unchanged; this follow-up is comment-only.

janhilgard (approved):
Re-approving. The comment lands exactly where a reader trips:

```python
cached_tokens = getattr(response, "cached_tokens", None)
if type(cached_tokens) is not int:
    cached_tokens = None
# Error responses continue above; normal token responses carry cache usage.
request.cached_tokens = cached_tokens
```

That unconditional overwrite is what made six of ten `MLLMBatchResponse(...)` sites omitting the field look like a bug until you trace the `finish_reason == "error"` branch that `continue`s before reaching it. One line saves that trace.

`test_api_models.py` and `test_request.py` still green. Nothing else changed since 5 Sep.

PR #776, opened 2026-09-06
fix(anthropic): convert image blocks instead of silently dropping them

azamamirza:
**Symptom.** Sending an image through `/v1/messages` returns HTTP 200 with a
confident text-only answer. The model never sees the picture, and nothing in
the response indicates that.

**Cause.** `_convert_message` in `api/anthropic_adapter.py` branches on
`text`, `tool_use` and `tool_result` with no `image` branch and no catch-all.
An image block parses cleanly (`AnthropicContentBlock` carries `source`, and
`test_anthropic_models.py::test_image_block` covers the parsing) and is then
discarded during conversion. vLLM's own Anthropic surface converts image
blocks, so this was also a parity gap.

**Fix.** Convert Anthropic `source` to an OpenAI `image_url` content part:

- `base64` sources become data URIs with the declared `media_type` preserved
  (not assumed PNG); `url` sources pass through.
- Text and image parts keep the author's interleaving order. For document
  work, "here is the page" before "now answer this" is a different task from
  the reverse.
- Unknown source types and empty payloads raise instead of dropping. A data
  URI with an empty payload looks well-formed and makes the model answer
  about nothing, which is the exact failure this change exists to end.
- The text-only path still produces a plain string, so the common path is
  unchanged.
- Also fixes an adjacent latent bug this exposed: the "no text and no tool
  results" guard appended a stray empty user message, which image-only
  requests newly triggered.

`Message.content` already accepts `str | list[ContentPart] | list[dict]`;
the adapter simply never produced the list form.

**Tests.** `tests/test_anthropic_adapter_images.py` (9 tests): base64 with
media_type preserved, url sources, text/image interleaving order, image-only
messages produce no stray empty message, unknown source type raises, empty
payload raises, text-only path still emits a plain string. Full suite:
3074 passed / 24 skipped.

### DIFF
--- a/tests/test_anthropic_adapter_images.py
+++ b/tests/test_anthropic_adapter_images.py
@@ -1,0 +1,123 @@
+"""Anthropic image blocks must reach the model.
+
+Before this, an image block parsed cleanly and was then dropped, so a vision
+request returned HTTP 200 with a confident text-only answer. vLLM's own
+Anthropic surface converts image blocks, so this was also a parity gap.
+"""
+
+import base64
+
+import pytest
+
+from vllm_mlx.api.anthropic_adapter import _convert_message
+from vllm_mlx.api.anthropic_models import AnthropicMessage
+
+PNG = base64.b64encode(b"\x89PNG\r\n\x1a\nfake").decode()
+
+
+def parts_of(msg):
+    """Normalize content parts to dicts (pydantic coerces them to ContentPart)."""
+    out = []
+    for p in msg.content:
+        out.append(p if isinstance(p, dict) else p.model_dump(exclude_none=True))
+    return out
+
+
+def url_of(part):
+    u = part["image_url"]
+    return u if isinstance(u, str) else u["url"]
+
+
+def _img(media_type="image/png", data=PNG):
+    return {
+        "type": "image",
+        "source": {"type": "base64", "media_type": media_type, "data": data},
+    }
+
+
+def _msg(role, content):
+    return AnthropicMessage(role=role, content=content)
+
+
+def test_image_block_becomes_an_image_url_part():
+    out = _convert_message(_msg("user", [_img()]))
+    assert len(out) == 1
+    parts = parts_of(out[0])
+    assert isinstance(parts, list), "content must be multimodal, not a string"
+    assert parts[0]["type"] == "image_url"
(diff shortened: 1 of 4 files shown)


Thump604 (changes requested):
Thanks for fixing this. I reviewed 60293c9 and the conversion direction makes sense, but two paths still need attention before merge:

1. The new ValueErrors for an empty/unknown image source escape `create_anthropic_message`: `anthropic_to_openai()` runs before the route's try block, with no ValueError handler. Please translate these input errors into a deliberate client-error response before engine acquisition, and add route-level coverage for both streaming and non-streaming requests. The direct adapter `raises` tests don't cover the HTTP behavior.

2. An assistant message containing both an image and `tool_use` still takes the `tool_calls_for_assistant` branch, which uses `combined_text` and never includes `content_parts`. The new image-only assistant test misses this. Please preserve the image alongside the tool call, or explicitly reject that combination if it is unsupported, with a regression for the combined case.

Separately, images nested inside `tool_result.content` still go through the existing text-only extraction. That predates this PR; it can remain a clearly documented follow-up rather than expanding this patch into a broader tool-message redesign.

This is an exact-source review; I have not run a local vision model for it. The ten checks are green, but they don't establish these missing paths.

azamamirza:
Both requested changes addressed. 

1. The route now translates adapter ValueErrors into a 400 before engine acquisition, with route-level tests for empty-data and unknown-source images non-streaming plus the empty-data case streaming, all through the real app.
2. An assistant message with an image and a tool_use now keeps the full content part list alongside the tool call. I checked that downstream media extraction is role-agnostic, so preserving is supported rather than rejecting, and the combined case has a regression test.
3. Agree on tool_result-nested images as a follow-up. I left a comment marking the extraction site. Full suite: 3078 passed / 24 skipped locally on Apple silicon.

Thump604 (approved):
Re-reviewed e2992eb4 against my previous requests. Invalid image sources now return HTTP 400 before engine acquisition, and assistant messages preserve image content alongside tool_use. The added route-level and adapter tests cover the requested paths. Images nested inside tool_result remain the separate documented follow-up we agreed on.

I found no remaining blocker in this scoped change. This was a source-only re-review; I have not rerun the tests or model inference locally. Thanks for addressing both points.

PR #774, opened 2026-09-05
Validate vllm-mlx as a local backend for seven coding agents

waybarrios:
## Summary

Add an isolated acceptance runner for **OpenCode, pi, Codex, Claude Code, GitHub Copilot CLI, Cline CLI, and OpenClaw** using vllm-mlx as their local inference backend. The task requires real model-issued tool calls, a file edit, returned tool results, and a completed answer containing a token that was absent from the prompt.

**Real-model validation: all seven clients completed acceptance with Qwen3-8B-6bit on an Apple M5 Max.** These were installed clients talking to actual MLX inference, with private profiles and one client at a time. The final seven-client run used the same source revision and exact client-version pins.

The validated follow-up fixes are published in [`00dbf22`](https://github.com/waybarrios/vllm-mlx/commit/00dbf22d6f64fdd64cf4ed99cb64525ba042aed2), signed by Wayner Barrios with a GitHub-verified SSH signature. [Real-model logs and source provenance](https://github.com/waybarrios/vllm-mlx/pull/774#issuecomment-5557174863) document all seven runs.

### Changes

- Add private client adapters, a bounded HTTP/SSE observation proxy, a process runner, structured JSON reports, portable tests, documentation, and a manual Apple Silicon workflow.
- Preserve stable streamed tool-call IDs, reject truncated HTTP bodies, close active proxy connections, and clean up process groups on timeout or SIGTERM.
- Fix the `qwen` streaming parser so trailing whitespace and final engine chunks cannot emit an already-completed call again with a different ID.
- Preserve individual tool results and assistant tool calls during message normalization, including their call IDs and contents.
- Give Codex command tools the standard system executable path without inheriting the operator's environment. Native sandboxing stays enabled.
- Disable OpenClaw startup respawning and the Node compile cache so its execution stays within the runner's process group through cleanup.
- Make the fixture task explicit: read both files, replace `pending` with the input token, preserve the final newline, and answer with the token. Exact-byte acceptance requirements are unchanged.

### Acceptance contract

Each client receives private HOME/XDG/configuration directories and a workspace containing a random token in `input.txt` and `pending` in `output.txt`. A pass requires the selected model/API, at least two inference requests, completed streaming, model-issued calls with matching returned results, a result and final answer containing the token, exact output bytes, unchanged input bytes, and successful client exit without protocol errors or timeout. Symlinks and non-regular fixture files are rejected.

[...]

PR #777, opened 2026-09-08
docs: provide a release-bound first-response path

Thump604:
## Summary

Give first-time users one isolated, version-pinned install-to-response path before introducing advanced features. Keep editable development installs separate and remove the 30-second setup claim.

The existing examples also used three unrecognized flags and an unconfigured `default` model alias. This corrects those against the published 0.4.1 wheel, adds health/model-ID checks, and distinguishes vendor artifacts from lower-memory derivatives without claiming equivalent quality or universal feature support. It links the existing coding-client work in #774 rather than adding another client setup system.

## Validation

- Checked the published wheel hash against PyPI, CLI flag definitions and health/model/chat routes.
- Executed the wheel's isolated model-name validator: unconfigured `default` returns 404; the example's actual served ID is accepted.
- Dependency resolution dry-run passes in an isolated Python 3.12 environment.
- Shell examples pass syntax checks; relative file links and `git diff --check` pass.

Documentation only. No fresh model download, model smoke test, inference-suite run, or performance/quality qualification is claimed. The server version pin is not a lock of every transitive dependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,11 +17,16 @@
 
 A vLLM-style inference server for Apple Silicon Macs. Unlike `Ollama` or `mlx-lm` used directly, it ships **continuous batching, paged KV cache, prefix caching, and SSD-tiered cache**, and exposes **both OpenAI `/v1/*` and Anthropic `/v1/messages`** from a single process. Run LLMs, vision models, audio, and embeddings on Metal with unified memory, no conversion step.
 
-## Quick start (30 seconds)
+## Quick start
+
+New users: follow the [isolated release install](docs/getting-started/installation.md#install-a-release)
+and [first-response check](docs/getting-started/quickstart.md#first-response).
+The example below assumes that environment is activated. Initial model
+download/loading time depends on the machine and connection.
 
 ```bash
-pip install vllm-mlx
-vllm-mlx serve mlx-community/Llama-3.2-3B-Instruct-4bit --port 8000 --continuous-batching
+python -m pip install 'vllm-mlx==0.4.1'
+vllm-mlx serve mlx-community/Llama-3.2-3B-Instruct-4bit --host 127.0.0.1 --port 8000
 ```
 
 **OpenAI SDK:**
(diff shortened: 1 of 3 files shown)


janhilgard (approved):
Approving. Docs PRs are worth reviewing by checking the claims rather than the prose, so I verified each correction against `main`:

| README said | PR says | in `cli.py` |
|---|---|---|
| `--mtp` | `--enable-mtp` | only `--enable-mtp` exists |
| `--spec-prefill` | `--specprefill` | only `--specprefill` exists |
| `--metrics` (Prometheus line) | removed | `--metrics` does not exist |
| `model="default"` | explicit model id | see below |

All four were wrong before and are right now.

The `model="default"` one is a real defect, not a style change. On the single-model path `_validate_model_name()` does:

```python
if _model_name and request_model != _model_name:
    raise HTTPException(status_code=404, …)
```

and `_model_name = served_model_name or model_name`, so the quickstart's own two commands, run in order, returned 404 on the very first request. There is a `_default_model_key = "default"` at `server.py:3621`, but it belongs to the model-manager path, which the quickstart does not use. A first-time user following the README hit a 404 and had no way to know the README was at fault.

Two notes, neither blocking.

**Pinning `vllm-mlx==0.4.1` will go stale.** It matches PyPI today (`0.4.1` is latest, ahead of `0.4.0`, `0.4.0rc1`), and I understand wanting a first run that cannot break on an unrelated release. But nothing updates it, so the quickstart will quietly install an old version some months from now — the failure mode being a user reporting a bug already fixed. If the intent is "a known-good version", a line saying so, or a release-checklist item to bump it, would keep it honest.

**The quickstart no longer passes `--continuous-batching`**, while the paragraph three lines above sells the project on *"continuous batching, paged KV cache, prefix caching"*. Simplifying the first command is defensible, but a reader who tries the headline feature is running without it. A sentence pointing at the flag would close that gap without complicating the command.

The `openai` install note is a good catch — that it is not pulled in by the server package is exactly the kind of thing that stops a first run one step after the server is up.

Thump604:
Thanks Jan, and thanks for checking the actual behavior. Addressed both notes in 2e6234f: the walkthrough now calls out the pinned reference release and how to update its pins and checks together, and the README/quickstart point to --continuous-batching after the first response. The minimal command stays unchanged. Documentation-only follow-up; diff checks passed, with no new model or inference claims.

PR #778, opened 2026-09-08
Fix Qwen SpecPrefill dispatch and MTP-off continuation

Thump604:
## Summary

Fix three compatibility checks that made Qwen draft-backed SpecPrefill fall
back to dense generation:

- Read `model.model_type` as well as the existing config-based identity, so
  mlx-lm's Qwen draft selects the gated Qwen query extractor.
- Check RoPE presence with `is not None`. An empty `nn.RoPE` module is falsey
  but still supplies positional encoding.
- Omit `mtp` from sparse continuation when MTP is off. Passing `mtp=False`
  reaches an unsupported `generate_step` keyword on mlx-lm 0.31.3.

The MTP-on call remains unchanged. No model allowlist, default, dependency,
token selection, or cache policy changes.

## Verification

74 focused SimpleEngine, cancellation, query-dispatch, and rotating-cache
tests pass on this branch. The dispatch/RoPE regressions and MTP-off
continuation assertion failed before the fixes. Black and diff checks pass.

Separately, a local Runtime composition with these changes completed one
Qwen3.6-35B / Qwen3.5-2B text retrieval canary: 2,277 of 11,237 prompt tokens
selected, all three keys correct, normal completion, no fallback. Before the
fixes, the correct answer came only from dense fallback. This is not an
upstream-head hardware qualification, speed claim, or coverage for media,
continuous batching, MTP coexistence, Qwen27, or Qwen Next.

### DIFF
--- a/tests/test_simple_engine.py
+++ b/tests/test_simple_engine.py
@@ -2113,7 +2113,8 @@ class TestSimpleEngineConcurrency:
         assert seen["tokens"] == [10, 11, 12, 13]
 
     @pytest.mark.anyio
-    async def test_specprefill_success_preserves_mtp_path(self):
+    @pytest.mark.parametrize("mtp", [False, True])
+    async def test_specprefill_success_preserves_mtp_path(self, mtp):
         """Successful sparse prefill should continue through the normal MTP path."""
         from types import SimpleNamespace
 
@@ -2148,13 +2149,13 @@ class TestSimpleEngineConcurrency:
         )
 
         text_model = MagicMock()
-        text_model.mtp = object()
+        text_model.mtp = object() if mtp else None
         text_model.make_mtp_cache.return_value = ["mtp-cache"]
 
         engine = SimpleEngine(
             "test-model",
             force_mllm=True,
-            mtp=True,
+            mtp=mtp,
             mtp_num_draft_tokens=4,
             specprefill_enabled=True,
             specprefill_threshold=1,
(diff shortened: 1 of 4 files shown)


janhilgard (approved):
Approving. Three fixes, and the `_get_rope()` one is more consequential than its two-line diff suggests.

## `_get_rope()` never returned a real `rope`

```python
# before
return getattr(attn, "rope", None) or getattr(attn, "rotary_emb", None)
```

`nn.Module` inherits from `dict`, so a module with no parameters is falsy. A RoPE is exactly that — precomputed or computed on the fly, nothing to learn. Checked on the installed MLX:

```
nn.Module inherits from: dict
mlx_lm RoPE:  bool = False
nn.RoPE(64):  bool = False, parameters = 0
```

So the `or` skipped every real `rope` and fell through to `rotary_emb`, which is absent on mlx_lm models — meaning `_get_rope()` returned `None` for them, always.

That is not cosmetic, because of the caller:

```python
if _get_rope(attn_obj) is not None:
    query_extractor = _llama_extract_queries
else:
    query_extractor = _nemotron_h_extract_queries
```

Every mlx_lm model outside `_EXTRACTOR_REGISTRY` was therefore routed to the **Nemotron-H** extractor regardless of its actual architecture. SpecPrefill then scored tokens with the wrong extractor and produced a plausible-looking but wrong selection — no error, just worse prefill. Ironic given the comment a few lines up:

> *explicit registry, not attribute sniffing — avoids silent misclassification on new models*

The attribute sniffing was the part that misclassified. `rope if rope is not None else …` is the right fix; a truthiness test on an `nn.Module` is never what you want.

## The other two

`model_type` now checked on the model before its config — mlx_vlm text wrappers carry it directly, so the registry lookup was falling through to the sniffing path above for exactly the models it was meant to classify. The two fixes compound: one sent models to the sniffer, the other made the sniffer always answer Nemotron-H.

`**({"mtp": True} if use_mtp else {})` rather than `mtp=use_mtp` — passing the kwarg only when enabled, so a `generate_step` that predates the parameter keeps working when MTP is off. Narrow and correct; MTP-on still requires the newer signature, which is the right place to demand it.

## Tests

`tests/test_specprefill_query_dispatch.py` covers the dispatch directly, which is the property that was silently wrong. Local run: 15 passed across the specprefill suites.

Worth a grep for other `or` chains over `getattr(module, …)` in this file and its neighbours — the same shape will misbehave anywhere a parameterless `nn.Module` is the expected value, and it reads as correct.

PR #780, opened 2026-09-08
fix(cli): resolve MLLM/LLM routing from config.json before engine selection

### LINKED ISSUE #748: [Bug]: is_mllm_model() silently loads the wrong engine for bare repo-id args (pattern list missing Qwen3.8)
### Environment

- Chip: Apple M5 Max
- RAM: 128 GB
- Python version: 3.14.6
- vllm-mlx version: `origin/main` @ `22efb47968f96e6e7a79704f4eb0184cda4e0068` — line numbers below verified fresh against this exact commit (`git fetch origin` immediately prior)
- mlx / mlx-lm / mlx-vlm / mlx-audio versions: mlx==0.32.2, mlx-lm==0.31.3, mlx-vlm==0.6.15 (per the #730 qualification environment this issue was surfaced in)

### Model and quantization

`mlx-community/Qwen3.8-27B-8bit`, snapshot `815b83c0df8ffd1d1b5244cf75fd6ef14fca9ef9`. Its `config.json` includes `vision_config`, so it is correctly MLLM/hybrid by content. Not specific to this checkpoint — reproduces for any bare-repo-id model whose name doesn't match an entry in `MLLM_PATTERNS`.

### Reproduction

```
vllm-mlx serve mlx-community/Qwen3.8-27B-8bit --continuous-batching --enable-prefix-cache --port 8091
```

Check the startup log for the MLLM/LLM routing decision (e.g. via `--mllm` omitted and inspecting engine construction logs / `mllm=` state).

### Expected behavior

Since the model's `config.json` declares `vision_config`, the server should detect it as MLLM and route to the MLLM engine path, or at minimum warn loudly that it could not determine this before falling back to a heuristic.

### Actual behavior

The bare HF repo-id invocation silently loads the non-MLLM engine path (`mllm=False`), with no warning — different scheduler, no MLLM features, and any benchmark or repro run this way is measuring the wrong engine.

**Root cause**, verified against `origin/main@22efb47`:

- `is_mllm_model()` (`vllm_mlx/api/utils.py:469-493`):
  ```python
  def is_mllm_model(model_name: str) -> bool:
      config = _try_read_config_json(model_name)
      if config is not None:
          return _config_indicates_vlm(config)
      return _check_legacy_string_patterns(model_name)
  ```
- `_try_read_config_json()` (`api/utils.py:408-434`) only resolves when `model_name` is a local directory (`candidate.is_dir()`). For a bare HF repo id like `mlx-community/Qwen3.8-27B-8bit`, that's always `False` before the weights are downloaded/cached under that literal string, so `config.json` is never consulted — the function falls straight through to substring matching.
- `_check_legacy_string_patterns()` (`api/utils.py:456-465`) checks the input against `MLLM_PATTERNS` (`api/utils.py:335-365`). The list includes `"Qwen3.5-"`/`"qwen3_5"` (added by #138 for Qwen3.5) but has no `"Qwen3.8"` entry, so `mlx-community/Qwen3.8-27B-8bit` matches nothing and `is_mllm_model()` returns `False`.
- This return value directly decides engine routing, with no config-based double-check afterward and no warning emitted: `self._is_mllm = force_mllm or is_mllm_model(model_name)` in both `BatchedEngine.__init__` (`vllm_mlx/engine/batched.py:229`) and `SimpleEngine.__init__` (`vllm_mlx/engine/simple.py:208`), where `model_name` is exactly the raw string passed on the CLI for the single-model serve path (`load_model()`, `server.py:3361` — see `_model_path = model_name` at `server.py:3486`, threaded straight into the engine constructor's `model_name`).

**Scope note**: local-directory paths (`vllm-mlx serve /path/to/model`) are unaffected — `_try_read_config_json` resolves `config.json` correctly there. The registry (`--models-config`) path is also unaffected for engine routing: registry-mode `force_mllm` comes from an explicit YAML `mllm:` key or a server-wide default (`model_registry.py:1238-1241`, `force_mllm = entry.force_mllm if entry.force_mllm is not None else self._defaults.force_mllm`), not from `is_mllm_model()`. (The registry path does call `is_mllm_model(source)` once, at `model_registry.py:1190`, but only to choose which files `ensure_model_downloaded` fetches — it doesn't feed engine routing.) The gap is specific to a bare HF repo-id argument on the single-model CLI path.

### Additional context

Surfaced during reporter-side qualification of #744 on #730 ([comment](https://github.com/waybarrios/vllm-mlx/issues/730#issuecomment-5469311195)): "`--mllm` must be forced for direct CLI invocation of this artifact. ... a bare HF repo-id argument (`vllm-mlx serve mlx-community/Qwen3.8-27B-8bit`) falls back to filename substring matching, which doesn't include `"Qwen3.8"` in its pattern list ... so it silently loads the non-MLLM engine and none of this shows up. My first pass hit exactly this and produced meaningless results until caught via `mllm=False` in the startup log."

Filed separately per maintainer request on that same issue ([comment](https://github.com/waybarrios/vllm-mlx/issues/730#issuecomment-5470926373)): "`is_mllm_model()` only inspects local config before falling back to patterns that do not include Qwen3.8. Please track those separately rather than broadening #744."

### Suggested direction

Resolve `config.json` via the HF cache/metadata (e.g. `huggingface_hub`'s cached-file lookup, which works for a bare repo id once cached, or a lightweight remote metadata fetch) before falling back to `MLLM_PATTERNS`, rather than requiring a local directory. Short of that, warn loudly at startup whenever the fallback pattern list — not config inspection — is what decided the routing, so this doesn't fail silently.

djacobsmeyer:
## Summary

Fixes #748. `is_mllm_model()` only inspects `config.json` for inputs that are already local directories. A bare HF repo id passed to the single-model `serve` CLI (e.g. `mlx-community/Qwen3.8-27B-8bit`) is never a local directory before its weights are downloaded, so detection always fell straight to `MLLM_PATTERNS` substring matching — which has no `"Qwen3.8"` entry — and silently loaded the text-only engine even though the model's own `config.json` declares `vision_config`.

**Shape implemented: (1), resolve-before-route, reusing the existing download resolution.** This was fully feasible: `serve_command` already calls `ensure_model_downloaded()` to pre-fetch the model before `load_model()`, but discarded the resolved local snapshot path it returned. This PR captures it instead.

### What changed

- `vllm_mlx/api/utils.py`: new `resolve_mllm_route(model_name, resolved_path, *, force_mllm)` — inspects the *resolved* snapshot's `config.json` (authoritative), falling back to the legacy `MLLM_PATTERNS` matcher only when `config.json` can't be read, and raising `MllmRouteUndetermined` (fail-closed, shape 2) in the residual case where a snapshot directory was genuinely resolved but its config is unreadable/missing *and* the name matches nothing in the pattern list either. `is_mllm_model()` itself is untouched — every existing caller (registry path, tokenizer fallback, etc.) is unaffected.
- `vllm_mlx/cli.py` (`serve_command`, the single-model CLI call site): captures the `Path` `ensure_model_downloaded()` already returns, calls `resolve_mllm_route()` against it, and threads the *resolved snapshot path* (not the raw repo id) into `load_model()` along with the resolved `force_mllm` decision — so the engine's own loader doesn't re-resolve the model a second time over the network, and can't silently re-derive a different answer via its own internal `is_mllm_model()` call. `served_model_name` now defaults to the original repo id so API-facing model names (`/v1/models`, etc.) are unaffected by loading from a local path.
- The routing decision and its source are now logged at startup: `Model route: MLLM (source=config)` / `... (source=explicit)` / `... (source=pattern)`.

No second network lookup is introduced — the pre-download that already happened is what's being reused.

## Test plan

Per the maintainer's acceptance criteria, tests bind the real single-model CLI call site (`cli.serve_command`), not just the lower-level helper:

- [x] `tests/test_cli.py::TestServeCommandMllmRouting::test_cached_bare_repo_id_routes_via_config_not_pattern` — (a) a bare repo id whose name matches no `MLLM_PATTERNS` entry but whose resolved `config.json` declares `vision_config` routes to MLLM, attributed to `source=config`.
- [x] `tests/test_cli.py::TestServeCommandMllmRouting::test_uncached_metadata_failure_fails_closed` — (b) a resolved snapshot with no readable `config.json` and a non-matching name fails closed (`SystemExit(1)`, actionable message, `load_model` never called) instead of silently defaulting to the text-only engine.
- [x] `tests/test_cli.py::TestServeCommandMllmRouting::test_ordinary_text_only_repo_routes_llm` — (c) an ordinary text-only repo (no VLM markers, no pattern match) still correctly routes to LLM.
- [x] `tests/test_cli.py::TestServeCommandMllmRouting::test_explicit_mllm_flag_short_circuits_config` — `--mllm` still overrides and is reported as `source=explicit`.
- [x] `tests/test_api_utils.py::TestResolveMllmRoute` — unit coverage for `resolve_mllm_route()` itself (explicit / config-true / config-false / fail-closed / pattern-fallback-without-resolution).
- [x] Updated the two existing `test_cli.py`/mocked-module tests that stub `vllm_mlx.api.utils` wholesale, so they expose the new `resolve_mllm_route`/`MllmRouteUndetermined` names.

Ran:
```
python -m pytest tests/test_api_utils.py tests/test_cli.py tests/test_lifecycle_cli.py tests/test_simple_engine.py tests/test_batched_engine.py tests/test_server.py -q
# 382 passed, 3 deselected

ruff check vllm_mlx/ tests/
# All checks passed!
```
No model inference or benchmark run, per the issue's scope note. The registry (`--models-config`) path is untouched, as specified.

https://claude.ai/code/session_01Nd2MuriqTy1XruWLGtjZak

### DIFF
--- a/tests/test_api_utils.py
+++ b/tests/test_api_utils.py
@@ -8,10 +8,13 @@ from vllm_mlx/api/utils.py. No MLX dependency.
 
 import json
 
+import pytest
+
 from vllm_mlx.api.models import ContentPart, ImageUrl, Message
 from vllm_mlx.api.utils import (
     MLLM_PATTERNS,
     SPECIAL_TOKENS_PATTERN,
+    MllmRouteUndetermined,
     _check_legacy_string_patterns,
     _config_indicates_vlm,
     _content_to_text,
@@ -20,6 +23,7 @@ from vllm_mlx.api.utils import (
     extract_multimodal_content,
     is_mllm_model,
     is_vlm_model,
+    resolve_mllm_route,
 )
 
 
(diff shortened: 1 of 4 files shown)


Thump604:
Thanks, this addresses the reported routing problem by using the resolved config before engine selection and retaining the original served model name. I checked the cold-download concern too: the reported Qwen3.8 snapshot uses files covered by the existing download patterns, so I found no blocker for that case.

One nonblocking follow-up: the initial download still selects LLM versus MLLM patterns from the name heuristic. For a different unknown VLM requiring files outside the LLM patterns, passing the resulting local directory could prevent the loader from fetching missing assets. Our synthetic fixture demonstrates that mechanism, not a failure on a supported model. I would track that separately rather than hold this routing fix or ask you to redesign it.

This was source/fixture review; I have not run model inference on this PR.

PR #782, opened 2026-09-09
fix(metrics): finish HTTP request metrics on stream exhaustion, not call_next return

### LINKED ISSUE #781: [Bug]: streaming /v1/chat/completions decrements vllm_mlx_http_requests_in_flight within milliseconds of request start, long before generation finishes
### Environment

- Chip: Apple M5 Max
- RAM: 128 GB
- Python version: 3.13.14
- vllm-mlx version: 0.4.1, `origin/main` @ `ec8e493` (verified byte-identical in the affected functions against the deployed pin, a divergent local checkout at `9c4c2dd`)
- mlx / mlx-lm / mlx-vlm / mlx-audio versions: mlx 0.32.2, mlx-lm 0.31.3, mlx-vlm 0.6.17, mlx-audio 0.5.1

### Model and quantization

Not model-specific — reproduces for any model on the streaming `/v1/chat/completions` path. Observed live with `mlx-community/Qwen3.8-27B-8bit`, served via `--models-config` (registry mode).

### Reproduction

```
vllm-mlx serve --models-config models.yaml --enable-metrics --port 8091 --api-key test

# Terminal 1: a streaming request whose generation takes several seconds (larger
# max_tokens / longer prompt make this easier to observe)
curl -N -X POST http://localhost:8091/v1/chat/completions \
  -H 'Authorization: Bearer test' -H 'Content-Type: application/json' \
  -d '{"model":"mlx-community/Qwen3.8-27B-8bit","messages":[{"role":"user","content":"Write a 2000 word essay on distributed systems."}],"max_tokens":4000,"stream":true}' &

# Terminal 2: poll the gauge every 200ms while terminal 1 is still streaming
while true; do
  curl -s -H 'Authorization: Bearer test' http://localhost:8091/metrics \
    | grep 'vllm_mlx_http_requests_in_flight{method="POST",path="/v1/chat/completions"}'
  sleep 0.2
done
```

### Expected behavior

`vllm_mlx_http_requests_in_flight{method="POST",path="/v1/chat/completions"}` should read `1.0` for the entire time the SSE stream is open — from request start until the final chunk (`data: [DONE]`) is sent — so it can be used as an "is this endpoint currently doing work" signal (e.g. for external monitoring/HUDs).

### Actual behavior

The gauge reads `1.0` only for a sub-second window at the very start of the request, then drops to `0.0` and stays there for the remaining duration of the stream, even though the client is still actively receiving real content chunks (potentially for minutes).

**Root cause**, found by direct source read (line numbers below are `origin/main` @ `ec8e493`):

- [`server.py#L1830-L1859`](https://github.com/waybarrios/vllm-mlx/blob/ec8e4932615acbcdd1dcefccbf095ff49b5b7b1e/vllm_mlx/server.py#L1830-L1859) — `_metrics_middleware`, an `@app.middleware("http")` wrapping `call_next`. It calls `_metrics.observe_http_start(...)` (`.inc()`s the gauge) before `await call_next(request)`, and `_metrics.observe_http_finish(...)` (`.dec()`s the gauge, and records the request-duration histogram + the `_total` counter) immediately after `call_next()` returns.
- Starlette's `BaseHTTPMiddleware.call_next()` does **not** wait for a `StreamingResponse`'s body to finish sending — it resolves as soon as the inner ASGI app has sent `http.response.start` (and made the first body chunk available). Everything the streamed body sends afterward happens outside the middleware's `try/finally`, so `.dec()` fires as soon as headers + the first chunk are available, not when the stream actually ends.
- [`server.py#L6544-L6571`](https://github.com/waybarrios/vllm-mlx/blob/ec8e4932615acbcdd1dcefccbf095ff49b5b7b1e/vllm_mlx/server.py#L6544-L6571) — `stream_chat_completion()` immediately `yield`s a content-free `role: assistant` preamble chunk (line 6571) *before* it ever awaits `engine.stream_chat(...)`. That preamble chunk is what makes the first body chunk available almost instantly (no real model work needed), which is what lets `call_next()` in the middleware resolve so early.
- [`server.py#L4663-L4704`](https://github.com/waybarrios/vllm-mlx/blob/ec8e4932615acbcdd1dcefccbf095ff49b5b7b1e/vllm_mlx/server.py#L4663-L4704) — `_build_chat_streaming_response()` constructs and returns the `StreamingResponse` around that generator without awaiting any of its content, confirming the handler returns to the middleware well before real generation starts.

Net effect: for every streaming chat/completions request (the default/common case for any interactive client — coding agents, chat UIs, etc.), the in-flight gauge, the `vllm_mlx_http_request_duration_seconds` histogram, and (less severely) the timing recorded against `vllm_mlx_http_requests_total` for that path are all effectively meaningless — they measure "time to first ASGI byte," not "time to serve the request."

**Real-world evidence**: in production use behind a tracing proxy that timestamps actual SSE content chunks (not the empty preamble), a genuine streaming generation ran for ~27 minutes end-to-end (ttft ~26.5 min to the first real content token, per the proxy's own instrumentation). Polling `GET /metrics` every 5s for the entire window, `vllm_mlx_http_requests_in_flight{method="POST",path="/v1/chat/completions"}` read `0.0` at every single poll.

### Additional context

**Workaround identified and already deployed downstream**: `vllm_mlx_scheduler_running_requests` / `vllm_mlx_scheduler_waiting_requests` (populated from `engine.get_stats()` in `_update_engine_gauges`) are *not* affected by this bug — they reflect the engine's own scheduler state for the entire processing window. They resolve through `ModelManager.get_metrics_engine()` (MRU-only in registry mode), so they can't attribute activity to a specific tier when multiple models are resident, but that's a much smaller caveat than "reads 0.0 for 99%+ of every streaming request's real duration."

**Cross-reference**: #738 reported these same scheduler/engine-state gauges as permanently stuck at their zero defaults in registry mode. That looks to already be fixed on `main` — [`server.py#L3802`](https://github.com/waybarrios/vllm-mlx/blob/ec8e4932615acbcdd1dcefccbf095ff49b5b7b1e/vllm_mlx/server.py#L3802) already reads `_model_manager.get_metrics_engine() if _model_manager is not None else _engine` rather than the bare `_engine` the issue describes — but #738 is still open, so flagging in case that fix hasn't been verified/closed yet. Confirmed live (registry mode, `mlx-community/Qwen3.8-27B-8bit`) that `vllm_mlx_scheduler_running_requests` does read a real nonzero value during active generation on this codebase state.

**Suggested fix direction**: decrement `http_requests_in_flight` (and record the duration/total-counter observations) when the `StreamingResponse`'s body generator is actually exhausted — e.g. in a `finally` around the SSE generator itself (`_ensure_sse_terminal`/`_disconnect_guard`) — rather than when `_metrics_middleware`'s `call_next()` returns. Non-streaming responses (which already await full generation before returning) are unaffected and should keep their current behavior.

djacobsmeyer:
## Summary

Fixes #781. `_metrics_middleware` finished HTTP request metrics (`http_requests_in_flight`, `http_request_duration_seconds`, `http_requests_total`) immediately after `call_next()` returned, but Starlette's `BaseHTTPMiddleware.call_next()` resolves as soon as the ASGI response has *started*, not when a `StreamingResponse` body finishes. For streaming `/v1/chat/completions` (the default for any interactive client), this decremented the in-flight gauge within milliseconds of request start, long before real generation happened, let alone finished.

Wraps the response's `body_iterator` (the lazy generator Starlette's `call_next()` always hands back) so metrics finish once the body is actually exhausted -- covering normal completion, a mid-stream error, and an early close, exactly once each. Non-streaming responses are unaffected: their body is already fully produced before `call_next()` returns.

## Type of change

- Bug fix

## Surface touched

- OpenAI API endpoints
- Metrics

## Test plan

```sh
pytest tests/test_metrics.py -v
pytest tests/test_server.py -k "metric or middleware or stream" -v
ruff check vllm_mlx/server.py tests/test_metrics.py --select E,F,W --ignore E402,E501,E731,F811,F841
black --check vllm_mlx/server.py tests/test_metrics.py
```

Also confirmed the new tests actually regress against the pre-fix middleware (`git stash` the `server.py` change, re-run `tests/test_metrics.py::TestMetricsMiddlewareStreamingTiming`, restore): 3 of the 4 new tests fail without the fix, for exactly the reasons the fix addresses.

## Test results

```
tests/test_metrics.py::TestMetricsEndpoint::test_metrics_endpoint_disabled_returns_404 PASSED
tests/test_metrics.py::TestMetricsEndpoint::test_metrics_endpoint_reports_dead_gauges_without_engine_or_manager PASSED
tests/test_metrics.py::TestMetricsEndpoint::test_metrics_endpoint_scrapes_registry_mode_engine PASSED
tests/test_metrics.py::TestMetricsEndpoint::test_metrics_endpoint_registry_mode_dead_gauges_when_idle PASSED
tests/test_metrics.py::TestMetricsEndpoint::test_metrics_endpoint_scrapes_runtime_stats PASSED
tests/test_metrics.py::TestMetricsEndpoint::test_metrics_collapse_unmatched_paths PASSED
tests/test_metrics.py::TestMetricsEndpoint::test_completion_request_updates_metrics PASSED
tests/test_metrics.py::TestMetricsEndpoint::test_streaming_chat_records_ttft_and_stream_metrics PASSED
tests/test_metrics.py::TestMetricsMiddlewareStreamingTiming::test_streaming_body_holds_gauge_until_exhausted PASSED
tests/test_metrics.py::TestMetricsMiddlewareStreamingTiming::test_error_mid_stream_decrements_exactly_once PASSED
tests/test_metrics.py::TestMetricsMiddlewareStreamingTiming::test_call_next_raising_before_any_response_finishes_once PASSED
tests/test_metrics.py::TestMetricsMiddlewareStreamingTiming::test_non_streaming_single_chunk_body_still_settles_immediately PASSED
============================== 12 passed in 2.16s ==============================

44 passed, 128 deselected  (tests/test_server.py -k "metric or middleware or stream")

ruff: All checks passed!
black: All done! 2 files would be left unchanged.
```

Environment: Apple M5 Max, 128 GB RAM, macOS 27.0, Python 3.12.13.

## Performance impact

N/A -- this only changes *when* a metrics call fires (once per request, same as before), not how often, and adds one thin async generator wrapper around each response body. No inference-path or kernel changes.

## Output parity

N/A -- no change to generation, sampling, scheduling, or tokenization.

## Risk notes

Relies on `body_iterator` being present on the response Starlette's `call_next()` returns, which is a private implementation detail (`starlette.middleware.base._StreamingResponse`). Guarded with `getattr(response, "body_iterator", None)`: if a future Starlette version drops or renames it, the middleware falls back to the old immediate-finish timing rather than raising -- same defensive posture this file already takes toward Starlette/uvicorn internals (see `_find_uvicorn_cycle`'s docstring). Verified the wrapped generator finishes metrics exactly once under normal completion, a mid-stream exception, and a manual early `aclose()` (see the new `TestMetricsMiddlewareStreamingTiming` tests).

### DIFF
--- a/tests/test_metrics.py
+++ b/tests/test_metrics.py
@@ -1,8 +1,10 @@
 # SPDX-License-Identifier: Apache-2.0
 """Tests for Prometheus server metrics."""
 
+import asyncio
 import platform
 import sys
+from types import SimpleNamespace
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


Thump604:
The normal-exhaustion timing fix looks right. I found one limit to the early-close claim: using the actual BaseHTTPMiddleware/_StreamingResponse path in supported Starlette 1.3.1, an ASGI send raising OSError after the first chunk left metrics unfinished at request exit and after another loop turn. They settled after later garbage collection, so this is delayed cleanup, not a demonstrated permanent leak.

The added test calls aclose() after a body exception has already terminated the iterator; that does not cover disconnect while the iterator is suspended at yield. Please add one focused send-failure/disconnect regression and either ensure settlement at request exit or explicitly document the remaining limitation. No broader middleware rewrite is requested.

This finding comes from a controlled ASGI test with the PR middleware, not model inference.

djacobsmeyer:
Thanks — reproduced exactly. Wrote a fake `_StreamingResponse` mirroring the real one's body-iteration loop (`async for chunk in self.body_iterator: await send(...)`, no `try/finally`), had `send()` raise `OSError` after the first chunk: confirmed the gauge stuck at 1.0 immediately post-exception against the previous commit, settling only after GC. Good catch that my `aclose()` test covered a different case (post-body-exception, iterator already terminated) than disconnect-while-suspended-at-yield.

Pushed `e2ed85d`: added `_MetricsFinishOnCallExit`, a thin ASGI-callable wrapper around the response so `finish()` also runs in a `finally` around the response's own call, not just around `body_iterator`. `finish()` is now idempotent, so whichever of the two paths fires first — body exhaustion/generator error, or the call itself ending — wins and the other is a no-op. This settles the send()-failure case synchronously at request exit rather than waiting on GC. No broader rewrite — additive to the existing `body_iterator` wrapping, which still handles normal exhaustion and generator-raised errors.

Added `test_send_failure_mid_stream_settles_at_request_exit`, using the same fake-response shape as the repro. Confirmed it fails against `0e2b201` (stuck at 1.0) and passes with `e2ed85d`. Full suite + ruff + black clean.

Thump604:
Thanks for following through on the exact case. I rechecked e2ed85d against 0e2b201 using the real Starlette 1.3.1 BaseHTTPMiddleware response path, with send() raising OSError on the first body chunk. The old head leaves metrics unfinished at request exit and after another loop turn; e2ed85d records one finish at request exit, with no second finish on the next loop turn.

That resolves my specific delayed-settlement finding. This was a controlled ASGI check with the exact middleware source, not model inference. No broader rewrite or additional model benchmark is needed for this finding.

PR #787, opened 2026-09-12
feat: OpenAI-compatible logprobs for chat and text completions (continuous batching)

### LINKED ISSUE #786: Feature: OpenAI-compatible logprobs / top_logprobs on chat and completions
### Current behaviour

`/v1/chat/completions` and `/v1/completions` accept `logprobs` / `top_logprobs` in the request, but the fields are silently ignored: the response is 200 and has no `logprobs`. Verified on 0.4.1 and `main` (`ec8e493`).

Both schedulers already compute a full-vocabulary `log_softmax` at every decode step (`scheduler.py` via mlx-lm's `GenerationBatch`, and `mllm_batch_generator.py`). The values are dropped before any `RequestOutput` is built, so exposing them is mostly plumbing.

### Why it matters

- **Confidence estimation and calibration.** Token-level confidence is the standard alternative to a model's self-reported confidence, for example when checking whether a small model's stated confidence is calibrated.
- **Constrained classification.** Combined with `response_format: json_schema`, the post-constraint logprob of the emitted label gives its probability among the allowed labels.
- **Compatibility.** OpenAI-compatible clients and eval harnesses (lm-eval, etc.) expect the field.

### Proposal

Return OpenAI-shaped logprobs from the continuous-batching engine, both the LLM and MLLM paths, streaming and non-streaming:

- values taken after logits processors and before temperature/top-p (processed logprobs);
- the stop token excluded;
- speculative drafting bypassed for requests that ask for logprobs.

Reject logprobs on engines that can't serve them with a clear 400 instead of silently dropping them.

I have an implementation with tests and live verification on an M2 Max. I'll open it as a PR linked here.

dhanesh:
## Summary

Adds OpenAI-compatible `logprobs` / `top_logprobs` to `/v1/chat/completions` (non-streaming and streaming) and `logprobs` to `/v1/completions`. They are served by the continuous-batching engine on both the LLM scheduler and the MLLM scheduler.

Today these request fields are silently ignored, even though both schedulers already compute a full-vocabulary `log_softmax` every decode step. The values are dropped at the first vllm-mlx output object. This PR carries them through for requests that ask for them.

Closes #786

## What changes

- **`vllm_mlx/logprobs.py` (new).**
  - `extract_token_logprob` reads the sampled token's logprob and the top-k alternatives (`argpartition`) on the engine thread, so no MLX arrays reach the API layer. It decodes token strings and clamps `-inf` from masked logits to `-9999.0`, OpenAI's value for unlikely tokens.
  - `record_step_logprobs` is shared by both schedulers.
  - It also provides formatters for the chat and legacy completions shapes.
- **Engine plumbing.**
  - `SamplingParams.logprobs` and `MLLMBatchRequest.logprobs` carry the request (`None` = off, `k` = top-k).
  - `RequestOutput` gains `new_logprobs` and `output_logprobs`, mirroring `new_token_ids` and `output_token_ids`. `RequestOutputCollector._merge_outputs` merges them.
  - `GenerationOutput` gains `logprobs` and `new_logprobs`, mirroring `text` and `new_text`.
- **Schedulers.** `scheduler.py` (LLM) and `mllm_scheduler.py` (MLLM) record one entry per content token. Stop tokens are excluded.
- **Speculative decoding.** In `mllm_batch_generator.py`, MTP/drafter speculation is bypassed for requests with logprobs, alongside the existing `logits_processors` bypass, because draft-emitted tokens report logprobs from a different distribution. A new `bypass_counts["logprobs"]` counter tracks this. The LLM path's `_install_mtp` hook is already inactive on mlx-lm >= 0.31.3; a comment at the call site notes that it would need the same bypass if it returns.
- **Server.**
  - The request flag is forwarded to the engine.
  - Engines that can't serve logprobs (`BaseEngine.supports_logprobs`, `False` except for `BatchedEngine`) get HTTP 400 up front, before any stream starts.
  - In streaming, reasoning and tool parsers can consume tokens without emitting a chunk. Their entries are buffered and attached to the next chunk sent, including the tool-call fallback and terminal chunks, so none are dropped. Existing `yield` lines are unchanged.
- **API models.** Adds `TopLogprob`, `ChatCompletionTokenLogprob`, `ChoiceLogprobs` and `CompletionLogprobs`. `top_logprobs` without `logprobs: true` is rejected, matching OpenAI.
- **Docs.** A new "Log probabilities" subsection in `docs/guides/server.md`.

## Semantics

- Values come after logits processors (JSON-schema constraints, `logit_bias`, penalties) and before temperature/top-p/top-k/min-p. That is the processed distribution at T=1, which is what mlx-lm's `generate_step` computes.
- Entries cover every generated token except the final stop token. The server post-processes text (reasoning/tool parsing, `response_format` JSON normalization, whitespace trimming), so `message.content` can differ from the concatenated tokens. For example, the model generates pretty-printed JSON and `content` is the compact re-serialization.
- `token` / `bytes` come from decoding each token alone. A partial-UTF-8 token therefore reports U+FFFD, as in vLLM.
- Batched forward passes reduce in a different order from single requests, so values can differ slightly between batched and unbatched runs. At an exact tie this can flip the greedy token. This is pre-existing batching behaviour that logprobs make visible; see Verification.
- **Cost.** Each logprobs request adds one small GPU sync per step (the sampled value plus a top-k over the vocabulary). Requests without logprobs do no extra work, although they share the step with any request that does. Gathering all rows into one `mx.eval` per step is a possible follow-up.

## Out of scope (possible follow-ups)

- **SimpleEngine** returns 400 for logprobs requests. Its generation routes differ (mlx-lm `stream_generate`, the mlx-vlm media path, SpecPrefill), so it is better handled in its own PR.
- `/v1/responses` and the Anthropic endpoint are unchanged.

## Verification

Hardware: Apple M2 Max, 32 GB, macOS.

- **New and updated tests:**
  - `tests/test_logprobs.py`: extraction, sorting, `-inf` clamping, placeholder rows, top-k cap, recording and stop-token exclusion, both output formats, the output-collector merge, and the LLM and MLLM schedulers' `_process_batch_responses`.
  - `tests/test_server_logprobs.py`: chat non-stream and stream, parser-consumed tokens (terminal chunk and end-of-stream flush), the 400 check, and completions non-stream and stream offsets.
  - `tests/test_batched_engine.py`: LLM and MLLM pass-through.
  - `tests/test_api_models.py`: request validation and bounds.
  - `tests/test_mllm_continuous_batching.py`: the MTP logprobs bypass.
- **Full suite:** 3073 passed, 40 skipped, 4 failed, 13 errors. The failures and errors are exactly the ones present on `main` (`ec8e493`) in this environment:
  - 2 in `test_lifecycle_cli`;
  - 2 in `test_prefix_cache_scheduler_parity`, a "no Stream(gpu)" thread error;
  - 13 in `test_streaming_detokenizer`, which needs to download a model.
- **Lint:** `black --check vllm_mlx/ tests/` and the CI ruff command (`ruff check vllm_mlx/ tests/ --select E,F,W --ignore E402,E501,E731,F811,F841`) are clean.
- **Live, `--continuous-batching --max-num-seqs 8`, T=0, JSON-schema `response_format`:**

  | Check | `mlx-community/Qwen3.5-4B-MLX-4bit` (MLLM path) | `mlx-community/Qwen3-0.6B-4bit` (LLM path) |
  |---|---|---|
  | Sampled token's logprob == top-1 logprob at every position | ✓ | ✓ |
  | Streaming vs non-streaming tokens/logprobs | identical (max abs diff 0.0) | identical (max abs diff 0.0) |
  | 8 concurrent requests vs solo | identical | identical except at an exact tie (note below) |
  | `logprobs` omitted → `"logprobs": null` | ✓ | ✓ |
  | `/v1/completions` `logprobs=3` (sampled token in `top_logprobs`) | ✓ | ✓ |

  Default (simple) engine: `logprobs: true` → HTTP 400 with an explanatory message. Requests without logprobs are unaffected.

  **Tie note (Qwen3-0.6B).** At one position the solo run has `'0'` and `'5'` tied at exactly −1.500. Under 8-way batching the order of floating-point reduction breaks the tie either way (for example `'5'` −1.494 vs `'0'` −1.509). As a result, 1–2 of every 8 concurrent requests generate `"confidence": 0.5` instead of `0.0`. Each response's logprobs match its own output, and the same flip happens without logprobs. Otherwise, batched logprob values stay within about 0.06 of the solo run.

https://claude.ai/code/session_013q1Gph3BbSN6DH4oNBSVen

### DIFF
--- a/docs/guides/server.md
+++ b/docs/guides/server.md
@@ -112,6 +112,36 @@ for chunk in stream:
         print(chunk.choices[0].delta.content, end="")
 ```
 
+#### Log probabilities
+
+With `--continuous-batching`, chat completions accept `logprobs: true` and
+`top_logprobs` (0-20), and text completions accept `logprobs` (0-5). The
+server then returns OpenAI-format per-token log probabilities, both streaming
+and non-streaming:
+
+```python
+response = client.chat.completions.create(
+    model="default",
+    messages=[{"role": "user", "content": "Hello!"}],
+    logprobs=True,
+    top_logprobs=5,
+)
+for token in response.choices[0].logprobs.content:
+    print(token.token, token.logprob)
+```
+
+Notes:
+
+- Values are taken after logits processors (JSON-schema constraints,
+  `logit_bias`, repetition and presence penalties) and before temperature,
+  top-p, top-k and min-p.
+- They cover every generated token except the final stop token. The server
+  post-processes text for reasoning and tool-call parsing, `response_format`
+  JSON normalization and whitespace trimming, so `message.content` can differ
+  from the concatenated tokens.
+- Requests that ask for logprobs skip speculative (MTP or drafter) decoding.
+- The default simple engine returns HTTP 400 for logprobs requests.
+
 ### Completions
 
 ```bash
--- a/docs/guides/server.md
+++ b/docs/guides/server.md
@@ -140,6 +140,9 @@ Notes:
   JSON normalization and whitespace trimming, so `message.content` can differ
   from the concatenated tokens.
 - Requests that ask for logprobs skip speculative (MTP or drafter) decoding.
+- Batched forward passes accumulate floating-point results in a different
+  order from single requests, so values can differ slightly between batched
+  and unbatched runs. At an exact tie this can also flip the greedy token.
 - The default simple engine returns HTTP 400 for logprobs requests.
 
 ### Completions
--- a/docs/guides/server.md
+++ b/docs/guides/server.md
@@ -139,7 +139,12 @@ Notes:
   post-processes text for reasoning and tool-call parsing, `response_format`
   JSON normalization and whitespace trimming, so `message.content` can differ
   from the concatenated tokens.
-- Requests that ask for logprobs skip speculative (MTP or drafter) decoding.
+- Requests that ask for logprobs never receive speculatively drafted tokens:
+  they skip multimodal MTP and drafter decoding. The LLM path's MTP hook is
+  inactive with mlx-lm 0.31 and later.
+- `token` and `bytes` come from decoding each token on its own, so a token
+  that is only part of a multi-byte UTF-8 character reports the replacement
+  character (U+FFFD), as in vLLM.
 - Batched forward passes accumulate floating-point results in a different
   order from single requests, so values can differ slightly between batched
   and unbatched runs. At an exact tie this can also flip the greedy token.
--- a/tests/test_api_models.py
+++ b/tests/test_api_models.py
@@ -475,6 +475,19 @@ class TestLogprobs:
                 top_logprobs=21,
             )
 
+    def test_chat_request_accepts_zero_top_logprobs(self):
+        req = ChatCompletionRequest(
+            model="test-model",
+            messages=[{"role": "user", "content": "Hi"}],
+            logprobs=True,
+            top_logprobs=0,
+        )
+        assert req.top_logprobs == 0
+
+    def test_completion_request_rejects_negative_logprobs(self):
+        with pytest.raises(ValidationError):
+            CompletionRequest(model="test-model", prompt="Hi", logprobs=-1)
+
     def test_completion_request_logprobs_bounds(self):
         req = CompletionRequest(model="test-model", prompt="Hi", logprobs=5)
         assert req.logprobs == 5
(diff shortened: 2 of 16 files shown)


Thump604:
I reviewed the API-to-scheduler plumbing and parser-buffered streaming at 0217e922. I did not find a blocking source-level issue in the paths inspected. This was a source review, not numerical or live-serving qualification.

Issue #785, opened 2026-09-11
[Bug]: MCP client reads SDK1 camelCase fields; connect() hangs against an SDK2 server


mmcco:
### Environment

- Chip (M1 / M2 / M3 / M4 / M5, including Pro / Max / Ultra): Apple M5 Max
- RAM: 128 GB
- macOS version: 26.6.2 (build 25G83)
- Python version: 3.12.11
- vllm-mlx version: 0.4.1 (same code on `main` @ ec8e493)
- mlx / mlx-lm / mlx-vlm / mlx-audio versions: 0.32.2 / 0.31.3 / 0.7.0 / 0.5.3

Relevant here: `mcp` and `mcp-types` are both **2.2.0**.

### Model and quantization

N/A — not model-dependent; no model is loaded to reproduce this.

### Reproduction

```shell
# 1. A minimal MCP stdio server (SDK 2.x API), saved as echo_server.py
cat > echo_server.py <<'EOF'
from mcp.server.mcpserver import MCPServer
app = MCPServer("echo")

@app.tool()
def echo(text: str) -> str:
    """Echo back the given text."""
    return text

if __name__ == "__main__":
    app.run()
EOF

# 2. Point vllm-mlx's MCP client at it
python - <<'PY'
import asyncio, logging, sys
logging.basicConfig(level=logging.INFO, stream=sys.stdout)
from vllm_mlx.mcp.client import MCPClient
from vllm_mlx.mcp.types import MCPServerConfig, MCPTransport

cfg = MCPServerConfig(name="echo", transport=MCPTransport.STDIO,
                      command=sys.executable, args=["echo_server.py"])
print(asyncio.run(MCPClient(cfg).connect()))
PY

# Logs the error below, then never returns -- the process must be killed.

# Sanity check that the server itself is fine: the SDK's own client connects to
# it and lists the tool successfully.
python - <<'PY'
import asyncio, sys
from mcp import ClientSession, StdioServerParameters
from mcp.client.stdio import stdio_client

async def main():
    p = StdioServerParameters(command=sys.executable, args=["echo_server.py"])
    async with stdio_client(p) as (r, w):
        async with ClientSession(r, w) as s:
            init = await s.initialize()
            print("initialize OK, protocol_version =", init.protocol_version)
            t = (await s.list_tools()).tools[0]
            print("tool:", t.name, "| has .input_schema:", hasattr(t, "input_schema"),
[...]

PR #788, opened 2026-09-12
fix(mcp): support SDK2 fields and clean up failed connections

Thump604:
## Summary

- Read SDK2 snake_case fields with SDK1 camelCase fallback, preserving empty schemas and false error flags.
- Close partially opened session/transport contexts after connection failure without reacquiring the held client lock, and clear stale tools.

Refs #785.

## Verification

- Eight focused compatibility/cleanup tests pass; Black, Ruff, Python compilation and diff checks pass.
- Bounded model-free stdio echo checks passed against `mcp==1.27.1` and separately pinned `mcp==2.2.0`/`mcp-types==2.2.0`: connect, list, call and disconnect.
- Injected initialization failure returned `False`/ERROR, cleared session/transport references, and the child exited in the bounded check.

## Scope

No model inference or serving-environment change. Real transport checks cover stdio; SSE behavior was not integration-tested. The tests do not certify every MCP server or dependency version.

### DIFF
--- a/tests/test_mcp_client_compat.py
+++ b/tests/test_mcp_client_compat.py
@@ -2,6 +2,7 @@
 """Focused MCP SDK compatibility and failed-connect cleanup tests."""
 
 from types import SimpleNamespace
+import asyncio
 
 import pytest
 
@@ -39,12 +40,11 @@ def _client() -> MCPClient:
 async def test_initialize_session_accepts_both_sdk_field_styles(result, caplog):
     client = _client()
     caplog.set_level("DEBUG")
-    client._session = SimpleNamespace(initialize=lambda: result)
 
     async def initialize():
         return result
 
-    client._session.initialize = initialize
+    client._session = SimpleNamespace(initialize=initialize)
     await client._initialize_session()
     assert "initialized" in caplog.text
 
@@ -56,7 +56,7 @@ async def test_initialize_session_accepts_both_sdk_field_styles(result, caplog):
         SimpleNamespace(
             name="snake",
             description="snake schema",
-            input_schema={"type": "object"},
+            input_schema={},
         ),
         SimpleNamespace(
             name="camel",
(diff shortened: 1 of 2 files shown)


janhilgard (approved):
Approving. Two independent fixes, both worth having.

## SDK field compatibility

`_field(obj, snake, camel, default)` reading snake_case first, then camelCase, then a default, is the right shape — and raising `AttributeError` when neither exists and no default was given is better than the previous pattern:

```python
# before
input_schema=(tool.inputSchema if hasattr(tool, "inputSchema") else {})
is_error=result.isError if hasattr(result, "isError") else False
```

Those silently substituted `{}` and `False` on an SDK that had merely renamed the field, so a v2 server's real `is_error=True` would have been read as `False` — a failed tool call reported as success. Requiring an explicit default at each call site makes that a decision rather than an accident, and the two sites that genuinely want one (`input_schema`, `is_error`) still pass it.

## Connection cleanup

The lock discipline is correct, which is the thing worth checking here since `asyncio.Lock` is not reentrant. `_close_connection()` is called at `client.py:134` (inside `connect()`'s `async with self._lock`) and `:233` (inside `disconnect()`'s), and does not take the lock itself — matching its docstring. No re-entrant deadlock.

The ordering inside the loop is the part I would call out as the real improvement:

```python
context = getattr(self, attr, None)
setattr(self, attr, None)          # cleared before the await
if context is None:
    continue
try:
    await context.__aexit__(None, None, None)
```

The old code assigned `None` *after* awaiting `__aexit__`, so a close that raised left a dead session still attached to the client. Clearing first means a failed close cannot leave a stale context behind for the next `connect()` to trip over.

Adding the cleanup to `connect()`'s failure path closes the matching leak: a server that failed during `initialize()` or tool discovery previously left its stdio subprocess and session object alive with `_state = ERROR`, and nothing ever collected them.

Per-context `try`/`except` means one bad context does not prevent the others from closing, and `disconnect()` still reaches its state transition even when a close fails — the old `finally` guaranteed that, and the guarantee survives the refactor.

Local run: 94 passed across the MCP suites; `tests/test_mcp_client_compat.py` is new and covers both field spellings.

Issue #783, opened 2026-09-11
[Bug]: build_text_model() drops the checkpoint's global quantization mode — MXFP8 Embedding crash and silent Linear misdecode


mmcco:
### Environment

- Chip (M1 / M2 / M3 / M4 / M5, including Pro / Max / Ultra): Apple M5 Max
- RAM: 128 GB
- macOS version: 26.6.2 (build 25G83)
- Python version: 3.12.11
- vllm-mlx version: 0.4.1 (also present on current `main`)
- mlx / mlx-lm / mlx-vlm / mlx-audio versions: 0.32.2 / 0.31.3 / 0.7.0 / 0.5.3

mlx device: Device(gpu, 0). mlx-embeddings 0.1.0.

### Model and quantization

mlx-community/gemma-4-12B-it-qat-mxfp8 (global MXFP8, group_size 32 / bits 8, plus 144 per-layer affine overrides at group_size 64 / bits 8)

### Reproduction

```shell
# Weight-free reproduction — no model download needed.
# Mirrors what build_text_model() does: quantize the skeleton WITHOUT the
# checkpoint's global `mode`, then load the checkpoint's scales into it.

python - <<'PY'
import mlx.core as mx, mlx.nn as nn
from mlx.utils import tree_flatten

# Source: quantized the way the checkpoint is (global mxfp8)
src = nn.Embedding(4, 32).to_quantized(group_size=32, bits=8, mode="mxfp8")
mx.eval(src.parameters())

# Derived: what build_text_model() builds today (no `mode` forwarded)
der = nn.Embedding(4, 32).to_quantized(group_size=32, bits=8)
der.load_weights(tree_flatten(src.parameters()), strict=False)
mx.eval(der(mx.array([0])))
# RuntimeError: [metal::Device] Unable to load kernel
#               affine_dequantize_uint8_t_gs_32_b_8
PY

# The same mismatch on nn.Linear does NOT raise — it silently misdecodes:
python - <<'PY'
import mlx.core as mx, mlx.nn as nn
from mlx.utils import tree_flatten
mx.random.seed(0)

def build(with_mode):
    class Net(nn.Module):
        def __init__(self):
            super().__init__(); self.a = nn.Linear(64, 64)
    n = Net()
    kw = dict(group_size=32, bits=8,
              class_predicate=lambda p, m: hasattr(m, "to_quantized"))
    if with_mode:
        kw["mode"] = "mxfp8"
    nn.quantize(n, **kw)
[...]

PR #791, opened 2026-09-12
fix(vlm): preserve global quantization mode for derived text models

Thump604:
Refs #783.

When `build_text_model()` derives a text model from a VLM checkpoint, its quantization setup forwards group size and bit width but drops the global mode. For an MXFP8 checkpoint, that constructs globally quantized layers as affine before loading the MXFP8 weights.

I now forward the checkpoint's global mode, defaulting to affine when it is absent. Explicit per-layer quantization settings remain authoritative.

I added regressions for MXFP8 and the affine default, plus a tiny random-weight model exercising the actual derived-text function. On MLX 0.32.0, all three focused tests pass; the numerical test checks exact embedding and linear outputs, including an explicit affine override.

I have not run the reported Gemma4 checkpoint or reproduced its MLX 0.32.2 result end-to-end. This PR is limited to the mode-plumbing fix and its bounded regression evidence.

### DIFF
--- a/tests/test_text_model_quantization_mode.py
+++ b/tests/test_text_model_quantization_mode.py
@@ -1,0 +1,121 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Source-level regression for derived text-model quantization plumbing."""
+
+import importlib
+import json
+from pathlib import Path
+import sys
+import types
+
+import pytest
+
+
+def _stub_module(name, *, package=False):
+    module = types.ModuleType(name)
+    if package:
+        module.__path__ = []
+    return module
+
+
+@pytest.mark.parametrize(
+    "global_mode, expected_mode", [("mxfp8", "mxfp8"), (None, "affine")]
+)
+def test_build_text_model_forwards_global_mode_without_overriding_layer_dicts(
+    tmp_path, monkeypatch, global_mode, expected_mode
+):
+    """Global mode reaches quantize while explicit layer dicts stay authoritative."""
+
+    mlx = _stub_module("mlx", package=True)
+    mlx_core = _stub_module("mlx.core")
+    mlx_nn = _stub_module("mlx.nn")
+    mlx_utils = _stub_module("mlx.utils")
+    mlx.core = mlx_core
+    mlx.nn = mlx_nn
+    mlx.utils = mlx_utils
+    monkeypatch.setitem(sys.modules, "mlx", mlx)
+    monkeypatch.setitem(sys.modules, "mlx.core", mlx_core)
+    monkeypatch.setitem(sys.modules, "mlx.nn", mlx_nn)
+    monkeypatch.setitem(sys.modules, "mlx.utils", mlx_utils)
+
+    mlx_utils.tree_flatten = lambda _parameters: [
+        ("global.scales", object()),
+        ("language_model.override.scales", object()),
+    ]
+
+    class FakeTextModelArgs:
+        @classmethod
+        def from_dict(cls, config):
+            return config
+
+    class FakeTextModel:
+        mtp = None
+
+        def __init__(self, _args):
+            self.loaded = []
+
+        def load_weights(self, weights, strict=False):
+            self.loaded.append((weights, strict))
+
+        def train(self, _mode):
+            return self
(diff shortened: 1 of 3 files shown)


mmcco:
I ran this PR end to end on the checkpoint from #783: `mlx-community/gemma-4-12B-it-qat-mxfp8` @ `d8989284`, M5 Max, MLX 0.32.2, mlx-lm 0.31.3. I compared `main` @ `ec8e493` against this head (`130ced3`); the installed trees differ only by the added `mode=` line.

**Layer parity, derived TextModel vs the source VLM (shared weights):**

| | base | this PR |
|---|---|---|
| Quantized modules with the wrong mode | 185 / 329 (every MXFP8 module built affine) | 0 / 329 |
| `layers.0.self_attn.q_proj` (global MXFP8), max \|Δ\| | 4,933,921.5, no error | 0.0 |
| `layers.0.mlp.gate_proj` (affine override), max \|Δ\| | 0.0 | 0.0 |
| `embed_tokens` | missing-kernel RuntimeError | 0.0 |
| Last-position logits, 3 prompts | RuntimeError | 0.0 |

**Serving, `--mllm`, SimpleEngine, text-only chat:** on base, 4/4 requests log `Text-only request → LLM path` and return HTTP 500 with the missing-kernel error. On this PR the same 4 return correct content (`Paris`, `144`, `Mercury`, and a valid strict-schema extraction). So the fix works on the real checkpoint.

Two things I hit that are outside this PR's scope:

1. **The derived route is only built when the model is served from a local directory.** Served by Hub repo id, neither build logs `MLLM text routing: text-only -> mlx_lm TextModel`, and text requests take the mlx-vlm path. The engine passes the repo id as `model_path`, and the only silent `None` return in `build_text_model()` is the `(model_path / "config.json").exists()` check. My first run "passed" on both builds for that reason. I served the snapshot directory with `--served-model-name` to exercise the fix.
2. **Once the route is reachable, it hits the missing end-of-turn stop that #610 fixes.** Unconstrained answers finish with `finish_reason: "length"` at the 256-token cap (`'Paris<turn|><turn|><turn|>thought\n<channel|>Paris<turn|>...'`), while strict-schema requests stop normally (`stop`, 63 tokens). I've added the details to #610. So an MXFP8 Gemma 4 text deployment needs both this PR and #610.

janhilgard (approved):
Approving. One line of source, and it fixes a silent-wrong-weights bug rather than a crash:

```python
mode=quantization.get("mode", "affine"),
```

`nn.quantize()` in MLX 0.32 takes `mode` as keyword-only with a default of `"affine"`, so a checkpoint carrying `"mode": "mxfp8"` (or any non-affine mode) was previously re-quantized as affine when the text model was derived from a VLM. Nothing errors — the weights are simply interpreted under the wrong scheme, which is the failure mode that costs the most time to find because the model still loads and still produces text.

The tests are the reason this is easy to approve. `test_mxfp8_global_and_affine_override_round_trip` does two things I would have asked for:

```python
assert derived.embedding.mode == "mxfp8"
assert derived.mxfp8_linear.mode == "mxfp8"
assert derived.affine_linear.mode == "affine"
…
assert bool(mx.allclose(actual, expected, atol=0, rtol=0)), name
```

It pins the global mode propagating **and** that a per-layer override is not clobbered by it — the obvious way to get this wrong is to force the global mode onto every layer — and it compares weights at `atol=0, rtol=0` rather than eyeballing shapes.

Mutation-checked. Removing the added line fails three of them:

```
FAILED …test_mxfp8_global_and_affine_override_round_trip
FAILED …test_build_text_model_forwards_global_mode_without_overriding_layer_dicts[mxfp8-mxfp8]
FAILED …test_build_text_model_forwards_global_mode_without_overriding_layer_dicts[None-affine]
```

with `KeyError: 'mode'` from the assertion, so the coverage is live rather than incidentally green.

The `[None-affine]` parametrisation is worth noting as the case that keeps the default honest: a checkpoint with no `mode` key still lands on affine, so this is not a behaviour change for the checkpoints that were already correct.

Local run: 28 passed, 5 skipped across the text-model and VLM-quantisation suites.

PR #796, opened 2026-09-15
fix(mllm): propagate authoritative prompt_tokens through batch respon…

### LINKED ISSUE #761: [Bug]: usage.prompt_tokens is always 0 on the MLLM path (0.3.0 → 0.4.1 regression for text-only workloads on VL checkpoints)
### Environment

- Chip: Apple M3 Ultra
- RAM: 256 GB
- macOS version: 26.5.2 (25F84)
- Python version: 3.12.12
- vllm-mlx version: 0.4.1 (release wheel from PyPI)
- mlx / mlx-lm / mlx-vlm / mlx-audio versions: mlx 0.32.2, mlx-lm 0.31.3, mlx-vlm 0.6.17, mlx-audio 0.5.1 (transformers 5.16.1)

### Model and quantization

Local directory `Qwen3.6-35B-A3B-mlx-8bit` (MLX 8-bit). Its `config.json` declares:

- `architectures: ["Qwen3_5MoeForConditionalGeneration"]`
- `model_type: "qwen3_5_moe"`
- VLM markers present: `vision_config`, `image_token_id`

The workload is **text-only** — chat completions and tool calling. No image or video input is ever sent.

### Summary

After upgrading 0.3.0 → 0.4.1, **`usage.prompt_tokens` is `0` on every chat completion**. `completion_tokens` remains correct, so `total_tokens` is understated by exactly the prompt length on every request.

This is a regression in *reporting* only — generation itself is correct, including multi-turn tool calling — but it silently zeroes input-token accounting for every downstream consumer (in our case an OpenAI-compatible proxy that forwards `usage` into a tracing backend, which now records 0 input tokens for all local traffic).

### Reproduction

```bash
vllm-mlx serve /path/to/Qwen3.6-35B-A3B-mlx-8bit \
  --served-model-name Qwen3.6-35B-A3B-mlx-8bit \
  --host 127.0.0.1 --port 8080

curl -s http://127.0.0.1:8080/v1/chat/completions \
  -d '{"model":"Qwen3.6-35B-A3B-mlx-8bit",
       "messages":[{"role":"user","content":"Say OK"}],
       "max_tokens":20}' | jq .usage
```

Measured on the **same model directory and the same prompts**, only the vllm-mlx version differing:

| prompt | 0.3.0 `prompt_tokens` | 0.4.1 `prompt_tokens` |
|---|---|---|
| short (~20 tok) | `20` | `0` |
| long (~600 tok) | `615` | `0` |

`completion_tokens` is correct in both.

### Root cause (as far as I traced it)

The model is now routed down the **MLLM path**, and that path does not populate per-request `prompt_tokens`.

`/health` shows the switch — same model directory, same flags:

- 0.3.0 → `"model_type":"llm"`
- 0.4.1 → `"model_type":"mllm"`

The reclassification itself looks **intentional and arguably more correct**: `is_mllm_model()` in `vllm_mlx/api/utils.py` changed from a substring match on the model name to reading `config.json` (`_config_indicates_vlm`). This checkpoint genuinely declares `vision_config` and a `…ForConditionalGeneration` architecture, so 0.4.1 classifies it as multimodal where 0.3.0 did not.

The server-side accounting code appears unchanged between versions — `prompt_tokens = output.prompt_tokens if hasattr(...) and output.prompt_tokens` — so `output.prompt_tokens` is simply arriving as `0`/unset from the MLLM generator, whereas the LLM path sets it.

This looks like the same class of gap already tracked for the MLLM path in #750 (per-request `cache_hit_type`/`cached_tokens` hardcoded as stubs for MLLM) and #756/#757 (per-request MTP accept/draft counters not attributed correctly), rather than a new subsystem.

### Expected behavior

`usage.prompt_tokens` reports the real prompt length on the MLLM path, as it does on the LLM path, so OpenAI-compatible clients get accurate `usage`.

### Notes / workaround

There does not appear to be a supported way to opt a text-only workload back onto the LLM path for a checkpoint whose config declares vision:

- `serve` accepts `--mllm` but not `--no-mllm` (that pair exists on another subcommand; passing it to `serve` exits with `unrecognized arguments`)
- in the model registry, `force_mllm` is resolved as `is_mllm = is_mllm_model(source) or bool(entry.force_mllm)` (`vllm_mlx/model_registry.py`), so a per-model `false` cannot force the text path

If a `--no-mllm` on `serve` (or a registry `mllm: false` that is honoured as a decision rather than only as "don't force") is a change you would accept, that would also give text-only users of VL checkpoints a way off the multimodal path entirely — but fixing `prompt_tokens` on the MLLM path is the more general fix, since multimodal users need correct usage too.

Happy to test a patch on this hardware.

riiji:
Fixes #761

## Summary

Fixes `usage.prompt_tokens` returning `0` (or under-reporting token counts) on the MLLM path for both text-only workloads on VL checkpoints and media-expanded requests.

### Root cause
`MLLMScheduler.add_request()` previously attempted to estimate `request.num_prompt_tokens` via `tokenizer.encode(prompt)` wrapped in `except Exception: pass`. This had two critical flaws:
1. If the tokenizer failed on multimodal templates/markers, the exception was silently swallowed and `num_prompt_tokens` remained `0`.
2. Even if it succeeded, it only tokenized raw text and could not account for media-expanded visual tokens (which are produced later by the processor).

### Fix
Per maintainer diagnosis on #761:
- Replaced the silent `except Exception: pass` in `MLLMScheduler.add_request` with a debug diagnostic logging the failure (`logger.debug(...)`).
- Propagated the authoritative processed token count (`req.input_ids.size`) from `MLLMBatchGenerator` via `MLLMBatchResponse.prompt_tokens` on request prefill/generation steps.
- In `MLLMScheduler._process_batch_responses()`, updated `request.num_prompt_tokens = response.prompt_tokens` and adjusted aggregate scheduler counters (`self.total_prompt_tokens += diff`) when authoritative counts arrive.
- Ensured consistent `usage.prompt_tokens` and `usage.total_tokens` across both normal and streaming `/v1/chat/completions` responses.

## Type of change

- Bug fix

## Surface touched

- OpenAI API endpoints
- Scheduler / batching
- Multimodal pipeline

## Test plan

Run regression tests verifying prompt token propagation and error logging:
```bash
pytest tests/test_mllm_prompt_tokens.py -v
```

Verification covers:
1. `MLLMBatchResponse` schema carrying `prompt_tokens`.
2. `MLLMBatchGenerator._next()` setting `prompt_tokens` directly from `req.input_ids.size`.
3. `MLLMScheduler._process_batch_responses()` updating `request.num_prompt_tokens` and adjusting `self.total_prompt_tokens`.
4. End-to-end propagation to `RequestOutput.prompt_tokens` matching `req.input_ids.size`.
5. Preprocessing failures cleanly maintaining 0 without crashing.

## Test results

image

## Performance impact

N/A, zero overhead. Reads `.size` from the already preprocessed `req.input_ids` array and assigns an integer field on `MLLMBatchResponse`.

## Output parity

Generated tokens and model outputs remain byte-for-byte identical to `main`. The only divergence is accurate `usage.prompt_tokens` and `usage.total_tokens` accounting in the response payload.

## Risk notes

- Low risk: `MLLMBatchResponse.prompt_tokens` is optional (`None` by default).
- Non-MLLM paths do not use `MLLMBatchGenerator` and are completely unaffected.
- The scheduler only updates `request.num_prompt_tokens` and adjusts aggregate counters if `response.prompt_tokens is not None` and differs from current accounting.

## Testing Artifacts

Before fix (183 tokens, because of the text prompt + model answer before) 

image

After the fix (945 tokens on the attached image, including the text prompt before) 

image


riiji:
Hi guys! I'm new here, researching this codebase for a couple of days now. This code is made by LLM since I barely understand what's going on. But did some manual testing to prove this code works. I hope for a couple of weeks I'll get used to it

riiji:
I came here because I made a small pet project that outperformed mlx-lm (I've written custom shaders in C++) and found that @Thump604 already did that and went here because the maintainers of mlx-lm didn't want to make mlx-lm fast 😅

PR #798, opened 2026-09-15
fix(registry): detect and reap dead engines instead of reporting loaded forever

djacobsmeyer:
## Problem

In registry mode, if an engine's background loop exits without going through the
engine's own `stop()` path, `ModelManager` never notices. `/v1/models` keeps reporting
the model as `loaded`, its memory stays charged against the budget so nothing else can
be admitted, and every subsequent `acquire()` hands out a lease on an engine that will
never make progress. The only recovery is restarting the server.

## Root cause

`vllm_mlx/model_registry.py` treats presence in `_loaded` as liveness. `list_models()`
derives state purely from that dict — line 753 (`"preempting" if loaded.preempting else
"loaded"`) and line 776 (`"loaded": loaded is not None`) — and `acquire()` calls
`_claim_loaded_locked()` (line 857) before considering anything else, so a stale entry is
re-leased forever. Nothing in the engine layer exposes whether the loop is still running.

## Fix

Give engines a health predicate and use it in the two places that lie:

- `BaseEngine.is_healthy()` defaults to `True` — engines with no persistent background
  task (SimpleEngine runs each request synchronously) cannot go stale this way.
  `EngineCore`, `MLLMScheduler` and `BatchedEngine` override it: a task that is `done()`
  while the engine still believes it is running means something fatal escaped the loop's
  own per-step `try/except`, which otherwise swallows exceptions and keeps looping.
- `list_models()` reports `state: "dead"` / `loaded: False` at read time and mutates
  nothing — a read endpoint should not have side effects.
- `acquire()` reaps the stale entry through the normal unload path, then loops back
  around and loads a fresh engine.

Reaping is gated on `active_requests == 0`, so an in-flight lease is never yanked out
from under its caller; a dead engine still holding requests is reaped on the next
`acquire()` once `release()` drops the count to zero.

## Tests

Four added to `tests/test_model_registry.py`:
`test_list_models_reports_dead_engine_truthfully` (reports, mutates nothing),
`test_acquire_reaps_dead_engine_and_reloads` (old engine stopped, fresh one leased),
`test_acquire_does_not_reap_dead_engine_with_active_requests` (the in-flight guard),
`test_healthy_engine_is_untouched` (happy path unchanged). That file is already in the
`test-apple-silicon` job list (`.github/workflows/ci.yml`, "Run MLX-dependent tests"), so
CI covers them on 3.11/3.13 — no CI change needed. The per-engine `is_healthy()`
overrides have no dedicated test; they are thin task-state predicates exercised
indirectly by `tests/test_engine_core_*`. Glad to add direct ones if you want them.

## Verification

At `edfcfac` (on `ec8e493`), arm64, Python 3.13, `uv venv` + `uv pip install -e
".[dev,vision,harmony]"` (matching the apple-silicon job):

```
pytest tests/test_model_registry.py           -> 64 passed
pytest -k "reap or dead_engine or healthy"    -> 6 passed
pytest <full test-apple-silicon job list>     -> 1049 passed, 36 deselected
pytest <EngineCore stream-affinity job list>  -> 27 passed
ruff check vllm_mlx/ tests/ --select E,F,W --ignore E402,E501,E731,F811,F841 -> passed
black --check vllm_mlx/ tests/                -> unchanged
```

https://claude.ai/code/session_01F71QnQEMDoVWuah88HmAfa


riiji (changes requested):
Hi @djacobsmeyer 

I've tested the code
Launch any model, and stop it on the decoding phase, the model will stuck

riiji:
wait, main works the same way

riiji:
okay, looks like this code makes things better actually but the problem with models being stuck still exists

PR #803, opened 2026-09-17
Honor temperature and top-p sampling in MLLM image requests

### LINKED ISSUE #799: [Bug]: temperature/top_p are silently dropped on the MLLM path (mllm.py passes temp= but mlx-vlm expects temperature=) — every image request decodes greedily
### Environment

- vllm-mlx **0.4.1** (also reproduced against `main` @ `ec8e493` by reading the source — the call sites are unchanged)
- mlx-vlm **0.7.1** and **0.6.17** (identical behaviour on both), mlx 0.32.2, mlx-lm 0.31.3
- macOS 26 / Apple M3 Ultra 256 GB, Python 3.12
- Launch: `vllm-mlx serve <Qwen3.6-35B-A3B-mlx-8bit dir> --served-model-name Qwen3.6-35B-A3B-mlx-8bit --embedding-model <Qwen3-Embedding-4B> --host 127.0.0.1 --port 8080` (model routed to the MLLM path via `config.json` — `/health` reports `"model_type":"mllm"`)

### Model and quantization

Qwen3.6-35B-A3B, MLX 8-bit (`Qwen3_5MoeForConditionalGeneration`, has `vision_config`). Any checkpoint that takes the MLLM path should reproduce.

### Reproduction

Send the **same image request three times with `temperature: 1.0`** and hash the content:

```python
import json, urllib.request, base64, hashlib
img = "data:image/png;base64," + base64.b64encode(open("red.png","rb").read()).decode()  # any small image
def run(content, extra):
    body = {"model": "Qwen3.6-35B-A3B-mlx-8bit", "max_tokens": 120,
            "messages": [{"role": "user", "content": content}],
            "chat_template_kwargs": {"enable_thinking": False}}
    body.update(extra)
    req = urllib.request.Request("http://127.0.0.1:8080/v1/chat/completions",
                                 data=json.dumps(body).encode(), headers={"Content-Type": "application/json"})
    d = json.load(urllib.request.urlopen(req))
    return hashlib.md5(d["choices"][0]["message"]["content"].encode()).hexdigest()[:8]

IMG = [{"type": "text", "text": "Write a 60-word whimsical story inspired by this colour."},
       {"type": "image_url", "image_url": {"url": img}}]
TXT = "Write a 60-word whimsical story inspired by the colour red."
print("image temp=1.0:", [run(IMG, {"temperature": 1.0, "top_p": 0.95}) for _ in range(3)])
print("text  temp=1.0:", [run(TXT, {"temperature": 1.0, "top_p": 0.95}) for _ in range(3)])
```

Observed:

```
image temp=1.0: ['c3f1d6cb', 'c3f1d6cb', 'c3f1d6cb']   <- identical: greedy
text  temp=1.0: ['5f12fa78', 'e6f29948', 'de21f959']   <- varies as expected
```

Also checked on the image path: `repetition_penalty` and `presence_penalty` DO change the output (they reach mlx-vlm), `top_k` and `frequency_penalty` do not.

### Expected behavior

`temperature` / `top_p` from the request are applied to image (multimodal) requests exactly as they are to text-only requests.

### Actual behavior

Every request that contains an image is decoded **greedily** regardless of the requested temperature. On a busy screenshot this produces the classic greedy failure mode — one line repeated until `max_tokens` (we hit a 32 770-token response of a single sentence from a chat client that sends no sampling params, and a `temperature: 1.0` request behaves the same).

**Cause (source):** `vllm_mlx/models/mllm.py` passes the sampling temperature to mlx-vlm as `temp=`:

- `main` @ `ec8e493`: lines **393**, **2044**, **2168**, **2475** — `temp=temperature,` into `stream_generate(...)` / `generate(...)`
- mlx-vlm's `GenerateKwargs` (`mlx_vlm/generate/types.py`) accepts **`temperature`**, `top_p`, `top_k`, `repetition_penalty`, `sampler` — there is no `temp` key, so the value is swallowed by `**kwargs` and mlx-vlm's `DEFAULT_TEMPERATURE = 0.0` (`mlx_vlm/generate/common.py`) applies. Same on mlx-vlm 0.6.17, so this is not a recent mlx-vlm rename.

**Fix that works here:** rename the four call-site kwargs `temp=temperature` → `temperature=temperature`. With that one-line change (applied locally to 0.4.1), three `temperature: 1.0` image requests give three different outputs, text-only greedy output is byte-identical to before, and embeddings are unchanged. A patch file is trivial; happy to open a PR if wanted.

### Additional context

- Related MLLM-path accounting issues: #761 (prompt_tokens 0), #750, #756 — this one is a correctness bug rather than accounting.
- Because the server's own fallback for a *missing* temperature is also 0.0 (`api/models.py`), a plain chat client that sends no sampling hits the same greedy path even on text; `--default-temperature` covers that case, but it cannot help image requests while the kwarg is dropped.

waybarrios:
Image requests could decode greedily even with a nonzero `temperature`, because four MLLM wrapper calls passed `temp` to mlx-vlm, which expects `temperature`. SimpleEngine also consumed `top_p` without forwarding it to MLLM chat generation.

Use the supported temperature keyword and preserve `top_p` across blocking and streaming MLLM chat dispatch. Add 17 regression cases covering the four wrapper routes, worker dispatch, greedy temperature, explicit nucleus sampling, and engine defaults. Include the wrapper regressions in the existing Apple Silicon CI job and document image sampling parameters.

Unit test examples:

The tests exercise the real wrapper and engine methods with inference stubs. The stub returns `temperature:top_p:top_k`, making dropped parameters directly observable without relying on random model output.

For example, using the `model` fixture in [test_mllm_sampling.py](https://github.com/waybarrios/vllm-mlx/blob/a58f5f68336198fc8f1248aac4c41b71f87bd392/tests/test_mllm_sampling.py):

```python
result = model.generate(
    prompt="Describe this image",
    images=["test-image"],
    temperature=0.8,
    top_p=0.37,
    top_k=17,
)
assert result.text == "0.8:0.37:17"
# Before the fix: "0.0:0.37:17"
```

The parametrized test checks `generate`, `stream_generate`, `chat`, and `stream_chat`, with both `temperature=0.0` and `temperature=0.8` (8 cases).

The [engine regression](https://github.com/waybarrios/vllm-mlx/blob/a58f5f68336198fc8f1248aac4c41b71f87bd392/tests/test_simple_engine_thread_pinning.py#L268) checks the parameters delivered to the model on blocking chat, streaming chat, and native-video worker dispatch (9 cases):

| Engine request | Received before fix | Expected after fix |
|---|---|---|
| `top_p=0.37` | `1.0` | `0.37` |
| `top_p` omitted | `1.0` | `0.9` (engine default) |
| `top_p=1.0` | `1.0` | `1.0` (control) |

These engine tests validate dispatch to the model stub, not native-video model inference.

Run the 17 new cases:

```bash
pytest tests/test_mllm_sampling.py tests/test_simple_engine_thread_pinning.py \
  -k 'sampling_parameters_reach_mlx_vlm or mllm_chat_sampling_reaches_generation_worker' -v
```

Validation:

- Before the runtime fix, 10 regression cases failed with the expected dropped-parameter values.
- Focused checks: 35 passed.
- Unit test selection on Python 3.13: 1,429 passed, 19 skipped, 17 deselected.
- Ruff 0.16.8, Black 26.5.1, and `git diff --check` passed.
- Confirmed the temperature keyword against mlx-vlm v0.6.5, the supported minimum, and v0.7.1.
- Regression tests use inference stubs; real Apple Silicon model inference remains unverified.

Fixes #799


riiji:
Tested and validated with script

```
python3 -c '
import io, base64
from PIL import Image
from vllm_mlx.models.mllm import MLXMultimodalLM

img = Image.new("RGB", (32, 32), color="red")
buf = io.BytesIO()
img.save(buf, format="PNG")
data_url = f"data:image/png;base64,{base64.b64encode(buf.getvalue()).decode()}"

print("Loading model...")
model = MLXMultimodalLM("mlx-community/Qwen2-VL-2B-Instruct-4bit", enable_cache=False)
model.load()

messages = [
    {
        "role": "user",
        "content": [
            {"type": "text", "text": "Write a 15-word imaginative poem inspired by this color."},
            {"type": "image_url", "image_url": {"url": data_url}}
        ]
    }
]

print("\nRunning 3 generations with temperature=1.0:")
outputs = []
for i in range(3):
    res = model.chat(messages, max_tokens=30, temperature=1.0)
    text = res.text.strip()
    outputs.append(text)
    print(f"Run {i+1}: {text}")

if len(set(outputs)) == 1:
    print("\nResult: Failure. All 3 runs produced identical output (greedy decoding; temperature=1.0 was ignored).")
else:
    print("\nResult: Success. Outputs vary across runs (sampling active).")
'
```


output (main branch): 

```
Run 1: Amber hues dance in the sky,
A warm glow that warms the heart.
A color that brings joy and cheer,
A symbol of love and
Run 2: Amber hues dance in the sky,
A warm glow that warms the heart.
A color that brings joy and cheer,
A symbol of love and
Run 3: Amber hues dance in the sky,
A warm glow that warms the heart.
A color that brings joy and cheer,
A symbol of love and

Result: Failure. All 3 runs produced identical output (greedy decoding; temperature=1.0 was ignored).
```

output (this branch)
```
Running 3 generations with temperature=1.0:                                                                                                                               |  0.00B /  0.00B            
Run 1: Red like flames and rustling leaves,
A secret language unspoken,
A fire-red passion dancing,
Harnessing magic in every frame.
Run 2: A skyward hue so bright:人间清澈见无疑.
Run 3: Amber hues gleam through the DVR like lifelike pixels, conveyance données flow.

Result: Success. Outputs vary across runs (sampling active).
```

docs/guides/multimodal.md
```diff
@@ -46,6 +46,11 @@ response = client.chat.completions.create(
 print(response.choices[0].message.content)
 ```
 
+Image requests honor `temperature` and `top_p` in both streaming and
+non-streaming responses. For sampling, pass values such as `temperature=0.8`
+and `top_p=0.9` to `client.chat.completions.create`; use `temperature=0.0`
+for greedy decoding.
+
```
riiji, line 53: should it be there? look like LLM's artifact

riiji (approved):

PR #805, opened 2026-09-17
feat(models): introduce native model implementations

riiji:
# feat(models): introduce native model implementations and architecture registry

## Summary

This PR introduces an architectural foundation for **native model implementations** in `vllm-mlx`, along with an opt-in registry and a first reference implementation for Qwen2 / Qwen2.5.

Why Qwen2? My small pet-project uses Qwen2

Currently, `vllm-mlx` relies directly on upstream `mlx-lm` model definitions. While functional, this limits optimization opportunities and forces bug fixes or adaptations to rely on runtime monkey-patching. With native models enabled via `--enable-native-models`, `vllm-mlx` gains full ownership over model architecture, layer execution, and weight loading.

Why we are doing this:
1. **Direct performance optimizations**: Full control over layer definitions allows optimizations like weight fusion ($Q, K, V \rightarrow QKV$ and $Gate, Up \rightarrow GateUp$), reducing Metal kernel dispatch overhead and improving prefill throughput on Apple Silicon.
2. **Clean architecture over monkey-patching**: Upstream `mlx-lm` quirks or missing features can be resolved cleanly inside native model classes instead of monkey-patching third-party modules at runtime.
3. **Foundation for custom Metal kernels**: Provides a clean entry point to introduce specialized Metal shaders, custom attention mechanisms, and kernel fusions in future PRs.

Why this approach:
1. **Aligned with vLLM design**: vLLM maintains its own native model definitions (`vllm/model_executor/models/`). Following the same pattern provides a scalable and predictable model execution system for `vllm-mlx`.
2. **Safe and opt-in**: Native models are gated behind `--enable-native-models`. If disabled, or if a model architecture is not registered, the system falls back to standard upstream `mlx-lm` without any behavioral change.
3. **Maintainable registry**: Native architectures are registered in `vllm_mlx/native_models/` and loaded dynamically based on model config.

## Type of change

- New feature
- Performance

## Surface touched

- CLI / serve command
- Metal kernels / v1 engine

## Test plan

Run unit and integration test suite:
```bash
pytest tests/test_native_qwen2.py
pytest tests/test_cli.py
```

Benchmark comparison (Qwen2.5-7B):
```bash
# Baseline (upstream mlx-lm)
python3 -m vllm_mlx.benchmark \
  --model mlx-community/Qwen2.5-7B-Instruct-4bit \
  --prompts 5 \
  --max-tokens 256

# With native models enabled
python3 -m vllm_mlx.benchmark \
  --model mlx-community/Qwen2.5-7B-Instruct-4bit \
  --prompts 5 \
  --max-tokens 256 \
  --enable-native-models
```

Interactive chat via serve:
```bash
python3 -m vllm_mlx.cli serve \
  mlx-community/Qwen2.5-7B-Instruct-4bit \
  --port 8000 \
  --enable-native-models
```

## Test results

Unit test suite results:
image

## Performance impact

Tested using the initial reference implementation (Qwen2.5-7B) to demonstrate the benefits of native layer execution:

- **Hardware**: Apple M4 Max (36 GB RAM)
- **Model**: `mlx-community/Qwen2.5-7B-Instruct-4bit`
- **Prefix cache cleared before test** (to be 100% sure I ran my branch first, main second)

#### Prefill (TTFT / prompt TPS):
| Prompt set | Concurrency | Prompt tokens | Baseline TTFT (ms) | Native TTFT (ms) | Delta |
|---|---|---|---|---|---|
| short | 1 | 43 | 120 - 125 ms | 125 - 126 ms | ~same |
| short | 4 | 43 | ~440 - 445 ms | ~440 - 443 ms | ~same |
| medium | 1 | 568 | ~742 - 747 ms | ~740 - 741 ms | ~same |
| medium | 4 | 568 | ~4100 - 4250 ms | ~3714 - 3720 ms | **11-13% faster** |
| long | 1 | 2513 | 3837 - 4245 ms | ~3175 ms | **~25% faster** |
| long | 4 | 2513 | 19784 - 22434 ms | 16533 - 17538 ms | **~20% faster** |

#### End-to-end latency:
- `long` prompt set at concurrency 4: dropped from **~33.0 s** down to **~27.7 s** (~16% faster overall).
- Generation step (TPOT) remains identical (~12.3 ms / token at concurrency 1), as single-token autoregressive decoding is strictly memory-bandwidth bound.

## Output parity

Verified in `tests/test_native_qwen2.py::test_qwen2_numerical_parity`:
- Greedy token predictions match upstream `mlx-lm` output byte-for-byte.
- Maximum absolute difference in float tensor logits is `< 1e-4` (caused by standard floating-point summation order in concatenated matrix operations).

## Risk notes

- **Zero regression risk by default**: Opt-in only via `--enable-native-models` (`default=False`). Existing production pipelines remain completely unaffected.
- **Graceful fallback**: If any error occurs while loading a native model or transforming weights, the loader catches the exception, logs a warning, and falls back to standard `mlx-lm` loading.
- **Strict architecture matching**: Unsupported model types (e.g. `qwen2_vl` or other non-registered families) are automatically directed to standard upstream loading.


riiji:
Hi! @waybarrios @Thump604 

I have an idea in my mind, instead of relying 100% on mlx, we can implement our own models like developers do  it in vLLM, this might help with a lot of issues (including compatibility of mlx-lm with other models)

what do you think about it? should be safe because it's under the flag and also gives some performance

waybarrios:
Pretty interesting! Sure thing it is possible. We could add metal kernels,  I was thinking on work in that part too. Check this documentation: https://developer.apple.com/documentation/metal 

Let me know what you think. Happy to discuss!

waybarrios:
Btw check this one too: https://github.com/jart/matmul It is pretty fast, this was added to llamafile (https://github.com/mozilla-ai/llamafile) purely based on c++

riiji:
Hi @waybarrios , this is a nice idea with matmul, actually, in my mind I have some kinda roadmap about this feature (I use framework called baby steps 😅 ) 

1. Add this feature under the --enable-native-models flag. So the current users shouldn't notice about this experimental flag. If this flag is enabled, we also have a fallback to the mlx-lm models
2. Start adding Native Models with criteria:

- Performance (Prefill or Decode) should be the same or better than without --enable-native-models flag
- Memory consumption should be the same or better than without --enable-native-models flag (or higher memory consumption can be explained)
- Accuracy must be within numerical tolerance (e.g. float16, float32, or something else)

3. If we're happy with Native Models, remove this flag, make it by default with a fallback to mlx-lm
4. Remove the patches folder once Native Models cover the functionality
5. Optimizations (including bringing C++ in with bindings, etc). I'm not actually sure we want to do it now because it will require users to install xcode and other stuff to make it work. I saw @Thump604 tried to get the C++ into this project. 

So yeah, in my mind I see https://github.com/jart/matmul  this in this codebase, but maybe, not at the current state. What do you think? 

p.s. I'm currently testing this code and implementation, later, I'll come with more benchmarks

riiji:
thought, If you want to bring C++ in this PR, I don't mind. we can try it, maybe I'll run some experimentations. This idea is very tempting 😄

riiji:
I've run the tests again, no, it doesn't work the way I want to. Mathematically it works perfectly and we reduce the number of dispatches but practically I see that the accuracy is dropping below expected + increased memory consumption. Still looking for some way to handle it 👀 I'm gonna try your idea with Metal

